PR #37, opened 2017-04-07
Add support for building both MPI and NOMPI libs in the same build.

chuckatkins:


chuckatkins:
@pnorbert @williamfgc  Does this work for you to build both mpi and nompi versions at once?

chuckatkins:
I have verified that this works on both "normal linux" and on Titan.

PR #45, opened 2017-04-07
Code structure changes and syntax

williamfgc:
Added utilities: include/utilities and source/utilities and modified all corresponding inclusions.

utilities/format/bp1
utilities/profiling/iochrono

To be added (by @JasonRuonanWang ):
include/utilities/comm/dataman
source/utilities/comm/dataman

Passing strings:
const std::string& str -> long ( 15 characters) string is common or going to STL container
const std::string str -> supposed to be a small string (  <= 15 characters )
std::string& str -> string to be modified
std::string str  -> rarely (if not) used 

Other fixes:
Functions must all start with upper case.


chuckatkins:
Can you please rebase off master to get rid of all the master merges?

CMakeLists.txt
```diff
@@ -52,20 +52,20 @@ include(CMakeDependentOption)
 # Setup shared library / -fPIC stuff
 get_property(SHARED_LIBS_SUPPORTED GLOBAL PROPERTY TARGET_SUPPORTS_SHARED_LIBS)
 cmake_dependent_option(ADIOS_BUILD_SHARED_LIBS
-  "Whether or not to build shared libraries" ON
+  "Whether or not to build shared libraries" OFF
```
chuckatkins: Why change this to default to off?

CMakeLists.txt
```diff
@@ -52,20 +52,20 @@ include(CMakeDependentOption)
...
   "SHARED_LIBS_SUPPORTED" OFF)
 if(SHARED_LIBS_SUPPORTED)
   cmake_dependent_option(ADIOS_ENABLE_PIC
-    "Build with Position Independent Code" ON
+    "Build with Position Independent Code" OFF
```
chuckatkins: same here, why change the defaults?

CMakeLists.txt
```diff
@@ -52,20 +52,20 @@ include(CMakeDependentOption)
...
     "NOT ADIOS_BUILD_SHARED_LIBS" ON)
 endif()
 set(BUILD_SHARED_LIBS ${ADIOS_BUILD_SHARED_LIBS})
 
 if(ADIOS_ENABLE_PIC)
-  set(CMAKE_POSITION_INDEPENDENT_CODE ON)
+  set(CMAKE_POSITION_INDEPENDENT_CODE OFF)
```
chuckatkins: This explicitly needs to be ON.

chuckatkins:
CMake changes are still problematic

CMakeLists.txt
williamfgc: We use the .a library as default as in adios1, I think it has to do with some systems having trouble with so libraries.

CMakeLists.txt
williamfgc: same as before

chuckatkins:
If rebasing isn't viable then squashing the branch into more self-contained commits should do it.

williamfgc:
Will rebase....default should be static library. I think only DataMan needs .so

PR #50, opened 2017-04-10
Fixing aggregation formatting in .bp/profiling.log

williamfgc:
Now meets the python dictionary format


williamfgc:
It's 'cause  rankLog is sent over MPI as MPI_CHAR to a vector<char>,  stringstream will need to convert to string and sent c_str()...I don't think stringstream is guaranteed to be contiguous in memory

PR #53, opened 2017-04-11
Fix FindDataMan.cmake missing include.

chuckatkins:
FindPackageHandleStandardArgs needs to be included in every find module
that the find_package_handle_standard_args is used.  I was getting
transitively included by othger find modules before but when all other
options are turned off except DataMan, it's missing.  This ensures it's
always included before calling find_package_handle_standard_args.


chuckatkins:
Note that this is the same change from #52 but it's just formatted a bit differently for how we want CMake code to be moving forward and has a more informative commit message.

PR #54, opened 2017-04-11
Move BP1Writer templates out of header

williamfgc:
Separating templates declarations (BP1Writer.h) and definitions (BP1Writer.tcc)


include/utilities/format/bp1/BP1Writer.h
```diff
@@ -527,4 +282,6 @@ class BP1Writer : public BP1Base
 } // end namespace format
 } // end namespace adios
 
+#include "utilities/format/bp1/BP1Writer.tcc"
```
chuckatkins: One of the reasons to separate the template implementation is to also not have to always have it included in user code.  I don't believe any of the template implementations here need to be public facing.  We should be able to drop this from the .h file and only include it in the .cpp

williamfgc: Once included in .cpp....how are the .tcc contents visible when calling the functions in .h?

PR #56, opened 2017-04-12
added DataMan into utilities

JasonRuonanWang:
Added DataMan into utilities

Changes:

1. Added DataMan.h and DataManager.h in include/utilities/realtime/dataman/DataMan.h
2. Added DataManager.cpp in source/utilities/realtime/dataman/DataManager.cpp
3. Changed DataManReader and DataManWriter engine to use adios::realtime::DataManager
4. Changed source/CMakeLists.txt to compile DataMan related files by default.
5. Changed ADIOS.cpp to remove ADIOS_HAVE_DATAMAN related stuff

Still need to do:

1. Remove ADIOS_HAVE_DATAMAN completely from all cmake files
2. Compile hello_datamanWriter and hello_datamanReader by default


JasonRuonanWang:
@chuckatkins Please have a look and let me know if there is any problems. Thanks.

include/utilities/realtime/dataman/DataManager.h
```diff
@@ -0,0 +1,44 @@
+#ifndef DATAMANAGER_H_
```
chuckatkins: Add license header

include/utilities/realtime/dataman/DataManager.h
```diff
@@ -0,0 +1,44 @@
...
+
+namespace adios
+{
+namespace realtime
+{
+
+class DataManager : public DataMan
+{
+public:
+    DataManager();
```
chuckatkins: Since both constructors and destructor have empty implementations, they should use `default` instead.  Also, always make the destructor of an inherited class virtual; i.e:
```
DataManager() = default;
virtual ~DataManager() = default;
```

include/utilities/realtime/dataman/DataMan.h
```diff
@@ -0,0 +1,468 @@
...
+#include <iostream>
+#include <memory>
+#include <string>
+#include <unistd.h>
+#include <vector>
+
+#include "external/json.hpp"
+
+using json = nlohmann::json;
+using namespace std;
```
chuckatkins: Never use `using namespace std`; always explicitly use `std::foo`

include/utilities/realtime/dataman/DataMan.h
```diff
@@ -0,0 +1,468 @@
...
+#include <functional>
+#include <iostream>
+#include <memory>
+#include <string>
+#include <unistd.h>
+#include <vector>
+
+#include "external/json.hpp"
+
+using json = nlohmann::json;
```
chuckatkins: Move the `using json ...` into the class definition to prevent it from polluting other classes when included

include/utilities/realtime/dataman/DataMan.h
```diff
@@ -0,0 +1,468 @@
+#ifndef DATAMAN_H_
```
chuckatkins: Add license header

include/utilities/realtime/dataman/DataMan.h
```diff
@@ -0,0 +1,468 @@
...
+#include <dlfcn.h>
+#include <functional>
+#include <iostream>
+#include <memory>
+#include <string>
+#include <unistd.h>
+#include <vector>
+
+#include "external/json.hpp"
+
```
chuckatkins: Try to logically group your includes and then sort within the groups.  Something like:
```
// std c headers
// system c headers
// std c++ headers
// other package headers
```
So in this case, it would look like
```
#include <cstdint>

#include <dlfcn.h>
#include <unistd.h>

#include <chrono>
#include <functional>
#include <iostream>
#include <memory>
#include <string>
#include <vector>

#include "external/json.hpp"
``` 

And once the implementation of members is moved to a seperate cpp file, do `dlfcn.h` and `unistd.h` need to be in the public header?  Try to move them to the cpp file as well if possible.

include/utilities/realtime/dataman/DataMan.h
```diff
@@ -0,0 +1,468 @@
...
+    std::function<void(const void *, string, string, string, vector<size_t>)>
+        m_callback;
+    map<string, shared_ptr<DataMan>> m_next;
+
+private:
+    json m_profiling;
+    chrono::time_point<chrono::system_clock> m_start_time;
+    chrono::time_point<chrono::system_clock> m_step_time;
+    bool m_profiling_enabled = false;
+};
```
chuckatkins: Since none of the functions involve templates then you can easily move thier implementation into a seperate cpp file.

include/utilities/realtime/dataman/DataMan.h
```diff
@@ -0,0 +1,468 @@
...
+class DataMan
+{
+public:
+    DataMan()
+    {
+        m_profiling["total_manager_time"] = 0.0f;
+        m_profiling["total_mb"] = 0.0f;
+        m_start_time = chrono::system_clock::now();
+    }
+    virtual ~DataMan() {}
```
chuckatkins: use `= default;` instead of providing an empty implementation

include/utilities/realtime/dataman/DataMan.h
```diff
@@ -0,0 +1,468 @@
...
+        else
+        {
+            for (auto i : m_next)
+            {
+                i.second->reg_callback(cb);
+            }
+        }
+    }
+
+    void dump(const void *p_data, json p_jmsg)
```
chuckatkins: Rather than dumping to cout, take an arbitrary ostream that defaults to cout:
```
void dump(const void *p_data, json p_jmsg, std::ostream &out = std::cout)
```

include/utilities/realtime/dataman/DataMan.h
```diff
@@ -0,0 +1,468 @@
...
+        if (dtype == "unsigned long")
+            return sizeof(unsigned long);
+        if (dtype == "float")
+            return sizeof(float);
+        if (dtype == "double")
+            return sizeof(double);
+        if (dtype == "long double")
+            return sizeof(long double);
+        if (dtype == "complex<float>")
+            return sizeof(float) * 2;
```
chuckatkins: You can safely use `sizeof(std::complex<float>)` here.

include/utilities/realtime/dataman/DataMan.h
```diff
@@ -0,0 +1,468 @@
...
+        }
+        else
+        {
+            logging("Dynamic library " + soname +
+                    " not found in LD_LIBRARY_PATH");
+        }
+        return nullptr;
+    }
+
+    string run_cmd(string cmd)
```
chuckatkins: This entire function is super dangerous.  Why is it needed?

source/ADIOS.cpp
```diff
@@ -25,10 +25,8 @@
 #include "engine/bp/BPFileReader.h"
 #include "engine/bp/BPFileWriter.h"
 
-#ifdef ADIOS_HAVE_DATAMAN // external dependencies
 #include "engine/dataman/DataManReader.h"
```
chuckatkins: This should still be guarded with an `#ifdef` since the DataMan code will only be supported for shared lib builds.

include/utilities/realtime/dataman/DataMan.h
```diff
@@ -0,0 +1,468 @@
...
+        }
+        p_jmsg["putbytes"] = product(p_jmsg["putshape"].get<vector<size_t>>(),
+                                     dsize(p_jmsg["dtype"].get<string>()));
+        p_jmsg["varbytes"] =
+            product(varshape, dsize(p_jmsg["dtype"].get<string>()));
+    }
+
+    inline shared_ptr<DataMan> get_man(string method)
+    {
+        string soname = "lib" + method + "man.so";
```
chuckatkins: Does this need to check `#ifdef __APPLE__` and use .dylib instead of .so?

source/utilities/realtime/dataman/DataManager.cpp
```diff
@@ -0,0 +1,70 @@
+#include "utilities/realtime/dataman/DataManager.h"
+
+namespace adios
+{
+namespace realtime
+{
+
+DataManager::DataManager() : DataMan() {}
+DataManager::~DataManager() {}
```
chuckatkins: Drop both of these in favor of "= default" in the header.

source/utilities/realtime/dataman/DataManager.cpp
```diff
@@ -0,0 +1,70 @@
+#include "utilities/realtime/dataman/DataManager.h"
```
chuckatkins: Add license header

include/utilities/realtime/dataman/DataMan.h
JasonRuonanWang: DataMan always makes .so libraries regardless the OS.

include/utilities/realtime/dataman/DataMan.h
JasonRuonanWang: It is needed because the MdtmMan uses named pipes to communicate with MDTMFTP. In this case MdtmMan needs to delete the pipes every time it is finished, which has to be done by running an external "rm" command as far as I know. However, this is only used in MdtmMan, so I can move it to the MdtmMan.cpp for the time being.

source/ADIOS.cpp
JasonRuonanWang: Actually in this case everything can be put into libadios2.a now. I have had a discussion with @williamfgc in the afternoon. Once DataMan is integrated into ADIOS2, we don't need to build shared libs anymore, but we will still need to keep the flexibility of loading arbitrary dynamic libraries in the DataMan engine.

include/utilities/realtime/dataman/DataMan.h
chuckatkins: Right, so that's​not always possible.  In the default Cray environment, the compiler can only build​ static libs.  I'd remove dataman as a user settable option but then make including it dependant on whether shared libs are being used.

include/utilities/realtime/dataman/DataMan.h
chuckatkins: You can use the unlink sys call instead.  Much safer than a function to execute an arbitrary shell command.

include/utilities/realtime/dataman/DataMan.h
JasonRuonanWang: This is just for run-time optimization. If something optimal can be loaded dynamically, then DataMan will use it. If not, then it should not affect how the basic static part of DataMan works. For example, whenever anything goes wrong, no matter it is because of the Cray environment or a wrongly built XxxMan.so, this function will simply return a nullptr, and then the main DataMan runtime optimizer will just ignore it and keep doing what is provided in the .a library.

source/ADIOS.cpp
chuckatkins: Still though, if built 'fully static' I don't think it will work.  So rather than condition it's inclusion in whether shared libs are being used, it should still be conditioned on whether shared libs are even supported by the platform.  See in the top level CMakeLists.txt how ADIOS_USE_PIC is set, it uses the same check.

JasonRuonanWang: I am completely fine if we keep the ADIOS_HAVE_DATAMAN and keep DataMan an optional thing. That's actually what I want too, because it will make me less nervous. But I have an impression that I once made something similar work -- loading a shared library in code built as a static library. I don't know if that static library is enough "fully static" though.

Anyway, I will keep DataMan as an optional thing.

chuckatkins: Not really optional, but conditional.  So a user won't be able to turn it on or off, but CMake will determine whether or not to enable it based on whether the platform supports it.  It will work fine from a normal static lib, but when it's built with the -static flag (default for Cray) then you'll get all sorts of warnings regarding how the dlopen functions won't work.

JasonRuonanWang: OK. I got it now. Sure. That makes perfect sense. So what I need to do is just reverse ADIOS.CPP to how it looked like before?

JasonRuonanWang: Also, what should I do after finishing these changes? Do another commit and pull request?

chuckatkins: Yep.  Keep the ifdefs.  Regarding workflow, just keep using this pr and just add more commits to the existing branch.  The or tracks your branch so when additional commits are added to the branch then the or is automatically updated.

include/utilities/realtime/dataman/DataManager.h
JasonRuonanWang: Done.

include/utilities/realtime/dataman/DataMan.h
JasonRuonanWang: Done.

include/utilities/realtime/dataman/DataMan.h
JasonRuonanWang: Done.

JasonRuonanWang:
@chuckatkins I have finished all changes and compiled successfully on my machine. But the CI build somehow failed. Could you please have a look? Thanks.

JasonRuonanWang:
@chuckatkins If I am going to put other managers into ADIOS2, then they can be built into the same libadios2.a too. But that only happens if those managers are well tested and proved safe to put into ADIOS2. The ability of loading dynamic libraries is really for loading some experimental managers for just quickly evaluating research ideas, or ones that are overly dangerous to put into, such as the one that uses named pipes extensively. For these managers I will prefer keep them separated from ADIOS2.

JasonRuonanWang:
@chuckatkins I am pretty sure that I ran clang-format for every file that I changed this time, but I am still getting this error...

chuckatkins:
The dataman examples all segfault right now because they can't find any manager SOs.  Should probably handle that a bit more gracefully that a seg fault.

JasonRuonanWang:
@chuckatkins The examples have been fixed.

chuckatkins:
Looks good to me.

PR #57, opened 2017-04-13
Add fixed width types and type trait info to the adios namespace.

chuckatkins:


chuckatkins:
@williamfgc please review

williamfgc:
@chuckatkins Looks great. One thing, can we make the unsigned default in 
template <size_t Bytes, bool Signed>
 +struct FixedWidthInt;?   
Most of the time I am using unsigned int for raw data....signed only if required for numeric data.
Thoughts?

williamfgc:
@chuckatkins Also, what do you think about this change: don't know if we can guarantee 32 and 64 always map besides 64-bit systems

using complex32_t = std::complex<float>;
using complex64_t = std::complex<double>;

using complexSP_t = std::complex<float>;
using complexDP_t = std::complex<double>;

include/ADIOSTypes.h
```diff
@@ -37,6 +42,96 @@ enum class IOMode
...
+
+// Not sure if we're really use these ones but we'll round it out for
+// completion
+using real32_t = float;
+using real64_t = double;
+using complex32_t = std::complex<float>;
+using complex64_t = std::complex<double>;
+
+// Get a fixed width integer type from a size specification
+template <size_t Bytes, bool Signed>
```
williamfgc: Can bool Signed = false be default?   unsigned is used for most raw data, signed only if numerical data requires.

chuckatkins:
RE FixedWidthInt template args:  I'd rather keep it explicit.  It's intended to be a low-level structure that will very rarely be used.  Mostly it will get used indirectly via TypeInfo<T>::IOType.  It was mainly for the use case of mapping `long int` and `long long int` to the same underlying type.

W.R.T the complex types, we can guarantee that float and double are 32 and 64 bit via IEEE-754, and then we can guarantee the `sizeof(std::complex<T>) == sizeof(T[2])`.  I explicitly don't include support for long double here in anticipation of dropping it entirely in the rest of the code as we discussed two weeks ago since it's a non-standardized type and its size is always implementation and even compiler dependent.

williamfgc:
OK, sounds good.

PR #66, opened 2017-04-18
Use KWSys dynamic loader to abstract dlopen calls

chuckatkins:


chuckatkins:
@JasonRuonanWang Please review

JasonRuonanWang:
@chuckatkins I had a try. But it gave me a set fault when the library file is found. It works for cases when the library file is not found though. 

For example, for the first run I put libdumpman.so in place. For the second run I removed libdumpman.so.


Mac-Pro:bin jason$ ./hello_datamanWriter 
[DataManager] Streaming method dump added
Segmentation fault: 11
Mac-Pro:bin jason$ ./hello_datamanWriter 
[DataManager] Streaming method dump added
[DataManager] Unable to locate the libdump.so library.
I am hooked to the DataMan library
putshape 1
varshape 1
offset 1
Rank: 0
Variable: myfloats
Type: float
Size: 10 elements
Payload: 40 bytes
Values (first 10 or max_size): 
0 1 2 3 4 5 6 7 8 9  ...

I am hooked to the DataMan library
putshape 1
varshape 1
offset 1
Rank: 0
Variable: myfloat
Type: float
Size: 1 elements
Payload: 4 bytes
Values (first 10 or max_size): 
1.12  ...

Mac-Pro:bin jason$

JasonRuonanWang:
@chuckatkins The latest commit worked on my side. Please go ahead and merge unless you still see seg faults on your side.

source/utilities/realtime/dataman/DumpMan.cpp
```diff
@@ -0,0 +1,83 @@
...
+namespace adios
+{
+namespace realtime
+{
+
+int DumpMan::init(json p_jmsg)
+{
+    if (p_jmsg["dumping"] != NULL)
+    {
+        m_dumping = p_jmsg["dumping"].get<bool>();
```
JasonRuonanWang: @chuckatkins If you want to keep DumpMan we probably need to change this NULL to nullptr, otherwise it does not compile using clang.

chuckatkins: I don't particularly care if we keep dumpman or not, it was just something I could add to test with.

JasonRuonanWang: It would be good to keep in the sense that I can use it as an example to add something else later as a dynamic library. Plus it is a quite trivial implementation and won't cause problems.

chuckatkins:
Melding with the CacheMan branch

PR #83, opened 2017-04-20
Bpthreads

williamfgc:
BP1Writer is sufficient to write in BP1 format from any engine
Allowing threads for buffer copy of large payload sizes.


williamfgc:
@chuckatkins when I try to rebase the Merge branch 'master' into bpthreads I get a lot of commits rather than one.

chuckatkins:
It doesn't look like a rebase is needed.  No conflicts, they were resolved in the last merge commit.

PR #84, opened 2017-04-21
Added all DataMan plugins

JasonRuonanWang:


JasonRuonanWang:
@chuckatkins Please review and let me know if there is anything I need to change. Thanks.

source/dataman/CacheMan.h
```diff
@@ -7,6 +7,7 @@
  *  Created on: Apr 18, 2017
  *      Author: Jason Wang
  */
+
 #ifndef CACHEMAN_H_
```
chuckatkins: In order to avoid issues with conflicting include guards, we're starting to use ifdef guards based on the actual path, not just the filename.  Please adjust all the include guards to use `#ifdef DATAMAN_FILENAME_H_`

source/dataman/CompressMan.h
```diff
@@ -0,0 +1,24 @@
...
+#ifndef COMPRESSMAN_H_
+#define COMPRESSMAN_H_
+
+#include "DataMan.h"
+
+class CompressMan : public DataManBase
+{
+public:
+    CompressMan() = default;
+    ~CompressMan() = default;
```
chuckatkins: Should be virtual

source/dataman/DataMan.cpp
```diff
@@ -61,7 +61,10 @@ void DataMan::add_stream(json p_jmsg)
         man->init(p_jmsg);
         this->add_next(method, man);
     }
-    add_man_to_path("zfp", method);
+    if (p_jmsg["compress_method"] != nullptr)
```
chuckatkins: Rather than test for nullptr, just treat the pointer as a bool, i.e. `if(p_jmsg["compress_method"])`

source/dataman/DataManBase.cpp
```diff
@@ -152,8 +152,10 @@ int DataManBase::put_end(const void *p_data, json &p_jmsg)
     m_profiling["manager_mbs"] =
         m_profiling["total_mb"].get<double>() /
         m_profiling["total_manager_time"].get<double>();
-    if (p_jmsg["compressed_size"] != nullptr)
+    if (p_jmsg["compressed_size"].is_number())
         p_jmsg["putbytes"] = p_jmsg["compressed_size"].get<size_t>();
```
chuckatkins: Still use `{` and `}` for single line `if` blocks

source/dataman/DumpMan.cpp
```diff
@@ -0,0 +1,77 @@
...
+ *
+ *  Created on: Apr 20, 2017
+ *      Author: Jason Wang
+ */
+
+#include "DumpMan.h"
+
+int DumpMan::init(json p_jmsg)
+{
+    if (p_jmsg["dumping"] != nullptr)
```
chuckatkins: Test for ptr as boolean, not by checking for `nullptr` equality

source/dataman/DumpMan.cpp
```diff
@@ -0,0 +1,77 @@
...
+int DumpMan::put(const void *p_data, json p_jmsg)
+{
+    put_begin(p_data, p_jmsg);
+
+    if (!m_dumping)
+    {
+        return 0;
+    }
+
+    std::string doid = p_jmsg["doid"];
```
chuckatkins: What happens if `p_jmsg["foo"]` is null here?  is that possible?

source/dataman/DumpMan.cpp
```diff
@@ -0,0 +1,77 @@
...
+    }
+    size_t putbytes = product(putshape, dsize(dtype));
+
+    std::cout << p_jmsg.dump(4) << std::endl;
+    std::cout << "total MBs = " << product(putshape, dsize(dtype)) / 1000000
+              << std::endl;
+
+    const void *data_to_dump;
+
+    std::vector<char> data;
```
chuckatkins: Rather than declare and resize, just use the sized constructor, i.e. `std::vector<T> foo(N);`

source/dataman/DumpMan.cpp
```diff
@@ -0,0 +1,77 @@
...
+    {
+        data_to_dump = data.data();
+    }
+    else
+    {
+        data_to_dump = p_data;
+    }
+
+    if (dtype == "float")
+        for (size_t i = 0; i < numbers_to_print; i++)
```
chuckatkins: Add `{` and `}` to both the `if` block and the `for` loop

source/dataman/CompressMan.h
```diff
@@ -0,0 +1,24 @@
...
+ * Distributed under the OSI-approved Apache License, Version 2.0.  See
+ * accompanying file Copyright.txt for details.
+ *
+ * CompressMan.h
+ *
+ *  Created on: Apr 20, 2017
+ *      Author: Jason Wang
+ */
+
+#ifndef COMPRESSMAN_H_
```
chuckatkins: Change header guard.  See above.

source/dataman/DumpMan.h
```diff
@@ -0,0 +1,36 @@
...
+ * Distributed under the OSI-approved Apache License, Version 2.0.  See
+ * accompanying file Copyright.txt for details.
+ *
+ * DumpMan.h
+ *
+ *  Created on: Apr 20, 2017
+ *      Author: Jason Wang
+ */
+
+#ifndef DUMPMAN_H_
```
chuckatkins: Change header guard.  See above.

source/dataman/DumpMan.h
```diff
@@ -0,0 +1,36 @@
...
+#define DUMPMAN_H_
+
+#include "DataMan.h"
+
+class DumpMan : public DataManBase
+{
+public:
+    DumpMan() = default;
+    virtual ~DumpMan() = default;
```
chuckatkins: Good !

source/dataman/MdtmMan.cpp
```diff
@@ -0,0 +1,250 @@
...
+ *      Author: Jason Wang
+ */
+
+#include "MdtmMan.h"
+
+#include <fcntl.h>
+#include <sys/stat.h>
+#include <unistd.h>
+
+#include "zmq.h"
```
chuckatkins: Use `<zmq.h>` instead of `"zmq.h"` since `zmq.h` is known to be an external header.  Only use `"foo.h"` when `foo.h` is known to be part of this project.

source/dataman/MdtmMan.cpp
```diff
@@ -0,0 +1,250 @@
...
+
+MdtmMan::~MdtmMan()
+{
+    if (zmq_ipc_req)
+        zmq_close(zmq_ipc_req);
```
chuckatkins: Add `{ }`

source/dataman/MdtmMan.cpp
```diff
@@ -0,0 +1,250 @@
...
+}
+
+int MdtmMan::init(json p_jmsg)
+{
+
+    StreamMan::init(p_jmsg);
+
+    if (p_jmsg["pipe_prefix"] == nullptr)
```
chuckatkins: Use ptr as bool instead of testing for `== nullptr`

source/dataman/MdtmMan.cpp
```diff
@@ -0,0 +1,250 @@
...
+
+    // ZMQ_DataMan_MDTM
+    if (m_stream_mode == "sender")
+    {
+        zmq_ipc_req = zmq_socket(zmq_context, ZMQ_REQ);
+        zmq_connect(zmq_ipc_req, "ipc:///tmp/ADIOS_MDTM_pipe");
+        char buffer_return[10];
+        zmq_send(zmq_ipc_req, pipe_desc.dump().c_str(),
+                 pipe_desc.dump().length(), 0);
+        zmq_recv(zmq_ipc_req, buffer_return, 10, 0);
```
chuckatkins: Use `sizeof(buffer_return)` here instead of `10` in case you ever want to change buffer_return's size.

source/dataman/MdtmMan.cpp
```diff
@@ -0,0 +1,250 @@
...
+    put_begin(p_data, p_jmsg);
+
+    std::vector<size_t> putshape =
+        p_jmsg["putshape"].get<std::vector<size_t>>();
+    std::vector<size_t> varshape =
+        p_jmsg["varshape"].get<std::vector<size_t>>();
+    std::string dtype = p_jmsg["dtype"];
+
+    int priority = 100;
+    if (p_jmsg["priority"] != nullptr)
```
chuckatkins: Use ptr as bool instead of testing for `== nullptr`

source/dataman/DumpMan.cpp
```diff
@@ -0,0 +1,77 @@
...
+            std::cout << ((float *)data_to_dump)[i] << " ";
```
chuckatkins: Prefer C++ cast and keep const modifier, i.e. `static_cast<const float*>(data_to_dump)`

source/dataman/DumpMan.cpp
```diff
@@ -0,0 +1,77 @@
...
+    if (dtype == "double")
+        for (size_t i = 0; i < numbers_to_print; i++)
+            std::cout << ((double *)data_to_dump)[i] << " ";
```
chuckatkins: same as above for both brackets and casting

source/dataman/MdtmMan.cpp
```diff
@@ -0,0 +1,250 @@
...
+        zmq_connect(zmq_ipc_req, "ipc:///tmp/ADIOS_MDTM_pipe");
+        char buffer_return[10];
+        zmq_send(zmq_ipc_req, pipe_desc.dump().c_str(),
+                 pipe_desc.dump().length(), 0);
+        zmq_recv(zmq_ipc_req, buffer_return, 10, 0);
+    }
+
+    // Pipes
+    mkdir(pipe_desc["pipe_prefix"].get<std::string>().c_str(), 0755);
+    for (int i = 0; i < pipe_desc["pipe_names"].size(); i++)
```
chuckatkins: This is well suited to convert to a range-for

source/dataman/MdtmMan.cpp
```diff
@@ -0,0 +1,250 @@
...
+}
+
+int MdtmMan::get(void *p_data, json &p_jmsg) { return 0; }
+
+void MdtmMan::on_recv(json jmsg)
+{
+
+    // push new request
+    jqueue.push(jmsg);
+    bqueue.push(NULL);
```
chuckatkins: `NULL` -> `nullptr` ?

source/dataman/MdtmMan.cpp
```diff
@@ -0,0 +1,250 @@
...
+    }
+
+    // for put
+    for (int outloop = 0; outloop < jqueue.size() * 2; outloop++)
+    {
+        if (jqueue.front()["operation"] == "put")
+        {
+            // allocate buffer
+            size_t putbytes = jqueue.front()["putbytes"].get<size_t>();
+            if (bqueue.front() == NULL)
```
chuckatkins: Use ptr as bool

source/dataman/MdtmMan.cpp
```diff
@@ -0,0 +1,250 @@
...
+                bqueue.front() = malloc(putbytes);
+
+            // determine the pipe for the head request
+            json msg = jqueue.front();
+            if (msg == nullptr)
```
chuckatkins: Use ptr as bool.  Also add `{}`

source/dataman/MdtmMan.cpp
```diff
@@ -0,0 +1,250 @@
...
+                }
+            }
+
+            // read the head request
+            int error_times = 0;
+            int s = iqueue.front();
+            putbytes = msg["putbytes"].get<int>();
+            while (s < putbytes)
+            {
+                int ret = read(pipes[pipeindex], ((char *)bqueue.front()) + s,
```
chuckatkins: Use C++ static_cast

source/dataman/DataMan.cpp
JasonRuonanWang: This does not work with the JSON library. I have tried that.

chuckatkins: Bummer!  That's okay.  Skip that one then

JasonRuonanWang: I have tried it again and I saw this:

libc++abi.dylib: terminating with uncaught exception of type std::domain_error: type must be boolean, but is null
Abort trap: 6

But I think I can replace them with if (msg["compress_method"].is_string())

chuckatkins: No, that's fine. Better to use nullptr

JasonRuonanWang: I think .is_string() is the way that the library encourage people to use. It returns false if that object is nullptr, or that object is not a string. In either case it is safe to use.

JasonRuonanWang:
@chuckatkins I have changed everything as requested. Please review.

JasonRuonanWang:
I found a few more places where I should have added brackets.

JasonRuonanWang:
@chuckatkins I fixed a few other problems which might cause errors at compile time. I wasn't able to check whether it compiles on the machine I was using in the afternoon because of the pthread thing not merged in. Sorry about that. This time it should compile and work.

chuckatkins:
@JasonRuonanWang I've rebased on master to eliminate all of the merges into this topic branch.  Please look at the dataman-rebased branch on my fork.  If it looks good to you then I'll force-push it onto yours to update the pr, and then ill merge it.

JasonRuonanWang:
@chuckatkins Yes, it looks all good to me. Please go ahead.

PR #87, opened 2017-04-23
MdtmMan worked with ZfpMan for automatic compression and decompression

JasonRuonanWang:


source/dataman/ZfpMan.h
```diff
@@ -24,7 +24,7 @@ class ZfpMan : public CompressMan
     virtual void flush();
     void *compress(void *p_data, json &p_jmsg);
```
chuckatkins: compress and decompress should have comments that document the memory they return needs to be `free`d by the caller else it will leak.

source/adios2/engine/dataman/DataManWriter.tcc
```diff
@@ -30,16 +30,17 @@ void DataManWriter::WriteVariableCommon(Variable<T> &variable, const T *values)
 
     // This part will go away, this is just to monitor variables per rank
 
+    if (variable.m_GlobalDimensions.empty())
+        variable.m_GlobalDimensions = variable.m_LocalDimensions;
```
chuckatkins: Add `{ }`

PR #94, opened 2017-04-26
Fixed a bug in CacheMan for cleaning variables after a time step is finished

JasonRuonanWang:


source/dataman/CacheMan.cpp
```diff
@@ -44,25 +44,20 @@ int CacheItem::put(const void *a_data, json a_jmsg)
...
-    }
-
     for (size_t i = 0; i < putsize; i += chunksize)
     {
         std::vector<size_t> p = one2multi(putshape, i);
         p = apply_offset(p, offset);
         size_t ig = multi2one(varshape, p);
-        std::copy((char *)a_data + i * dsize,
-                  (char *)a_data + i * dsize + chunksize * dsize,
+        std::copy(const_cast<char *>(static_cast<const char *>(a_data)) +
```
chuckatkins: The input iterators here should not be un-const-'d.  The input components to almost any STL algorithm should be const.  The general idea is to avoid using const_cast unless you absolutely have to as it's usage usually means there's something borked in the API usage (not always of course, but usually).  An example might be if you're using a library what doesn't have any const correctness, like ZFP, where even the input data is non-const.

In this case, however, the input pointers can stay const.

source/dataman/CacheMan.cpp
```diff
@@ -106,75 +101,73 @@ void CacheMan::push()
...
-void CacheMan::clean(std::string a_mode)
-{
-    for (auto i : m_cache)
-    {
-        for (auto j : i.second)
-        {
-            j.second.clean(a_mode);
-        }
-    }
+    m_cache.push(std::vector<char>(varbytes));
```
chuckatkins: Use `emplace(varbytes)` instead of `push(std::vector<char>(varbytes))` here to create the buffer in-place on the queue instead of copying it.  You could also explicitly move the value with `push(std::move(std::vector<char>(varbytes)))` but really jsut calling `emplace` is much cleaner.

source/dataman/DataMan.cpp
```diff
@@ -58,14 +59,14 @@ void DataMan::add_stream(json a_jmsg)
 
     if (m_tolerance.size() < m_num_channels)
     {
-        for (int i = 0; i < m_num_channels; i++)
+        for (int i = 0; i < m_num_channels; ++i)
```
chuckatkins: You can replace this loop entirely with `m_tolerance.insert(m_tolerance.end(), m_num_channels, 0);` to insert `m_num_channels` copies of `0`

source/dataman/CacheMan.h
```diff
@@ -52,7 +52,7 @@ class CacheItem
                             const std::vector<size_t> &p)
     {
         size_t index = 0;
-        for (int i = 1; i < v.size(); i++)
+        for (int i = 1; i < v.size(); ++i)
```
chuckatkins: Good use of prefix increment!

source/dataman/CacheMan.h
```diff
@@ -41,7 +41,7 @@ class CacheItem
                                             const std::vector<size_t> &o)
     {
         std::vector<size_t> g;
-        for (int i = 0; i < p.size(); i++)
+        for (int i = 0; i < p.size(); ++i)
```
chuckatkins: Can replace loop with std::transform:
```
std::vector<size_t> g(p.size());
std::transform(p.begin(), p.end(), o.begin(), g.begin(),
               [](size_t a, size_t b) { return a + b; });
```

JasonRuonanWang: @chuckatkins This certainly looks more clean. I am just curious how the efficiency compares between a for loop and std::transform + lambda function. I know this case is trivial, but which should I use for large functions?

chuckatkins: Typically you want to think of the functors you pass to stl algorithms as small, compact, kernels.  It's not always the case but it's a good rule of thumb.  If the body of your loop is only a few lines then it's probably well suited to a functor in an stl algorithm.  If the iterations are independent of one another then it can also be beneficial to factor out to a functor applied by an stl algorithm.

However, if your loop body is large and complex with a lot of branching then it's probably best to leave in explicit loop form.

PR #102, opened 2017-05-05
Variables3

pnorbert:


pnorbert:
@williamfgc @guj @JasonRuonanWang Please check if merging this branch into your branch does not mess up your code and examples. The biggest changes are to DefineVariable() and SheduleRead()/Read(). The order of arguments in DefineVariable() changed: global dimensions (now called Shape), offsets (now called Start) and local dimensions (now called Count).

source/adios2/core/adiosFunctions.cpp
```diff
@@ -663,6 +663,27 @@ void ConvertUint64VectorToSizetVector(const std::vector<std::uint64_t> &in,
     }
 }
 
+void Uint64ArrayToSizetVector(const std::size_t nElements, const uint64_t *in,
+                              std::vector<std::size_t> &out)
+{
+    out.resize(nElements);
```
chuckatkins: Good spot for std::transform`:
```
std::transform(in, in + nElements, out.begin(), [](uint64_t value) { return static_cast<size_t>(value); })
```

examples/heatTransfer/read/heatRead_adios2.cpp
```diff
@@ -60,12 +60,14 @@ int main(int argc, char *argv[])
...
     }
 
     auto bpReader = ad.Open(inputfile, "r", mpiReaderComm, bpReaderSettings);
 
     if (bpReader == nullptr)
-        throw std::ios_base::failure("ERROR: failed to open ADIOS bpReader\n");
+        throw std::ios_base::failure("ERROR: failed to open " +
+                                     std::string(inputfile) + "\n");
 
     unsigned int gndx;
```
williamfgc: prefer 
unsigned int gndx = vgnx->m_Data[0]; as in line 79
See: https://github.com/ornladios/ADIOS2/wiki/Variables-Scope,-Functions,-and-Namespaces
2. Variable declaration

source/adios2/core/VariableBase.h
```diff
@@ -148,11 +148,14 @@ class VariableBase
...
-    int GetNSteps() { return m_nsteps; }
+    unsigned int GetNSteps() { return m_nsteps; }
+
+    ///< Should only be called by read engines
+    void SetNSteps(unsigned int steps) { m_nsteps = steps; }
 
 private:
     /* Values filled by InquireVariable() */
-    int m_nsteps =
+    unsigned int m_nsteps =
```
williamfgc: unsigned int m_Steps

source/adios2/core/VariableBase.h
```diff
@@ -145,18 +145,40 @@ class VariableBase
...
+        m_ReadNSteps = nSteps;
+    }
+
     /** Return the number of steps available for the variable
      *  @return Number of steps
      */
-    unsigned int GetNSteps() { return m_nsteps; }
+    unsigned int GetNSteps() { return m_NStepsAvailable; }
 
     ///< Should only be called by read engines
```
williamfgc: doxygen ///<  documents text before  the message. 
Use /**   documentationa */ for the line below

pnorbert:
Don't merge. Step-by-step reading does not work in HeatTransfer

source/adios2/ADIOSTypes.h
```diff
@@ -137,6 +114,35 @@ struct TypeInfo<T, typename std::enable_if<std::is_same<
     using ValueType = typename T::value_type;
 };
 
+const size_t UnknownDim = 0;
+const size_t JoinedDim = SIZE_MAX - 1;
```
pnorbert: C++ max number represented by std::size_t

williamfgc: it should be (size_t)-1

source/adios2/core/Engine.h
```diff
@@ -242,9 +250,10 @@ class Engine
      * must use Read(variable) instead intentionally
      */
     template <class T>
-    void Read(Variable<T> &variable, const T *values)
+    void Read(Variable<T> &variable, T *values)
     {
-        Read(variable, values);
+        ScheduleRead(variable, values);
+        PerformReads(PerformReadMode::BLOCKINGREAD);
```
pnorbert: This is not clear yet. Yes we separate them out, so ScheduleRead() and PerformRead() are official. Question if we will have Read() as schedule+perform(in blocking mode) or rather we also eliminate Write() and have ScheduleWrite() instead. We are sill undecided on this.

source/adios2/core/Engine.h
```diff
@@ -353,19 +365,68 @@ class Engine
...
      */
-    void ScheduleRead(const std::string variableName)
-    {
-        ScheduleRead(variableName, nullptr);
-    }
+    virtual void ScheduleRead(const std::string &variableName);
 
     /**
      * Unallocated unspecified version, ADIOS will receive any variable and will
      * allocate space for incoming data
```
pnorbert: I am not sure what you mean by that. The highlighted function is the one that allows the user say hey adios allocate the space for me to read this variable in.

examples/heatTransfer/read/heatRead_adios2.cpp
pnorbert: This is just because the Read() does not work, so instead it is using InquireVariable there and the read is commented out. It should be the latter when it works.

source/adios2/ADIOS.h
```diff
@@ -97,13 +97,34 @@ class ADIOS
...
-    Variable<T> &DefineVariable(const std::string &name,
-                                const Dims localDimensions = Dims{1},
-                                const Dims globalDimensions = Dims{},
-                                const Dims offsets = Dims{});
+    Variable<T> &DefineVariable(const std::string &name, const Dims shape = {},
+                                const Dims start = {}, const Dims count = {},
+                                bool constantShape = false);
+
+    template <class T>
+    Variable<T> &DefineLocalValue(const std::string &name)
```
williamfgc: Not a fan, the less public API functions to remember the better...just like MPI or OpenMP core functionality. How is the returned Variable<T>& different from DefineVariable<T>(name) and DefineLocalValue<T>(name)?   I understand the separation for DefineVariableCompound as it requires more member functions to set the offsets, but not for DefineVariable<T>.   As a user I don't like remembering more than a few things, otherwise they think it's "complicated".

source/adios2/ADIOS.h
```diff
@@ -97,13 +97,34 @@ class ADIOS
...
+
+    template <class T>
+    Variable<T> &DefineLocalArray(const std::string &name, bool isJoined,
+                                  Dims shape = {})
+    {
+        return DefineVariable<T>(name, shape, {}, {}, false);
+    }
+
+    template <class T>
+    Variable<T> &DefineArray(const std::string &name, Dims shape = {},
```
williamfgc: Array + Dims is confusing, I will keep it as Variable since we only deal with 1D arrays.  As a first-time user I'd think I can pass a double**  with a shape = {Nx, Ny}; since we "DefineArray"

source/adios2/ADIOS.cpp
```diff
@@ -394,7 +405,8 @@ void ADIOS::CheckMethod(std::map<std::string, Method>::const_iterator itMethod,
 // Explicitly instantiate the necessary template implementations
 #define define_template_instantiation(T)                                       \
     template Variable<T> &ADIOS::DefineVariable<T>(                            \
-        const std::string &, const Dims, const Dims, const Dims);              \
+        const std::string &, const Dims, const Dims, const Dims,               \
+        bool constantShape);                                                   \
```
williamfgc: const bool constantShape

examples/hello/timeBP/timeBPWriter.cpp
```diff
@@ -48,13 +48,13 @@ int main(int argc, char *argv[])
...
-            adios.DefineVariable<double>("myDoubles", {Nx});
+            adios.DefineLocalArray<double>("myDoubles", false, {Nx});
 
         adios::Variable<float> &ioMyMatrix =
-            adios.DefineVariable<float>("myMatrix", {rows, columns});
+            adios.DefineLocalArray<float>("myMatrix", true, {rows, columns});
 
         adios::Variable<float> &ioMyMatrix2 =
-            adios.DefineVariable<float>("myMatrix2", {rows, columns});
+            adios.DefineLocalArray<float>("myMatrix2", false, {rows, columns});
```
williamfgc: See my comment below about localArray......from looking at it I think I am expecting-->  float array[rows][columns];

source/adios2/ADIOSTypes.h
chuckatkins: Or even better, `std::numeric_limits<size_t>::max()`

williamfgc: the problem with the std is that it doesn't provide compile-time limits. Just like vector.max_size(), numeric_limits only provides runtime solutions. So far,  (size_t )-1 does the job at compile time and it's portable.

source/adios2/core/VariableBase.h
```diff
@@ -33,80 +34,138 @@ class VariableBase
 {
```
williamfgc: I will work on this, the same for Variable template separation

source/adios2/ADIOSTypes.h
chuckatkins: That was generally seen as an oversight and defect in c++98. c+11 made these all constexpr so they are evaluated at compile time.

williamfgc: I remember numeric_limits wasn't well supported by C++11 compilers as constexpr a few years ago, so we end up using (size_t)-1 . If that has changed I think we should use numeric_limits<size_t> instead.

source/adios2/core/adiosFunctions.cpp
williamfgc: Good one!

source/adios2/core/Engine.h
chuckatkins: Please disregard

source/adios2/ADIOSTypes.h
chuckatkins: I just checked with GCC4.8, which is as far back as we will support, and it's constexpr

williamfgc: Great! Thanks, Chuck! std::numeric_limits<size_t> it is.

examples/hello/timeBP/timeBPWriter.cpp
pnorbert: changed

source/adios2/ADIOS.cpp
pnorbert: fixed

source/adios2/ADIOSTypes.h
pnorbert: fixed

PR #103, opened 2017-05-08
Move jsonhpp to thirdparty

chuckatkins:


chuckatkins:
@JasonRuonanWang Just a heads up that json.hpp is moving to /thirdparty

PR #105, opened 2017-05-09
Fixed a compiling problem for Mac which is caused by conflicts between size_t and uint64_t

JasonRuonanWang:


JasonRuonanWang:
@pnorbert @williamfgc Could someone review and merge? I don't have write access...

PR #110, opened 2017-05-11
added direct callback mode to dataman receiver to bypass cache

JasonRuonanWang:


chuckatkins:
There's no need for the extra master-merge if you drop the example change.  I'll push the updated branch and merge it.

PR #124, opened 2017-06-02
Refactor

williamfgc:
All, please read the commit messages and review your implementation.
Thanks,
William


source/adios2/ADIOSTypes.h
```diff
@@ -11,36 +11,169 @@
...
+};
+
+enum class OpenMode
+{
+    Undefined,
+    Write,
+    Read,
+    Append,
+    ReadWrite,
+    w,
```
chuckatkins: I'd get rid of the duplicate single letter versions.  The entire API is designed around words so lower-case single letters don't really fit, especially when there's already the word versions of the enums.

source/adios2/ADIOSTypes.h
```diff
@@ -11,36 +11,169 @@
...
+/** Used to set IO class */
+enum class IOMode
+{
+    Independent, ///< all I/O operations are independent per rank
+    Collective   ///< expect collective I/O operations
+};
+
+enum class OpenMode
+{
+    Undefined,
```
chuckatkins: It might be a good idea to give these explicit values so the bit-wise operations make sense:
```
Undefined = 0b00000000,
Write     = 0b00000001,
Read      = 0b00000010,
Append    = 0b00000101,
ReadWrite = 0b00000011
```
This allows for things like:
```
// Write test, valid for Write, ReadWrite, and Append
if (val & OpenMode::Write)
```
and
```
// Read test, valid for Read and ReadWrite
if (val & OpenMode::Read)
```

source/adios2/ADIOSTypes.h
```diff
@@ -11,36 +11,169 @@
...
+    Read,
+    Append,
+    ReadWrite,
+    w,
+    r,
+    a,
+    rw
+};
+
+typedef enum {
```
chuckatkins: Use enum classes and CameCase.  The `typedef enum { } Foo;` syntax is for C.

source/adios2/ADIOSTypes.h
```diff
@@ -11,36 +11,169 @@
...
+} ReadMultiplexPattern;
+
+typedef enum {
+    NOWAITFORSTREAM = 0,
+    WAITFORSTREAM = 1
+} StreamOpenMode; // default: wait for stream
+
+enum class TransformType
+{
+    bzip2,
```
chuckatkins: CamelCase

source/adios2/core/ADIOS.cpp
```diff
@@ -0,0 +1,88 @@
...
+ADIOS::ADIOS(const std::string configFile, const bool debugMode)
+: m_ConfigFile(configFile), m_DebugMode(debugMode)
+{
+}
+
+ADIOS::ADIOS(const std::string configFile, MPI_Comm mpiComm,
+             const bool debugMode)
+: m_MPIComm(mpiComm), m_ConfigFile(configFile), m_DebugMode(debugMode)
+{
+    if (m_DebugMode == true)
```
chuckatkins: Drop the `== true` since `m_DebugMode` is already a bool

source/adios2/core/ADIOS.cpp
```diff
@@ -0,0 +1,88 @@
...
+    }
+}
+
+IO &ADIOS::DeclareIO(const std::string ioName)
+{
+    auto itIO = m_IOs.find(ioName);
+
+    if (itIO != m_IOs.end()) // exists
+    {
+        if (m_DebugMode == true)
```
chuckatkins: Drop the `== true` since `m_DebugMode` is already a bool

source/adios2/core/ADIOS.cpp
```diff
@@ -0,0 +1,88 @@
...
+        {
+            if (itIO->second.InConfigFile() == false)
```
chuckatkins: `if (foo == false)` -> `if (!foo)`

source/adios2/core/ADIOS.cpp
```diff
@@ -0,0 +1,88 @@
...
+    {
+        CheckMPI();
+    }
+    // XML to be implemented later
+    // InitXML( m_XMLConfigFile, m_MPIComm, m_DebugMode, m_HostLanguage,
+    // m_Transforms, m_Groups );
+}
+
+ADIOS::ADIOS(MPI_Comm mpiComm, const bool debugMode)
+: m_MPIComm(mpiComm), m_DebugMode(debugMode)
```
chuckatkins: This should work as a delegating constructor:
```
ADIOS::ADIOS(MPI_Comm mpiComm, const bool debugMode)
: ADIOS("", mpiComm, debugMode)
{ }
```

source/adios2/core/ADIOS.cpp
```diff
@@ -0,0 +1,88 @@
...
+                    " previously declared, name must be unique "
+                    " , in call to DeclareIO\n");
+            }
+        }
+        return itIO->second;
+    }
+
+    // doesn't exist, then create
+    m_IOs.emplace(ioName, IO(ioName, m_MPIComm, false, m_DebugMode));
+    return m_IOs.at(ioName);
```
chuckatkins: Rather than look this up again, could you just use the iterator returned from `emplace`?

source/adios2/core/Engine.inl
```diff
@@ -0,0 +1,129 @@
...
+ *      Author: William F Godoy godoywf@ornl.gov
+ */
+
+#ifndef ADIOS2_CORE_ENGINE_INL_
+#define ADIOS2_CORE_ENGINE_INL_
+#ifndef ADIOS2_CORE_ENGINE_H_
+#error "Inline file should only be included from it's header, never on it's own"
+#endif
+
+#include "Engine.h"
```
chuckatkins: Remove include.  The previous `ifdef` check guarantees this is always being included by `Engine.h`

source/adios2/core/Engine.cpp
```diff
@@ -5,525 +5,209 @@
...
-void Engine::Advance(AdvanceMode /*mode*/, float /*timeout_sec*/) {}
+void Engine::Advance(const float /*timeout_sec*/) {}
+void Engine::Advance(const AdvanceMode /*mode*/, const float /*timeout_sec*/) {}
 void Engine::AdvanceAsync(
     AdvanceMode /*mode*/,
     std::function<void(std::shared_ptr<adios::Engine>)> /*callback*/)
 {
 }
 
+AdvanceStatus Engine::GetAdvanceStatus() { return m_AdvanceStatus; }
```
chuckatkins: Should be a const method

source/adios2/core/IO.cpp
```diff
@@ -0,0 +1,279 @@
...
+       const bool debugMode)
+: m_Name(name), m_MPIComm(mpiComm), m_InConfigFile(inConfigFile),
+  m_DebugMode(debugMode)
+{
+}
+
+void IO::SetEngine(const std::string engineType) { m_EngineType = engineType; }
+void IO::SetIOMode(const IOMode ioMode) { m_IOMode = ioMode; };
+
+VariableCompound &IO::GetVariableCompound(const std::string &name)
```
chuckatkins: Should we make all of these `Get` methods const?

source/adios2/core/IO.cpp
```diff
@@ -0,0 +1,279 @@
...
+
+std::shared_ptr<Engine> IO::Open(const std::string &name,
+                                 const OpenMode openMode)
+{
+    return Open(name, openMode, m_MPIComm);
+}
+
+// PRIVATE Functions
+unsigned int
+IO::AddTransportParameters(const std::string type,
```
chuckatkins: It would also be good to have a signature that could accept an std::map.  That would allow you to bypass the vector -> map parsing and call it with `{"key1": "val1", "key2" : "val2"}` syntax.

source/adios2/core/IO.cpp
```diff
@@ -0,0 +1,279 @@
...
+                                  "type must be the first argument\n");
+        }
+    }
+
+    mapParameters["transport"] = type;
+    m_TransportsParameters.push_back(std::move(mapParameters));
+    return static_cast<unsigned int>(m_TransportsParameters.size() - 1);
+}
+
+unsigned int IO::GetVariableIndex(const std::string &name)
```
chuckatkins: Should this be a size_t ?

source/adios2/core/IO.inl
```diff
@@ -0,0 +1,67 @@
...
+ *      Author: William F Godoy godoywf@ornl.gov
+ */
+
+#ifndef ADIOS2_CORE_IO_INL_
+#define ADIOS2_CORE_IO_INL_
+#ifndef ADIOS2_CORE_IO_H_
+#error "Inline file should only be included from it's header, never on it's own"
+#endif
+
+#include "IO.h"
```
chuckatkins: Remove circular include

bindings/python/include/ADIOSPy.h
```diff
@@ -23,7 +23,7 @@
 #include "adios2/pybind11/pybind11.h"
 #endif
 
-#include "adios2/ADIOS.h"
+#include "../../../source/adios2/core/ADIOS.h"
```
williamfgc: @chuckatkins Thanks! I haven't worked on the Python bindings. I am actually working on it now.

source/adios2/core/ADIOS.cpp
williamfgc: I always debate between making a bool check explicit or implicit

source/adios2/core/ADIOS.cpp
williamfgc: Got it!

source/adios2/core/ADIOS.cpp
williamfgc: Should we make all bool checks implicit?  I like the distinction between an actual bool and bool operator (from shared_ptr or fstream)....I'm fine with either way

source/adios2/core/ADIOS.cpp
williamfgc: Good call!

source/adios2/core/Engine.inl
williamfgc: Don't I need the definitions from Engine.h?  It's more of Engine.inl understanding Engine.h members

source/adios2/core/Engine.cpp
williamfgc: Good one, the entire Read API needs to be reworked

source/adios2/core/ADIOS.cpp
chuckatkins: Typically you want to use a bool as-is and avoid the extra test.  It also works better when the coding guidelines enforce "boolean variable names should be verbs" rules, i.e. `m_DebugMode` becomes `m_IsDebugMode` or something similar; the boolean methods already do this and it makes reading the if statement very natural: "if obj is ready then do thing" becomes `if (obj->IsReady()) { DoThing(); }`.  It may be something to consider implementing in the near future, I wouldn't push for it at this point though.  Maybe wait till things settle down a bit.

source/adios2/core/Engine.inl
chuckatkins: Yes, but nothing can actually include `Engine.inl` because of the enforcing `#ifdef` check so it can never show up stand-alone.  The only way it gets included is by `Engine.h` including it so the definitions are guaranteed to exist.

source/adios2/core/ADIOS.cpp
chuckatkins: I think all bools should be used as bools and drop the unnecessary equality test.  I believe it helps with readability, especially if the variables are names appropriately.

examples/experimental/multistep/reader_allsteps.cpp
```diff
@@ -183,24 +184,22 @@ int main(int argc, char *argv[])
...
-        vNparts->SetStepSelection(0, vNparts->GetNSteps());
+            Make2DArray<unsigned int>(vNparts->m_AvailableSteps, Nwriters);
+        vNparts->SetStepSelection(0, vNparts->m_AvailableSteps);
         bpReader->Read<unsigned int>(*vNparts, Nparts[0]);
         if (rank == 0)
-            Print2DArray(Nparts, vNparts->GetNSteps(), Nwriters, "Nparts");
+            Print2DArray(Nparts, vNparts->m_AvailableSteps, Nwriters, "Nparts");
         Delete2DArray(Nparts);
-#ifdef ADIOS2_HAVE_MPI
         MPI_Barrier(MPI_COMM_WORLD);
```
williamfgc: @pnorbert is this part of mpidummy ?

source/adios2/core/Engine.inl
williamfgc: Got it!

examples/experimental/multistep/reader_allsteps.cpp
chuckatkins: It is but mpidummy.h is private and not part of the public interface.

pnorbert:
We are not using mpidummy in user codes. I wanted a single source for both
MPI and non-MPI build, hence I use the #ifdef.
Only MPI_COMM_WORLD is available from including adios2.h

source/adios2/ADIOSTypes.h
```diff
@@ -11,36 +11,169 @@
 #ifndef ADIOS2_ADIOSTYPES_H_
 #define ADIOS2_ADIOSTYPES_H_
 
+/// \cond EXCLUDE_FROM_DOXYGEN
```
williamfgc: <c*> is still C++ only . I'd separate it if was the actual C header stddef.h and stdint.h

source/adios2/ADIOSTypes.h
williamfgc: Will do

source/adios2/ADIOSTypes.h
williamfgc: Changing all enum to enum class

source/adios2/ADIOSTypes.h
chuckatkins: It is but the idea is to separate the C headers from C++ headers, so for the C headers, also separate actual C headers from C++ wrapped C headers.  See https://github.com/ornladios/ADIOS2/wiki/File-Header-Structure-and-Includes

source/adios2/ADIOSTypes.h
williamfgc: Good catch, I think CamelCase should be also enforced for string parameters  (e.g. AddTransport("File", ...);   not AddTransport("file"....)  ) for consistency. Thoughts?

chuckatkins: Certainly.  I'm looking at implementing some clang-tidy checks that can enforce variable naming for members, data members, classes, enums, local variables, etc., but the config file is really confusing for that particular rule.  Until then, we'll catch it via code review.

source/adios2/ADIOSTypes.h
williamfgc: I'm missing something. Usually OpenMode goes to a m_OpenMode member (of Engine and Transport) of enum class OpenMode type and comparison would be `if( m_OpenMode == OpenMode::Write)`

chuckatkins: Scratch this, don't worry about it.  The enum classes make this messier and more trouble than it's worth.

source/adios2/core/IO.cpp
williamfgc: Changing them

source/adios2/core/IO.cpp
williamfgc: Agree. I want to add this once we support more config file formats (e.g. json). For now it matches the XML pair format so the compile time API and runtime config file are consistent. The user won't have to learn more than one syntax.

source/adios2/core/IO.cpp
williamfgc: I use unsigned int for key values, it's not a size or location per-se.

source/adios2/core/IO.inl
williamfgc: Doing it in all inl files

bindings/python/include/ADIOSPy.h
williamfgc: On a 2nd thought this will come after the current changes

source/adios2/ADIOSTypes.h
williamfgc: My bad, will change this. Thanks!

source/adios2/core/ADIOS.cpp
williamfgc: Changed all bool tests to implicit

source/adios2/core/IO.cpp
williamfgc: Those returning an internal reference can't be const. I will change the others, though.

bindings/python/include/MethodPy.h
```diff
@@ -20,7 +20,7 @@
 #include "adios2/pybind11/pybind11.h"
 #endif
 
-#include "adios2/core/Method.h"
+#include "../../../source/adios2/core/IO.h"
```
williamfgc: I'd make the Python bindings a separate branch after resolving this PR. Today we are not compiling it anyways.

PR #126, opened 2017-06-09
Python Bindings and their required internal code changes

williamfgc:
Enable adios2 bindings for Python 2.7.x, 3.5 with PyBind11 
Removed boost.python dependencies
Replaced IO class variadic template arguments for vector<string> ("key=value") and map<string,string> {"key", "value"} versions.

Changes to make core headers completely "public" as needed by external bindings:
Variable.inl to Variable.tcc
VariableCompound.inl now independent (removed adios2/helper/adiosFunctions.h dependency) 

To do:
Add tests with other Engines


williamfgc:
@chuckatkins I added the nompi portion for the Python bindings. Need changes in cmake. Thanks!

chuckatkins:
Rather than a seperate nompi file, I think just using an ifdef would be more appropriate.  I'll make the pr shortly.

williamfgc:
That's fine.

PR #134, opened 2017-06-14
CMake: Allow auto-detection of default options

chuckatkins:


williamfgc:
@chuckatkins Let me know when it's ready so we can merge into master

PR #137, opened 2017-06-19
Enable config xml file

williamfgc:
Tested C++ examples and Python bindings examples in serial and MPI compilations.


chuckatkins:
I'm really not comfortable having our own home-grown XML parser.  I'd much rather see us use something like libxml2 which is a small, lightweight, validating XML parser and widely available on most systems.

Also, this needs unit tests before it's ready to merge.

williamfgc:
We are not parsing XML.....the only thing the XML library would do for us is get <tag key1=val1 key2=val2>, and that's it, it's extremely trivial with . We are writing our own binary parser. I think we can handle std::string and don't need a C library for this.  I am not against using a external parser as long as it doesn't "crash" compilation.  Of course, I added tests. We can test more, but the underlying functionality is provided by the C++ string library.

chuckatkins:
> the only thing the XML library would do for us is get "< tag key1=val1 key2=val2 >", and that's it, it's extremely trivial with C++ map and std::string
...
We are writing our own binary parser (bp)

Parsing XML is much more complicated than parsing the BP format.  BP is very narrowly scoped while XML has an extensive set of standards behind it allowing for much more complicated files than "< tag key1=val1 key2=val2 >"  If that's the only case we handle then we can't really call it "XML", but jsut something that looks like XML.

> Of course, I added tests. We can test more, but the underlying functionality is provided by the C++ string library.

I mean testing the string operations.  There should be several unit tests (i.e. in the testing directory, and run by make test) that read known XML config files and then explicitly verify that the correct adios objects were created with the correct settings.

eisenhauer:
I'm going to agree with Chuck here...  Both on the use of an actual XML parser and the testing scenarios.

williamfgc:
In fact, you're right we read something that looks like XML, it's a very simple and trivial schema with only strings.  I disagree with the BP format claim, debugging a binary format is always harder than a human readable string format. XML after all it's a string format at its core.  Yes, I will adding more tests. The string operations are mostly C++ std::string API, the added logic is to distinguish actual tag information from values in quotes.

williamfgc:
Can someone provide an XML library and guarantee it's 100% portable? If so, please provide an initial implementation. I want someone else owning this part. I am not against it, I just don't think I want to spend time debugging compilation issues. So please, any volunteer to own this part is more than welcome to take over.

eisenhauer:
The complexity in parsing XML is the complexity of the standard, not the binary vs. string part.  I wouldn't have any portability worries about libxml2...  I would worry more about making a homebrew thing be 100% XML...

chuckatkins:
>  debugging a binary format is always harder than a human readable string format. XML after all it's a string format at its core

I may have misspoke as I was not trying to imply otherwise.  Debugging binary I/O is certainly much more opaque and difficult that ascii I/O.  I was instead trying to point out that the range of what constitutes valid BP input is much more constrained than what constitutes valid XML input, hence the desire to use an existing XML parser that already effectively addresses the wide array of valid input.  

> Can someone provide an XML library and guarantee it's 100% portable?

CMake and VTK have both been using Expat and LibXML2  for decades on everything from Linux and Windows to QNX and VxWorks.  Either of them are good options supported on more platforms than ADIOS will need to be.  They are also both MIT licensed so fully compatible with us.

> I am not against it, I just don't think I want to spend time debugging compilation issues.

I'll add libxml2 into thirdparty and adjust the CMake accordingly so it can be readily consumed by a config parser implementation.

williamfgc:
@eisenhauer @chuckatkins Thanks. Like I said, it's not a big deal, if you want to take over please go ahead. I need to focus on a long list of items in the development, and I thought having a build-in minimal parser based on C++11 will take away all the compilation issues and slowdown shown in adios1 with mxml. I have been playing with this for months. We will go through the same conversation when we want to add support for a JSON config file. If we look at the details, which unfortunately we haven't talked about, most of the parsing is in the parameters inside the tags anyways and must be homebrew.

williamfgc:
I am going to merge this if there is no further discussion on the code details, and please do a PR replacing the parser when it's ready as we can always replace it. After all, this is an alpha version. I need, at least basic functionality to test on Titan and Summit-dev from master with runtime options. Also, we should reserve github discussions for code review and pointing out details in the code. I take responsibility for this implementation.

PR #165, opened 2017-06-21
WIP: Added working python types test

williamfgc:
Added .pyc .bp and .bp.dir to .gitignore
Verified types with adios1 bpls

To do: adios2NPTypes.py and adios2.so must be in PYTHONPATH


chuckatkins:
@williamfgc There's now a python_add_test CMake function that handles the necessary pieces for you.

williamfgc:
@chuckatkins Great! Thanks! I will add more cases as I continue.

PR #168, opened 2017-06-22
Makes missing xml file permissive

williamfgc:
If file doesn't exist, continue with run
Fixed bug if space before and after "=" in: \<tag attribute1= "value1" \>"

To do: need to discuss policy for config files, xml config files still experimental

### DIFF
--- a/source/adios2/helper/adiosString.cpp
+++ b/source/adios2/helper/adiosString.cpp
@@ -20,15 +20,13 @@
 namespace adios2
 {
 
-std::string FileToString(const std::string &fileName)
+std::string FileToString(const std::string &fileName) noexcept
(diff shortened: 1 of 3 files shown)


chuckatkins:
To enable the "auto-closing" of the issue, please place the "Fixes #blah" message in the actual commit message, not the PR description.

williamfgc:
The agreement is to implement XML and JSON in beta and that XML is experimental (only for our internal use). I need this capability today to avoid recompiling every single time for large scale MPI tests and to burn hours on Titan before the end of June. We can remove it (like we did with the examples). There is no new development in this PR, it's bug fixes and policy change.

PR #169, opened 2017-06-22
Use kwsys as object lib

chuckatkins:


chuckatkins:
This is currently pending design approval upstream and will remove WIP once that's settled.

chuckatkins:
We'll use this as-is for now and update again once upstream decides how it want's to proceed.

PR #177, opened 2017-06-23
Switch to pugixml

chuckatkins:
This was intended to get pushed back until after the alpha release but it was a fairly minor effort and since the XML processing is already in the release branch now then this should make it more reliable.

### DIFF
--- a/source/adios2/core/ADIOS.cpp
+++ b/source/adios2/core/ADIOS.cpp
@@ -79,6 +79,18 @@ IO &ADIOS::DeclareIO(const std::string ioName)
     return ioPair.first->second;
 }
 
+IO &ADIOS::GetIO(const std::string name)
+{
+    auto itIO = m_IOs.find(name);
+    if (itIO == m_IOs.end())
+    {
+        throw std::invalid_argument(
+            "ERROR: Unable to find previously defined IO object with name \"" +
+            name + "\" in call to GetIO.");
+    }
+    return itIO->second;
+}
+
 // PRIVATE FUNCTIONS
 void ADIOS::CheckMPI() const
 {
--- a/source/adios2/core/IO.cpp
+++ b/source/adios2/core/IO.cpp
@@ -45,6 +45,8 @@ void IO::SetIOMode(const IOMode ioMode) { m_IOMode = ioMode; };
 
 void IO::SetParameters(const Params &parameters) { m_Parameters = parameters; }
 
+const Params &IO::GetParameters() const { return m_Parameters; }
+
 unsigned int IO::AddTransport(const std::string type,
                               const std::vector<std::string> &parametersVector)
 {
--- a/source/adios2/helper/adiosXML.cpp
+++ b/source/adios2/helper/adiosXML.cpp
@@ -11,258 +11,107 @@
 #include "adiosXML.h"
 
 /// \cond EXCLUDE_FROM_DOXYGEN
-#include <sstream>
 #include <stdexcept> //std::invalid_argument
 /// \endcond
 
+#include "adios2/ADIOSMPI.h"
 #include "adios2/ADIOSTypes.h"
 #include "adios2/helper/adiosString.h"
+
+#include <pugixml.hpp>
 
 namespace adios2
 {
 
-void RemoveCommentsXML(std::string &currentContent) noexcept
+Params InitParametersXML(pugi::xml_node node, bool debugMode)
 {
-    std::string::size_type startComment(currentContent.find("<!--"));
-
-    while (startComment != currentContent.npos)
+    Params params;
+    for (pugi::xml_node paramNode : node.children("parameter"))
     {
-        std::string::size_type endComment(currentContent.find("-->"));
-        currentContent.erase(startComment, endComment - startComment + 3);
-        startComment = currentContent.find("<!--");
-    }
-}
-
-TagXML GetTagXML(const std::string tagName, const std::string &content,
-                 std::string::size_type &position)
-{
-    auto lf_SetPositions =
-        [](const std::string input, const std::string &content,
-           std::string::size_type &position) -> std::string::size_type {
-
-        const std::string::size_type inputPosition =
-            GetStringPositionXML(input, content, position);
-
-        if (inputPosition != std::string::npos)
+        pugi::xml_attribute attrKey = paramNode.attribute("key");
+        if (!attrKey)
         {
-            position = inputPosition + input.size();
-        }
-        return inputPosition;
-    };
-
-    TagXML tagXML;
-
-    std::string name(tagName);
-    if (name.back() == ' ')
-    {
-        name.pop_back();
-    }
-
-    auto openingStart = lf_SetPositions("<" + name, content, position);
-    auto openingEnd = lf_SetPositions(">", content, position);
-
-    if (openingStart == std::string::npos || openingEnd == std::string::npos)
-    {
-        tagXML.IsFull = false;
-        return tagXML;
-    }
-
-    tagXML.Header = content.substr(openingStart, openingEnd + 1 - openingStart);
-
-    auto closingStart =
-        GetStringPositionXML("</" + name + ">", content, position);
-
-    if (closingStart == std::string::npos)
-    {
-        throw std::invalid_argument(
-            "ERROR: could not find closing tag </" + name +
-            "> in XML config file, in call to ADIOS constructor\n");
-    }
-    tagXML.IsFull = true;
-    tagXML.Elements =
-        content.substr(openingEnd + 1, closingStart - (openingEnd + 1));
-
-    //    std::cout << "START..." << tagXML.Header << "...";
-    //    std::cout << tagXML.Elements << "...END\n";
-
-    return tagXML;
-}
-
-std::string::size_type
-GetStringPositionXML(const std::string input, const std::string &content,
-                     const std::string::size_type &startPosition) noexcept
-{
-    std::string::size_type foundPosition(content.find(input, startPosition));
-    if (foundPosition == content.npos)
-    {
-        return foundPosition;
-    }
-
-    // check if it is not inside " " or ' '
-    std::string::size_type currentPosition(startPosition);
-
-    while (foundPosition != content.npos)
-    {
-        const std::string::size_type singleQuotePosition(
-            content.find('\'', currentPosition));
-        const std::string::size_type doubleQuotePosition(
-            content.find('\"', currentPosition));
-
-        if ((singleQuotePosition == content.npos &&
-             doubleQuotePosition == content.npos) ||
-            (singleQuotePosition == content.npos &&
-             foundPosition < doubleQuotePosition) ||
-            (doubleQuotePosition == content.npos &&
-             foundPosition < singleQuotePosition) ||
-            (foundPosition < singleQuotePosition &&
-             foundPosition < doubleQuotePosition))
-        {
-            break;
-        }
-        // find the closing corresponding quote
-        std::string::size_type closingQuotePosition;
-
-        if (singleQuotePosition != content.npos &&
-            doubleQuotePosition == content.npos)
-        {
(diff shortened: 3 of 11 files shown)


chuckatkins:
Additional checks for things like duplicate engine entries and duplicate attribute names are currently ignored and just skipped over.  Rather than implement additional verification in the parser, I'd opt for us defining an actual schema to allow external tools to validate with.

source/adios2/core/ADIOS.cpp
```diff
@@ -79,6 +79,18 @@ IO &ADIOS::DeclareIO(const std::string ioName)
     return ioPair.first->second;
 }
 
+IO &ADIOS::GetIO(const std::string name)
+{
+    auto itIO = m_IOs.find(name);
+    if (itIO == m_IOs.end())
```
williamfgc: if( m_DebugMode )
{
   if (itIO == m_IOs.end())
   {
       ...
    }
}

source/adios2/core/ADIOS.cpp
```diff
@@ -79,6 +79,18 @@ IO &ADIOS::DeclareIO(const std::string ioName)
...
+    {
+        throw std::invalid_argument(
+            "ERROR: Unable to find previously defined IO object with name \"" +
+            name + "\".");
```
williamfgc: throw std::invalid_argument(
             "ERROR: Unable to find previously defined IO object " +
            name + ", in call to GetIO\n");

source/adios2/core/IO.cpp
```diff
@@ -45,6 +45,8 @@ void IO::SetIOMode(const IOMode ioMode) { m_IOMode = ioMode; };
 
 void IO::SetParameters(const Params &parameters) { m_Parameters = parameters; }
 
+const Params &IO::GetParameters() { return m_Parameters; }
```
williamfgc: the const will trigger a warning on some compilers: it's meaningless at the beginning of a function

chuckatkins: It's not meaningless, it's part of the return type, which is const ref.  Which now that I think about it, the whole function should be const as well.

Which compilers does it warn on?

source/adios2/helper/adiosXML.cpp
```diff
@@ -11,473 +11,162 @@
...
-            auto quotePosition =
-                currentTag.find_first_not_of(" \t\n", equalPosition + 1);
-
-            const char quote = currentTag.at(quotePosition);
-            if (quote == '\'' || quote == '"')
+            if (debugMode)
             {
-                value = lf_GetQuotedValue(quote, quotePosition, currentTag);
+                throw std::runtime_error(
+                    "No \"value\" attribute found on <parameter> element.");
```
williamfgc: throw std::runtime_error("ERROR: in XML config file, no \"value\" attribute found on <parameter> element" + tagHeader + " in call to ADIOS constructor\n");

source/adios2/core/ADIOS.cpp
chuckatkins: Done

source/adios2/helper/adiosXML.cpp
```diff
@@ -11,473 +11,162 @@
...
-            break;
-        }
-        InitIOXML(ioXML, mpiComm, debugMode, transforms, ios);
+        params.emplace(attrKey.value(), attrValue.value());
     }
+    return params;
 }
 
-void InitIOXML(const TagXML &ioXML, const MPI_Comm mpiComm,
+void InitIOXML(pugi::xml_node ioNode, const MPI_Comm mpiComm,
```
williamfgc: Should it be const pugi::xml_no ioNode ?

source/adios2/helper/adiosXML.cpp
```diff
@@ -11,473 +11,162 @@
...
-                ioXML.Header +
-                ", check XML config file, in call to ADIOS constructor\n");
-        }
-
-        if (ios.count(ioName) == 1) // io exists
+        if (debugMode)
         {
-            throw std::invalid_argument("ERROR: io name " + ioName +
-                                        " must be unique in XML config file, "
-                                        "in call to ADIOS constructor\n");
```
williamfgc: Let's be consistent in the exception error messages. These are for the user to report bugs. The more informative the better, use std::invalid_argument to be more specific and that this is coming from the XML config file and the ADIOS constructor. Keep in mind that adios is typically just one component of many. It's nice to know where the ERROR comes from in the user code.

source/adios2/core/ADIOS.cpp
chuckatkins: This one is a bit unclear.  If the only way to handle the name not found is to throw an exception since there's no way to return a "null" value for a reference object.  Give that the Declare method is also a reference object then I think it makes sense to keep this one as well.  If we would rather switch those to pointers instead of references, which I think is reasonable, so a null return is meaningful in the case of an error but no exception thrown, thhen I think it should be done in a seperate PR.

source/adios2/core/IO.cpp
williamfgc: functions don't decide constness of their return types.  intel compiler will issue those.  And no, you can't make the function const when returning an internal reference... just try if you're not convinced. 

const Foo foo = GetFoo();
Foo foo = GetFoo();

source/adios2/helper/adiosXML.cpp
chuckatkins: The error messages can be changed to clearly reflect they are coming from XML parsing.

chuckatkins: Done

source/adios2/core/ADIOS.cpp
williamfgc: m_DebugMode = true --> we try to catch as many exceptions as possible from user input
m_DebugMode = false --> the user has run the same thing over and over again and it works, no need for the same checks over and over again.

source/adios2/helper/adiosXML.cpp
williamfgc: and the ADIOS function the user calls.  What if an application has their own XML inputs for other things....e.g. VTK input files.

source/adios2/helper/adiosXML.cpp
chuckatkins: It doesn't really matter as much since it's an internal helper function anyways, but sure.

source/adios2/core/ADIOS.cpp
williamfgc: Let's avoid pointers if nullptr is not a valid state. References were created for that purpose.

chuckatkins: > the user has the choice to turn off extra checks, not us.

Sure, but some check's aren't extras.  In this case, what should the behavior be if the requested object is not found, i.e. what get's returned.

williamfgc: the STL will throw its own exception if itIO is the end iterator.  Then then user will have to turn on m_DebugMode (which is the recommended mode for development) if ADIOS catches this and provides more context.  That's how I am handling all exception checks. For well established flows it's unnecessary to do the same checks. Also, we are providing the option to other library developers to handle exceptions their own way and not always from adios.

source/adios2/helper/adiosXML.cpp
chuckatkins: InitXML has no idea where it's being called from.  Really none of the helper functions do since they're stand-alone and not members.  Now, if the parsing was in a private member function of the ADIOS class, which would make sense if nothing else is supposed to call to call it, then it would be feasible to do say the error was coming from the ADIOS class.  As a stand alone non-member function, it's an inversion of ownership though.  If the parsing moves to the ADIOS class then I'll support changing the messages but that's outside the scope of this PR.

williamfgc: InitXML is specifically created as a helper to the ADIOS class. The content has very specific components and would make a full-blown private function in ADIOS.cpp, that's why a helper come in handy.  Helpers are not always stand-alone, in fact they can be whatever we want to decongest classes. Can we guarantee we are calling it somewhere else?

williamfgc: " If the parsing moves to the ADIOS class then I'll support changing the messages but that's outside the scope of this PR "

You are changing original behavior that had a certain purpose....that's inside the scope of the original code that the PR is changing.

source/adios2/core/IO.cpp
chuckatkins: It's part of the return type, i.e. const ref, so it forces you to treat it as such.  For instance:

```
class Bar
{
  Foo m_Foo;
  const Foo& GetFoo() const { return m_Foo};
};
const Foo& foo1 = Bar.GetFoo(); // fine
Foo& foo2 = Bar.GetFoo(); // Error
```
The lvalue of GetFoo cannot be assigned to a mutable reference, it has to be const, which forces const access to the m_Foo member, even as it's reference is exposed publicly.

Also verified no warnings with Intel 17 compiler.

source/adios2/helper/adiosXML.cpp
```diff
@@ -11,473 +11,166 @@
 #include "adiosXML.h"
 
 /// \cond EXCLUDE_FROM_DOXYGEN
+#include <cstring> // std::strcmp
```
williamfgc: not used....let's stick with std::string for string operations

chuckatkins: That was leftover from an initial pass.  I'll drop it.

source/adios2/core/ADIOS.cpp
chuckatkins: It doesn't throw anything.  Dereferencing an end() iterator is undefined and is effectively an access violation.

source/adios2/helper/adiosXML.cpp
chuckatkins: done

chuckatkins:
So, I believe the only issue now is what to do about the GetIO method when the requested name cannot be found.  If in debug mode then obviously we throw an exception, but if not, we can't just segfault with an access violation.  As far as I can see, there's no valid exit path for the function outside of throwing an exception.

source/adios2/core/ADIOS.cpp
williamfgc: This is the kind of discussion that end with testing code. Yeah, dereferencing is undefined behavior that's won't show until you try accessing the dereferenced iterator within C++. That's when the exception will be thrown, typically as bad_alloc.  Anyways, the point here is that we don't guarantee anything if m_DebugMode == false. That's the user being confident on its own flow.

chuckatkins: There won't be an exception since there's no code inside libstdc++ that's testing for an error case and throwing one, which is different from the new operator which checks for success or failure of memory allocation and can throw a bad_alloc exception when detected.  I don't think that we can allow an access violation when we know it's going to occur.  In this case, our software stack is in a known state and can easily proceed and recover.

williamfgc:
C++ gives you 2 options to access a vector by index, at(i)  (bounds check exception) and operator [i] (doesn't do bounds check, undefined behavior). We are giving exactly the same options.

source/adios2/core/ADIOS.cpp
williamfgc: Please test with code before making assumptions.

source/adios2/helper/adiosXML.cpp
williamfgc: Const correctness matters all over the code. Provides intent and compile time checks. Has nothing to do with scope.

source/adios2/core/ADIOS.cpp
williamfgc: You are returning a reference therefore there is always a constructor involved, lvalue or rvalue

source/adios2/core/IO.cpp
williamfgc: Why const? What is the user expected to do with Params? Just read in?

Issue #180, opened 2017-06-25
Limit File Writes to 2Gb


williamfgc:
Without this limit it fails on Titan for POSIX write

chuckatkins:
Fixed by #180

PR #181, opened 2017-06-25
Solve Issue #180 Modified file transports to enable write loop in size > 2Gb

williamfgc:
Checking std::ios_base::failure exceptions
IOChrono is off by default, each owner turns it on explicitly
Modified FileDescriptor and FilePointer destructors
Added Tests BPWriteRead for stdio and fstream
Removed IO AddTransport with vector (shoud have been there?) to enable
AddTransport in Tests

### DIFF
--- a/examples/experimental/multistep/reader_allsteps.cpp
+++ b/examples/experimental/multistep/reader_allsteps.cpp
@@ -115,7 +115,7 @@ int main(int argc, char *argv[])
             bpReaderSettings.SetEngine(
                 "ADIOS1Reader"); // BP is the default engine
             // see only one step at a time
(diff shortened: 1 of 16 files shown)


pnorbert:
Looks good to me. BTW in adios1 the limit is set to 2GB-16MB, and that works too. I don't know if it would make any difference in performance to have ~1G or ~2G chunks. 
adios_internals.h:#define MAX_MPIWRITE_SIZE 2130706432    /* 2GB - 16MB */

williamfgc:
@pnorbert  Thanks! Yeah, I don't know about performance wrt XG chuck sizes, we'll have to measure.

PR #185, opened 2017-06-25
Fix find adios1cmake

pnorbert:


pnorbert:
Sorry, wrong target again.

PR #190, opened 2017-06-27
Memory

williamfgc:
Increased Default Maximum Memory to 200Mb --> Changed to Unlimited per time step.
Added config.xml for the heat transfer example for memory management
(among other parameters) at runtime 

To do:
Copy config.xml to /build/bin, right now the heat transfer example won't
run if config.xml is not found


pnorbert:
So with these changes, you buffer one output step as long as you don't run out of memory. How does time aggregation behave?

williamfgc:
@pnorbert the same way, time aggregation only updates the time_index as we go forward. Or there is something else time aggregation does?

chuckatkins:
See https://github.com/ornladios/ADIOS2/commit/f6477a1a633eb5f5212da1d808268a1504cd596a for how to deal with the location of the xml config

chuckatkins:
Sorry if I'm missing it but I see what the branch is doing but what bug or problem is it trying to fix?

PR #198, opened 2017-06-28
Async I/O implementation of adios Write in heat transfer examples

williamfgc:
Added an extra argument to the heat transfer example to make I/O truly asynchronous using C++11 std::future and std::async. 

I can see some overlap:
sync: ./heatTransfer_write_adios2 ht2 1 1 2500 2500 500 10  --> 136 s

async : ./heatTransfer_write_adios2 ht2 1 1 2500 2500 500 10 on ---> 96 s

### DIFF
--- a/examples/heatTransfer/write/CMakeLists.txt
+++ b/examples/heatTransfer/write/CMakeLists.txt
@@ -15,7 +15,7 @@ if(ADIOS2_HAVE_MPI)
   target_include_directories(heatTransfer_write_adios2
     PRIVATE ${MPI_C_INCLUDE_PATH}
   )
-  target_link_libraries(heatTransfer_write_adios2 adios2 ${MPI_C_LIBRARIES})
+  target_link_libraries(heatTransfer_write_adios2 adios2 ${MPI_C_LIBRARIES} pthread)
   target_compile_definitions(heatTransfer_write_adios2 PRIVATE
    -DDEFAULT_CONFIG=${CMAKE_CURRENT_SOURCE_DIR}/config.xml)
 
@@ -33,7 +33,7 @@ if(ADIOS2_HAVE_MPI)
       PRIVATE ${MPI_C_INCLUDE_PATH}
     )
     target_link_libraries(heatTransfer_write_adios1
-      adios1::adios ${MPI_C_LIBRARIES}
+      adios1::adios ${MPI_C_LIBRARIES} pthread
     )
   endif()
 
@@ -51,7 +51,7 @@ if(ADIOS2_HAVE_MPI)
       PRIVATE ${MPI_C_INCLUDE_PATH} ${HDF5_C_INCLUDE_DIRS}
     )
     target_link_libraries(heatTransfer_write_hdf5
-      ${MPI_C_LIBRARIES} ${HDF5_C_LIBRARIES}
+      ${MPI_C_LIBRARIES} ${HDF5_C_LIBRARIES} pthread
     )
   endif()
 
@@ -70,7 +70,7 @@ if(ADIOS2_HAVE_MPI)
       PRIVATE ${MPI_C_INCLUDE_PATH} ${HDF5_C_INCLUDE_DIRS}
     )
     target_link_libraries(heatTransfer_write_ph5
-      ${MPI_C_LIBRARIES} ${HDF5_C_LIBRARIES}
+      ${MPI_C_LIBRARIES} ${HDF5_C_LIBRARIES} pthread
     )
   endif()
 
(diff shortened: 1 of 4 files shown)


pnorbert:
1. Is pthread always required to link with the application when using C++11 futures? Or is this system specific? 
2. Did you see any performance difference with the async run? Is this going into the release or master?

pnorbert:
Does this work on Titan at all? Multithreaded MPI is not working everywhere. I know right now ADIOS2 is doing N-to-N pattern and no metadata, so it may work at the moment, but aggregation will run into trouble. So I would not put this into the release because we don't have an agreement on asynchronous I/O and I don't want to push ideas onto users at this point.

williamfgc:
@pnorbert yeah, on POSIX systems, C++11 threads always depend on libpthread (in fact I'd say all of them depend on libpthread: OpenMP, TBB, C11, Java threads, etc.) @chuckatkins I don't know on Windows or other systems.  @chuckatkins Will do the change. Yeah, performance looks nice! Thanks!

williamfgc:
@pnorbert, this is independent of what ADIOS is doing....this is just overlapping calculation and IO and making sure IO is done when it should be for the next step.  On Titan we can enable threads per MPI node, this is not MPI threads per-se, more OpenMP style.

pnorbert:
You mean, only the buffering is overlapped? The advance() function and the actual disk IO is not?
I meant, in general, you can't have two threads doing MPI communication at the same time, so the computation must not communicate while the IO thread is doing any communication. Metadata collection is an example where the IO thread would need to do communication.

williamfgc:
everything inside async and the rest of the code in the t loop overlap. Depending on the adios buffer memory size we might or not be writing to disk. For this case we are in every time step. 

 for (unsigned int t = 1; t <= settings.steps; ++t)
        {
            if (rank == 0)
                std::cout << "Step " << t << ":\n";
            for (unsigned int iter = 1; iter <= settings.iterations; ++iter)
            {
                ht.iterate();
                ht.exchange(mpiHeatTransferComm);
                ht.heatEdges();
            }

            if (settings.async)
            {
                futureWrite.get();
                futureWrite = std::async(std::launch::async, &IO::write, &io, t,
                                         std::ref(ht), std::ref(settings),
                                         mpiHeatTransferComm);
            }
}

pnorbert:
Okay, looking at the source now, not the diff. I see that ADIOS Advance() and ht.exchange() overlaps. Both of them uses MPI (will use in the future). What guarantees that this works on every machine?

williamfgc:
@pnorbert the only guarantee we have is C++11 portability....as of today of course there is zero guarantee until we test. Still, nice performance bump.

pnorbert:
Yes, nice performance bump with very simple work but I don't want to see this example in the release. Async IO is a new feature and we don't do new features in the last week, do we? We need a thorough discussion on asynchronous IO.

chuckatkins:
> What guarantees that this works on every machine?

I don't know that we have that guarantee given the MPI threading issues some systems have.  However, it is an optional setting that is off by default so you'd have to explicitly enable it.  And I think it's good to show in an example how it be done.

williamfgc:
@pnorbert this is in the heat transfer example.....the agreement is that we don't add anything new in the source code of adios, only examples and tests. This is an example that an application need to manage from its side, not ours.  Yes, and it's nice to start the discussion with numbers, not just words.

examples/heatTransfer/write/CMakeLists.txt
```diff
@@ -15,13 +16,15 @@ if(ADIOS2_HAVE_MPI)
...
-  target_link_libraries(heatTransfer_write_adios2 adios2 ${MPI_C_LIBRARIES})
+  target_link_libraries(heatTransfer_write_adios2 adios2 ${MPI_C_LIBRARIES} 
+                        ${CMAKE_THREAD_LIBS_INIT})
   target_compile_definitions(heatTransfer_write_adios2 PRIVATE
    -DDEFAULT_CONFIG=${CMAKE_CURRENT_SOURCE_DIR}/config.xml)
 
   if(ADIOS2_HAVE_ADIOS1)
     find_package(ADIOS1 REQUIRED)
     find_package(MPI COMPONENTS C REQUIRED)
+    find_package(Threads REQUIRED)
```
chuckatkins: `find_package(Threads REQUIRED)` only needs to be called once at the top level.

williamfgc: Let me do the change!

chuckatkins:
Other than the redundant `find_pacakge(Threads` calls, it looks fine to me.  The `find_pacakge(MPI` calls are also redundant but that was already there so not your issue to fix (unless you want to, then great, but it's not a blocker for me)

williamfgc:
yeah, I followed the MPI call.... @chuckatkins  your call, I just made the requested change.

chuckatkins:
Looks good to me.  Please merge.

PR #201, opened 2017-06-28
Updating some documentation, removed a warning and h5 file name

williamfgc:
Warning triggered when -DCMAKE_BUILD_TYPE=Release on Mdtman

### DIFF
--- a/ReadMe.md
+++ b/ReadMe.md
@@ -15,9 +15,20 @@ Copyright.txt for more details.
 * source - Main ADIOS source
 * testing - Tests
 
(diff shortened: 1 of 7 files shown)


chuckatkins:
Thanks for removing the warning change.  In future PRs, rather than adding a commit to revert a change in a previous commit that's already part of the pr, it would be preferable to just re-write that previous commit.  It's not something I'm going to nit-pick over now though, just something to keep in mind in the future.

Looks good to me.  Please merge.

PR #207, opened 2017-07-18
Moved profiling.log to profiling.json format

williamfgc:
Added test for generating profiling.json
Fixed inclusion for NLohmannJson json.hpp in Test
Tested resulting profiling.json on R (jsonlite), Python (json)
and C++ (json for modern C++)

### DIFF
--- a/source/adios2/engine/bp/BPFileWriter.cpp
+++ b/source/adios2/engine/bp/BPFileWriter.cpp
@@ -71,37 +71,12 @@ void BPFileWriter::Close(const int transportIndex)
                                    m_BP1Writer.m_HeapBuffer.m_DataPosition,
                                    transportIndex);
 
-    // close them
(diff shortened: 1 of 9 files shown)


chuckatkins:
Can you verify whether or not this breaks the windows+visual studio build?  I'm not sure if the problem was dataman itself or json.hpp

williamfgc:
I don't have access to a Windows machine, @pnorbert @JasonRuonanWang @chuckatkins do you have access? json.hpp should be supported on Windows: https://github.com/nlohmann/json#supported-compilers

JasonRuonanWang:
@williamfgc It looks good to me. But I never use the .value method of this JSON implementation, so I am not sure if additional exception catching would be needed, in case the function does not return the correct type. 

I don't have a Windows machine either.

pnorbert:
This PR does not break my windows build.

Issue #208, opened 2017-07-19
Tests fail on Mac for long int


williamfgc:
long int not compatible with int64_t , it's long long int on Mac

PR #209, opened 2017-07-19
Fixing Issue #208

williamfgc:
Using language primitives in test data instead of fixed width integers.

casting size_t to unsigned long in mpidummy to remove warning on Mac

Tested on Mac and Linux (Ubuntu 16.04)


testing/adios2/engine/SmallTestData.h
```diff
@@ -10,21 +10,21 @@
...
         {128, 129, 130, 131, 132, 133, 134, 135, 136, 137}};
-    std::array<uint16_t, 10> U16 = {
+    std::array<unsigned short, 10> U16 = {
         {32768, 32769, 32770, 32771, 32772, 32773, 32774, 32775, 32776, 32777}};
-    std::array<uint32_t, 10> U32 = {
+    std::array<unsigned int, 10> U32 = {
         {2147483648, 2147483649, 2147483650, 2147483651, 2147483652, 2147483653,
          2147483654, 2147483655, 2147483656, 2147483657}};
-    std::array<uint64_t, 10> U64 = {
+    std::array<unsigned long int, 10> U64 = {
```
chuckatkins: The decision to use fixed-width types instead of primitives in the test data was explicit and on purpose to make sure we explicitly test the right data sizes.  "unsigned long int", for instance, is often a 32-bit int, but could be 64.  uint64_t makes sure that we're testing 64-bit ints.

Is there a reason to change the test data?

source/adios2/mpidummy.cpp
```diff
@@ -329,9 +329,10 @@ int MPI_File_read(MPI_File fh, void *buf, int count, MPI_Datatype datatype,
     if (bytes_read != bytes_to_read)
     {
         std::snprintf(mpierrmsg, MPI_MAX_ERROR_STRING,
-                      "could not read %" PRId64 " bytes. read only: %" PRId64
+                      "could not read %llu bytes. read only: %llu"
                       "\n",
-                      bytes_to_read, bytes_read);
+                      (unsigned long long)bytes_to_read,
```
chuckatkins: Can these use `static_cast<unsigned long long>` instead?  Other than that, it's fine.

testing/adios2/engine/SmallTestData.h
williamfgc: Fixed width types need to be consistent across the tests. For example the uint64_t data is passed to a variable defined with unsigned long. We either change the fixed width types or the DefineVariable<T> calls, I just picked one way.

source/adios2/mpidummy.cpp
williamfgc: This is from C code, I have no preferences.

testing/adios2/engine/SmallTestData.h
williamfgc: @chuckatkins I am going to merge this, please feel free to make everything fixed width type or primitive type, I have no preference as fixed width types get mapped automatically by the compiler. We just can't mix both (data: int64_t pointer with test: DefineVariable<unsigned long long>) as in the original tests. I am just providing a fix for testing on Mac, also see my PR for Windows. I finally have access to a VM. Thanks!

Issue #211, opened 2017-07-20
Generated BP File on Windows is corrupted


williamfgc:

williamfgc:
Solved with PR #212.

Root cause:
Missed binary mode in opening files.
Reinterpret_cast between size_t* and uint_64t* showed non-zero bytes on Windows. Replaced with local variable + static_cast

PR #212, opened 2017-07-21
Solve Issue #211 and made changes due to Windows warnings and bugs

williamfgc:
Bugs: 1) Missing binary mode in FilePointer fopen and FStream ofstream
transports was adding characters to bp files becoming corrupted.
      2) Replacing reinterpret_cast of size_t to uint_64, not the same size
on Windows 64-bit

Removed some warnings generated by Visual Studio

Added static_cast for type conversions

Refactor XML reader to restore original check on single engine
conditions and eliminate warnings. Test added.

Allow ctime and fopen "not secure" warnings: now using binary fstream as
default file IO library on Windows

### DIFF
--- a/source/adios2/helper/adiosXML.cpp
+++ b/source/adios2/helper/adiosXML.cpp
@@ -6,6 +6,7 @@
  *
  *  Created on: May 17, 2017
  *      Author: William F Godoy godoywf@ornl.gov
+ *              Chuck Atkins chuck.atkins@kitware.com
  */
 
 #include "adiosXML.h"
@@ -16,6 +17,7 @@
 
 #include "adios2/ADIOSMPI.h"
 #include "adios2/ADIOSTypes.h"
+#include "adios2/helper/adiosMPIFunctions.h"
 #include "adios2/helper/adiosString.h"
 
 #include <pugixml.hpp>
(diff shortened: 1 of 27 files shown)


source/adios2/helper/adiosXML.cpp
```diff
@@ -122,49 +155,43 @@ void InitXML(const std::string configXML, const MPI_Comm mpiComm,
...
-                "ERROR: XML: No <adios-config> element found");
+                "ERROR: XML: No <adios-config> element found in file " +
+                configXML + ", in call to ADIOS constructor\n");
         }
-        return;
     }
 
-    ios.clear();
-    for (pugi::xml_node ioNode : configNode.children("io"))
+    for (const pugi::xml_node ioNode : configNode.children("io"))
```
chuckatkins: I think the `ios.clear()` still needs to be there.

source/adios2/toolkit/capsule/heap/STLVector.cpp
```diff
@@ -45,7 +45,8 @@ void STLVector::ResizeData(const size_t size)
         {
             throw std::runtime_error("ERROR: bad_alloc detected when resizing "
```
chuckatkins: Would these be better suited to use `std::throw_with_nested`?
```cpp
            std::throw_with_nested(
                std::runtime_error("ERROR: bad_alloc detected when resizing "
                                   "data buffer with size " +
                                   std::to_string(size) + "\n"));
```

source/adios2/toolkit/format/bp1/BP1Writer.cpp
```diff
@@ -56,7 +56,8 @@ void BP1Writer::WriteProcessGroupIndex(
     WriteNameRecord(name, dataBuffer, dataPosition);
 
     // processID in metadata,
-    const uint32_t processID = static_cast<uint32_t>(m_BP1Aggregator.m_RankMPI);
+    const uint32_t processID =
```
chuckatkins: Why change the cast?  The base type is `int` so the cast to `uint32_t` ensures it's the right size and constness is in the assignment via the lvalue type, so it already has const correctness.

source/adios2/toolkit/transport/file/FilePointer.cpp
```diff
@@ -48,16 +48,16 @@ void FilePointer::Open(const std::string &name, const OpenMode openMode)
 
     if (m_OpenMode == OpenMode::Write)
     {
-        m_File = std::fopen(name.c_str(), "w");
+        m_File = fopen(name.c_str(), "wb");
```
chuckatkins: Good catch

williamfgc: This is the root cause for a corrupted file on Windows. I am switching to fstream binary, fopen triggers a non safe warning in favor of fopen_s

chuckatkins: the _s functions are a windows thing I believe.  Other OSs have thier own versions (OpenBSD has a different version).  You should be able to remove the warning though with:
```cpp
#ifdef _MSC_VER
#define _CRT_SECURE_NO_WARNINGS
#endif
```
at the top.  I do think it would still be useful to keep the C stdio implementation since the C stdio operations tends to be much more performance than the C++ iostream.  If instead, though, you don't want to bother with it and just delete the backend completely that's fine too, but since it's already there I think it's useful to keep around.

source/adios2/helper/adiosXML.cpp
williamfgc: It has no effect, this is only called from ADIOS constructors, IOs are always empty at that stage.

source/adios2/toolkit/capsule/heap/STLVector.cpp
williamfgc: Good catch.

source/adios2/toolkit/format/bp1/BP1Writer.cpp
williamfgc: I don't recall if this caused a Windows warning, without static_cast for sure. Doesn't hurt to keep the intend of the cast. It will fail (or at least throw a warning) if uint32_t = static_cast<const uint32_t>...

source/adios2/toolkit/transport/file/FilePointer.cpp
williamfgc: I tried #define _CRT_SECURE_NO_WARNINGS on my Windows 10 and Visual Studio 17. It would still give the warning for fopen and ctime. C++ ifstream binary at worse is a thin wrapper around C stdio + 1 or 2 pointer redirections for write(buffer, size), at best it will go deeper into the kernel IO implementation. I've never experience any slowdown as both will pick up the system buffer size. The bad reputation comes from the fstream as a stringstream where the overloaded << >> for casting types will slow things down in heap, but not for system IO writes and reads (which are probably orders of magnitude larger).

williamfgc:
@JasonRuonanWang thanks. I will take a look.

source/adios2/toolkit/transport/file/FilePointer.cpp
chuckatkins: So, it looks like the `#define` has changed across different versions of visual studio.  Some used `_CRT_SECURE_NO_WARNINGS` and others used `_CRT_SECURE_NO_DEPRECATE`.  There's also another set of deprecation warnings you can run into that are essentially the same thing that need the `_SCL` prefix to turn off.  Defining both should suffice:
```cpp
#ifdef _MSC_VER
#define _CRT_SECURE_NO_WARNINGS
#define _CRT_SECURE_NO_DEPRECATE
#define _SCL_SECURE_NO_WARNINGS
#define _SCL_SECURE_NO_DEPRECATE
#endif
```
It's a pretty common code block for cross-platform C and C++ to address this very problem.

williamfgc: Let me give it a shot! Thanks @chuckatkins

williamfgc: FYI. The only thing that worked was removing the warnings explicitly with pragmas:

#ifdef _WIN32
#pragma warning(disable : 4996) // fopen
#endif

williamfgc:
@chuckatkins Can you verify this is the right setup for merging this branch? Looks like master was merged into windowsbug, not the other way around. Thanks!

PR #215, opened 2017-07-25
Add plugin engine

chuckatkins:


chuckatkins:
I still need to add tests and examples, but the implementation is there now.

source/adios2/helper/adiosDynamicBinder.h
```diff
@@ -0,0 +1,38 @@
...
+
+public:
+    DynamicBinder(std::string libName);
+    ~DynamicBinder();
+
+    VoidSymbolPointer GetSymbol(std::string symbolName);
+
+private:
+    struct Impl;
+    std::unique_ptr<Impl> m_Impl;
```
williamfgc: Keep in mind the copy constructor will get deleted implicitly due to std::unique_ptr.

williamfgc:
Great job! More than reviewing it, I'd like to understand the implementation once you're done. If this is real-time I'd think the bool m_DebugMode we discussed before has probably little to no benefit, and exceptions might have to be always checked. I don't know enough to evaluate, though.

Issue #216, opened 2017-07-25
ADIOS Copy Constructor issue with Python Bindings


williamfgc:

williamfgc:
Tested on Mac and Linux (Ubuntu)

PR #217, opened 2017-07-26
Fixing Issue #216 Python bindings

williamfgc:
Copy constructor was implicitly removed for ADIOSPy due to m_ADIOS.

PyBind11 needed the copy constructor.

m_ADIOS is now a shared pointer to avoid copy constructor removal

Separated MPI nonMPI test as ctest failed for MPI

### DIFF
--- a/bindings/python/ADIOSPy.cpp
+++ b/bindings/python/ADIOSPy.cpp
@@ -19,7 +19,8 @@ namespace adios2
 
 ADIOSPy::ADIOSPy(const std::string configFile, MPI_Comm mpiComm,
                  const bool debugMode)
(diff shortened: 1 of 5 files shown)


chuckatkins:
Since the adios python object is the only thing that access the adios c++ object, should it be a unique_ptr instead of a shared_ptr?

williamfgc:
@chuckatkins good point....std::unique_ptr has its copy constructor deleted by definition. Clang will do the same thing and delete ADIOSPy copy constructor since m_ADIOS doesn't exist. I use shared_ptr most of the time as we don't know what we end up doing later (putting this inside a thread, using with libraries, etc.) std::unique_ptr has too much "static"-like behavior.

PR #219, opened 2017-07-28
Implemented Transform class API, and BZip2, Zfp wrappers

williamfgc:
Wrapper classes CompressBZip2 and CompressZfp implemented
Under transform/compress and adding examples and tests
Added/Modified cmake files

To do:
Add more tests
Add BP format support


williamfgc:
@JasonRuonanWang thanks for reporting this. Can you verify ZFP's version? I tested this on Linux and Mac with the latest ZFP from their master.

JasonRuonanWang:
Right. I tried the newest version and it all worked for me. Sorry about that. I didn't know that they change the API.

chuckatkins:
@williamfgc @JasonRuonanWang is there a minimum version required?  If so I can update the find module to properly deal with versioning to make sure we only find a compatible version.

JasonRuonanWang:
@chuckatkins I think in this case the minimum version required is 0.5.1. I was using 0.5.0 when I was the error.

examples/hello/bpBZip2Wrapper/CMakeLists.txt
```diff
@@ -0,0 +1,18 @@
...
+  add_executable(hello_bpBZip2Wrapper helloBPBZip2Wrapper.cpp)
+  target_include_directories(hello_bpBZip2Wrapper PRIVATE ${MPI_C_INCLUDE_PATH})
+  target_link_libraries(hello_bpBZip2Wrapper ${MPI_C_LIBRARIES})
+  
+else()
+  add_executable(hello_bpBZip2Wrapper helloBPBZip2Wrapper_nompi.cpp)
+  
+endif()
+
+target_link_libraries(hello_bpBZip2Wrapper adios2 ${BZIP2_LIBRARIES})
```
chuckatkins: bzip2 is a private dependency of ADIOS.  Since it's not directly being used then the libraries shouldn't be included as a dependency.  Also, if you were to use them (which you shouldn't need, they need to be dropped) then you'd want to use the imported targets instead of the libraries directly, il.e use `BZip2::BZip2` instead of `${BZIP2_LIBRARIES}`

examples/hello/bpZfpWrapper/CMakeLists.txt
```diff
@@ -0,0 +1,18 @@
...
+  add_executable(hello_bpZfpWrapper helloBPZfpWrapper.cpp)
+  target_include_directories(hello_bpZfpWrapper PRIVATE ${MPI_C_INCLUDE_PATH})
+  target_link_libraries(hello_bpZfpWrapper ${MPI_C_LIBRARIES})
+  
+else()
+  add_executable(hello_bpZfpWrapper helloBPZfpWrapper_nompi.cpp)
+  
+endif()
+
+target_link_libraries(hello_bpZfpWrapper adios2 zfp::zfp)
```
chuckatkins: Similar to bzip2, zfp is a private dependency of adios2 and isn't being used directly by the executable.

source/adios2/CMakeLists.txt
```diff
@@ -71,15 +71,16 @@ endif()
   
 if(ADIOS2_HAVE_BZip2)
   find_package(BZip2 REQUIRED)
-  target_sources(adios2 PRIVATE transform/compression/BZip2.cpp)
-  target_link_libraries(adios2 PRIVATE BZip2::BZip2)
+  target_sources(adios2 PRIVATE transform/compress/CompressBZip2.cpp)
+  target_include_directories(adios2 PUBLIC ${BZIP2_INCLUDE_DIR})
```
chuckatkins: Drop the use of the BZIP2 CMake variables and restore the use of the `BZip2::BZip2` imported target instead.

source/adios2/CMakeLists.txt
```diff
@@ -71,15 +71,16 @@ endif()
...
 endif()
 
 if(ADIOS2_HAVE_ZFP)
   find_package(ZFP REQUIRED)
-  message("ADIOS ZFP support not yet implemented")
-#  target_sources(adios2 PRIVATE transform/compression/ZFP.cpp)
-#  target_link_libraries(adios2 PRIVATE zfp::zfp)
+  target_sources(adios2 PRIVATE transform/compress/CompressZfp.cpp)
+  target_include_directories(adios2 PUBLIC ${ZFP_INCLUDE_DIRS})
+  target_link_libraries(adios2 PUBLIC zfp::zfp)
```
chuckatkins: Drop the include dirs, they should already be provided by the imported target.  zfp should also remain a private dependency.

source/adios2/helper/adiosString.cpp
```diff
@@ -129,4 +129,83 @@ void SetParameterValue(const std::string key, const Params &parameters,
...
+        }
+    }
+    else
+    {
+        valueDouble = std::stod(value);
+    }
+    return valueDouble;
+}
+
+double StringToUInt(const std::string value, const bool debugMode,
```
chuckatkins: return type double?

source/adios2/helper/adiosString.h
```diff
@@ -67,6 +67,40 @@ GetParametersValues(const std::string &key,
...
+
+/**
+ * function that cast a string to unsigned int verifying validity of the cast
+ * with exceptions in debugMode
+ * @param value string to be casted
+ * @param debugMode check for string conversion
+ * @param hint passed for extra debugging info if exception is thrown
+ * @return value as unsigned int
+ */
+double StringToUInt(const std::string value, const bool debugMode,
```
chuckatkins: Wrong return type

source/adios2/transform/compress/CompressBZip2.h
```diff
@@ -0,0 +1,76 @@
...
+ * CompressBZip2.h : wrapper to BZip2 compression library
+ *
+ *  Created on: Jul 24, 2017
+ *      Author: William F Godoy godoywf@ornl.gov
+ */
+
+#ifndef ADIOS2_TRANSFORM_COMPRESSION_COMPRESSBZIP2_H_
+#define ADIOS2_TRANSFORM_COMPRESSION_COMPRESSBZIP2_H_
+
+#include <bzlib.h>
```
chuckatkins: bzlib.h can be moved to the cpp file since it's not used anywhere in the public interface for this class.

source/adios2/transform/compress/CompressBZip2.h
```diff
@@ -0,0 +1,76 @@
...
+{
+
+public:
+    /**
+     * Unique constructor
+     * @param debugMode
+     */
+    CompressBZip2(const bool debugMode);
+
+    ~CompressBZip2() = default;
```
chuckatkins: Does the destructor need to be virtual here?

source/adios2/transform/compress/CompressBZip2.h
```diff
@@ -0,0 +1,76 @@
...
+
+    size_t BufferMaxSize(const size_t sizeIn) const final;
```
chuckatkins: Do these need to be final?  Why not let them be derived from?

source/adios2/transform/compress/CompressZfp.h
```diff
@@ -0,0 +1,101 @@
...
+     * @param dimensions
+     * @param type
+     * @return size of decompressed data in dataOut
+     */
+    size_t Decompress(const void *bufferIn, const size_t sizeIn, void *dataOut,
+                      const Dims &dimensions, const std::string type,
+                      const Params &parameters) const final;
+
+private:
+    /**
```
chuckatkins: Are these helper functions using any data members?  If not then they can easily be made standalone instead of members and live entirely in the cpp file, out of the header.  That would also let you move the zfp include the cpp file to minimize include pollution in the interface.

testing/adios2/transform/CMakeLists.txt
```diff
@@ -0,0 +1,19 @@
...
+# Distributed under the OSI-approved Apache License, Version 2.0.  See
+# accompanying file Copyright.txt for details.
+#------------------------------------------------------------------------------#
+
+
+if(ADIOS2_HAVE_BZip2)
+  add_executable(TestBZip2Wrapper TestBZip2Wrapper.cpp)
+  target_link_libraries(TestBZip2Wrapper adios2 gtest gtest_main ${BZIP2_LIBRARIES})
+
+  gtest_add_tests(TARGET TestBZip2Wrapper ${BZIP2_LIBRARIES})
```
chuckatkins: gtest_add_tests only needs the target, not any link libraries

testing/adios2/transform/TestBZip2Wrapper.cpp
```diff
@@ -0,0 +1,125 @@
+#include <cstdint>
+#include <iostream>
+#include <numeric> //std::iota
+#include <stdexcept>
+
+#include <adios2.h>
+
+#include <gtest/gtest.h>
+
```
chuckatkins: This is a good start for tests.  It would be even better to add 2 more, one that compresses natively with bzip2 api and decompresses with adios wrapper, then another that compresses with adios wrapper and decompresses with native bzip2 api.

chuckatkins:
@JasonRuonanWang @williamfgc See #220

examples/hello/bpBZip2Wrapper/CMakeLists.txt
williamfgc: I knew I was doing something wrong. Thanks @chuckatkins, I'll make the changes.

source/adios2/CMakeLists.txt
williamfgc: Is this the right way to do it?: 

if(ADIOS2_HAVE_BZip2)
  find_package(BZip2 REQUIRED)
  target_sources(adios2 PRIVATE transform/compress/CompressBZip2.cpp)
  target_link_libraries(adios2 PRIVATE BZip2::BZip2)
endif()

source/adios2/helper/adiosString.cpp
williamfgc: Good catch!

source/adios2/transform/compress/CompressBZip2.h
williamfgc: Good catch!

source/adios2/transform/compress/CompressBZip2.h
williamfgc: No need, since it's a final class.

source/adios2/transform/compress/CompressBZip2.h
williamfgc: I don't a good case scenario for that, this is a wrapper to the bzip2 library so it's a final class. Any new implementation, which would be orthogonal to bzip2, should extend Transform directly.

testing/adios2/transform/TestBZip2Wrapper.cpp
williamfgc: @chuckatkins Agree with the test...in that case, bzip2 becomes a public dependency for TestNativeBZip2.cpp?

source/adios2/transform/compress/CompressZfp.h
williamfgc: Good point, I think I am only passing m_DebugMode, but that can be easily rephrased. Will add a helper function and work on that.

testing/adios2/transform/CMakeLists.txt
williamfgc: Got it!

williamfgc:
Marking, this as WIP since there are more changes to be done. Thanks @JasonRuonanWang and @chuckatkins for looking into this.

williamfgc:
Yes, #220 is what we need.

source/adios2/transform/compress/CompressBZip2.h
chuckatkins: Even if this class is, the parent class isn't so if the destructor isn't virtual then the parent destructor won't get called.  There's also a compelling case to not make this class final.  Imagine a pbzip2 transform that derives from this and it's implementation calls the parent bzip2 in parallel on chunks of input data.

testing/adios2/transform/TestBZip2Wrapper.cpp
chuckatkins: Correct!  Although it doesn't need to be a separate test, just 2 other test cases for the current test.  Much in the same way the engine tests have test cases for adios2 -> adios2, native -> adios2, adios2 -> native.  Either way it doesn't need to be a blocker for this branch, just a to-do after it's merged.  If you don't want to deal with it in this branch that's fine, I can just add an issue for the missing tests after the merge.

source/adios2/transform/compress/CompressBZip2.h
chuckatkins: See previous comment on the virtual destructor.

source/adios2/transform/compress/CompressZfp.h
chuckatkins: Would it make more sense to have debug mode be passed to the base constructor and held as a data member in the base class instead of having to pass to every "apply" call?

source/adios2/transform/compress/CompressBZip2.h
williamfgc: Parent classes can never be final, hence the destructor of Transport is virtual, but marking anything virtual in a final class like CompressBZip2 has no effect. https://isocpp.org/wiki/faq/virtual-functions#virtual-dtors   PbZip2 already exists http://compression.ca/pbzip2/ and is orthogonal to this class as it's a library on its own. Now, the final keyword reflects today's intention (ADIOS.cpp will create a CompressBZip2), making this class a Base class will be as simple as removing the final keyword and adding virtual to the destructor and intended functions. I am not against your idea, but we should do that when we cross that bridge.

source/adios2/transform/compress/CompressBZip2.h
williamfgc: See my comment about it. We should mark it final as it's truly final today. We can always change that (it's literally replacing text) when an actual implementation exists for extending this class.

source/adios2/transform/compress/CompressZfp.h
williamfgc: @chuckatkins let me rephrase it as I confused myself, too. m_DebugMode is the only member variable used by these functions (not passed). The structure you describe is exactly how it is (m_DebugMode is a base class member). If I put these functions in a helper I will have to pass m_DebugMode  as const bool in the arguments.

testing/adios2/transform/TestBZip2Wrapper.cpp
williamfgc: Yes, @JasonRuonanWang needs this branch that's why I started it. In fact, ZFP is basically a refactoring from his zfpman. We are checking for original data recovery in the first test. But I agree we can can make it an adios2 vs native test.

source/adios2/transform/compress/CompressZfp.h
chuckatkins: If it makes sense to keep it as a member function that's fine.  It was more that if zfp header could be taken out of the public interface that'd be great.  If they make more sense as member functions with zfp types in the interface though then that's fine too.  It was a question to consider, not a 'needs changing'

williamfgc: @chuckatkins I agree, check that CompressBZip2.h and CompressZfp.h are not part of the public interface, I am following what we decided with Engine and its derived classes. They are polymorphic.

williamfgc:
I think it's ready to merge, unless you have more reviews/comments. I tried answering all the reviews above. Thanks @chuckatkins and @JasonRuonanWang

PR #221, opened 2017-07-31
Set debugMode to true by default.

pnorbert:
Let sophisticated users turn off the debugMode flag off. This change causes that examples/basics/value and examples/basics/globalArray to throw an exception. The examples are correct. The library needs to be fixed to handle the dimension definition API correctly to make the examples work again.

### DIFF
--- a/source/adios2/core/ADIOS.h
+++ b/source/adios2/core/ADIOS.h
@@ -45,7 +45,7 @@ public:
      * @param debugMode true: extra exception checks (recommended)
      */
     ADIOS(const std::string configFile, MPI_Comm mpiComm,
-          const bool debugMode = false);
+          const bool debugMode = true);
 
     /**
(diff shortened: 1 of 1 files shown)


chuckatkins:
What do you think about this being able to be set by an environment variable?  Something like:
```cpp
ADIOS(const std::string configFile, const bool debugMode = CheckEnvVarBoolWithDefault<true>("ADIOS2_DEBUG_MODE"));
```
where `CheckEnvVarBoolWithDefault` would need to be implemented as well.  That would allow users to not have to change thier code, just an environment varibale, to switch debug mode on or off.

eisenhauer:
I like it…

g

williamfgc:
+1

williamfgc:
@chuckatkins let's make that part of a separate PR....I like the idea as it changes debug mode at runtime using the cheapest available way (a const bool).

PR #223, opened 2017-07-31
fixed a size_t / uint64_t problem for macOS 10.12 and Clang 8.1.0

JasonRuonanWang:


JasonRuonanWang:
@chuckatkins I can change vector<size_t> to adios2::Dims if needed. But there has to be two sets of these variables, because some of the later code requires uint64_t, while some requires size_t.

examples/heatTransfer/read/heatRead_adios2.cpp
```diff
@@ -109,7 +111,7 @@ int main(int argc, char *argv[])
     double *T = new double[vT->m_AvailableSteps * readsize[0] * readsize[1]];
```
williamfgc: adios2::Dims is already std::vector<size_t>, just change std::vector<uint64_t> to adios2::Dims, no need to create a separate dimension vector.

JasonRuonanWang: The problem is that the printData function is used by heatRead_adios1 and heatRead_adios2 at the same time. If I change the printData to using adios2::Dims then I have to add two sets of these variables in heatRead_adios1 instead of adding them into heatRead_adios2.

JasonRuonanWang: @williamfgc I have pushed another commit doing so. Is this what I should do?

chuckatkins:
I think @williamfgc was also suggesting that by fixing the `printData` signature you no longer need the separate set of arrays and can just declare `readsize` and `offset` as `adios2::Dims` variables and only use those.

examples/heatTransfer/read/heatRead_adios2.cpp
williamfgc: You can always make PrintData<T> a template function that takes const T*....PrintData should accommodate to the examples, not the other way around.

JasonRuonanWang:
@chuckatkins I totally understand. But if I do so, then I am essentially moving the problem from heatRead_adios2 to heatRead_adios1. Because, firstly, heatRead_adios1 is based on ADIOS 1 API which requires uint64_t, and secondly, it uses the printData function, which is now changed to size_t. So whichever type I use, I have to keep two sets of these variables in either heatRead_adios1 or heatRead_adios2.

chuckatkins:
Ok, I understand the issue now in trying to marry the code base to work with adios1 and adios2 APIs.  I didn't quite see that as I was only looking at it in isolation.

williamfgc:
@JasonRuonanWang See my comment about making PrintData < T > a template function

examples/heatTransfer/read/heatRead_adios2.cpp
JasonRuonanWang: @williamfgc Yeah, I kind of know this. I am just trying to avoid making too many changes. Because this is not my code, if I make too much modification then it might break something else while fixing the problem I have now. Do you want me to close cancel this pull request? I can use Linux temporarily until someone fixes it.

williamfgc: Naaaah, template is fine...in fact, templates are supposed to be used in this kind of situations.  You won't break anything as the adios1 example will be intact. The template type will get resolved automatically. It's less invasive than switching from uint64_t to size_t

JasonRuonanWang:
@williamfgc Yeah, I kind of know this. I am just trying to avoid making too many changes. Because this is not my code, if I make too much modification then it might break something else while fixing the problem I have now. Do you want me to close cancel this pull request? I can use Linux temporarily until someone fixes it.

williamfgc:
I don't think @pnorbert would mind if @JasonRuonanWang changes PrintData to PrintData<T> to support uin64_t* and size_t* as that will solve an issue on Mac that we don't see on Linux.

JasonRuonanWang:
@williamfgc Then I need to move the printData function from .cpp to .h, or create another .tcc file. In the meantime, there is only one function in the .cpp file, so if I do so, I probable need to remove the .cpp file then. You see, there is a lot of things... which I am not sure any of you would mind.

williamfgc:
@JasonRuonanWang Then overload the function with size_t , it's a copy and paste thing.

JasonRuonanWang:
@williamfgc Done.

williamfgc:
Thanks @JasonRuonanWang  !

PR #224, opened 2017-07-31
Fix the variable definition initialization rules and add test for Def…

pnorbert:
…ineVariable.

Fixed and tested defining local values, joined arrays and global arrays with correct parameters.

### DIFF
--- a/source/adios2/core/Engine.inl
+++ b/source/adios2/core/Engine.inl
@@ -29,7 +29,7 @@ void Engine::Write(Variable<T> &variable, const T *values)
 {
     if (m_DebugMode)
     {
-        variable.CheckDims("in call to Write");
+        variable.CheckDimsBeforeWrite("Write(" + variable.m_Name + ")");
     }
 
     DoWrite(variable, values);
--- a/source/adios2/core/VariableBase.cpp
+++ b/source/adios2/core/VariableBase.cpp
@@ -168,14 +168,6 @@ void VariableBase::InitShapeType()
             }
             m_ShapeID = ShapeID::JoinedArray;
         }
-        else if (std::count(m_Shape.begin(), m_Shape.end(), JoinedDim) > 1)
-        {
-            throw std::invalid_argument(
-                "ERROR: variable can't have more than one "
-                "JoinedDim in shape argument, in call to "
-                "DefineVariable " +
-                m_Name + "\n");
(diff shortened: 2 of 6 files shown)


pnorbert:
Do not merge yet. Let me move some checks into CheckDims() instead of InitShapeType().

williamfgc:
@pnorbert mark it as WIP....let me know when you're ready

pnorbert:
Ready for merging

source/adios2/core/Engine.inl
```diff
@@ -29,7 +29,7 @@ void Engine::Write(Variable<T> &variable, const T *values)
 {
     if (m_DebugMode)
     {
-        variable.CheckDims("in call to Write");
+        variable.CheckDimsBeforeWrite("Write(" + variable.m_Name + ")");
```
williamfgc: @pnorbert This might create confusion since the user might rush to look for a Write(const string &name, T* ) signature, (e.g.  Write("myvar", var.data() ); )...I would keep it as "in call to Write" since the Variable.m_Name is already printed by the exception throw by either CheckDimsCommon or CheckDimsBeforeWrite. Still, this shouldn't stop the PR.

source/adios2/core/VariableBase.cpp
```diff
@@ -265,6 +257,12 @@ void VariableBase::InitShapeType()
     }
 
     /* Extra checks for invalid settings */
+    if (m_DebugMode)
+        CheckDimsCommon("DefineVariable(" + m_Name + ")");
```
williamfgc: Like in here, DefineVariable("myvar") is what the user needs to look for. This is a good debugging message as it's non-ambiguous.

source/adios2/core/VariableBase.h
```diff
@@ -137,9 +137,14 @@ class VariableBase
     /** Registered transforms */
     std::vector<TransformInfo> m_TransformsInfo;
 
+    /** Self-check dims according to type, called right after DefineVariable and
+     * SetSelection.
+     * @param hint extra debugging info for the exception */
+    void CheckDimsCommon(const std::string hint) const;
```
williamfgc: Nicely put common part

williamfgc:
@pnorbert Great job! and thanks for adding the test cases.

PR #227, opened 2017-08-02
templatize PrintData() so that it can work with all types

pnorbert:


pnorbert:
Jason please check if this builds on your Mac. Thanks.

PR #229, opened 2017-08-08
Initial C bindings

williamfgc:
Makefile compiles creating libadios2_c.so and hello example runs and creates bp file

To do:
CMake and Testing infrastructure (separate branches)
Verify struct adios2_Engine  forward declaration...Done!


williamfgc:
@JasonRuonanWang please try again.

chuckatkins:
Is there a compelling reason to make the c bindings optional vs always on?

chuckatkins:
Rather than a separate libadios2_c library, the c bindings can be built into libadios2 and just use a single library to contain both the C and C++ interfaces.

williamfgc:
@pnorbert @chuckatkins @JasonRuonanWang  OK, finally I resused @pnorbert and @chuckatkins  cmake infrastructure to figure out library, examples and testing for the C bindings. I updated the ReadMe.md adding the ADIOS2_USE_C to turn OFF and ON the C bindings. Please try it and run ctest. Thanks!

williamfgc:
@chuckatkins I have no preferences, as long as the user has a clear interface separation between C and C++. Many projects are C++ only...no need for the C bindings per-se, unless they decide to use the C API, which is a waste of features if the projects is C++. In fact, C is just a way to other languages, I am not aware of any scientific application using C, it's either C++ or Fortran.  Python is nicely covered directly with PyBind11. Your call, I have no preference. We'll have libadios2_f when we get to the Fortran bindings.

williamfgc:
@chuckatkins @pnorbert @JasonRuonanWang I am going to merge this PR if there is no more comments. If we decide to unify the library, that can be a separate branch/PR. This is an initial C bindings working implementation as the PR says. For now I want to avoid conflict with future branches touching the same files. Thanks!

PR #236, opened 2017-08-17
Fix misc mac issues

chuckatkins:
@guj @eisenhauer this should fix your mac build issues


eisenhauer:
This fixes my issues...

PR #242, opened 2017-08-24
C: Make C bindings always on and embedded into libadios2

chuckatkins:


chuckatkins:
Sorry for the big "mega" commit.  The various pieces needed to happen all at once so it wasn't clear how to separate them.

chuckatkins:
Just to verify, you can see the C bindings in teh libadios2.so here:

williamfgc:
Sounds good to me. Thanks for getting this done.  I like how the void* got replaced by structs. I tried doing it from adios2, but it's in a namespace

chuckatkins:
Yeah, I just made them forward declared structs with no implementation.  It works out because we only use them as opaque pointers so there's no actual need to have a full definition.  This preserves the type safety and makes sure you can't pass an adios2_ADIOS pointer in place of an adios2_IO pointer because they're not both just void*.

Issue #147, opened 2017-06-20
Unit tests don't run when MPI is enabled


chuckatkins:

chuckatkins:
Fixed

PR #246, opened 2017-08-28
Fix global array definitions in ADIOS1 test for 2D Arrays

HaochengLIU:
Close #246 .

Partially address #147 . I would fix it totally this week.


HaochengLIU:
@chuckatkins 
Would you please kindly review this MR? It address issue #246 .

PR #251, opened 2017-08-29
Reduce verbose debug info in CircleCI

chuckatkins:
@scottwittenburg looks like I had some leftover debug output in the environment modes CMake script.


scottwittenburg:
Ah, that explains all that extra output then.  Going back to -VV for the cmake verbosity in this case?

PR #252, opened 2017-08-30
Make all Bp engine tests able to run in parallel

### LINKED ISSUE #230: Path in bp file name fails
/path/to/file doesn't produce /path/to/file.bp.dir/file.bp.0

HaochengLIU:
Now BPWriteProfilingJSON would use a BP filename with path. Close #230 

BPWriteReadAttribute: Since collective meta generation has not landed yet, I have to use
same attribute for all processes. Ideally attribute per process
should be unique to each other for testing purpose.

### DIFF
--- a/testing/adios2/engine/bp/CMakeLists.txt
+++ b/testing/adios2/engine/bp/CMakeLists.txt
@@ -4,38 +4,19 @@
 #------------------------------------------------------------------------------#
 
 # MPI versions of the test are not properly implemented at the moment
-if(NOT ADIOS2_HAVE_MPI)
-  
-  if(ADIOS2_HAVE_ADIOS1)
-    add_executable(TestBPWriteRead TestBPWriteRead.cpp)
-    target_link_libraries(TestBPWriteRead adios2 gtest adios1::adios)
 
-    add_executable(TestBPWriteReadAttributes TestBPWriteReadAttributes.cpp)
-    target_link_libraries(TestBPWriteReadAttributes adios2 gtest adios1::adios)
-  
-    add_executable(TestBPWriteReadstdio TestBPWriteReadstdio.cpp)
-    target_link_libraries(TestBPWriteReadstdio adios2 gtest adios1::adios)
-
-    add_executable(TestBPWriteReadfstream TestBPWriteReadfstream.cpp)
-    target_link_libraries(TestBPWriteReadfstream adios2 gtest adios1::adios)
-    
-    gtest_add_tests(TARGET TestBPWriteRead)
-    gtest_add_tests(TARGET TestBPWriteReadAttributes)
-    gtest_add_tests(TARGET TestBPWriteReadstdio)
-    gtest_add_tests(TARGET TestBPWriteReadfstream)
-  endif()
-    
-  add_executable(TestBPWriteProfilingJSON TestBPWriteProfilingJSON.cpp)
-  target_link_libraries(TestBPWriteProfilingJSON
-    adios2 gtest gtest_main NLohmannJson
-  )
-  
-  gtest_add_tests(TARGET TestBPWriteProfilingJSON)
-endif()
-
-# Testing CMake code
 add_executable(TestBPWriteRead TestBPWriteRead.cpp)
 target_link_libraries(TestBPWriteRead adios2 gtest adios1::adios)
+
+#add_executable(TestBPWriteReadAttributes TestBPWriteReadAttributes.cpp)
+#target_link_libraries(TestBPWriteReadAttributes adios2 gtest adios1::adios)
+
+#add_executable(TestBPWriteReadstdio TestBPWriteReadstdio.cpp)
+#target_link_libraries(TestBPWriteReadstdio adios2 gtest adios1::adios)
+
+#add_executable(TestBPWriteReadfstream TestBPWriteReadfstream.cpp)
+#target_link_libraries(TestBPWriteReadfstream adios2 gtest adios1::adios)
+
 if(ADIOS2_HAVE_MPI)
   target_include_directories(TestBPWriteRead PRIVATE ${MPI_C_INCLUDE_PATH})
   target_link_libraries(TestBPWriteRead ${MPI_C_LIBRARIES})
@@ -46,3 +27,24 @@ if(ADIOS2_HAVE_MPI)
 endif()
 
 gtest_add_tests(TARGET TestBPWriteRead ${extra_test_args})
+#gtest_add_tests(TARGET TestBPWriteReadAttributes)
+#gtest_add_tests(TARGET TestBPWriteReadstdio)
+#gtest_add_tests(TARGET TestBPWriteReadfstream)
+
+add_executable(TestBPWriteProfilingJSON TestBPWriteProfilingJSON.cpp)
+target_link_libraries(TestBPWriteProfilingJSON
+  adios2 gtest gtest_main NLohmannJson
+)
+
+if(ADIOS2_HAVE_MPI)
+  target_include_directories(TestBPWriteProfilingJSON PRIVATE ${MPI_C_INCLUDE_PATH})
+  target_link_libraries(TestBPWriteProfilingJSON ${MPI_C_LIBRARIES})
+  set(extra_test_argsJSON
+    EXEC_WRAPPER
+      ${MPIEXEC} ${MPIEXEC_NUMPROC_FLAG} ${MPIEXEC_MAX_NUMPROCS}
+  )
+endif()
+
+gtest_add_tests(TARGET TestBPWriteProfilingJSON ${extra_test_argsJSON})
+
+
--- a/testing/adios2/engine/bp/CMakeLists.txt
+++ b/testing/adios2/engine/bp/CMakeLists.txt
@@ -14,8 +14,13 @@ target_link_libraries(TestBPWriteReadAttributes adios2 gtest adios1::adios)
 add_executable(TestBPWriteReadstdio TestBPWriteReadstdio.cpp)
 target_link_libraries(TestBPWriteReadstdio adios2 gtest adios1::adios)
 
-#add_executable(TestBPWriteReadfstream TestBPWriteReadfstream.cpp)
-#target_link_libraries(TestBPWriteReadfstream adios2 gtest adios1::adios)
+add_executable(TestBPWriteReadfstream TestBPWriteReadfstream.cpp)
+target_link_libraries(TestBPWriteReadfstream adios2 gtest adios1::adios)
+
+add_executable(TestBPWriteProfilingJSON TestBPWriteProfilingJSON.cpp)
+target_link_libraries(TestBPWriteProfilingJSON
+  adios2 gtest NLohmannJson
+)
 
 if(ADIOS2_HAVE_MPI)
   target_include_directories(TestBPWriteRead PRIVATE ${MPI_C_INCLUDE_PATH})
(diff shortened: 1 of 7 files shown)


testing/adios2/engine/bp/CMakeLists.txt
```diff
@@ -4,31 +4,57 @@
...
-  
-  gtest_add_tests(TARGET TestBPWriteProfilingJSON)
 endif()
+
+gtest_add_tests(TARGET TestBPWriteRead ${extra_test_args})
+gtest_add_tests(TARGET TestBPWriteReadAttributes ${extra_test_args})
+gtest_add_tests(TARGET TestBPWriteReadstdio ${extra_test_args})
+gtest_add_tests(TARGET TestBPWriteReadfstream ${extra_test_args})
+
+add_executable(TestBPWriteProfilingJSON TestBPWriteProfilingJSON.cpp)
```
chuckatkins: Can the JSON test be grouped with all the others?

testing/adios2/engine/bp/TestBPWriteRead.cpp
```diff
@@ -102,83 +151,105 @@ TEST_F(BPWriteReadTest, ADIOS2BPWriteADIOS1Read1D8)
...
                                "verbose=3");
 
         // Open the file for reading
-        ADIOS_FILE *f =
-            adios_read_open_file((fname + ".dir/" + fname + ".0").c_str(),
-                                 ADIOS_READ_METHOD_BP, MPI_COMM_WORLD);
+        std::string index = std::to_string(mpiRank);
+        ADIOS_FILE *f = adios_read_open_file(
+            (fname + ".dir/" + fname + "." + index).c_str(),
+            ADIOS_READ_METHOD_BP, MPI_COMM_WORLD);
```
chuckatkins: @williamfgc @pnorber ADIOS1 can process the parallel files from ADIOS2 right?  And if that's the case, ADIOS1 is reading back each rank's file independently, not in parallel.  In which case, should we initialize ADIOS1 with MPI_COMM_SELF instead?

testing/adios2/engine/bp/CMakeLists.txt
HaochengLIU: Done.

pnorbert:
Yes, we should. The open statement is collective operation in adios1.

testing/adios2/engine/bp/TestBPWriteRead.cpp
HaochengLIU: This test is about asking multiple processes write data into a 1D array using `BPFileWriter`. Ex. process 0 would write index 0\~5, process 1 would write index 6\~9 etc.. The result would be a 1D array with size as mpiSize*5.

When reading the data back, if I'm using`MPI_COMM_SELF` then the bounding box selection for each process would be [mpiRank * 5, 5](start and count). With `MPI_COMM_WORLD` the bounding box selection would be [0,5]. Here I'm really confused... Why `MPI_COMM_SELF` would need a global starting index while `MPI_COMM_WORLD` would use a local starting index?

pnorbert:
This test I believe is invalid, because we are not supposed to read from a
subfile directly, when a global array is defined across the subfiles. The
writer should produce the .bp file with all metadata, which one opens with
MPI_COMM_WORLD.

Nevertheless, in the current situation, with either communicator, the
correct selection should be [mpiRank * 5, 5]. [0.5] should result in error
on all non-zero rank.

HaochengLIU:
The failing test is about stream error from [ADIOS2BPWriteProfilingJSON](https://open.cdash.org/testDetails.php?test=578969710&build=5042101).
@chuckatkins We discussed about it in the morning. Let's see if it would fail this time.

chuckatkins:
lgtm but I'd like @pnorbert or @williamfgc to verify that adios1 is being used correctly for the read verification

testing/adios2/engine/bp/CMakeLists.txt
```diff
@@ -4,31 +4,51 @@
...
+
+add_executable(TestBPWriteReadstdio TestBPWriteReadstdio.cpp)
+target_link_libraries(TestBPWriteReadstdio adios2 gtest adios1::adios)
+
+add_executable(TestBPWriteReadfstream TestBPWriteReadfstream.cpp)
+target_link_libraries(TestBPWriteReadfstream adios2 gtest adios1::adios)
+
+add_executable(TestBPWriteProfilingJSON TestBPWriteProfilingJSON.cpp)
+target_link_libraries(TestBPWriteProfilingJSON
+  adios2 gtest gtest_main NLohmannJson
```
chuckatkins: You can remove gtest_main since we define our own main loop

testing/adios2/engine/bp/CMakeLists.txt
```diff
@@ -4,31 +4,51 @@
...
-  
-  gtest_add_tests(TARGET TestBPWriteProfilingJSON)
 endif()
+
+gtest_add_tests(TARGET TestBPWriteRead ${extra_test_args})
+gtest_add_tests(TARGET TestBPWriteReadAttributes ${extra_test_args})
+gtest_add_tests(TARGET TestBPWriteReadstdio ${extra_test_args})
+gtest_add_tests(TARGET TestBPWriteReadfstream ${extra_test_args})
+gtest_add_tests(TARGET TestBPWriteProfilingJSON ${extra_test_args})
+
```
chuckatkins: Remove extra trailing whitespace

testing/adios2/engine/bp/TestBPWriteProfilingJSON.cpp
```diff
@@ -17,6 +17,7 @@
 #include <stdexcept>
 
 #include <adios2.h>
+#include "adios2/ADIOSTypes.h"
```
chuckatkins: No need to include this directly since it's already included by `adios2.h`

testing/adios2/engine/bp/TestBPWriteProfilingJSON.cpp
```diff
@@ -17,6 +17,7 @@
...
 
 #include <gtest/gtest.h>
 #include <json.hpp> //This fails to be included
```
chuckatkins: > `  //This fails to be included`

I'm assuming this doesn't fail to be included anymore right?

testing/adios2/engine/bp/TestBPWriteProfilingJSON.cpp
```diff
@@ -40,47 +41,73 @@ class BPWriteProfilingJSONTest : public ::testing::Test
 // ADIOS2 write, native ADIOS1 read
 TEST_F(BPWriteProfilingJSONTest, ADIOS2BPWriteProfilingJSON)
 {
-    std::string fname = "ADIOS2BPWriteProfilingJSON.bp";
+    // Use a relative path + file name to test path in file name capability
+    std::string fname;
+    fname = "foo" + std::string(1, adios2::PathSeparator) + "ADIOS2BPWriteProfilingJSON.bp";
```
chuckatkins: @williamfgc I don't think we need the platform-dependent path separator.  The fopen calls should work correctly, even on windows, with foward-slash separators.

testing/adios2/engine/bp/TestBPWriteReadAttributes.cpp
```diff
@@ -65,93 +107,97 @@ TEST_F(BPWriteReadAttributeTest, ADIOS2BPWriteADIOS1ReadSingleTypes)
...
         adios_read_init_method(ADIOS_READ_METHOD_BP, MPI_COMM_WORLD,
                                "verbose=3");
 
         // Open the file for reading
-        ADIOS_FILE *f =
-            adios_read_open_file((fname + ".dir/" + fname + ".0").c_str(),
-                                 ADIOS_READ_METHOD_BP, MPI_COMM_WORLD);
+        ADIOS_FILE *f = adios_read_open_file(
+            (fname + ".dir/" + fname + "." + mpiRankString).c_str(),
+            ADIOS_READ_METHOD_BP, MPI_COMM_WORLD);
```
chuckatkins: `MPI_COMM_WORLD` -> `MPI_COMM_SELF`

testing/adios2/engine/bp/TestBPWriteRead.cpp
```diff
@@ -544,93 +742,118 @@ TEST_F(BPWriteReadTest, ADIOS2BPWriteADIOS1Read2D4x2)
...
 
-// Read test data using ADIOS1
-#ifdef ADIOS2_HAVE_MPI
-    // Read everything from rank 0
-    int rank;
-    MPI_Comm_rank();
-    if (rank == 0)
-#endif
     {
         adios_read_init_method(ADIOS_READ_METHOD_BP, MPI_COMM_WORLD,
```
chuckatkins: `MPI_COMM_WORLD` -> `MPI_COMM_SELF`

testing/adios2/engine/bp/TestBPWriteRead.cpp
```diff
@@ -318,93 +440,118 @@ TEST_F(BPWriteReadTest, ADIOS2BPWriteADIOS1Read2D2x4)
...
```
chuckatkins: `MPI_COMM_WORLD` -> `MPI_COMM_SELF`

testing/adios2/engine/bp/TestBPWriteRead.cpp
```diff
@@ -102,83 +151,105 @@ TEST_F(BPWriteReadTest, ADIOS2BPWriteADIOS1Read1D8)
...
```
chuckatkins: `MPI_COMM_WORLD` -> `MPI_COMM_SELF`

testing/adios2/engine/bp/TestBPWriteReadAttributes.cpp
```diff
@@ -65,93 +107,97 @@ TEST_F(BPWriteReadAttributeTest, ADIOS2BPWriteADIOS1ReadSingleTypes)
...
 
-// Read test data using ADIOS1
-#ifdef ADIOS2_HAVE_MPI
-    // Read everything from rank 0
-    int rank;
-    MPI_Comm_rank();
-    if (rank == 0)
-#endif
     {
         adios_read_init_method(ADIOS_READ_METHOD_BP, MPI_COMM_WORLD,
```
chuckatkins: `MPI_COMM_WORLD` -> `MPI_COMM_SELF`

pnorbert:
A question. Can we call the bpmeta executable from the test to generate the
.bp global metadata file and do the read test on it, with comm world?
Eventually adios2 will create this file and then we don't need to rewrite
the tests, just remove the bpmeta call.

testing/adios2/engine/bp/CMakeLists.txt
HaochengLIU: Done.

testing/adios2/engine/bp/CMakeLists.txt
HaochengLIU: Done.

testing/adios2/engine/bp/TestBPWriteProfilingJSON.cpp
HaochengLIU: Done.

testing/adios2/engine/bp/TestBPWriteReadAttributes.cpp
HaochengLIU: Done.

testing/adios2/engine/bp/TestBPWriteReadAttributes.cpp
HaochengLIU: May I ask does github has `close discussion` functionality like gitlab?

testing/adios2/engine/bp/TestBPWriteProfilingJSON.cpp
HaochengLIU: Reference: https://msdn.microsoft.com/en-us/library/yeby3zcb.aspx

HaochengLIU:
`MPI_COMM_WORLD` has all been replaced with `MPI_COMMON_SELF` whenever needed in reading.

williamfgc:
@chuckatkins @HaochengLIU  about the path separator, please implement what you think makes more sense. This was just a quick thing to test on my Windows VM.

PR #253, opened 2017-08-30
Finish image buildout

scottwittenburg:


scripts/circle/Dockerfile
```diff
@@ -1,36 +1,64 @@
...
+      gcc gcc-c++ gcc-gfortran \
+      git make autoconf automake libtool\
+      bzip2-devel zeromq-devel hdf5-devel \
+      python-devel numpy
+
+# Install gcc7 OpenHPC packages
+RUN yum -y install \
+      lmod-ohpc gnu7-compilers-ohpc openmpi-gnu7-ohpc phdf5-gnu7-openmpi-ohpc \
+      python-numpy-gnu7-ohpc
+
```
chuckatkins: It might not matter much but it's probably worth doing a `yum clean all` after all packages are installed to remove all the cached metadata.  May save a few MB

scottwittenburg: Good idea, I'll add that.

scripts/circle/Dockerfile
```diff
@@ -1,36 +1,64 @@
...
-        ohpc-base \
-        openmpi-gnu7-ohpc \
-        phdf5-gnu7-openmpi-ohpc \
-        netcdf-gnu7-openmpi-ohpc
+FROM centos:centos7
 
-WORKDIR /home/adios2
+# Install extra repos
+RUN yum -y install epel-release https://github.com/openhpc/ohpc/releases/download/v1.3.GA/ohpc-release-1.3-1.el7.x86_64.rpm && \
+    yum -y upgrade
```
scottwittenburg: Is this what we want instead of `yum -y update`?  The dockerfile "best practices" documentation says to avoid upgrade, though they give the `apt-get` example rather than the `yum` one.

https://docs.docker.com/engine/userguide/eng-image/dockerfile_best-practices/#run

chuckatkins: We can drop it entirely.  apt and yum work differently in this regard in that `apt-get update` is first required to update repo metadata before `apt-get install` will work.  `yum` on the other hand will automatically update it's metadata as needed.  The `yum upgrade` step is the equivalent of `apt-get dist-upgrade` and can be safely removed.

scottwittenburg: Ok, I'll remove that.  Thanks.

scottwittenburg:
@chuckatkins I think this is ready.  I'll start a new branch from the tail of this one, and let you push the merge button if you're ok with it.

PR #267, opened 2017-09-12
Add upstream FindMPI

chuckatkins:
Upstream CMake has a major re-write of FindMPI that should fix a wide array of issues people have been encountering.  This brings in a local copy of the new upstream FindMPI and ports everything to use the new import targets instead of variables.


chuckatkins:
@eisenhauer I believe this branch will address the MPI  issues on Theta

PR #268, opened 2017-09-12
TESTING: Make binding tests able to run in parallel

HaochengLIU:
Use google test for C binding tests and simplify the logic.
In the future, we should also add validation for binding tests.


HaochengLIU:
@chuckatkins 
Would you please kindly review this MR?

testing/adios2/bindings/C/TestBPWriteTypes.cpp
```diff
@@ -8,20 +8,32 @@
  *      Author: wgodoy
```
chuckatkins: Author?

testing/adios2/bindings/C/TestBPWriteTypes.cpp
```diff
@@ -8,20 +8,32 @@
...
  */
 
+#ifdef ADIOS2_HAVE_MPI
```
chuckatkins: The `#ifdef` should be after including `adios2_c.h` since none of the `ADIOS2Config.h` that actually defines the options isn't included until then.

testing/adios2/bindings/python/CMakeLists.txt
```diff
@@ -4,9 +4,12 @@
 #------------------------------------------------------------------------------#
 
 if(NOT ADIOS2_HAVE_MPI)
-    python_add_test(PythonBPWrite TestBPWriteTypes_nompi.py)
+    python_add_test(TARGET PythonBPWrite SCRIPT TestBPWriteTypes_nompi.py)
```
chuckatkins: The `TARGET` option should be `NAME` since what your passing is the test name, not a CMake target.  The gtest functions actually used the CMake target information, but in this case, it's just a test name.

cmake/ADIOSFunctions.cmake
```diff
@@ -22,11 +22,20 @@ function(message_pad msg out_len out_msg)
...
+function(python_add_test)
+  set(options)
+  set(oneValueArgs
+      TARGET
   )
-  set_property(TEST ${name} PROPERTY
+  set(multiValueArgs
+      EXEC_WRAPPER
+      SCRIPT
+  )
```
chuckatkins: Since there's only a few of these they can easily fit on a single line.  The multi-line breakout is really to deal with long lines > 80c.

```cmake
set(options)
set(oneValueArgs TARGET)
set(multiValueArgs EXEC_WRAPPER SCRIPT)
```

That being said, is `SCRIPT` actually a multi-value argument?  What would happen if you passed several scripts here?  it seems things would break on multiple levels.  This would likely be better suited to dropping the `TARGET` option entirely and just having the multi-value `EXEC_WRAPPER` and `SCRIPTS` arguments.  Then loop through and add a test for each one by extracting the test name from filename by dropping the extension (see [get_filename_component](https://cmake.org/cmake/help/v3.6/command/get_filename_component.html)).

testing/adios2/bindings/C/TestBPWriteTypes.cpp
HaochengLIU: Done.

testing/adios2/bindings/C/TestBPWriteTypes.cpp
HaochengLIU: Done.

HaochengLIU:
@chuckatkins 
Would you please kindly review the updated code?

PR #269, opened 2017-09-12
Get AppVeyor builds working

scottwittenburg:


scottwittenburg:
@HaochengLIU Can you take a look at the changes to `testing/adios2/interface/`, and @chuckatkins can you look at the changes to `testing/adios2/xml/`?  

I ran into trouble turning on Python under windows, so I was thinking I'd leave it off for now.  As for the time differences we see between the two appveyor builds over on CDash, I have had some initial indication that the problem may need to be solved within CDash itself.

chuckatkins:
rather than create a separate XML file for Windows, just specifying the fstream library should work for all platforms.  just change the current config1.xml so it will work with all platforms.

HaochengLIU:
LTGM. Thanks scot! Once it's merged I would rebase my MR.

scottwittenburg:
Thanks for the feedback @chuckatkins, that does make more sense.  So unless there are any other requests for changes, or other comments, I think we could merge this.  I see a green merge button, so that make me  think I have the power to merge :wink:

PR #275, opened 2017-09-28
added Read() to Transport class

JasonRuonanWang:


williamfgc:
@JasonRuonanWang Thanks for adding this. This is fine for now if this allows you to continue your work. I am planning to add Read( char* buffer, size_t size, size_t start = std::numeric_limits<size_t>() ), that way start should demote to the current position of the stream if not passed. It's will be the same for Write.

williamfgc:
@chuckatkins I'll merge this, I will then make the changes mentioned above.

PR #280, opened 2017-10-10
Fix zmq cmake config

chuckatkins:


chuckatkins:
@JasonRuonanWang Please check that this addresses your issue

PR #316, opened 2017-11-23
Sst integration

eisenhauer:
OK, trying again.  This time everything also builds on Titan with the evpath configuration that we've been using for SC demos.  (It might have a problem if you have adios1 enabled because both of those pull in evpath and potentially result in duplicate dependency arguments.  I turned off adios1 for testing.)


pnorbert:
This builds on my virtual machine as well. It builds together with ADIOS1 because we just use a lean adios1 build without any dependencies.

PR #326, opened 2017-12-02
Refactored heat transfer example. A config file is now command line a…

pnorbert:
…rgument. Reader's decomposition is also command line argument.

This example works with BPFileWriter/BPFileReader except for no end-of-stream error is occuring. 
Reader only deals with array T, gndx and gndy values are not read in. 
Reader can be run with arbitrary decomposition (DataMan requires to run it with the same decomp as the writer). 
Config file is also given in command line, not baked in, so we can have a separate config prepared for each engine.


williamfgc:
@guj  BeginStep is something we must implement for each engine. The return type just reflect checking if we reached the final step.

PR #351, opened 2017-12-12
python example for heat transfer

ORNLChu:
Found the missing of attribute "InConfigFile" in Python binding class "adios2::IO"


williamfgc:
@ORNLChu this need to be functional before we can do a pull request

PR #354, opened 2017-12-15
 heat transfer example worked with DataManWriter with callback based DataManReader

JasonRuonanWang:


pnorbert:
Why do you have a separate heatTransfer_dataman executable with IO_dataman.cpp source? Does DataMan not work with the IO_adios2.cpp implementation?

JasonRuonanWang:
@pnorbert Oh. I thought that's the idea because I saw the POSIX, HDF5 and ADIOS1 ones. I can change it back. It should work with the adios2 example.

PR #359, opened 2017-12-19
Fix warnings from python and GNU headers.

chuckatkins:
Due to some squireliness of how python headers work, we need to include
any python binding headers first, before any C library headers.  This is
contrary to the projects coding and style guidelines but necessary
since python.h will otherwise re-define core system feature macros.


chuckatkins:
@williamfgc just a heads up that this is explicitly against our guidelines for header organization but the commit should explain why

PR #362, opened 2017-12-23
Preliminary FFS-based MxN data transfer for global arrays

eisenhauer:
works with heatTransfer.

### DIFF
--- a/source/adios2/engine/sst/SstReader.cpp
+++ b/source/adios2/engine/sst/SstReader.cpp
@@ -8,6 +8,7 @@
  *      Author: Greg Eisenhauer
  */
 
+#include "adios2/helper/adiosFunctions.h"
 #include <cstring>
 #include <string>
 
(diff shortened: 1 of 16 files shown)


pnorbert:
Same typename trick is required on line 61
/home/adios/work/worktree/ADIOS2/release/source/adios2/engine/sst/SstReader.cpp: In lambda function:
/home/adios/work/worktree/ADIOS2/release/source/adios2/engine/sst/SstReader.cpp:61:31: error: ‘Reader’ was not declared in this scope
         SstReader::SstReader *Reader =

eisenhauer:
Also added to line 61.

source/adios2/engine/sst/SstReader.cpp
```diff
@@ -58,8 +58,8 @@ SstReader::SstReader(IO &io, const std::string &name, const Mode mode,
         std::vector<size_t> VecStart;
         std::vector<size_t> VecCount;
         std::string Type(type);
-        SstReader::SstReader *Reader =
-            reinterpret_cast<SstReader::SstReader *>(reader);
+        typename SstReader::SstReader *Reader =
+            reinterpret_cast<typename SstReader::SstReader *>(reader);
```
chuckatkins: Why would this need `typename` when there are no template arguments.

pnorbert: My Ubuntu 16.04, g++ 5.4.0 did not like that line. ```typename``` fixed it. Is there a better fix?

chuckatkins:
Most recent push was a rebase on current release

Issue #379, opened 2018-01-05
BPFile writer aborting on Titan


pnorbert:
$ aprun -n 1 ./heatTransfer_write_adios2 ./heat_bpfile.xml heat.bp 1 1 5 6 1 1
Step 1:
Total runtime = 0.00466585s

It segfaults when gathering global metadata. In this example, the subfile is readable, but at larger runs, they are also invalid BP files somehow. 

(gdb) bt
#0  0x0000000000000000 in ?? ()
#1  0x00000000004f3f6c in __gthread_equal (__t1=0, __t2=0) at /opt/gcc/4.9.3/snos/include/g++/x86_64-suse-linux/bits/gthr-default.h:680
#2  0x00000000004fcffc in std::operator== (__x=..., __y=...) at /opt/gcc/4.9.3/snos/include/g++/thread:84
#3  0x00000000004fd0e9 in std::thread::joinable (this=0xe16ea0) at /opt/gcc/4.9.3/snos/include/g++/thread:162
...
#20 0x00000000004f81b3 in adios2::format::BP3Serializer::DeserializeIndicesPerRankThreads (this=0xe11790, serialized=...)
    at /ccs/proj/e2e/pnorbert/ADIOS/ADIOS2/source/adios2/toolkit/format/bp3/BP3Serializer.cpp:781
#21 0x00000000004f7a2f in adios2::format::BP3Serializer::AggregateMergeIndex (this=0xe11790, indices=...)
    at /ccs/proj/e2e/pnorbert/ADIOS/ADIOS2/source/adios2/toolkit/format/bp3/BP3Serializer.cpp:652
#22 0x00000000004f5548 in adios2::format::BP3Serializer::AggregateCollectiveMetadata (this=0xe11790)
    at /ccs/proj/e2e/pnorbert/ADIOS/ADIOS2/source/adios2/toolkit/format/bp3/BP3Serializer.cpp:234
#23 0x00000000004d2109 in adios2::BPFileWriter::WriteCollectiveMetadataFile (this=0xe11750)
    at /ccs/proj/e2e/pnorbert/ADIOS/ADIOS2/source/adios2/engine/bp/BPFileWriter.cpp:200

williamfgc:
@pnorbert let me know if this fixes the issue. As far as I remember, I needed to pass 2 additional parameters on Titan for using threads, number of NUMA threads and MPI processes per node to avoid thread overflow on a single node (might be the case here).

pnorbert:
[...]

PR #380, opened 2018-01-06
Issue #379 async not call if threads = 1

williamfgc:


pnorbert:
Thanks for fixing this. It also broke the insitumpi engine and I assume
then the dataman engine as well.

PR #381, opened 2018-01-06
Issue #379 on Read side, no async if threads=1

williamfgc:


pnorbert:
Read works on Titan with this fix.

PR #385, opened 2018-01-10
[WIP] Add RDMA SST DP

philip-davis:
Adds RDMA-based data plane to SST. This is implemented using libfabric. This adds an external dependency on libfabric.

Caveats:
-Working only for the sockets provider so far.
-Uses CManager from EVPath for out-of-band signaling of read requests.


eisenhauer:
Philip and I chatted at the end of the AHM.  I think this is not the time to merge this.  There are really two parts here: adding dependencies for the build stuff and committing the non-dummy data plane.  The latter works over the libfabric sockets provider, but not over any RDMA networks yet,  so it doesn't yet add capabilities and might interfere with testing.  The CMake stuff might change with Chuck's work, and further, I think we'll need a mechanism to enables/disables this based not only on the presence of the libfabric library, but also what providers are available.  (Because we've seen that libfabric RDMA code is very sensitive to the actual underlying provider and likely won't run on anything it hasn't been vetted-on/ported-to.)  So, best to close this PR for now.

philip-davis:
In anticipation of the milestone, I am reopening this PR.

eisenhauer:
@chuckatkins The plan, when this is integrated, is to have the initial default to *not* use this, but to use the EVPath-based data transport that is already in play.  That way it doesn't break anything in regression tests (that isn't already broken).  Because of that, my principal concerns about integrating this would be that the dependencies are handled reasonably.  I expect that the problem would be that this depends upon libfabric, which itself may have different dependencies on different platforms.  It's an autoconf-based project, so dealing with those automagically may be an issue.  Anyhow, tagging you here Chuck as our dependencies guy.  Also, @philip-davis we may want the default data plane to be platform dependent, at least defaulting to the WAN/EVPath transport if EVPath is found but libfabric isn't found, etc.  Just need to munge the CMake stuff so that all this happens.  (And maybe it already is here, I haven't looked yet.)

PR #392, opened 2018-01-11
A test for insitumpi engine

pnorbert:
Parameterized, parallel test.


eisenhauer:
So, this is in an engine-specific testing directory, but it should be a regression test that the SST engine (or any streaming engine that supports MxN) would pass too.  Should we be thinking of a set of more generic streaming engine tests that we run with all appropriate engines?

eisenhauer:
I did a quick test, copying this and changing all the InsituMPI text to read Sst and the test succeeds...

pnorbert:
DO NOT MERGE. Another PR will bring a staging test common for SST and InSituMPI

PR #394, opened 2018-01-11
WIP: Test staging common

pnorbert:
This test currently aborts when running with the SST engine. It works fine with InSituMPI.


pnorbert:
Now the test checks the received values and this breaks the SST runs:
The very first value in the array is 0, the next one is 0.001, but on the receiver side the very first value of the array will be the 0.001.

pnorbert:
@chuckatkins Chuck, can you please look at this build? CircleCI timeouts on the staging-common/TestStagingMPMD test in the gcc-openmpi build. This test has 4 separate cases. Even the short ones (SingleStep and MultiStep) timeout at 300 seconds. I can run those under 10s on my VM with 2 CPU cores, forcing to run with 16 MPI tasks to have the full test (12 params). The other two (SlowWriter, SlowReader) run longer, right now under 40 seconds on my VM.

chuckatkins:
@pnorbert try rebasing this on master now

PR #395, opened 2018-01-12
SZ compression

jychoi-hpc:
I finished adding SZ compression as an operator in Adios2. This includes
- SZ cmake find module
- SZ operator
- An example

### DIFF
--- a/cmake/FindSZ.cmake
+++ b/cmake/FindSZ.cmake
@@ -1,0 +1,60 @@
+#------------------------------------------------------------------------------#
+# Distributed under the OSI-approved Apache License, Version 2.0.  See
+# accompanying file Copyright.txt for details.
+#------------------------------------------------------------------------------#
+#
+# FindSZ
+# -----------
+#
+# Try to find the SZ library
+#
+# This module defines the following variables:
+#
+#   SZ_FOUND        - System has SZ
+#   SZ_INCLUDE_DIRS - The SZ include directory
+#   SZ_LIBRARIES    - Link these to use SZ
+#
+# and the following imported targets:
+#   SZ::SZ - The SZ compression library target
+#
+# You can also set the following variable to help guide the search:
+#   SZ_ROOT - The install prefix for SZ containing the
+#              include and lib folders
+#              Note: this can be set as a CMake variable or an
+#                    environment variable.  If specified as a CMake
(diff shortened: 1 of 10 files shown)


cmake/FindSZ.cmake
```diff
@@ -0,0 +1,60 @@
...
+    set(SZ_INCLUDE_OPTS HINTS ${SZ_ROOT}/include NO_DEFAULT_PATHS)
+    set(SZ_LIBRARY_OPTS
+      HINTS ${SZ_ROOT}/lib ${SZ_ROOT}/lib64
+      NO_DEFAULT_PATHS
+    )
+  endif()
+
+  find_path(SZ_INCLUDE_DIR sz.h ${SZ_INCLUDE_OPTS})
+  find_library(SZ_LIBRARY NAMES SZ ${SZ_LIBRARY_OPTS})
+  find_library(ZLIB_LIBRARY NAMES zlib ${SZ_LIBRARY_OPTS})
```
chuckatkins: Is zlib also a dependecy?  If so this should be dropped and instead added via find_package(zlib)

jychoi-hpc:
SZ distribute their library together with ZLIB. I will check if we can use different version of ZLIB installed in a system.

Thanks,
Jong

-- 
Jong Youl Choi
Scientific Data Group
Computer Science and Math Division
Oak Ridge National Laboratory
Homepage: http://www.ornl.gov/~jyc/

williamfgc:
@jychoi-hpc thanks for the great work! To pass the travis-ci check we use the clang-format (v3.8 recommended on Mac and llvm style). It's just a binary that need to be applied to your files, see our document: https://github.com/ornladios/ADIOS2/wiki/Clang-Format-Text-Style Looking forward to SZ support. Thanks!

jychoi-hpc:
How can I get this "clang-format" installed on Mac? Can I install with macport or homebrew?

cmake/FindSZ.cmake
jychoi-hpc: After a bit of investigation, it looks like SZ uses their own zlib library and include in their distribution along with SZ. Then, if this is the case, we don't need to use another "find_package(zlib)", right?

eisenhauer:
http://brewformulas.org/ClangFormat

Issue #396, opened 2018-01-12
GathervVectors() does not work when data is null everywhere


pnorbert:
source/adios2/helper/adiosMPIFunctions.inl:
template <class T>
void GathervVectors(const std::vector<T> &in, std::vector<T> &out,
                    size_t &position, MPI_Comm mpiComm,
                    const int rankDestination)

rank 0 returns without entering MPI_Gatherv() if the total size to be collected is 0, leaving the rest of processes hanging in the call. 

The bug only presents itself on >128 processes on Titan. Any example without attributes would show the bug.

pnorbert:
Bug was fixed with an extra MPI_Bcast(). However, that is not the scalable fix, so this needs to be re-examined. Also multi-threading with this operation was not tested yet.

PR #398, opened 2018-01-12
Issue #396 extra check for empty GathervVectors

williamfgc:


williamfgc:
@pnorbert can you check if this works on Titan?

PR #399, opened 2018-01-13
Sst common testing fix

eisenhauer:
Fix some Sst issues that pop up with the common testing prog.


eisenhauer:
Closing this, my opened version of Norbert's WIP, because the SST fix is going in another patch.

Issue #405, opened 2018-01-16
Make Close free MPI Communicator in all Engines


williamfgc:
1) Remove MPI_Comm_free from destructor 
2) Make all engines free MPI communicator at Close after each derived class DoClose function

PR #406, opened 2018-01-16
Issue #405 make Close call MPI_Comm_free in all Engines

williamfgc:
Close now calls a derived class DoClose (private) function and free the Communicator if all transports are closed.
Remove MPI_Comm_free from Engine destructor.
Modified all derived engines to pass tests


chuckatkins:
Shouldn's we remove the `MPI_Comm_free` entirely since we don't have ownership of the communicator?  We're simply passed a communicator in whose lifetime is managed externally.  I would think we should only call `MPI_Comm_free` on communicators that we create and manage.

williamfgc:
@chuckatkins that's right, IO Open does a MPI_Comm_dup so each Engine owns the duplicated communicator and `MPI_Comm_free` is needed. This PR is just to move the free from the destructor to Close for all engines.

Issue #401, opened 2018-01-13
BPFile writer with heat transfer: missing data in BP file 


pnorbert:
Data is 0 where it should not be. Bug present in VM as well as on Titan. 

$ mpirun -n 1 ./bin/heatTransfer_write_adios2 ../examples/heatTransfer/heat_bpfile.xml heat1.bp 1 1 100 100 10 10

$ bpls -l heat1.bp
  double   T     11*{100, 100} = -6.40663 / 7.19752 / null  / null 

$ bpls -l heat1.bp -d T -n 100 > T1.txt
$ less -S T1.txt
steps 3, 5, 7, 8, 9 have lots of zeros in the data

The statistics looks fine:

$ bpls -l heat1.bp -D
  double   T     11*{100, 100} = -6.40663 / 7.19752 / null  / null 
        step  0: 
          block 0: [ 0:99,  0:99] = -6.40663 / 7.19752/ N/A / N/A 
        step  1: 
          block 0: [ 0:99,  0:99] = -5.52777 / 5.90105/ N/A / N/A 
        step  2: 
          block 0: [ 0:99,  0:99] = -4.93104 / 5.05211/ N/A / N/A 
        step  3: 
          block 0: [ 0:99,  0:99] = -4.47786 / 4.49254/ N/A / N/A 
        step  4: 
          block 0: [ 0:99,  0:99] = -4.09603 / 4.09298/ N/A / N/A 
        step  5: 
          block 0: [ 0:99,  0:99] = -3.78051 / 3.80745/ N/A / N/A 
        step  6: 
          block 0: [ 0:99,  0:99] = -3.52656 / 3.60721/ N/A / N/A 
        step  7: 
          block 0: [ 0:99,  0:99] = -3.30453 / 3.44875/ N/A / N/A 
        step  8: 
          block 0: [ 0:99,  0:99] = -3.11012 / 3.31992/ N/A / N/A 
        step  9: 
          block 0: [ 0:99,  0:99] = -2.94833 / 3.22147/ N/A / N/A 
        step 10: 
          block 0: [ 0:99,  0:99] = -2.80852 / 3.13586/ N/A / N/A 

The variable size is 100x100 above. The bug is not present with 10x10.

williamfgc:
[...]

Issue #411, opened 2018-01-22
PG name should be the IO object's name


pnorbert:
void BP3Serializer::PutProcessGroupIndex()
line 55    const std::string name(std::to_string(m_RankMPI));

In ADIOS1 the name of the PG comes from the user given name, which is used in the XML to identify a group. This name is the same for all writers. Currently ADIOS2 just uses the rank for the name, which makes the PGs have different names although they belong together. 

I found one place where this apparently matters: adios 1.x's old read api, and the plotter tool using that api. It does not find the variables in the file. The PG naming issue may or may not cause this issue but in debugging I found looping over the many groups with zero variables instead of just getting one group with a variable.

williamfgc:
Good one!

PR #429, opened 2018-01-31
Issue #401 saving IO name as pg name

williamfgc:


williamfgc:
Should solve issue #411

PR #433, opened 2018-02-01
DataMan worked for asynchronous workflows which allows reader side to disconnect and reconnect freely

JasonRuonanWang:


pnorbert:
Some things are not working with the tutorial example that works in hello_dataman
- Reconnect - the second run of reader connects but seems to be hanging in the read
- Blocking=no parameter does not seem to work in the tutorial example, the writer is still waiting for the reader to connect
- Killing the reader sometimes hangs the writer (if it is ahead with a lot of steps) - then it continues when the reader is back
- CurrentStep starts from 0 in the second run of the reader


- N-to-1 does not work, not even in hello_dataman

To reproduce: install adios2, get [ADIOS2-Examples](https://github.com/ornladios/ADIOS2-Examples) and build it. Then
```
ADIOS2-Examples/cpp/heat$ mpirun -n 1 ./heatSimulation runtimecfg/dataman.xml sim.bp 1 1  200 200 100 10  
```
```
ADIOS2-Examples/cpp/heat$ mpirun -n 1 ./heatVisualization runtimecfg/mixed.xml sim.bp
```

Note that two different xmls are used in the test to make them use dataman (skipping the middle program heatAnalysis)

JasonRuonanWang:
@pnorbert I will look at the tutorial code. But on my machine N-to-1 definitely worked. I was running 2 MPI ranks on the writer and 2 threads on the reader. 

mpirun -np 2 ./hello_dataManWriter
./hello_dataManReader 2

JasonRuonanWang:
@pnorbert "Blocking = no" is an engine parameter rather than a transport parameter, because it needs different engine logic to cache data that is received. So if you move that XML line from the transport section to the engine section it should work in the non-blocking mode.

JasonRuonanWang:
@pnorbert In order to use N-to-1 you need to add multiple transports on the receiver side, the same number of transports as the number of MPI ranks on the sender side. I have only tried adding them through the no-XML API. But XML should work the same way.

Issue #440, opened 2018-02-13
Add updated User Guide documentation


williamfgc:

williamfgc:
Latest master docs in readthedocs: adios2-adaptable-io-system-version-2.readthedocs.io/en/latest/index.html

PR #441, opened 2018-02-13
Initial documentation 

williamfgc:
User Guide using Sphinx
API using doxygen
Addressing Issue #440


williamfgc:
Merging this so everyone can follow up on their own engine.

PR #446, opened 2018-02-14
Eliminated Bcast to rely on MPI Gatherv on zero sized arrays

williamfgc:


williamfgc:
Related to issue #396

PR #448, opened 2018-02-15
Fix step count in BP3 deserializer.

chuckatkins:
Count the number of available steps rather than assume it's equal to
the largest available step number.  This addresses issues where the step
numbers are non-consecutive and do not start at 0 (i.e. in the file
series written by XGC).

### DIFF
--- a/source/adios2/toolkit/format/bp3/BP3Deserializer.cpp
+++ b/source/adios2/toolkit/format/bp3/BP3Deserializer.cpp
@@ -12,6 +12,7 @@
 #include "BP3Deserializer.tcc"
 
 #include <future>
+#include <unordered_set>
 #include <vector>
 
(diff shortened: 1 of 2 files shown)


chuckatkins:
@williamfgc I'm pretty sure this is fine but I want to make sure it's not breaking any other BP logic.

williamfgc:
@chuckatkins This is good. For a more formal definition valid for non-contiguous pieces we use variable->m_IndexStepBlockStarts which is a map whose keys track the current steps, and values track the payload position for each block within a step. I will stick with that at the Engine level.

chuckatkins:
Thanks.  This came up in the XGC files since the step number is specified in each file but xgc.3d.00004.bp has step number 4 so the deserializer was interpreting it as 4 steps, rather than just 1 step.

Issue #430, opened 2018-01-31
adios2_config is broken because of readlink.


eisenhauer:
The readlink command does not uniformly support the "-f" option.   See:
https://stackoverflow.com/questions/1055671/how-can-i-get-the-behavior-of-gnus-readlink-f-on-a-mac
This causes adios2_config to fail to generate appropriate output on OSX.

eisenhauer:
Feb 15 commit fixed this issue.

PR #450, opened 2018-02-15
A vastly more complicated substitute for where readlink doesn't behav…

eisenhauer:
A vastly more complicated substitute for where readlink doesn't behave the way we want.  Does this work everywhere that has bash?

This PR is to address [Issue 430](https://github.com/ornladios/ADIOS2/issues/430).


pnorbert:
It works on my Ubuntu VM.

PR #457, opened 2018-02-23
Support for Fortran local variables

williamfgc:
Added adios2_null_dims parameter for shape and start of local variables in adios2_define_variable


williamfgc:
Merging as I need to change a few things on top of this

PR #462, opened 2018-02-28
Make tests run serially and force max mpi procs to 16

scottwittenburg:


scottwittenburg:
@chuckatkins I think this is ready to be merged so @pnorbert can rebase #394 on top of it.

PR #467, opened 2018-03-02
Use CMake's builtin FindPythonLibs instead of pybind11's

### LINKED ISSUE #306: CMake doesn't detect missing Python.h
Python-dev is required by PyBind11. Compilation fails even when numpy and mpi4py are installed. This is not detected by cmake.
```
workspace/ADIOS2/thirdparty/pybind11/pybind11/include/pybind11/detail/common.h:111:20: fatal error: Python.h: No such file or directory
 #include <Python.h>
                    ^
compilation terminated.
```

chuckatkins:
The FindPythonLibNew module in pybind11 is lacking any validation of
actually finding the component's it's looking for.

Fixes #306

### DIFF
--- a/cmake/FindPythonFull.cmake
+++ b/cmake/FindPythonFull.cmake
@@ -15,6 +15,7 @@
 #   PYTHONINTERP_FOUND
 #   PYTHON_EXECUTABLE
 #   PYTHONLIBS_FOUND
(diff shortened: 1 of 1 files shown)


chuckatkins:
@williamfgc please check if this fixes the issue for you

PR #468, opened 2018-03-02
Fix configure error in bpSZWrapper example

scottwittenburg:


scottwittenburg:
Ping @chuckatkins for review

chuckatkins:
It looks like this is an MPI only implementation so rather than drop the link line, it needs to be moved inside the if block instead.

scottwittenburg:
Oh duh, will fix.

chuckatkins:
Either that, or even better, move add_executable to before the if block and fix the c++ code to check the appropriate #ifdef and work with both mpi and no-mpi

scottwittenburg:
Yes I was looking at some other examples that do that kind of thing.  That's probably better, and easy to do.

scottwittenburg:
@chuckatkins Do you know why some examples link with `MPI::MPI_C`, while others refer to a CMake variable `${MPI_C_LIBRARIES}`?

chuckatkins:
Legacy.  New code should use the target, not the variables.

scottwittenburg:
Sounds good.  Now, it seems I don't have SZ installed on my machine, and it seems neither does our Docker container where we run the CircleCI tests.  So 1) what's the best way for me to install it locally on my machine so I can test with it, and 2) (maybe the same answer) how should we install it into our Docker container?

pnorbert:
from github:  https://github.com/disheng222/SZ/releases

scottwittenburg:
@chuckatkins With my latest push, I was able to build and run the bpSZWrapper example both with and without MPI enabled.

examples/hello/bpSZWrapper/helloBPSZWrapper.cpp
```diff
@@ -9,7 +9,6 @@
  *  Created on: Jul 26, 2017
  *      Author: William F Godoy godoywf@ornl.gov
  */
-#include <mpi.h>
```
chuckatkins: Keep the mpi.h include, just move it to after adios2.h and guard with the ifdef.  Otherwise everything else looks good.

scottwittenburg: I can definitely do that, but I removed it because we have tests with MPI where we don't include it.  And it worked without it.

pnorbert:
adios2.h includes mpi.h or mpidummy.h depending on the build

examples/hello/bpSZWrapper/helloBPSZWrapper.cpp
chuckatkins: It works because it's getting pulled in transitively by adios2.h but we shouldn't expect or rely on transitive includes.  If we explicitly make MPI calls then we should explicitly include mpi.h.  The other tests should be changed at some point.

Thanks for fixing this!

scottwittenburg:
@chuckatkins I think this thing is all set.

PR #472, opened 2018-03-06
Add a static analysis check to travis build matrix

scottwittenburg:


source/adios2/engine/insitumpi/InSituMPIFunctions.cpp
```diff
@@ -27,6 +27,8 @@ namespace insitumpi
 std::vector<int> FindPeers(const MPI_Comm comm, const std::string &name,
                            const bool amIWriter, const MPI_Comm commWorld)
 {
+    int garbage = 7 / 0;
```
chuckatkins: Does it really not find anything else?

scottwittenburg: No it found a bunch of stuff, now that I'm using it correctly, see the Travis output for the `analyze` build.  I just added this as a validation that it would now check code that isn't turned on unless MPI is enabled.

scottwittenburg: By the way, all the extra packages needed by `openmpi-bin` and `libopenmpi-dev`, plus the extra files to analyze, drove the time on that build up around 11 minutes.  Maybe still ok as long as it still beats `appveyor`?

chuckatkins: You might want to try mpich instead, it's likely to have fewer dependencies.

scottwittenburg: Will do, thanks.

scottwittenburg: And in either case, do I actually need the runtime stuff, or just the development libraries?  So in the case of openmpi, maybe I could have skipped installing `openmpi-bin`?

chuckatkins: You do, otherwise you won't have `mpiexec`.  The runtime packages are set up as a dependency of the dev packages anyways so even if you don't explicitly install them they'll still get pulled in.

scottwittenburg: But all I want to do is compile, not run anything, do I still need `mpiexec`.  I guess its academic based on your comment about the dev packages pulling them in anyway.

scottwittenburg: But just for my enlightenment, if I only want to compile, do I still need `mpiexec`?

chuckatkins: No, but you'll need the runtime libraries, even if not the executables.  The runtime package will include `libmpi.so.1.2`, the actual library file, and the `libmpi.so.1 -> libmpi.so.1.2` symlink since `libmpi.so.1` is the SONAME that the runtime linker looks for.  The dev package, however, will have the `libmpi.so -> libmpi.so.1.2` symlink since `libmpi.so` is what the compile time linker looks for.  You basically need three files:
  * ACTUAL_LIBRARY_FILE
  * RUNTIME_SYMLINK -> ACTUAL_LIBRARY
  * COMPILE_TIME_SYMLINK -> ACTUAL_LIBRARY

scottwittenburg:
@chuckatkins In that last commit ([2714720](https://github.com/ornladios/ADIOS2/pull/472/commits/271472006e6979d13ee7f4e86b04dd63c76df8c1)) where I used CMake 3.10 to run cppcheck automatically, it seems like the way CMake does it (by printing a "Warning:" at the beginning) keeps CDash from seeing the errors I'm having `cppcheck` print , so we're back to warnings instead of errors.  That bit seems [hard-coded](https://github.com/ornladios/ADIOS2/pull/472/commits/271472006e6979d13ee7f4e86b04dd63c76df8c1) in CMake.

So do you think I should go back to the compiler wrappers approach?  At least that way we were getting errors in the CDash report.  Or maybe you have an idea how we can get that back with CMake static check functionality.

scottwittenburg:
Ok, I pushed a different branch to back up that commit that uses CMake 10 to do `cppcheck` automagically.  For now I'm going back to the compiler wrappers approach in this topic because of the extra degree of control it provides.

scottwittenburg:
Ping @chuckatkins  By exiting 0, the two static analysis checks are passing from Travis perspective now, but still showing up as errors on CDash.  Also, they're are now submitting to the "Analysis" track, so we could be all set here.

PR #480, opened 2018-03-15
fixed a install rule in c bindings

JasonRuonanWang:


JasonRuonanWang:
This fix still does not work properly, as any file coming after adios2/bindings/C/adios2/adios2_c_FILE.h still does not install

@williamfgc

williamfgc:
Issue #481

williamfgc:
Fixed #481

PR #483, opened 2018-03-15
fixed a linker warning complaining about a library path not found in …

JasonRuonanWang:
fixed a linker warning complaining about a library path not found in gtest build dir.


chuckatkins:
We can also add a warning suppression so it doesn't show up on the dashboard, or if it's just not feasible to address upstream then we can work the repo, make the change there, and and then update the thirdparty copy from the fork.  Making local changes to the third party repos though prevents us from easily updating them.

JasonRuonanWang:
@chuckatkins Yes, I had a look and it seems that the latest googletest already fixed this problem.

PR #490, opened 2018-03-19
WIP: Adding BP marshaling into SST engine

JasonRuonanWang:


chuckatkins:
@JasonRuonanWang is this still active or can it be merged or closed?  If it is still active then please rebase off the current master to get some updated in the CI infrastructure.

JasonRuonanWang:
@chuckatkins I am still working on it

JasonRuonanWang:
@chuckatkins I got this error in one of the checks. 

source/adios2/engine/sst/SstReader.cpp
Variable 'rank' is assigned a value that is never used.
402  rank = subFileIndexPair.first;

But actually I have used that variable later in the function in a macro, see Line 428 of source/adios2/engine/sst/SstReader.cpp

eisenhauer:
Yeah, Codacy gets stuff wrong quite a bit.  Take it with a grain of salt.  It isn't required to pass that to merge.

chuckatkins:
Enabling Codacy was a bit of an experiment.  I didn't know what type of false alarms we'd end up seeing.  It seems like it may be a bit too much noise and too many false alarms so I'll probably end up disabling it in the near future as a PR check.  Either way, the red X shows up but it won't black the merge.

chuckatkins:
That being said, you will want to rebase on the current master to get the updated circle CI builds to run.

JasonRuonanWang:
@chuckatkins Rebased. But there is a CI error which I don't quite understand what it is. Could you have a look?

chuckatkins:
That one test has been a bit flakey on icc+impi.  It's ok to ignore it for now.  I've disabled the icc+impi build as required because of that so it will still run but the failure isn't a merge blocker.  The others are required and are passing.

chuckatkins:
@eisen this is the failure I was referring to.

JasonRuonanWang:
@eisenhauer @chuckatkins @williamfgc @pnorbert 

This is pretty much finished for now.

I have done a lot of merges in this PR, so I think I should check with you guys in case I broke anything.

eisenhauer:
As a matter of form, you probably want to separate the updates that are not related to BP serialization in SST (e.g.  all the clang-format whitespace changes in files that are not involved here), and not merge the debug stuff (Dump, hardcoding marshaling selection, etc.).  I assume that the current one remote read per variable per writer and Wait after each one is just WIP and you're going to aggregate those, yes?

JasonRuonanWang:
@eisenhauer Just removed dump. The hardcoding marshaling selection should be already removed in previous commits. Now I am using engine parameters from the hello_sstWriter and Reader to control marshaling selection. 

I have also moved the Wait to the end of PerformGets in this commit. But I am worried that this will only work for reads that happen to contiguous memory. If the read schedules are not dealing with contiguous memory, then I might need to wait for each read to save memory allocation, otherwise it is going to double or even multiply the read buffers since I will need to allocate a map and a number of vectors in it to store the received data. In some cases (when the read schedules are very sparse in memory), doing this will result in allocating orders of magnitude more memory than the user application variables do. In this case maybe one wait per read is going to be much more efficient. What do you think?

PR #494, opened 2018-03-21
Update third party

chuckatkins:


chuckatkins:
@JasonRuonanWang this goes through and updates all the third party libs to their current upstream releases including googletest.  Please check to see if the this update fixes your issue from #483

JasonRuonanWang:
It's fixed.

PR #497, opened 2018-03-21
Fix Fortran module race condition in build of tests

chuckatkins:
Place SmallTestData_mod.f90 into an object library to avoid multiple copies of
small_test_data.mod from overwriting one another in parallel builds.  CMake
should be able to handle this on it's own a bit more inteligently but at the
moment it does not so the object library approach used here is really just a
workaround but should be sufficient for what we need.


williamfgc:
I only saw this on Scott's laptop, but it's great to identify the root cause. 
Thanks @chuckatkins

chuckatkins:
It's been coming up quite a bit on the dashboards and is the main reason Fortran isn't well tested in the nightlies.  This should fix that though.

PR #498, opened 2018-03-22
fix an error reported by static analysis

pnorbert:
Asking for a review because I needed to completely erase the whole build directory and rebuild from scratch, otherwise the recompiled ManyVars test aborted. But built from scratch, the test (and all tests) ran fine. 

Something in Google Test gets confused otherwise. But this PR only replaces a char * with an std::string string in the test code.

### DIFF
--- a/testing/adios2/performance/manyvars/manyVars.cpp
+++ b/testing/adios2/performance/manyvars/manyVars.cpp
@@ -138,7 +138,7 @@ public:
     int REDEFINE =
         0; // 1: delete and redefine variable definitions at each step to
            // test adios_delete_vardefs()
-    char *FILENAME = nullptr;
+    std::string FILENAME;
 
     /* Variables to write */
     int *a2 = nullptr;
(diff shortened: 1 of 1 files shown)


chuckatkins:
You could even do this with a single `snprintf:`
```c++
snprintf(FILENAME, sizeof(FILENAME), "manyVars.%d_%d_%d%s.bp", NVARS, NBLOCKS, NSTEPS, REDEFINE ? "_redefine" : "");
```
That's up to you though.  Either way, I think this is much cleaner.

pnorbert:
Do not merge this before the release. Something is strangely bad with the build. After configuring and building the whole adios code this test runs fine. Any changes to the source and then make, will cause this test to abort (Cmake strangely reruns and executables are re-linked). 

111: TestManyVars: ../nptl/pthread_mutex_lock.c:81: __pthread_mutex_lock: Assertion `mutex->__data.__owner == 0' failed.
111: [adiosVM:17242] *** Process received signal ***
111: [adiosVM:17242] Signal: Aborted (6)
111: [adiosVM:17242] Signal code:  (-6)

chuckatkins:
@pnorbert What's the state of this PR?  Is ok to merge, still broken, can be closed, etc.? If it is still active can you please rebase off current master to use some updates in CI?

PR #501, opened 2018-03-25
Add RDMA SST DP

philip-davis:
Adds RDMA-based data plane to SST. This is implemented using libfabric. This adds an external dependency on libfabric.

Caveats:
-Tested for ibverbs and sockets providers so far.
-Uses CManager from EVPath for out-of-band signaling of read requests.
-Adds the third-party dependency of libfabric to use RDMA transport
-Tested against libfabric 1.5.3

The default data plane for the SST engine remains "evpath". Using the rdma dataplane requires:
`<parameter key="DataTransport" value="rdma"/>` in the engine declaration.

If cmake is unable to find libfabric at compile time, selecting the rdma data plane will result in an error at runtime. `LIBFABRIC_PREFIX` can be used to provide a hint to cmake to find the libfabric install.


philip-davis:
@eisenhauer @chuckatkins 

I did a number of things to detect libfabric and conditionally include the rdma dp. I'm sure there are better ways to do some of these things and this will need some refactoring.

Currently, selecting the rdma data plane if it was not built results in an error message `Unknown DP interface rdma, load failed`, followed by a crash. The crash is not ideal, hopefully is okay for the moment.

eisenhauer:
I've had a look, and didn't notice any major issues.  Did wonder about including stdio.h in sst.h though.  Was there something that required that?

eisenhauer:
Whoops, I meant stdlib.h in sst.h @philip-davis

philip-davis:
I get complaints about the size_t type otherwise.

Also, I somehow read your earlier comment as in rdma_dp.c, which is the reason for the latest two commits, which cancel each other out. d'oh.

eisenhauer:
Try stddef.h instead?

philip-davis:
yes, that is better. done. Thanks!

eisenhauer:
OK, I'm going to go ahead and pull this in...

PR #506, opened 2018-03-27
Fixed missing target include for libfabric

philip-davis:
Fixes a build failure on Summit


chuckatkins:
This should be using the imported target instead.  That being said, please cancel this PR.  I'm about to push one up that has several; more updates, including this.

Issue #471, opened 2018-03-05
Add ADIOS2.FlushAll() 


pnorbert:
We need a function where the user can ask ADIOS to flush/sync all open file streams when e.g. they close the file for the checkpoint. This ensures that at the checkpoint, all other data streams are synced to the disk.

PR #507, opened 2018-03-27
Enhancement #471 ADIOS and IO FlushAll, Engine Flush

williamfgc:
Explicit Flush functionality
FlushAll are single calls to flush engines within the adios or io object
scope
Added test for reading between FlushAll and Close at Write

### DIFF
--- a/source/adios2/core/Engine.h
+++ b/source/adios2/core/Engine.h
@@ -63,6 +63,8 @@ public:
      */
     IO &GetIO() noexcept;
 
+    Mode OpenMode() const noexcept;
+
     StepStatus BeginStep();
 
     /**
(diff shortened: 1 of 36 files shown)


source/adios2/core/Engine.h
```diff
@@ -63,6 +63,8 @@ class Engine
      */
     IO &GetIO() noexcept;
 
+    Mode OpenMode() const noexcept;
```
chuckatkins: We usually do properties as `GetFoo()` rather than `Foo()` don't we?

williamfgc: Good point.....Engine has the non-const GetIO, GetSync and GetDeferred .... while the more informative CurrentStep and OpenMode are const.

williamfgc:
Marking WIP since I need to add these to the bindings

PR #511, opened 2018-03-29
Addressing static analysis from coverity scan

williamfgc:


testing/adios2/bindings/C/TestBPWriteTypes.cpp
```diff
@@ -120,8 +120,16 @@ int main(int argc, char **argv)
     MPI_Init(nullptr, nullptr);
 #endif
 
-    ::testing::InitGoogleTest(&argc, argv);
-    int result = RUN_ALL_TESTS();
+    int result = -1;
+    try
```
chuckatkins: Why swallow the exception here rather than propagate it?

source/adios2/helper/adiosType.h
```diff
@@ -163,6 +163,14 @@ std::string VectorToCSV(const std::vector<T> &input) noexcept;
 template <class T>
 std::string ValueToString(const T value) noexcept;
 
+/**
+ * Checks if input pointer is nullptr ad throws an exception
+ * @param pointer input to be checked against nullptr
+ * @param hint additional exception information
+ */
+template <class T>
+void CheckForNullptr(const T *pointer, const std::string hint);
```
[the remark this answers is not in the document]
williamfgc: this function is inlined...T can be anything. I don't want to introduce void* unless it's the only option.

testing/adios2/bindings/C/TestBPWriteTypes.cpp
williamfgc: Coverity is complaining about RUN_ALL_TEST not being caught.

chuckatkins: That's okay.  It's allowed to complain about some things.  Test in particular we want to crash if an exception makes it that fer since that means we didn't catch it internally where we should have.  I's rather keep the exception propagation here and teach coverity to ignore that specific issue

williamfgc: Let's do the latter....I will change this back.

chuckatkins: It's part of getting used to both false alarms and acceptable errors with automated analysis tooling.  100% pass rate is never expected and in this case, even if we can't easily suppress it, I'd say it's an acceptable error.

williamfgc: Contaminates the report.... let's get rid of this false positive.

williamfgc: @chuckatkins so I just rerun the tests with intentional failures and the try-catch doesn't affect the end result. This is just catching an exception thrown by google-test itself, not the tests generated exceptions.

chuckatkins: Still though, we don't want to swallow exceptions, especially in test code.  If an exception get's thrown by the test framework then the test should crash.  Coverity flagging it as an issue is a generic "this could crash" type error, which in this case, is known and intended.

williamfgc: I think we are going in circles here. If you can't provide an example on where this fails and if it can't be fixed with commenting out 2 lines of code then I don't think it's worth spending more time on this.

PR #519, opened 2018-03-30
Address various memory leaks inside SST

chuckatkins:


eisenhauer:
These look fine.  I've got a larger WIP that covers dynamic analysis, but I'm holding it until it's a bit more complete (and after the release).  I think these fixes are there too, but I'll manage the conflicts.

PR #539, opened 2018-04-07
nlohmann_json: allow system package

ax3l:
Add the CMake `ADIOS2_USE_SYSTEM_JSON[=OFF]` option. With this option, package system maintainers can decide to use an external [nlohmann-json](https://github.com/nlohmann/json) library, e.g. to provide bug fixes.

Defaults as before to the internally shipped json library.

### Refactoring

- **CMake:**  the self-defined name `NLohmannJSON` differed from [the official CMake target name](https://github.com/nlohmann/json/blob/v3.1.2/CMakeLists.txt#L7).
I updated the old CMake targets to stay consistent with the "upstream" name which avoids introducing aliases when using the library from an external location.
- **Include:** the official API is included via `<nlohmann/json.hpp>`. Keep the prefix to stay consistent between internally shipped and externally installed headers.

### To Do

- [ ] see inline question about robot min-version update
- [x] generally, `ADIOS2_USE_SYSTEM_...=ON` was broken since found packages were not visible in the parent scope (root `CMakeLists.txt`) -> Issue identified, fix added to last commit: [SO link](https://stackoverflow.com/a/48741931/2719194)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -156,7 +156,7 @@ install(FILES ${PROJECT_BINARY_DIR}/source/adios2/ADIOSConfig.h
 #------------------------------------------------------------------------------#
 include(CTest)
 mark_as_advanced(BUILD_TESTING)
-add_subdirectory(thirdparty)
+include(thirdparty/dependencies.cmake)
 
 #------------------------------------------------------------------------------#
 # Main library source
--- a/source/adios2/CMakeLists.txt
+++ b/source/adios2/CMakeLists.txt
@@ -104,7 +104,7 @@ target_sources(adios2 PRIVATE
     engine/dataman/DataManWriter.h
     engine/dataman/DataManWriter.tcc
 )
-target_link_libraries(adios2 PRIVATE NLohmannJson)
+target_link_libraries(adios2 PRIVATE nlohmann_json)
 
 
 if(ADIOS2_HAVE_BZip2)
--- a/thirdparty/CMakeLists.txt
+++ b/thirdparty/CMakeLists.txt
@@ -1,7 +1,6 @@
 include(CMakeDependentOption)
 
 add_subdirectory(KWSys)
-add_subdirectory(NLohmannJson)
 add_subdirectory(pugixml)
 
 if(BUILD_TESTING)
@@ -11,6 +10,9 @@ endif()
 if(ADIOS2_HAVE_Python)
   add_subdirectory(pybind11)
 endif()
+
+option(ADIOS2_USE_SYSTEM_JSON
+  "Use an externally supplied nlohmann-json library" OFF)
 
 cmake_dependent_option(ADIOS2_USE_SYSTEM_EVPath
   "Use an externally supplied EVPath library" OFF
@@ -32,6 +34,12 @@ cmake_dependent_option(ADIOS2_USE_SYSTEM_ENET
   "Use an externally supplied ENET library" OFF
   "NOT ADIOS2_USE_SYSTEM_EVPath" OFF
 )
+
+if(ADIOS2_USE_SYSTEM_JSON)
+  find_package(nlohmann_json 3.1.2 REQUIRED)
+else()
+  add_subdirectory(NLohmannJson)
+endif()
 
 if(ADIOS2_HAVE_SST)
   if(NOT ADIOS2_USE_SYSTEM_EVPath)
--- a/source/adios2/engine/dataman/DataManCommon.h
+++ b/source/adios2/engine/dataman/DataManCommon.h
@@ -17,7 +17,7 @@
 #include "adios2/toolkit/format/bp3/BP3.h"
 #include "adios2/toolkit/transportman/dataman/DataMan.h"
 
-#include <json.hpp>
+#include <nlohmann/json.hpp>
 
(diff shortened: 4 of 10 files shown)


thirdparty/CMakeLists.txt
```diff
@@ -33,6 +35,12 @@ cmake_dependent_option(ADIOS2_USE_SYSTEM_ENET
   "NOT ADIOS2_USE_SYSTEM_EVPath" OFF
 )
 
+if(ADIOS2_USE_SYSTEM_JSON)
+  find_package(nlohmann_json 3.1.2 REQUIRED)
```
ax3l: might need an `NLohmannJson` CMake target alias here...

ax3l: fixed. the self-defined name `NLohmannJSON` differed from the official CMake target.
I updated the old CMake targets to stay consistent with the "upstream" name which avoids introducing aliases when using the library from an external location,

ax3l: - [ ] **Question**: how can we auto-update (`sed?`) the `3.1.2` min-version here? Can we include this in the `update.sh`?

source/adios2/CMakeLists.txt
```diff
@@ -104,7 +104,7 @@ target_sources(adios2 PRIVATE
     engine/dataman/DataManWriter.h
     engine/dataman/DataManWriter.tcc
 )
-target_link_libraries(adios2 PRIVATE NLohmannJson)
+target_link_libraries(adios2 PRIVATE nlohmann_json)
```
ax3l: For some reason, the `nlohmann_json` target is not defined here when build with external

ax3l: Found the reason: `add_subdirectory` keeps found CMake Config/Module packages in the current dir or below... it's not intended to move it to parent (use `include` instead):
https://stackoverflow.com/a/48741931/2719194

ax3l: This is a general bug with `ADIOS2_USE_SYSTEM` packages - will add a fix

ax3l: Issue identified, fix commited: [SO link](https://stackoverflow.com/a/48741931/2719194)

ax3l: (side note: SYSTEM is the wrong naming, it should be EXTERN(AL) or something :-) - in most cases it's hopefully a package manager not the system path ;-) )

chuckatkins: I agree that the `PackageName_USE_SYSTEM_DependencyName' doesn't really convey what we want here but it's a fairly established convention in the CMake ecosystem for how to specify this concept.

chuckatkins:
I don't really see a compelling need to support an external json library since this is header only and purely a private dependency.  If it was a public dependency in either interface or library I would certainly be on board with it, but it's not.  In addition, upstream also requires gcc 4.9 while ADIOS2 supports 4.8 (the EL7 system compiler).  Our internal version is patched to accommodate that.

CMakeLists.txt
```diff
@@ -156,7 +156,7 @@ install(FILES ${PROJECT_BINARY_DIR}/source/adios2/ADIOSConfig.h
 #------------------------------------------------------------------------------#
 include(CTest)
 mark_as_advanced(BUILD_TESTING)
-add_subdirectory(thirdparty)
+include(thirdparty/dependencies.cmake)
```
chuckatkins: This should stay add_subdirectory.  Switching to include carries a fair number of scope implications.

chuckatkins:
I'm working on getting some of the extra pieces for package configs and namespacing moved upstream now so once that's done this can be pulled in the same way as the other third party infrastructure.  I'm going to close this for now and once merged upstream then I'll fix the external third party support.

CMakeLists.txt
ax3l: did you test if `ADIOS2_USE_SYSTEM...` works for the other dependencies? It should not, since `add_subdirectories` only includes newly defined `INTERFACE` libs into the main scope, but not `find_package` modules/config packages: [SO link](https://stackoverflow.com/a/48741931/2719194)

ax3l:
> I don't really see a compelling need to support an external json library since this is header only and purely a private dependency. If it was a public dependency in either interface or library I would certainly be on board with it, but it's not.

There is always a need to expose internal dependencies. Systems might require patched versions (which can be shared via modules/package managers like spack/etc.), developers might want newer versions, etc.

> In addition, upstream also requires gcc 4.9 while ADIOS2 supports 4.8 (the EL7 system compiler). Our internal version is patched to accommodate that.

uff...

> I'm working on getting some of the extra pieces for package configs and namespacing moved upstream now so once that's done this can be pulled in the same way as the other third party infrastructure. I'm going to close this for now and once merged upstream then I'll fix the external third party support.

Sounds good, no prob.

chuckatkins:
>> I'm working on getting some of the extra pieces for package configs and namespacing moved
>> upstream now so once that's done this can be pulled in the same way as the other third party 
>> infrastructure. I'm going to close this for now and once merged upstream then I'll fix the external 
>> third party support.

> Sounds good, no prob.

See nlohmann/json#1048

PR #543, opened 2018-04-10
Update json

chuckatkins:


chuckatkins:
@ax3l this should accommodate what you're looking for.  However, the external nlohmann_json won't properly work until the packaging changes make their way upstream.

chuckatkins:
@ax3l I think this finally does it (I had to filter out a lot of stuff added to the upstream repo).  The packaging changes have been merged upstream as well so they'll be available in the next nlohmann_json release, whenever that happens.  Once a new release is available I'll adjust the find_package call to require that minimum version.  In the mean time, with this branch, an external version will work if it's from the current nlohmann_json master.

PR #544, opened 2018-04-10
fixed a few type warnings for printf in manyvar

JasonRuonanWang:


chuckatkins:
Either way this is good to go after the formatting fixes.  If you want to fix the remaining warnings that'd be great, but it's fine if not.

JasonRuonanWang:
@chuckatkins Yes. I would definitely do it. But these are the only warnings I have seen. What else warnings did you see and how can I reproduce them?

chuckatkins:
Check the dashboard with the [cdash] link in the status

JasonRuonanWang:
I got this error which does not seem like a formatting thing.

Error: COMMIT_RANGE is empty or undefined

chuckatkins:
Awesome!

PR #558, opened 2018-04-18
Added Object based Cxx03 bindings and refactored all bindings

williamfgc:
object based API using adios2::cxx03 namespace
separate libadios2_cxx03 library
separate public header interface after make install
automatic redirection from adios2.h


chuckatkins:
The CMake for this doesn't work right at the moment because the public link dependency on adios2 will force the c++11 flags to get used.  I'm not sure yet at the moment how to get those dropped.  I'll think about it a bit.  At the moment though, this will still build with C++11 flags, breaking your ABI when trying to use it with cxx98/03

williamfgc:
I tested this with a C++03 external program using openmpi and gcc 5.3.1 without the -std=c++11 flag and works fine, but I didn't check further.  The CMake line: set_property(TARGET adios2_cxx03 PROPERTY CXX_STANDARD 98)  seems not to be working and the ABI still C++11 for libadios2_cxx03, which seems strange. Maybe I just got lucky?

chuckatkins:
It works if you do it manually but not if you do it from CMake since CMake will pull in the target usage requirements from the adios2 target.  The best option I see here is to break the C bindings out into their own library, with a private dependency on the `adios2` cxx11 library, and then have the adios2_cxx98 library (there is no separate cxx03 std option so really this makes more sense to make it explicitly a cxx98 library than `03` especially since `03` was really just a bug fix release anyways and still,l the default for most c++ compilers).

chuckatkins:
Without the split it's broken inside CMake

williamfgc:
@chuckatkins would making adios2 private to adios2_cxx98 work? I can always use opaque pointers in the public interface for adios2_cxx98 rather than including adios2_c.h explicitly.

chuckatkins:
It still won't work right.  There's no mechanism in CMake to only propagate some of the usage requirements, i.e. libraries and includes, but not all of them, i.e. compile features.  What's needed is to have an adios2_c library privately depend on adios2 and then adios2_cxx03 (or ideally adios2_cxx98) can publicly depend on adios2_c.  That will ensure the cxx11 usage requirement doesn't propagate to adios2_cxx98.  It still might not work at all though for static libs since all the link dependencies are forced to be public so you'd have to link a cxx98 and cxx11 library in the same executable, which I don't think can work correctly (maybe, not sure).  It should be fine for shared libs though.

chuckatkins:
Also if you decide to keep the cxx98 interface, strings should be passed via const ref since you don't have the optimized string guarantee or move construction you get with cxx11.

williamfgc:
That's right, linking would have to be done for -ladios2 -ladios2_c and -ladios2_cxx98 in any case (private or public). Since this is specific for MFEM (maybe LAMMPS) I'd like to give this a try since any further change won't add any benefit at this point for running statically on Titan. I'd certainly discourage the use of this API for projects supporting C++11 features.

Issue #560, opened 2018-04-19
Python feature requests


suchyta1:
William, Norbert, and I have been starting to consider Python ease of use, and are opening the conversation about Python enhancements. A couple things I see that could make things simpler:

1. Introspection: The ability to return a "file object" that has a list of the variables, their sizes, etc.
2. Default selections: When reading, if the user doesn't give bounds it just tries to return the whole array.

Python users, feel free to add to the discussion...

williamfgc:
adding @kshitij-v-mehta @jychoi-hpc @JasonRuonanWang @pnorbert

williamfgc:
3. Add configFile.xml option to Python high-level API

suchyta1:
I've noticed that there doesn't seem to be a way to get attributes from the Python bindings. Something like InquireAttribute() seems natural in the low level API, as an analog to InquireVariable().

williamfgc:
@suchyta1 please take a look at PR #672  the python test (testing/adios2/bindings/pythonTestBPWriteReadTypes.py) for types will show the added functionality. Use the IO AvailableAttributes function to get the attribute elements, not InquireAttribute. Thanks!

williamfgc:
I will close this for now. For new requests please open a new issue. Thanks!

PR #563, opened 2018-04-19
#560 adding Python features in high-level API

williamfgc:
1) Added available_variables() function that returns a dict[var_name,
dict[key_meta,value_meta] ]

2) Default selection is current shape for global variables

3) Added open functions that takes config file and ioNameInConfigFile to
reuse xml config files

Omitting mix/max for single values in IO::GetAvailableVariables


suchyta1:
@williamfgc Looks reasonable to me.

PR #570, opened 2018-04-23
Fix adios2_reorganize to handle single values

pnorbert:


chuckatkins:
@pnorbert is this good to go now?

PR #579, opened 2018-04-25
Fix duplicate filenames used by different tests

chuckatkins:
Having the same filename used by different tests will cause data corruption
and false test failures when the two tests are run in parallel.


chuckatkins:
Closing and adding commit to existing circleci pr

PR #584, opened 2018-04-26
Treat warnings as errors in CI builds

chuckatkins:


chuckatkins:
Pending upstream merging of evpath fixes.

PR #587, opened 2018-04-26
just to test CI

eisenhauer:
Not for merging, but just to get a CI failure on SST if possible.


chuckatkins:
Hah!  It actually passed this time, go figure

eisenhauer:
The initial commit was a dummy, so starting over with real mods.

PR #592, opened 2018-04-27
RDMA DP Finalize Clean-up

philip-davis:
Fix a memory corruption issue that causes reads to fail.


source/adios2/toolkit/sst/dp/rdma_dp.c
```diff
@@ -692,7 +692,8 @@ static void RdmaProvideTimestep(CP_Services Svcs, DP_WS_Stream Stream_v,
     sprintf(Info->CheckString, "Rdma info for timestep %ld from rank %d",
             Timestep, Stream->Rank);
     Info->CheckInt = Stream->Rank * 1000 + Timestep;
-    Entry->Data = Data;
+    Entry->Data = malloc(sizeof(*Data));
```
chuckatkins: Is this ever freed?

philip-davis: No, it is not. I am getting a commit ready adding cleanup/deallocation for the DP, but I wanted to start a PR for this since the RDMA DP is currently broken without the above change.

philip-davis: Now it is

philip-davis:
This is ready to be reviewed, it should be cleaning up the data plane resources.

PR #597, opened 2018-04-30
Fix race condition in threaded deserialization

chuckatkins:


chuckatkins:
So, I believe this is clean now.  I'm going to push a few empty commits though to cycle through the ci a couple of times to make sure it's not failing anymore.

chuckatkins:
Thanks for the review!

PR #605, opened 2018-05-07
Convert &[0] to .data() to prevent null dereference for empty vectors

chuckatkins:


williamfgc:
@chuckatkins .data() is C++11. I purposely eliminated any reference to C++11

chuckatkins:
Ah, I didn't realize `.data()` was c++11.  THat's fine., they can be changed to `.begin()` in the cxx98 bindings and are equally valid.

PR #606, opened 2018-05-08
Tweak for EVPath shutdown that impact SST heisen-bug.

eisenhauer:
Testing a possible fix to the heisen-bug affecting some platforms during SST shutdown.


eisenhauer:
Cautiously optimistic...

PR #608, opened 2018-05-09
Turning aggregation substream exceptions to warnings

williamfgc:


source/adios2/toolkit/format/bp3/BP3Base.cpp
```diff
@@ -732,14 +733,30 @@ void BP3Base::InitParameterSubStreams(const std::string value)
...
-                description + "\n, in call to Open\n");
+                "ERROR: can't convert parameter value in substreams=value "
+                " to an integer number, in call to Open\n");
+        }
+
+        if (subStreams < 1)
+        {
+            if (m_RankMPI == 0)
+            {
+                std::cout << "Warning substreams is less than zero, using "
```
pnorbert: Warning substreams is less than ONE

williamfgc: Good catch....

PR #609, opened 2018-05-09
Add getopt for windows

chuckatkins:


chuckatkins:
@pnorbert once merged, you should be able to use this in the `CMakeLists.txt` with:
```
if(WIN32)
  target_link_libraries(myTargetThatNeedsGetOpt mingw-w64::getopt)
endif()
```

PR #616, opened 2018-05-11
Moved all MPI_Wait to Wait function in serializer and corrected MPI aggregation load balance

williamfgc:


pnorbert:
Worked with small job.

pnorbert:
Running on 512 nodes fine, much improved performance. Thanks.

PR #632, opened 2018-05-16
Add test for BP marshalling 

eisenhauer:
Add test for BP marshalling by specifying engine parameters on the command line


eisenhauer:
Not really a review, but just wanted to let you guys know that when this is merged, we'll be testing BP marshaling in CI.

Issue #635, opened 2018-05-18
bpls uint64_t and size_t printf specifier problem on Mac and Clang 9.1.0


JasonRuonanWang:
/Users/jason/dropbox/tmp/ADIOS2/source/utils/bpls/bpls.cpp:1521:42: warning: format specifies type 'unsigned long long' but the argument has type
      'std::__1::vector<unsigned long, std::__1::allocator<unsigned long> >::value_type' (aka 'unsigned long') [-Wformat]
            PRINT_DIMS_UINT64("  start", startv.data(), tdims - tidx, j);
            ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/Users/jason/dropbox/tmp/ADIOS2/source/utils/bpls/bpls.cpp:462:32: note: expanded from macro 'PRINT_DIMS_UINT64'
        printf("%" PRIu64 " ", v[loopvar]);                                    \
                ~~~            ^~~~~~~~~~
/Users/jason/dropbox/tmp/ADIOS2/source/utils/bpls/bpls.cpp:849:18: note: in instantiation of function template specialization
      'adios2::utils::readVar<std::__1::basic_string<char> >' requested here
        retval = readVar(fp, io, variable);
                 ^
/Users/jason/dropbox/tmp/ADIOS2/source/utils/bpls/bpls.cpp:651:42: note: in instantiation of function template specialization
      'adios2::utils::printVariableInfo<std::__1::basic_string<char> >' requested here
                ADIOS2_FOREACH_TYPE_1ARG(declare_template_instantiation)
                                         ^
/Users/jason/dropbox/tmp/ADIOS2/source/utils/bpls/bpls.cpp:1522:42: warning: format specifies type 'unsigned long long' but the argument has type
      'std::__1::vector<unsigned long, std::__1::allocator<unsigned long> >::value_type' (aka 'unsigned long') [-Wformat]
            PRINT_DIMS_UINT64("  count", countv.data(), tdims - tidx, j);
            ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/Users/jason/dropbox/tmp/ADIOS2/source/utils/bpls/bpls.cpp:462:32: note: expanded from macro 'PRINT_DIMS_UINT64'
        printf("%" PRIu64 " ", v[loopvar]);                                    \
[...]

PR #636, opened 2018-05-18
fixed type warning in bpls

JasonRuonanWang:
#635


source/utils/bpls/bpls.cpp
```diff
@@ -1528,7 +1528,7 @@ int readVar(Engine *fp, IO *io, Variable<T> *variable)
         {
             if (verbose > 2)
             {
-                printf("set Step selection: from %zu read %zu steps\n", s[0],
+                printf("set Step selection: from %llu read %llu steps\n", s[0],
```
pnorbert: This format should be "... %" PRIu64 " steps\n", because s is uint64_t.

PR #653, opened 2018-05-30
added DataMan format into toolkit and refined DataMan engine accordingly

JasonRuonanWang:


JasonRuonanWang:
@chuckatkins I got errors from visual studio build, which I don't know how to fix. Could you have a look when you have time? Thanks.

chuckatkins:
Adding dataman files needs to be conditional on whether or not dataman is enabled.  The failure is happening because on the windows builds it gets disabled because we can't find a suitable zmq, so the Jason library is never added as a dependency.  There should be a section later on in the cmakelists.txt that is conditional on ADIOS2_HAVE_DataMan.  Move the addition of the dataman source files to the adios2 target to inside that if block.

PR #658, opened 2018-05-30
Enable Fortran type checks

williamfgc:
Implemented logical handler%valid field


williamfgc:
@pnorbert good catch, added checks for ierr before setting types to a valid state

PR #694, opened 2018-06-27
Fix a problem with the Column-major extraction code 

eisenhauer:
and add tests that exercise the issue


eisenhauer:
Chuck, this should go in release if possible.  Jong, please check that this fixes your problem.

jychoi-hpc:
This work on my mac. But, I got NAN values on Titan. I will double check things and continue to test on Cori or Theta. Anyway, I think it is good to merge.

PR #704, opened 2018-07-06
Make DataMan support no longer optional

chuckatkins:


chuckatkins:
@JasonRuonanWang please check to see if this affects the Titan build issues.

Issue #712, opened 2018-07-11
Need attributes saved from non-zero ranks


pnorbert:
In situ viz libraries need this feature. They may create variables and associated attributes on the local ranks but attributes only saved from rank 0. We need to merge in attributes from other ranks if they don't exist on rank 0. 

I have created a test for this in PR #711

williamfgc:
@pnorbert  from PR #716 check the size of the bp file using mpirun -n 10 for our attributes tests with `ls -all`:

Before:
```drwxr-xr-x 4 wfg users  4096 Jul 12 08:20 .
drwxr-xr-x 3 wfg users  4096 Jul 12 08:20 ..
-rwxr-xr-x 1 wfg users **11957** Jul 12 08:20 ADIOS2BPWriteAttributeReadArrayTypes.bp
drwxr-xr-x 2 wfg users  4096 Jul 12 08:20 ADIOS2BPWriteAttributeReadArrayTypes.bp.dir
-rwxr-xr-x 1 wfg users  **8298** Jul 12 08:20 ADIOS2BPWriteAttributeReadSingleTypes.bp
drwxr-xr-x 2 wfg users  4096 Jul 12 08:20 ADIOS2BPWriteAttributeReadSingleTypes.bp.dir```


After:
```drwxr-xr-x 4 wfg users 4096 Jul 12 08:17 .
drwxr-xr-x 3 wfg users 4096 Jul 12 08:17 ..
-rwxr-xr-x 1 wfg users **1886** Jul 12 08:17 ADIOS2BPWriteAttributeReadArrayTypes.bp
drwxr-xr-x 2 wfg users 4096 Jul 12 08:17 ADIOS2BPWriteAttributeReadArrayTypes.bp.dir
-rwxr-xr-x 1 wfg users **1530** Jul 12 08:17 ADIOS2BPWriteAttributeReadSingleTypes.bp
drwxr-xr-x 2 wfg users 4096 Jul 12 08:17 ADIOS2BPWriteAttributeReadSingleTypes.bp.dir```

pnorbert:
Thanks. You can also verify this with bpdump

PR #715, opened 2018-07-11
#712 enable attributes in all ranks

williamfgc:
Needed by insitu-viz libs

### DIFF
--- a/source/adios2/toolkit/format/bp3/BP3Serializer.cpp
+++ b/source/adios2/toolkit/format/bp3/BP3Serializer.cpp
@@ -642,7 +642,7 @@ void BP3Serializer::SerializeDataBuffer(core::IO &io) noexcept
 
     // attributes are only written once
 
-    if (m_RankMPI == 0 && !m_MetadataSet.AreAttributesWritten)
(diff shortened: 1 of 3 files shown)


pnorbert:
Does this fix only store 1 instance of the attribute if multiple writers define it? We don't want to store N copies of values for each attribute if all N writers define the attributes.

williamfgc:
@pnorbert that would imply adding a collective, N-to-N comparisons, and sorting operations  per attribute. It would be easier to: 1) define attributes  in certain ranks at the user level, similar to global values. `if (rank==0) { io.DefineAttribute...}`.  or, 2) add a flag to DefineAttribute so the user can tell us if it's a constant attribute across all ranks (default), or will vary across ranks (special case).

pnorbert:
No, there is no need for that. Attributes cannot have different values
across processes by definition. We only need to store one value for each
attribute. We need to send the attributes from each rank to the master but
instead of merging like variables, we need to filter and add only
attributes to the metadata list that don't already exist.

williamfgc:
OK, I get it now. Let me work on the deserialization part.

PR #720, opened 2018-07-17
added tests for dataman

JasonRuonanWang:


chuckatkins:
The suse-pgi build error is a bug in the docker image.  I've fixed that so it should now build.  wrt to your other errors, all the tests are now executed via mpiexec, even for tests that don't use MPI.  Some MPI runtimes will error out if they are not.  So I believe your tests are failing or hanging because they are being run with mpiexec and dataman is likely not expecting that.

chuckatkins:
Although I guess that doesn't really explain the intel no-mpi failure...

chuckatkins:
Do you know what was causing the hang and why the sleep is warranted?

JasonRuonanWang:
@chuckatkins I think I got some idea. The writer thread writes too fast and it finishes before the reader thread starts, so the reader thread always waits for the time steps to come.

chuckatkins:
@JasonRuonanWang rather that putting arbitrary sleeps in to make the timing work, would it make more sense to instead allow the writer to retry the connection attempt?  Something like allowing x number of retries with y milliseconds in between.  That way it wouldn't matter which side starts up when.

JasonRuonanWang:
@chuckatkins This test works in the publication / subscription mode of DataMan, which means the writer writes arbitrarily without the awareness of any readers. In any workflows using this mode, the writer would not need to retry or check anything in principle. The idea is that the reader can be attached or detached freely in the workflow. It's the reader that should retry all the time to receive any time step that has been published by the writer. It's a completely asynchronous workflow so it's normal that in a real world workflow, a reader is waiting for a long time for the writer to start or re-start. 

I know what you mean, but the only thing I can do is to replace the while loop with a for loop. But in that case it's even harder to specify the number of iterations in order to pick up any steps from the writer. It's very easy to make the reader not read anything before it quits, if it's a for loop.

There are two arbitrary sleeps. One of them is put between launching the two threads. This is to ensure that the reader thread runs first and finishes it's initialization, so that the writer side will only start after the reader side finishes it's initialization stage. I could remove this, but then I need to put more time steps on the writer side, which increases the chance that the very slow CI VM runs out of the 300 second limit. The other arbitrary sleep is put in the writer's BeginStep - EndStep loop. This is to simulate the actual data production process. If this is not there, then it's very easy to have the writer side write so fast that it writes all steps at once while the reader side does not have time to receive anything. So I don't think there is an easy way to replace either of them with retries.

JasonRuonanWang:
@chuckatkins BTW I think this commit should pass, but I am getting some errors which do not seem to be relevant to what I was doing. Could you have a look?

chuckatkins:
@JasonRuonanWang Thaks for the explanation on the thread timing.  While it's not exactly the most elegant solution, I think it's fine for the sake of testing.  Also, I reran the travis build (it was a repo timeout issue).  You're all set!

PR #723, opened 2018-07-17
Provide multiblock variable inspection functions at the user-level

williamfgc:


williamfgc:
@pnorbert I will remove the Shape as we spoke

PR #726, opened 2018-07-17
fixed a return type warning in Engine class

JasonRuonanWang:


JasonRuonanWang:
@williamfgc I've got a very long list of warnings on Mac, which is caused by a typed function not returning anything. Could you have a look and let me know if this is a fair fix? Thanks.

PR #738, opened 2018-07-23
Added tests for DataMan Serializer and Deserializer

JasonRuonanWang:


JasonRuonanWang:
@chuckatkins 

I got a very weird problem. I have my DataManSerializer and DataManDeserializer classes both using nlohmann/json.hpp The serializer class found the header file but the deserializer didn't. They are using exactly the same configuration in Cmake. 

This only happens on Linux. My Mac is building fine. 


[ 88%] Building CXX object testing/adios2/engine/dataman/CMakeFiles/TestDataManFormat.dir/TestDataManFormat.cpp.o
In file included from /home/adios2/source/source/adios2/toolkit/format/dataman/DataManDeserializer.tcc:14:0,
                 from /home/adios2/source/testing/adios2/engine/dataman/TestDataManFormat.cpp:17:
/home/adios2/source/source/adios2/toolkit/format/dataman/DataManDeserializer.h:14:29: fatal error: nlohmann/json.hpp: No such file or directory
 #include <nlohmann/json.hpp>
                             ^
compilation terminated.
[ 89%] Linking CXX executable ../../../../bin/TestBPWriteReadVector

JasonRuonanWang:
@chuckatkins Never mind. I think I know the reason. This error is from the test, not the deserializer class. I need to add json as a dependency in the cmake script for this test.

PR #762, opened 2018-08-01
added flush function in sst to prevent exceptions when applications call it

JasonRuonanWang:


JasonRuonanWang:
@eisenhauer I am not sure if you want to leave it empty for long term. Because we have some applications which never call Engine.Close(). Instead it calls Flush() after every EndStep(). So this Flush() is supposed to clean all local buffers for the application to safely exit. I don't know if there is any such local buffers in SST which work asynchronously with the API so you need to kick them to finish. In DataMan there is. But for now an empty function should be good until we get another ticket...

eisenhauer:
Or we could print a message that says "Calling this in a streaming environment is a really bad idea if you care about performance, we suggest that you stop..."

JasonRuonanWang:
Yeah that's a good idea indeed. But for now it's empty anyway. I will probably add this message once I have a real Flush in DataMan.

PR #767, opened 2018-08-02
added decompression in sst reader for bp marshaling

JasonRuonanWang:


JasonRuonanWang:
@eisenhauer Sure. I will merge this first because the functionality is needed. I will work on the optimization very soon in another PR.

PR #777, opened 2018-08-03
Test HeatTransfer SST with both BP and FFS marshaling

chuckatkins:


chuckatkins:
@eisenhauer @JasonRuonanWang

JasonRuonanWang:
This should be already fixed in later pull request. I will restart the tests and see it they got any luck. If not then a rebase should work.

PR #786, opened 2018-08-08
removed all blocking transfers in SST BP

JasonRuonanWang:
trying to optimize SST BP so that there is no more blocking reads.


JasonRuonanWang:
@eisenhauer I see what you mean. That should require too much change on the engine side. Once you have that functionality I can definitely make changes accordingly.

PR #792, opened 2018-08-10
Compile SST without MPI present.  

eisenhauer:
Run 1x1 tests without MPI, all with MPI.


eisenhauer:
This branch has seriously messed up thirdparty stuff.  Starting over.

PR #796, opened 2018-08-14
Fix bpls.cmake location to work for multiconfig generators

chuckatkins:


williamfgc:
@chuckatkins I have my Windows laptop at home, I will check and let you know. Thanks!

PR #870, opened 2018-09-20
Improve FixedSchedule optimization for In-situ MPI engine

keichi:
If the schedule is fixed in In-situ MPI engine, the metadata buffer received at the first timestep is kept around over the subsequent timesteps. However, `InSituMPIReader` parses the metadata buffer at every timestep.

We can reuse the parsed metadata and reduce expensive calls to: `BP3Deserializer::PerformGetsVariablesSubFileInfo()` and `BP3Deserializer::GetSubFileInfo()` .


keichi:
@pnorbert Could you please review this?

PR #886, opened 2018-09-27
Added tests

philip-davis:
Added some tests for the helper string routines.


philip-davis:
Some of them are hit by the public API, but most are not, based on the report.

williamfgc:
If we can come up with calls to the public api that triggers those internal calls is a win win. If we can touch them coming from the public API it is dead code we should remove

pnorbert:
So do you want to merge this in? These are good unit tests if those functions stay around.

williamfgc:
Too late, my point is that a more helpful use of the gcov report is to help us identify: i) dead code and ii) missing user level corner case scenarios that touch deeper functionality. Once these tests are added it becomes hard to identify the previous scenarios. 

Also, a lot of these helper functions are already tested and many are just a thin layer around well-tested C++ and MPI functionality. For example, adding a config.xml  in the ADIOS constructor will touch FileToStream. If a helper function (or an internal std::invalid_argument) is never used and we can't come up with a test to touch it, I think it should be removed. The latter happened due to code duplication, future deprecation, etc.

PR #907, opened 2018-10-02
change the default timeout for BeginStep to max float

eisenhauer:


eisenhauer:
Jason, please make sure I have appropriately modified the DataMan use of this.

PR #916, opened 2018-10-08
C api

williamfgc:


williamfgc:
@pnorbert please check this doesn't break the Matlab bindings

Issue #914, opened 2018-10-05
Preserve string array attribute for single element arrays in C API


williamfgc:

williamfgc:
@tkurc

PR #918, opened 2018-10-09
Enable attribute string value and array in C bindings

williamfgc:
Addresses #914


williamfgc:
Codacy doesn't consider small std::string optimization, size()<15 is put in the stack, when using C++11

PR #925, opened 2018-10-13
Modify dataplane selection process so that we can do run-time viability determination.

eisenhauer:


eisenhauer:
@pnorbert, I gave Philip the skinny via Slack PM, but for your info as well...  This PR creates a mechanism through which we can do run-time viability checks for the dataplane and choose them via a priority mechanism (overkill for now since we just have 2, but for the future).  This still needs Philip to write the actual libfabric code to see if RDMA could run, so right now I'm just returning a constant and doing no run-time checks.  My initial thought was to enable RDMA by default if libfabric was found in the build, with the idea that if libfabric was found outside CI it was probably because it was on a platform where it was known to work and if it was found inside CI, that should be working already.  But, apparently we have libfabric on (at least) two of the CI VMs, because changing the default to be to use RDMA causes failures.  So, I've changed the default back to where it will not use RDMA by default regardless, and I'll merge the PR when Philip has seen it.  I suspect that we don't actually have RDMA hardware on those CI machines, so when Philip writes run-time code to check to see if we should use libfabric he'll see that there is not a supported RDMA "provider" in libfabric and return the value that disables that dataplane.  But he might want to look at these failures anyway, because I would have expected the "sockets" provider in libfabric to work for this.

@philip-davis, you can look at the Circle CI failures I saw here:
https://circleci.com/gh/ornladios/ADIOS2/tree/pull%2F925
However, probably the thing to do is to get this PR merged to master, then you can do your own WIP PR that re-enables the RDMA DP so that you can poke at those machines.

Issue #725, opened 2018-07-17
Some tests using MPI fail due to insufficient number of slots


keichi:
These tests fail on my laptop with the recent version of ADIOS2:

```
The following tests FAILED:
	  1 - HeatTransfer.BPFile.Write.MxM (Failed)
	  2 - HeatTransfer.BPFile.Read.MxM (Failed)
	  3 - HeatTransfer.BPFile.Dump.MxM (Failed)
	  4 - HeatTransfer.BPFile.Validate.MxM (Failed)
	  5 - HeatTransfer.BPFile.Write.MxN (Failed)
	  6 - HeatTransfer.BPFile.Read.MxN (Failed)
	  7 - HeatTransfer.BPFile.Dump.MxN (Failed)
	  8 - HeatTransfer.BPFile.Validate.MxN (Failed)
	  9 - HeatTransfer.BPFile.Write.Mx1 (Failed)
	 10 - HeatTransfer.BPFile.Read.Mx1 (Failed)
	 11 - HeatTransfer.BPFile.Dump.Mx1 (Failed)
	 12 - HeatTransfer.BPFile.Validate.Mx1 (Failed)
	 13 - HeatTransfer.SST.MxM (Failed)
	 14 - HeatTransfer.SST.Dump.MxM (Failed)
	 15 - HeatTransfer.SST.Validate.MxM (Failed)
	 16 - HeatTransfer.SST.MxN (Failed)
	 17 - HeatTransfer.SST.Dump.MxN (Failed)
	 18 - HeatTransfer.SST.Validate.MxN (Failed)
	 19 - HeatTransfer.SST.Mx1 (Failed)
	 20 - HeatTransfer.SST.Dump.Mx1 (Failed)
	 21 - HeatTransfer.SST.Validate.Mx1 (Failed)
	 22 - HeatTransfer.InsituMPI.MxM (Failed)
	 23 - HeatTransfer.InsituMPI.Dump.MxM (Failed)
	 24 - HeatTransfer.InsituMPI.Validate.MxM (Failed)
	 25 - HeatTransfer.InsituMPI.MxN (Failed)
	 26 - HeatTransfer.InsituMPI.Dump.MxN (Failed)
	 27 - HeatTransfer.InsituMPI.Validate.MxN (Failed)
	 28 - HeatTransfer.InsituMPI.Mx1 (Failed)
	 29 - HeatTransfer.InsituMPI.Dump.Mx1 (Failed)
[...]

PR #938, opened 2018-10-17
#725 enable test with mpi oversubscribe

williamfgc:
OMPI_MCA_rmaps_base_oversubscribe=yes must be either passed to cmake or be already an
env variable when running cmake and ctest


chuckatkins:
The entire logic block here can be reworked to be much simpler I think.  The use case is to oversubscribe when < 4 processors are detected and the OpenMPI env var explicitly says it's okay, right?  FindMPI already sets MPIEXEC_MAX_NUMPROCS to the detected number of physical cores.  Given that, the whole logic block can be reduced to:
```cmake
if(ADIOS2_HAVE_MPI)
  if(MPIEXEC_MAX_NUMPROCS LESS 4 AND "$ENV{OMPI_MCA_rmaps_base_oversubscribe}")
    message(STATUS "OpenMPI oversubscribe detected: raising MPIEXEC_MAC_NUMPROCS to 4 for testing")
    set(MPIEXEC_MAX_NUM_PROCS 4 CACHE STRING "" FORCE)
  endif()
  ...
endif()
```
I think that's right.  Would that work?

williamfgc:
@chuckatkins cmake had issues with MPIEXEC_MAX_NUMPROCS  https://gitlab.kitware.com/cmake/cmake/merge_requests/1405 which were replicated on  my Mac using different cmake versions. That's why I have to use something more backwards compatible like `cmake_host_system_information`

chuckatkins:
We include a private copy of the upstream FindMPI module that should already be patched for this: https://github.com/ornladios/ADIOS2/blob/cefab223f7bad996449c081d7b9805e7ca13bc01/cmake/upstream/FindMPI.cmake#L1291-L1292
so the `cmake_host_system_information` workaround in the main `CMakeLists.txt` shouldn't be necessary anymore.  Does it still not work correctly for you?

williamfgc:
@chuckatkins yes, that works using cmake 3.12 on my Mac (we were forced to upgrade).

PR #954, opened 2018-10-25
New ffs upstream

eisenhauer:


eisenhauer:
Was hoping to avoid the Travis errors, but they're still there...

PR #980, opened 2018-11-01
Stream read is now consist with BP3.

guj:


pnorbert:
This needs to be reviewed. Why do we have Steps() returning > 1 in streaming mode?

williamfgc:
Steps() is global info per variable....is `const` so it is only for info purposes

PR #1013, opened 2018-11-26
WIP: Big endian support in BP3 format, using git lfs 

williamfgc:
Added tests with pre-saved bp files written in big endian


chuckatkins:
We don't actually run executable tests on travis, only formatting and static analysis, so having the travis scripts grab the lfs data is unnecessary and will end up counting against the fairly limited bandwidth allotted to us by github.  So we'll want to keep it restricted to the circle instances.

williamfgc:
Hi @chuckatkins the issue is with the git commands that retrieve the source code that fail due to lack of portable lfs support. I see the same issues on my Mac, lfs is not seamlessly integrated into git for existing code. Not sure if it's worth adding it into the adios2 repo. In any case, this PR is just to explore our options, I won't merge it into master.

williamfgc:
Closing current PR until we have a better idea how to handle git LFS

PR #1015, opened 2018-11-27
Add git-lfs to CI docker images

chuckatkins:
@williamfgc once this is merged into master then the CI images will get automatically rebuilt by dockerhub (typically takes anywhere from an hour to a day for the new images to be available).  At that point we can update the circle scripts to use lfs, but not until then.


williamfgc:
@chuckatkins will this affect how users download adios2 locally? I mean, I had a lot of issues with git commands locally on my Mac and my Linux VM, not sure if `lfs` is worth the trouble in the adios2 repo.

chuckatkins:
This won't, no.  All it's doing is making sure that git-lfs is installed and available on our CI docker images.  It will allow us flexibility to use lfs in CI but doesn't actually do anything to our repo.

PR #1016, opened 2018-11-27
temporarily disabled dataman callback

JasonRuonanWang:
temporarily disabled dataman callback because the test behaves randomly and affects other pull requests


JasonRuonanWang:
@chuckatkins This should fix the hanging problem. I have temporarily removed the callback routine in DataMan since it's causing too many problems and no apps are really using it.

PR #1019, opened 2018-11-28
Add LibFabric tests

philip-davis:
Adds HeatTransfer-based tests, if compiled with LibFabric support. These run the MxN tests, with DataTransport set to "rdma".


philip-davis:
@eisenhauer Thanks for checking it out. My goal is to get the testing infrastructure to exercise the LibFabric DP code...even if it's just using the sockets fabric. I think the default choice of the EVPath DP makes sense if we don't have an RDMA fabric, so I added a test that specifies the DataTrasport attribute to be 'rdma'.

eisenhauer:
OK, that makes sense....  Lets run with it...

PR #1043, opened 2018-12-07
[WIP] Check to see if local memory registration fails in RDMA DP

philip-davis:
Also create a PR to see if I get the same test failures as Greg


chuckatkins:
@philip-davis is this still active?

PR #1047, opened 2018-12-10
Inline engine

aronhelser:
Add an Inline engine. Based on the engine skeleton, so has lots of logging available when the 'verbose' parameter is set. 

Adds the `GetBlock` methods to Engine, so the data pointer provided by the writer can be shared with the reader, without copying. 

Currently:
* keeps around the blockInfo, needs to be deleted at EndStep
* ignores selection, just uses the most recent blockInfo

@chuckatkins here's where I'm at.


aronhelser:
Thanks @williamfgc  for taking a look - I agree with your comments and I will address them shortly.

chuckatkins:
@williamfgc, @pnorbert and I just iterated with @aronhelser on this a bit.  I think the main changes we discussed are to pass through the data pointer in the block info structure to the pubic API and to replace ```void Engine::GetBlock<T>(var, ptr, mode)``` with ```Variable<T>::SetSelection(int blockId)``` followed by ```void Engine::Get<T>(var, mode)``` that populates the data pointer in the appropriate block info.  We can discuss it in more detail offline if you'd like.

williamfgc:
@chuckatkins there is already something implemented in master for the public API. It's `void Variable<T>::SetBlockSelection(const size_t blockID);` I had to implement this for local variables, which are somehow supported now (need more tests).

aronhelser:
Looks like one of the ideas from our meeting needs to be modified. I'm going to have to pass in an Info struct to the `Get<T>(var, info)` call, because of how the Info structs are always passed/returned by value.

aronhelser:
My new test is getting an error, that I don't reproduce locally either on windows or ubuntu:

> 154: [8a0f033d55e3:12532] [[INVALID],INVALID] ORTE_ERROR_LOG: A system-required executable either could not be found or was not executable by this user in file ../../../../../orte/mca/ess/singleton/ess_singleton_module.c at line 458
> 154: [8a0f033d55e3:12532] [[INVALID],INVALID] ORTE_ERROR_LOG: A system-required executable either could not be found or was not executable by this user in file ../../../../../orte/mca/ess/singleton/ess_singleton_module.c at line 166
> 154: --------------------------------------------------------------------------
> 154: Sorry!  You were supposed to get help about:
> 154:     orte_init:startup:internal-failure
> 154: But I couldn't open the help file:
> 154:     /proj/pgi/linux86-64/2018/mpi/openmpi-2.1.2/share/openmpi/help-orte-runtime: No such file or directory.  Sorry!
> 154: --------------------------------------------------------------------------
> 154: *** An error occurred in MPI_Init
> 154: *** on a NULL communicator
> 154: *** MPI_ERRORS_ARE_FATAL (processes in this communicator will now abort,
> 154: ***    and potentially your MPI job)
> 154: [8a0f033d55e3:12532] Local abort before MPI_INIT completed completed successfully, but am not able to aggregate error messages, and not able to guarantee that all other processes were killed!
> 

This is on the suse-pgi-openmpi build only. Does this look familiar? Ideas?

testing/adios2/engine/inline/CMakeLists.txt
```diff
@@ -0,0 +1,15 @@
...
+# accompanying file Copyright.txt for details.
+#------------------------------------------------------------------------------#
+
+add_executable(TestInlineWriteRead TestInlineWriteRead.cpp)
+target_link_libraries(TestInlineWriteRead adios2 gtest)
+
+if(ADIOS2_HAVE_MPI)
+
+  target_link_libraries(TestInlineWriteRead MPI::MPI_C)
+
```
williamfgc: add here `set(extra_test_args EXEC_WRAPPER ${MPIEXEC_COMMAND})`  before any call to `gtest_add_tests`

aronhelser: Thanks! I'll try it

aronhelser:
Better, tests passed, thanks!

If you'd like to take a look, @chuckatkins , @pnorbert  and/or @williamfgc you can see that I had to add an `Info` param to the engine `Get` call. It acts kind of like a token, so that after `PerformGets` or `EndStep`, the user can use it to get the data pointer.

aronhelser:
@williamfgc and @chuckatkins, do you think you can take a look?

aronhelser:
Caught up with Chuck, and a few minor things to do:
* document that passing Info into Get is experimental, and subject to change
* rebase on master, and test again.

aronhelser:
All done, @chuckatkins please take a look and merge when ready, thanks!

aronhelser:
pinging you again @chuckatkins  to take a look and merge.

PR #1072, opened 2018-12-18
WIP ADIOS2 with a client-limited version of ENET

eisenhauer:


eisenhauer:
This PR was created to see if the many-clients mod to ENET might be causing the odd protocol problems.  The protocol problems still occurred without those mods, to presumably the answer is no.

PR #1081, opened 2018-12-31
Fix export configs

chuckatkins:


andraslibal:
I just downloaded the latest lammps distribution available 2019 aug 7 tried to run any simple example that I could find (it is very sad that the official lammps documentation does not provide an example *.in file to try lammps) so I got the examples from the git examples directory. Nothing works now because everything is in a package. Ok. Then I try rebuilding it with all_on or most ...

make -C ../cmake/presets/all_on.cmake ../cmake

and then the error comes in:

Could not find a package configuration file provided by "ADIOS2" with any of the following names:
    ADIOS2Config.cmake
    adios2-config.cmake

Apparently this fix fixed it. Well I have to say that it did not.

I don't even care about this package but I think that it would be great if lammps came out of the box with one example that actually works - I really don't care which example, just give the users a simple *.in file that is tested it is right there in the documentation page after the download and the install instructions and it works. We'll do the rest from there but it is very disheartening to not even be able to get lammps running on a simple example out of the box.

pnorbert:
Hi, it is not clear from your comment if you have an issue with the ADIOS2 software. If you have problem installing LAMMPS on its own (without ADIOS2), and use its examples, then please open an issue on the LAMMPS site. LAMMPS should build without ADIOS2 by default. 

If you wanted to build it with ADIOS2 and have an issue with it then please explain what's wrong. 
Thank you

andraslibal:
My issue is that when I try to build lammps with all packages, it finds a problem in the ADIOS2 configuration. Maybe because it is the first one? I don't know but this prevents lammps from building with packages and if I can't build it with packages then I can't run any example code. 

Whether this is an error specifically from the package or from a generic lammps configuration setting is not visible to the user, the user will try and search the last error message that is displayed. That is why I am here.

akohlmey:
> My issue is that when I try to build lammps with all packages, it finds a problem in the ADIOS2 configuration. Maybe because it is the first one? I don't know but this prevents lammps from building with packages and if I can't build it with packages then I can't run any example code.
> 
> Whether this is an error specifically from the package or from a generic lammps configuration setting is not visible to the user, the user will try and search the last error message that is displayed. That is why I am here.

This is **definitely** not an ADIOS issue, but it is a user issue of expecting a behavior in compiling LAMMPS that is different from what is documented in the LAMMPS manua. If you request to configure and compiled LAMMPS with all optional packages enabled it is **your** responsibility as the person compiling to make certain that all documented prerequisites are fulfilled. If the configuration step fails because of enabling a feature for which the prerequisites are not given, it is the correct behavior. It is documented in the LAMMPS manual how you can selectively augment the presets by enabling or disabling additional/optional packages.

PR #1087, opened 2019-01-02
Disable the currently flaky tests from running in CI dashboards

chuckatkins:
@eisenhauer @philip-davis These tests will still run on the command line via `ctest -R` but by adding then to CTestCustom exclusion then they will be ignored by the automated CI scripts.  When they are able to be reliably enabled again, please remove them from this list.

### DIFF
--- a/CTestCustom.cmake.in
+++ b/CTestCustom.cmake.in
@@ -17,3 +17,17 @@ list(APPEND CTEST_CUSTOM_COVERAGE_EXCLUDE
   ".*/testing/.*"
   ".*/examples/.*"
 )
+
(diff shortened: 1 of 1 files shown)


eisenhauer:
Hmm.  Odd that this set doesn't include the tests that I've most see fail while trying to debug the ENET issues.

chuckatkins:
This isn't a particularly exhaustive list, i just grabbed the ones that seemed to be holding up CI over the past week or two (i.e. re-run a few times until it passes).  Also this list doesn't actually seem to be excluding anything so I need to tinker with it a little.  I've no problem adding items to the list.

PR #1129, opened 2019-02-02
Add another staging engine, WDM, optimized specifically for WDM project requirements

JasonRuonanWang:


williamfgc:
@JasonRuonanWang nice job! Can you add some tests, so our coverage doesn't go down. It is also more reliable when we deploy libzmq based engines in conda.Thanks.

PR #1139, opened 2019-02-05
Expand staging-common test suite with moved and reworked SST tests.

eisenhauer:
Move most of SST engine tests to staging-common.  Rewrite tests to take engine and filename parameters so that they can be used by other engines.  Filename parameter helps avoid conflicts between tests.  Eliminate _nompi fortran versions in favor of conditional compilation.


eisenhauer:
@pnorbert and @JasonRuonanWang 
OK guys, this is passing everything now.  I consider this to be a start on a common staging test set.  The things that I'd love to see that this does *not* have is some mechanism for classifying tests according to features required, features tested, etc.  That would more easily support a mechanism where we could characterize engines vs. tests that they *should* pass.  As it is, we probably have to build up test lists on a per-engine basis.  But it is a start...

PR #1146, opened 2019-02-07
Python and C++ High Level API read and write attributes

williamfgc:


williamfgc:
@isosc @suchyta1 @pnorbert @jychoi-hpc merging changes to high-level Python API that will break your existing code

PR #1151, opened 2019-02-08
Utilize MPI MPMD mode to run common staging tests where possible...

eisenhauer:


eisenhauer:
Jason and Norbert...  OK, I've modified the simple Read and Write tests (not the Server/Client) to do an appropriate MPI_Comm_Split.  I've modified the run_staging_test script to use MPI MPMD mode when it can (not when there are zero readers, or when the reader is delayed).   This should mean that it's possible to add InSituMPI tests to the set, assuming the engine supports what's being tested.  I naively tried to add one and found it didn't support stuff, so I'll let Norbert poke at that.  MPMD should be a bit faster than doing two MPI execs too...

Issue #1150, opened 2019-02-08
Inconsistent types between DefineVariable<T> and InquireVariable<T>


germasch:
To reproduce (on latest master):
```cxx

#include <adios2.h>

void test1()
{
  using T = char;

  adios2::ADIOS ad(MPI_COMM_WORLD, adios2::DebugON);

  {
    adios2::IO io_writer = ad.DeclareIO("writer");
    auto var = io_writer.DefineVariable<T>("var");
    auto writer = io_writer.Open("test.bp", adios2::Mode::Write);
    
    T test = 99;
    writer.Put(var, test);
    writer.Close();
  }

  {
    adios2::IO io_reader = ad.DeclareIO("reader");
    auto reader = io_reader.Open("test.bp", adios2::Mode::Read);
    auto var = io_reader.InquireVariable<T>("var");
    assert(static_cast<bool>(var));
    
    T test;
    reader.Get(var, test);
    reader.Close();

    assert(test == 99);
  }
}

int main(int argc, char** argv)
{
  MPI_Init(&argc, &argv);
  test1(); 
  MPI_Finalize()
  return 0;
}
```

This test writes a single variable, and then tries to read it back again. The type of the variable can be changed with the `using T = ...` directive. The test succeeds with `using T = int;`, but `using T = char;` fails.

It also fails with `using T = std::size:t` and `using T = unsigned long`.

germasch:
I've tracked this down to `core/IO.tcc`. In `InquireVariable`, thos if statement triggers because of inconsistent types.
```cxx11
    if (itVariable->second.first != helper::GetType<T>())
    {
        return nullptr;
    }
```
In particular, the variable from the file ends up being `signed char` (as is correct on Linux/MacOS), but while `signed char` and `char` behave exactly the same, they are distinct types, and `signed char` != `char`.

This is not pretty, but can be worked around relatively easily by using `signed char` or `unsigned char` for the variable type, though I don't think it's desirable since just `char` is very commonly used.

The `unsigned long` problem is even worse, though. (And `size_t` is an alias for `unsigned long` on 64-bit systems, so that's essentially the ame issue).

[...]

PR #1154, opened 2019-02-09
Converts the cxx11 Variable interface to map to/from stdint types

germasch:
First of all, this is just a proof-of-concept, so I'm not asking for it to actually be merged. For example, it only handles Variables, but not Attributes the same way.

This PR is a proposed solution to the problem discussed in #1150. It leaves the core of adios2 unchanged, but it changes the cxx11 interface to be more symmetric with respect to reading back variables using the same type used write them.

Even though there's been some discussion already in #1150, let me repeat some background -- sorry for this being so lengthy, but I'm trying to make this PR kinda self-contained.

* C++11 inherited from C a kinda messy type system. 
  * There are type aliases (same as typedef), which are just new names for the same type. E.g., `using size_t = unsigned long` (or `typedef unsigned long size_t`).
  * The integer types are not well defined w.r.t their actual representation. E.g., a `long int` is 32 bits on 32-bit Linux, but 64 bits on 64-bit system. There are, however, type aliases defined to help alleviate this ambiguity, e.g., `uint64_t` is an alias to a fundamental type that's unsigned and 64 bits wide. On Mac OS 64 bit (which I'll use as an example, because that's what's on my laptop, though I believe 64 bit Linux is the same), `uint64_t` is an alias for `unsigned long long`, though it could have been an alias for `unsigned long`, as both of these are 64 bit unsigned integer types. However, it is important to realize that while `unsigned long` and `unsigned long long` are both 64 bit unsigned types, they are distinct types.
 * I will call two types "equivalent" if they have the same signedness and the same number of bit. On my 64 bit MacOS, `unsigned long` and `unsigned long long` are equivalent, though on, e.g., 32-bit Linux they are not. They are, however distinct types (not aliases) as far as the language is concerned.
 * For another example, `char` and `signed char` are equivalent on 64 bit MacOS, but they are distinct types.
 * Disregarding that there may be non-native 128 bit integer types, too, there are 8 different representations of integers (signed, unsigned x 8, 16, 32, 64 bit). But in C++11, there are 11 distinct integer types, which map to those representations, in a not necessarily standard way.

*Back to adios2:* When using the cxx11 interface for defining a variable and writing it to disk with bp3, any of the 11 integer types can be used, or any of their aliases (including the stdint type aliases), and they will be mapped to their corresponding, possible machine-dependent actual representation. E.g., on Linux-32, `unsigned long` will be written as an unsigned 32 bit quantity (because that's what it actually is). On Linux-64/MacOS-64, it will be written as an unsigned 64 bit quantity (again, because that's what it actually is). One lesson to be learned here is that if one wants to write portable files, one should not `unsigned long`, but rather `uint32_t` or `uint64_t`, depending on what's actually needed. However making a variable of `unsigned long` type is supported and works, just with a machine-dependent result.

However, InquireVariable / reading is not really symmetric in that for writing, one can use any 64-bit unsigned type to create a 64-bit unsigned value in the file. When reading back that quantity, one cannot use any 64-bit type, though, it has to be fundamental type underlying `uint64_t` or an alias for that type.

In my opinion, this asymmetry is not good. If adios2 allows me to write an `unsigned long` variable, it should allow me to read the value back into an `unsigned long` variable. One way to deal with the asymmetry could be to only allow reading and writing of the 8 stdint types, though it's not enforceable to use only the stdint types, since those are just aliases not distinct types, so when allowing to write `uint64_t`, one would automatically allow to also write `unsigned long long` on MacOS-64, but one would prohibit to write `unsigned long` or `size_t`. In addition to being difficult to enforce, it also makes it a hassle to use adios2 for certain use cases, like adding checkpoint / restart to existing codes. In this case, it makes a lot of sense to write, e.g., `size_t` or `char` variables to a file, even though those types are ambiguous in that different system will have different representations of such types. However, with the goal being to read the data back on the very same system, portability of checkpoint files isn't much of a concern, while having to deal with converting, e.g, an array of `size_t` to a stdint type of choice so that it can be unambiguously written/restored could be quite painful.

This PR implements an alternate proposed solution. As before, writing of any, e.g., 64 bit integer type is allowed (e.g., both `unsigned long` and `unsigned long long` on MacOS-64). But it is now also allowed to read such a value back into any 64 bit integer type, ie., it can be read back into `unsigned long` or `unsigned long long` (or `uint64_t` or `size_t`) on MacOS-64.

This is all done at the cxx11 interface level, ie., the core is unchanged (though I think the core should probably be changed to just support the 8 stdint types in the first place then). A `Variable<unsigned long long>` is a distinct type from a `Variable<unsigned long>`, as it needs to be because how the language works, however, they internally will both just be wrappers to `core::Variable<uint64_t>` (again, on MacOS-64).

The necessary mapping to do this actually already existed, by means of `adios2::TypeInfo`. This proof-of-concept only does make the change for Variables, not for Attributes, though I don't think there's any reason that Attributes couldn't be handled the same way.

As far as I can tell, none of the automated tests fail that weren't already broken.


williamfgc:
@germasch this is very cool, thanks. My only fear (going back to the trade-offs) is that we are allowing non-portable read code to work (e.g. `InquireVariable<long int>` or `InquireVariable<size_t>`). I don't see why is painful to map size_t to a fixed size, especially since writer code is usually not as portable as reader code in HPC (I'd actually discourage and question the use of `size_t` for any kind of large variable data in any application), our uses are for very small data sizes (just like in the STL). In any case, C++ and C backwards compatibility is a can of worms we have to deal with one way or another. 

As maintainers, it will hit us back at some point when people writing analysis code don't use fixed width types and realize their C++ analysis code is non-portable (adios2 will get the blame, regardless). 

BTW, the `char` problem is different, signedness is also a property of the compiler flags, so there is no guarantee `InquireVariable<char>` will work even on the same platform. @chuckatkins , @pnorbert , thoughts?

germasch:
> My only fear (going back to the trade-offs) is that we are allowing non-portable read code to work (e.g. `InquireVariable<long int>` or `InquireVariable<size_t>`).

Yes, there is an argument that can be made that only stdint types should be supported. But if so, one should at least also only support stdint types in `DeclareVariable`. Right now, using `DeclareVariable<unsigned long>` works to write data, but it can't be read back with `InquireVariable<unsigned long>`. And right now, the failure will happen at run time, and maybe only on some systems. If you want to force users to use stdint types, one should either not provide the template instantiation for non-stdint types, which would lead to link-time errors, or, preferredly use some static_assert that triggers on compilation of non-stdint instantiations. It would still end up being confusing that `Declare/InquireVariable<unsigned long long>`  would still work, but `Declare/InquireVariable<unsigned long>` doesn't (and maybe vice versa on another system).

I think there are two "right" choices: Support all integer types on both write and read, or support only stdint types (that will include their underlying primitive types and other aliases) for both write and read.

Personally, I find the 2nd option preferable, and I should also note that going to the first option would be backwards incompatible. It does give the user more ability to shoot themselves in the foot, but if they do, it'll at least won't be a silent error. If someone decides to write `char` data into a checkpoint file, and then recompiles the code with `-funsigned-char` or whatever it's called and tries to read their data back as `char`, they'll get (essentially) a nullptr from `InquireVariable` -- adios2 will still type safe in terms of type width and signedness.

> I don't see why is painful to map size_t to a fixed size, especially since writer code is usually not as portable as reader code in HPC (I'd actually discourage and question the use of `size_t` for any kind of large variable data in any application), our uses are for very small data sizes (just like in the STL). In any case, C++ and C backwards compatibility is a can of worms we have to deal with one way or another.

So one thing here is to not necessarily focus on `size_t` as such, maybe I can present a case where a particle simulation code uses `unsigned long` to present how many particles are in the simulation, to support a large number of particles on 64-bit systems but < 4G on a 32-bit system. Does that make much sense? Maybe not, because one might just give up 4 bytes of memory and use uint64_t always. More generally, though, there is a lot of code out there that uses primitive C types all over, and where one doesn't want to change those to stdint types because who knows what will break. And for checkpoint / restart, one really usually doesn't care if checkpoint files are not portable, and that's a use case that I believe is important for adios2 to support.

So I've encountered this issue in the first place while working on checkpoint/restart in my code. And that's generally quite the pain, because C++ doesn't allow introspection (yet). But one of the things one can do is use `declval()` to simplify things a bit, that is to figure out the type of a given variable and use that type for `DeclareVariable()`. So that's the reason why I find it useful to be able to just write and read any primitive variable. Of course, I could do the mapping that's in the cxx11 interface in this PR in my own code instead, but I feel like it's something that could be useful for other users as well.

williamfgc:
```More generally, though, there is a lot of code out there that uses primitive C types all over, and where one doesn't want to change those to stdint types because who knows what will break.```   They don't need to change their current computation code that lives in memory. When integration with adios2 one can do a simple mapping at the glue layer `if( sizeof(unsigned long) == 8 ) io.DefineVariable<uint64_t>.... else io.DefineVariable<uint32_t>`  in fact, this is the recommended way to write to any file system if you want your files to be portable regardless of adios (assuming from your statement that using non-portable `long int` is intentional).

```And for checkpoint / restart, one really usually doesn't care if checkpoint files are not portable, and that's a use case that I believe is important for adios2 to support.``` I wouldn't generalize based on your use case or application. C/P can be used/converted for analysis and we are currently working with LAMMPS, see for example: https://lammps.sandia.gov/doc/Run_options.html#restart2data

One thing is that our CI tests are based on fixed-width types. I'd look into https://github.com/ornladios/ADIOS2/blob/master/testing/adios2/interface/TestADIOSDefineVariable.cpp   GetType<unsigned int>, and other non-portable types, will likely still break. 

Keep in mind that we are also talking about maintenance cost. Ultimately, do we want to maintain non-portable types at the library layer? I personally don't and I'd ask any user to move to fixed width types, but I'd like to hear from everyone in team.

**Comment by germasch** [comment]
I rebased my PR (the first commit is otherwise still identical), and added some more commits that almost complete mapping integer types to stdint types at the cxx11 interface level.

The one thing I haven't touched is the Operator stuff, since I'm not familiar with that stuff at all yet, but I don't think there's any major problem with doing the same there.

I have a bunch more changes sitting here, converting the C interface and also the backends (ie, Engines) to only interface to the core layer by the 8 stdint types. A similar discussion can be had about the C interface, though there are even more options on how to handle it, since it's not as language-type constrained. (E.g., one could decide not to have an `adios2_type_long` at all, as right now the primitive type are considered distinct from the stdint types in that enum, as opposed to C++, where there's no way to distinguish aliased types).

So from what I have I can tell that it's possible to have the core layer only be instantiated for the 8 stdint types rather than the 11 primitive C++ integer types, and none of the tests break, which I guess isn't too surprising since the tests are generally only testing the 8 stdint types, which are just mapped to themselves in these changes (at compile time, so there's no performance penalty in this mapping).

Diving a bit deeper in the code I think gave me a bit more of a perspective on how things are. So I would describe this as the core being built around the 11 primitive integer types, see e.g. ADIOS2_FOREACH_TYPE_1ARG and many other places that also repeat these 11 types. So if the goal is for the core to be written around the 11 primitive types, it'd obviously make sense to also have the C/C++ interfaces be based around these 11 types, and there's no argument for the mapping I'm doing in this PR. However, that mapping still happens, which is why the `DefineVariable<unsigned long>` won't be found with `InquireVariable<unsigned long>` when reading back, the issue that triggered this whole discussion.

The mapping right now happens in the backends, e.g., in the bp3 engine. I was actually wondering whether other engines might even handle things differently. the bp3 engine (on MacOS-64) will map `unsigned long` to what it calls `type_unsigned_long` (which is a somewhat unclear name for unsigned 64-bit) when writing. And when reading back, it'll map `type_unsigned_long` to `uint64_t` a.k.a. `unsigned long long`. So on my system there is no way bp3 could possibly ever create `unsigned long` variable from a file. However, the hdf5 code has
```cxx
    else if (H5Tequal(H5T_NATIVE_ULONG, h5Type))
    {
        AddVar<unsigned long>(io, name, datasetId, ts);
    }
```
so in theory it could create an `unsigned long` variable when reading from a HDF5 file. However, in practice, my unsigned long variable gets written as `H5T_STD_U64LE`, which maps to a `uint64_t` when read, so it does behave exactly like bp3 -- which is good.

Having the engines determine the exact behavior as it comes to type could be seen as a flexible design -- however, it could also be considered dangerous, as the application might get subtle behavior changes as a consequence of changing backends. It is my understanding that a goal of ADIOS(2) is to make backend configuration changes as transparent as possible. With that in my mind, I don't think this flexibility at the backend to decide how to treat the 11 primitive types is actually a good thing.

germasch:
Let me start with this:

> Keep in mind that we are also talking about maintenance cost. Ultimately, do we want to maintain non-portable types at the library layer? I personally don't and I'd ask any user to move to fixed width types, but I'd like to hear from everyone in the team.

So I think one thing we're strongly in agreement about is that at the core, adios2 should just support the 8 stdint types, and that it shouldn't have to deal with the ambiguity of having multiple types representing the same actual data type (ie, the core should support `uint64_t`, not `unsigned long` and `unsigned long long` (again, I'm using 64-bit mac/linux as example). My patches on top of this PR actually make this happen, ie., I'm switching things to 
```cxx
#define ADIOS2_FOREACH_STDTYPE_1ARG(MACRO)				\
    MACRO(std::string)							\
    MACRO(int8_t)							\
    MACRO(uint8_t)							\
    MACRO(int16_t)							\
    MACRO(uint16_t)							\
    MACRO(int32_t)							\
    MACRO(uint32_t)							\
    MACRO(int64_t)							\
    MACRO(uint64_t)							\
    MACRO(float)							\
    MACRO(double)							\
    MACRO(long double)							\
    MACRO(std::complex<float>)						\
    MACRO(std::complex<double>)
```
throughout the core / engines. (It's mostly complete as far as the core goes, but the engines could certainly benefit of some clean up as a consequence of now only having to deal with a set of unique, well-defined types).

So if we do indeed agree that the core / engine should just deal with the 8 unique stdint types, the open issue is really only about the cxx11 and c interfaces.

One of the questions, in general, is about how stable adios2 is considered at this point. The user-visible changes in this PR are quite small, but it does change behavior (e.g, it changes that `DefineVariable<unsigned long>` can be found via `InquireVariable<unsigned long>` after a return trip through a file (or across the network, I suppose). And there's probably more that I didn't think of -- in fact, I'll mention something when I get to the tests.

I think it would be perfectly fine for adios2 to only support the 8 stdint types at the c/cxx11 interface level as well. In fact, that would be safer to use to some extent, as making a `unsigned long` Variable just wouldn't work at all. However, as you pointed out, due to C++ limitations, it would support all primitive types but `char`, `long` and `unsigned long` (on MacOS-64), not because it wants to, but because those are the ones that the stdint types are aliased to. So it won't really prevent users from using those types, and once users do so, their code will still be non-portable. It's even possible that on some other system, `unsigned long` would work, but `unsigned long long` wouldn't, depending on how the system defines `uint64_t`. In the C interface, fundamentally the same issue exists, but one could be more explicit about which types are supposed to be used, ie., not providing a `adios2_type_unsignd_long`, or `adios2_type_unsigned_long_long`, just `adios2_type_uint64_t`.
That wouldn't prevent the user from saving a `unsigned long` value in a `adios2_type_uint64_t` variable, but well, there isn't even anything preventing the user from saving a `short` value there. either.

The alternative is to say, considering we're talking about a cxx11 interface, so we're going to support C++11 primitive types, even though that means mapping to out 8 stdint-type internal types, even though that'll introduce some amount of ambiguity there. That is my preferred solution, and it's definitely the more backwards compatible one. I think that it's possible to define proper semantics to go with it, but obviously if the user chooses to uses primitive types in interfacing with adio2, they have to deal with portability issues,. That's really just the same kind of portability issues an application has that uses, e.g., `unsigned long` in the first place, but doesn't use adios at all.

germasch:
> One thing is that our CI tests are based on fixed-width types, hence we are not testing your PR. I'd look into https://github.com/ornladios/ADIOS2/blob/master/testing/adios2/interface/TestADIOSDefineVariable.cpp and https://github.com/ornladios/ADIOS2/blob/master/testing/adios2/engine/bp/TestBPWriteReadAttributesADIOS2.cpp
> `GetType<unsigned int>`, and other non-portable types, will likely still break as of today....so I am not sure if adios2 will be silent as you mentioned.... the Variable will be false at read if using `<unsigned long>` at write. This is how you found the non-symmetric `unsigned long` at first.

So you're right that there are still issues to be addressed, and I haven't even looked into all of this. There's definitely an issue with GetType:
* Current behavior:
  *  If one does `DefineVariable<unsigned long>`, then  `VariableType() == GetType<unsigned long>()`
   * If one manages to read back that variable, it'll have `VariableType() != GetType<unsigned long>()` (because it's now `unsigned long long`)
* Behavior after this PR: 
  *  If one does `DefineVariable<unsigned long>`, then  `VariableType() != GetType<unsigned long>()`
(type is now `unsigned long long`)
  * One will manage to read the variable back easily with `InquireVariable<unsigned long>`, but still `VariableType() != GetType<unsigned long>()` (because it is `unsigned long long`)

So while it was half broken before, it's now entirely broken, which is a step backwards. However, that's essentially just a bug in my work: What `GetType<..>()` should do is to return `uint64_t`, etc. That will make both cases succeed, and variables behave the same before and after the round trip to disk.

This would be a very quick thing to do given that I just need to use the same mapping to `TypeInfo::IOType` in `GetType`, however there is a bit of a fundamental separation issue in that there are no separate cxx11 vs core versions of `GetType`, so if I change `GetType` as is, it'll affect all of the core/engines. I'm actually fairly sure that I could do it without breakage after the end of my patch series, since the core/engines then only use the 8 stdint types in the first place. But really, there should be a separate GetType for the cxx11 interface. I, by the way, also think that the core GetType shouldn't return a string in the first place, but just an enum class, as is used for the rest of the adios types, and maybe even that can go away eventually, since the type is already encoded in the Variable<...> type, anyway. (One could make it easier to recover with something like a `using value_type = T;` kind member in the class.) I actually don't see much point in having `GetType` or `VariableType` in the cxx11 interface, as the user definitely already knows the type because they couldn't have made a `Variable<T>` in the first place otherwise. And if they just want it as a string for debugging or something, `typeid(Var::value_type).name()` will do (okay, so maybe that's worth a convenience function).

Anyway, in summary, my point is, it is possible to make a cxx11 interface based on the semantics "every unsigned 64-bit type" will behave the same way. So whether you use `unsigned long long`, `uint64_t`, or `unsigned long` for your value and for the `Variable<>` template argument, the behavior will be identical (on Linux/MacOS-64). If you used `uint64_t`, your code will be portable, and if you don't, you're on your own.

germasch:
> `More generally, though, there is a lot of code out there that uses primitive C types all over, and where one doesn't want to change those to stdint types because who knows what will break.` They don't need to change their current computation code that lives in memory. When integration with adios2 one can do a simple mapping at the glue layer, this is for write, but is the same idea for read: `if( sizeof(unsigned long) == 8 ) io.DefineVariable<uint64_t>.... else io.DefineVariable<uint32_t>` in fact, this is the recommended way by the standards committee https://isocpp.org/wiki/faq/newbie#choosing-int-size which applies to writing to any file system if you want your files to be portable regardless of adios (assuming from your statement that using non-portable `long int` is intentional).

I basically said this before: Having adios2 support only stdint types is one reasonable option. In that case having me do my own mapping (which I can even steal from your `TypeInfo`) is not a problem. Also, to explain where it comes from: On each proc, I get the number of patches on that proc by essentially `patches.size()`, where `patches` is a `std::vector`, and I want to write that into my file so that I can recover the specific domain decomposition I was using. I actually know this will always fit into an int, so I could use int, or more explicitly, int32_t. But I find it generally preferable to not muck with types if I don't have to because the "natural" type is good enough -- and `size()` return a `size_t`, so that's what I used. These were my first steps with adios2, and what happened was rather non-intuitive, I could make a `Variable<size_t>`, but I could not read it back. I know understand the why very well, but that interface (allowing to create a Variable of a certain type, but not letting me read it back) is broken in my opinion. If I had gotten an error when using `Variable<size_t>`, I still wouldn't have liked that interface, but it's a choice ("only use stdint types, and unfortunately when you use primitive types, some will work because the stdint types are aliased to them, and others are not") that's reasonable given the limitations of the language. But `size_t` is admittedly a borderline case, because it's definitely not portable. However, `char` is also broken the same way, and would stop working completely when using stdint types only. Obviously, `char` is a rather commonly used type, and I don't think one that's unreasonable to use, even though it's strictly speaking also not portable.

> `And for checkpoint / restart, one really usually doesn't care if checkpoint files are not portable, and that's a use case that I believe is important for adios2 to support.` I wouldn't generalize based on your use case or application. C/P can be used/converted for analysis and we are currently working with LAMMPS, see for example: https://lammps.sandia.gov/doc/Run_options.html#restart2data

I'm not saying that it wouldn't be better to have portable checkpoint files -- of course it is. But it's also much better to have high-performance non-portable checkpointing with adios than having non-portable file-per-processor checkpointing or none at all. So the question in my opinion is, do you want to say "you can only use adios2 if you make your checkpointing completely portable"? I think, nicely put, most HPC codes in research aren't exactly perfect... 
To the particular point with my code, my plan is to use adios2 "the right way" as far as writing out the actual data in my checkpoint, ie., making a global array with all the pieces ending up in the right spot, which isn't trivial for my application, it'd be easier to just kinda do the "LocalArray" approach, every proc dump their stuff. So that I can actually (portably) get my data out from my checkpoint files if it makes sense for analysis etc. But the part about which patch lived on which proc is only used for actual restart, so I don't really care about portability. Also, while I haven't used the python interface, I would assume that it'll allow me to open a variable by name without having to provide the datatype in the first place, so I would get my number of patches back as 32-bit or 64-bit depending on where I ran the code, but I'd have it, anyway. Just in case I could actually managed to find a 32-bit computer to run on... ;)

williamfgc:
@germasch 

```What `GetType<..>()` should do is to return `uint64_t`, etc. That will make both cases succeed, and variables behave the same before and after the round trip to disk.``` I like this solution a lot!

I think we are getting somewhere, I agree with all your points, but two:

```I actually don't see much point in having GetType or VariableType in the cxx11 interface, as the user definitely already knows the type because they couldn't have made a Variable< T > in the first place otherwise.```
At Reading variables are defined internally at run-time, while at writing DefineVariable< T > is resolved at compile-time. So we need to use the MACRO for casting first if we don't know an incoming file. Having something like GetType<T> in a MACRO will simplify the if statements for casting. bpls uses this heavily as this is an interactive command-line utility. Sure we could have used an enum, but for the most part GetType is used for information purposes and a string naturally fits into std::cout or std::exception message, or prints from bpls. An enum wouldn't buy us much, especially since C++11 small strings are optimized and stored in the stack, not heap.

```Also, while I haven't used the python interface, I would assume that it'll allow me to open a variable by name without having to provide the datatype in the first place``` There is 2 independent levels for the Python interface, high-level and low-level. The high-level will return a numpy array from a read statement, the user doesn't require the type or pre-allocation. The low-level has the same requirements as C++11, C, Fortran low-level APIs: requires the right pre-allocated numpy array and type. Sounds like the high-level API would do the job for what you need, we use it heavily in analysis code. The low-level is still required for workflows on the big-machines and in-memory staging engines. But for files, it won't make a difference in performance.

```But it's also much better to have high-performance non-portable checkpointing with adios than having non-portable file-per-processor checkpointing or none at all.``` I think we are mixing two things:  1) what's important, 2) the scenarios we must support as a library. In fact, I will go as far to say that most of the time c/p is not even used, but discarded. That doesn't mean we don't have to support other scenarios. Also, I wouldn't worry about mixing performance and the types topic, as you said everything is for free, done at compile-time. Performance is adios2 is given by the algorithms, IO strategies and flexible parameters at the user interface.

**Comment by germasch** [comment]
> `` What `GetType<..>()` should do is to return `uint64_t`, etc. That will make both cases succeed, and variables behave the same before and after the round trip to disk. `` I like this solution a lot!

I've implemented this, and I'll push it in a minute. As most everything I've been doing, it's only lightly tested, but then again it's also a pretty straightforward change.

> `I actually don't see much point in having GetType or VariableType in the cxx11 interface, as the user definitely already knows the type because they couldn't have made a Variable< T > in the first place otherwise.`
> At Reading variables are defined internally at run-time, while at writing DefineVariable< T > is resolved at compile-time. So we need to use the MACRO for casting first if we don't know an incoming file contents when coding. Having something like GetType in a MACRO will simplify the if statements for casting. bpls uses this heavily as this is an interactive command-line utility. Sure we could have used an enum, but for the most part GetType is used for information purposes and a string naturally fits into std::cout or std::exception message, or prints from bpls. An enum wouldn't buy us much, especially since C++11 small strings are optimized and stored in the stack, not heap.

Well, I don't see a need to change what `GetType()` returns right now, so one could have a discussion about it later.

> `Also, while I haven't used the python interface, I would assume that it'll allow me to open a variable by name without having to provide the datatype in the first place` There is 2 independent levels for the Python interface, high-level and low-level. The high-level will return a numpy array from a read statement, the user doesn't require the type or pre-allocation. The low-level has the same requirements as C++11, C, Fortran low-level APIs: requires the right pre-allocated numpy array and type. Sounds like the high-level API would do the job for what you need, we use it heavily in analysis code. The low-level is still required for workflows on the big-machines and in-memory staging engines. But for files, it won't make a difference in performance.

That sounds very reasonable.

> `But it's also much better to have high-performance non-portable checkpointing with adios than having non-portable file-per-processor checkpointing or none at all.` I think we are mixing two things: 1) what's important, 2) the scenarios we must support as a library. In fact, I will go as far to say that most of the time c/p is not even used, but discarded. That doesn't mean we don't have to support other scenarios. Also, I wouldn't worry about mixing performance and the types topic, as you said everything is for free, done at compile-time. Performance is adios2 is given by the algorithms, IO strategies and flexible parameters at the user interface.

Okay, point taken. I didn't mean to say that supporting stdint vs primitive types will have a performance impact, what I meant is just that supporting not all the native C++ types is a potential barrier to adaption of ADIOS2, hence making it harder to switch whatever home-grown solution to the high performance that adios can provide.

**Comment by germasch** [comment]
I pushed some more changes to make the cxx11 level `GetType<T>()` consistent with the rest
of the C++ type -> IOType mapping.

This branch is now failing to merge to the updated master, but I think I'll skip rebasing at this time i order to not mess up this thread.

I decided, for now to stick with the existing strings for compatibility. That is, while `unsigned long` gets mapped to `uint64_t`, `uint64_t` is an alias for `unsigned long long`, that's why the existing `GetType<uint64_t>()` returns "unsigned long long int".

Changing that would break various tests, e.g.: in `TestADIOSDefineAttribute.cpp`
```
    EXPECT_EQ(attributeU32.Type(), "unsigned int");
```

It would actually be more logical to change both the test and the GetType to return "uint32_t", because that's the type used when defining the attribute in the test, but it's a backward incompatible change, so I don't know what your stance is on that.

Also noteworthy: For `uint64_t` there is no test of the type vs an expected string, because, I suppose, that type might be implementation dependent. I think that's an argument for going to "uint64_t" type strings. Though, I suppose, really these tests should be using `GetType<uint64_t>()` rather checking the strings directly in the first place.

Still, changing the strings to stdint based names is something that is much more easily done at the end of this work, when core is completely converted to use only stdint types.

There is some more behavior change that might be worth a discussion:
`GetType<T>()` previously returned "compound" for other (non-specialized) types, and "unknown" for `void`.

The new mapped `GetType<T>()` should give a compile time error when T is not one of the standard supported primitive types. The special `case` for void could be easily put back, but it doesn't really make any sense to me.

The "compound" case is harder for me to judge, because I don't really know where the compound support is going. Right now, I don't think there's any case where `GetType<>` is used on such a type, because it would have broken. Returning "compound" could be put back in with some std::enable_if hackery. But I'm not sure it's really the right thing in the first place. Do we want
`GetType<struct A> == GetType<struct B>`? Or similarly, if "varA" is a variable to handle `struct A`, do we want `VariableType("varA") == GetType<struct B>()`? That doesn't sound like such a good idea.

I haven't thought this through, but in the C++11 interface, one might consider some kind of `traits` approach to have ADIOS2 learn about compound types that one wants to be supported, and having that could be a mechanism to return "struct A" or "struct B" rather than always "compound".

Anyway, finally, I added a commit that extends one of the tests to make sure that `VariableType() == getType()` for non stdint types, which is really the case which is fixed by this latest set of changes, and which was broken by the -> IOType mapping earlier in this PR.

williamfgc:
```It would actually be more logical to change both the test and the GetType to return "uint32_t", because that's the type used when defining the attribute in the test, but it's a backward incompatible change, so I don't know what your stance is on that.```

Please go ahead and change both (GetType ->"int32_t" and tests), my stance is that, ideally, I wouldn't want to deal with the ambiguity of non-portable types, and this is a good solution to get rid of that. 

```The "compound" case is harder for me to judge, because I don't really know where the compound support is going.```
Let's not worry about compound types, what you found is boiler-plate for a potential fit, but no one cares in HPC (at least the ones we faced) to justify a full implementation. They are POD so they don't add much performance value.

germasch:
> Please go ahead and change both (GetType ->"int32_t" and tests), my stance is that, ideally, I wouldn't want to deal with the ambiguity of non-portable types, and this is a good solution to get rid of that.

I will do this, but in the end, that is, after converting core to only use the stdint based types, which will take more work (which I've mostly done here, locally). Doing it now is too much of a mess, since core still supports the 11 types (though they aren't really all used). So core should keep using its helper::GetType which needs to use primitive types for now, and while I could change the cxx11 GetType, type strings are passed back from core to cxx11, which would then need to be converted...

> Let's not worry about compound types, what you found is boiler-plate for a potential fit, but no one cares in HPC (at least the applications we face) to justify a full implementation. They are POD so they don't add much performance value. It's a nice to have, but not really a critical requirement.

Well, for what it's worth, I'd actually like to see POD support. Considering I'm dealing with particle codes, where each particle needs a mix of float and integer (indexing) data, there's really no nice way of saving them currently.

germasch:
Alright, assuming I can get the latest version I force pushed to pass the CI tests here, I'll declare it ready to merge.

The set of changes affects almost exclusively the cxx11 interface only, mapping between primitive and stdint types at the border to core. There are a couple minor fixe for core to use `helper::GetType`, and I snug in a fix for some warnings in bpls which didn't seem worth a separate PR.

This project seems to be moving at lightning speed, so I'll try to get more follow-up patches in soon, as otherwise it seems I'm dealing with fixing rebase conflicts every day.

williamfgc:
@germasch this is great, thanks! Should fix your problem. I agree, we can tackle core later (it's an internal thing, anyways).

germasch:
@williamfgc, I saw a revert-1154-cxx11_map_to_stdint branch, so what did I break?

By the way, your changes to the C interface broke the matlab interface, I have fixes for that.

williamfgc:
@germasch it's all good...I just took the easiest path to see if your PR broke adios2 after `make install`. It was another PR.

germasch:
@williamfgc no, I'm not using (or even building) the matlab interface, but just grepping for the old c interface types turned this up. A (untested) fix is in PR #1169.

PR #1160, opened 2019-02-11
Bump release version to 2.3.1

chuckatkins:


chuckatkins:
This should be held as the last commit before the release

PR #1164, opened 2019-02-12
WIP: support local arrays with bpls. Distinct dimension format printing: s…

pnorbert:
…teps*[blocks]*{dim0,...,dimN}, and print __ for varying values in each position

WIP:
- Waiting for Block Selection to work on global arrays (then bpls code can be simplified)
- Waiting for solving an issue with potential zero number of blocks in a step


pnorbert:
Was merged into release instead of master

PR #1169, opened 2019-02-13
C bindings fixes / updates

germasch:
Originally, this work had meant to do the same kind of type mapping as the cxx11 interface, that is,
map, e.g., `adios2_type_unsigned_long_long` and `adios2_type_unsigned_long` (on Mac/Linux-64) to the same underlying `uint64_t` for core adios2.

That part is actually not needed anymore, since in master those primitive variable based types are mostly gone. So this is mostly some cleanup and consistency improvements.

* fix the Matlab interface which got broken by changes in master that removed the primitive adios2_types (untested, though, and it looks like the CI doesn't build it, either.)
* remove 'adios2_type_{un,}signed_char`, since all the other primitive types are already gone
* get rid of the type-string to adios_type maps, which were at least slightly broken because of a hardcoded mapping of `unsigned long` to always `uint64_t`
```cxx
       {"unsigned long int", {adios2_type_uint64_t}},
```
* change the various macro if-loops to only iterate over the actual stdint-based types, rather than primitive types
* reintroduce `adios2_type_long_double`. I'm not sure this was intended to be dropped, I don't personally care about it, but it's supported by core and cxx11, so why just drop it from the C interface?

### DIFF
--- a/bindings/C/c/adios2_c_attribute.cpp
+++ b/bindings/C/c/adios2_c_attribute.cpp
@@ -12,30 +12,11 @@
 #include "adios2/ADIOSMacros.h"
 #include "adios2/core/AttributeBase.h"
 #include "adios2/helper/adiosFunctions.h"
+#include "adios2_c_internal.h"
 
 #ifdef __cplusplus
 extern "C" {
 #endif
-
-namespace
-{
-const std::map<std::string, std::vector<adios2_type>>
-    adios2_attribute_types_map = {
-        {"int", {adios2_type_int32_t}},
-        {"float", {adios2_type_float}},
-        {"double", {adios2_type_double}},
-        {"signed char", {adios2_type_int8_t}},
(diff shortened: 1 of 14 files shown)


germasch:
FWIW, it's actually very little work to implement the primitive -> stdint type mapping on top of this, which would allow to support `adios2_type_{primitive}`, including `adios2_type_char`. There are argument for and against it, so I made this PR independent of that discussion.

williamfgc:
@germasch `long double` is not a standard type and very corner case, it doesn't justify maintenance costs so we dropped. Do you mind dropping it in your PR before I can merge? If not, not a big deal, I can remove it later.  Thanks!

germasch:
@williamfgc I just force-pushed the branch after rebasing onto latest master and dropping the `long double` commit (which was already the last commit, but the order that github shows commits gets confused be my rebasing...) 

Assuming the checks pass, it should be good to go.

PR #1205, opened 2019-02-18
stdint-type updates

germasch:
This are a couple of relatively small changes that aim to avoid the use of hardcoded type strings
within adios2. For example, replace `"unsigned long long int"` by `GetType<uint64_t>()`.

If this were done in all possible places, things shouldn't actually break if one were to rename `"unsigned long long int"` to `"uint64_t"` for clarity eventually, however, there are still more places that use hardcoded strings, and they're actually difficult to identify, since there also libraries (e.g. ffs) that also use strings for types, though not necessarily the same ones.

This patch changes behavior in one spot, where it now allows DataMan to use ZFP compression for `int64_t`, which was previously broken as it checked for `"int64_t"` and `"long"` type. Both have never been actual type strings, there used to be `"long int"`, but that's also gone now.


germasch:
So adding tests back of 64 bit quantities was actually useful in that it showed an issue where bpls
would show different type names for 64-bit quantities, depending on whether uint64_t is an alias for unsigned long or unsigned long long. I believe the output from bpls may have been pretty much the only thing affected, since internally `GetType<>()` is used almost everywhere, so that actual string doesn't matter as long as both sides of the comparison use `GetType<>()` based strings.

PR #1216, opened 2019-02-20
More stdint-type updates

germasch:
I'm continuing my quest to explicitly use only stdint-based types in core and below.

This series updates format/bp3, format/bp4 and interop/hdf5 correspondingly, while also avoiding
a good amount of repetitiveness by using ADIOS2_FOREACH_....

diffstat speaks for itself: `16 files changed, 321 insertions(+), 1412 deletions(-)`  ;)

### DIFF
--- a/source/adios2/toolkit/format/bp3/BP3Base.h
+++ b/source/adios2/toolkit/format/bp3/BP3Base.h
@@ -360,6 +360,12 @@ protected:
     };
 
     /**
+     * Maps C++ type to DataTypes enum
+     */
+    template <typename T>
+    struct TypeTraits;
+
+    /**
      * Characteristic ID in variable metadata
      */
     enum CharacteristicID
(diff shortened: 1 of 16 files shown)


germasch:
I wish I could ever make a pull request which passes checks on the first try... If anyone knows what I'm doing wrong with clang-format, let me know:

When I ran `git clang-format master` before pushing the branch, it had no changes, but the CI complained about missed formatting in two files. When I ran `clang-format -i` on those files, they were reformatted all over, instead of the two smallish patches suggested by the CI...

williamfgc:
@germasch we use clang-format v3.8 , it's a bit old, but clang-format is not backwards compatible for the most part. As a side note, I use the Eclipse IDE with the CppStyle plugin so every time a file is saved clang-format is applied directly. There might be a similar plugin for your development environment of choice. I haven't had issues since then.

chuckatkins:
I'd actually like to bump the version of clang format we use to something current.  It'll depend on what's easily available to people though.  That's a different conversation entirely though.

PR #1218, opened 2019-02-22
introduce DataType as separate type

germasch:
Alright, so this is going to be lengthy (both the code and this description). I can break this up into pieces that could be merged one at a time. (And actually, there are a couple more things that could/should be done on top of this, too).

I hadn't really planned to dive into ADIOS2, and I'll probably mostly disappear from here again soon, but I feel like I would like to see through what I started. This all began when I hit problems due to adios2 internally being templated for 11 integer types, while it really only could properly distinguish the 8 stdint types `({u,}int{8,16,32,64}_t`. The PRs I provided so far, which moved most of adios2's internals to the stdint-based types have already been merged. Actually, in making those PRs, I had already used something akin to the below, which made it possible to trace where and how the various types are used and what the expected values are.

The primary objective of this PR is to introduce a `class DataType` as a replacement for where right now `std::string` is used. For most of this series, `DataType` actually is just a strongly typed wrapper around `std::string`, which serves to document where strings are used to designate an adios2 types vs just some other string vs, e.g., a FFS data type. It then also narrows down `DataType` to be only one out of a set of predefined strings (and doing that, it turns out while we tested again `"unknown"`, there's actually no way a data type string would ever have been set to `"unknown"`. Finally, it actually replaces `DataType` to be just a wrapper around an `int`, rather than around `std::string`. That last part could / should be continued to actually use an enum, and doing this smartly, one could use that same enum in the C interface and in bpls, avoiding some mapping code.

This PR touches a lot of code, though most of it in fairly trivial ways. So I feel I better justify it:
* None of the bindings are affected. (Though it does open opportunities to improve them)
* `DataType` is strongly typed (it converts only explicitly to/from a string where needed)
* `DataType` enforces a predefined subset of types, so no typos or accidental new inventions are possible. (This had happened, though not in ways that actually caused problems, e.g., comparing a type string to `"int64_t"`, which is not a value it ever could have had.
* `DataType` gets us close to being able to rename the type strings from the kinda unpleasant primitive-based names (`"unsigned long long int"`) to the much clearer `"uint64_t"`.
* adios2 uses enum classes for pretty much every other choice, only the data type is represented by `std::string` right now.
* While performance is obviously going to be mostly dominated by network and file system, the code passes type strings around a lot, and compares them even more (all the ADIOS2_FOREACH_* loops). Some of them are actually in places which I think are somewhat performance sensitive, like in `adiosMemory.h`. Small string optimization means that for the most part dynamic memory allocation is avoided, at least, though I believe `std::string("unsigned long long int")` actually triggers a malloc on gcc. Comparing strings is definitely more complex than comparing integers, too.
* As a somewhat arbitrary indication, I ran `size libadios2.2.3.1.dylib` on my mac:
```
master:
__TEXT	__DATA	__OBJC	others	dec	hex
16019456	98304	0	6811648	22929408	15de000

after this PR:
__TEXT	__DATA	__OBJC	others	dec	hex
15056896	98304	0	6815744	21970944	14f4000
```
  That's actually a rather noticeable (1MB) decrease in __TEXT (actual machine code). I believe this mostly because I statically allocate the possible DataTypes, rather than creating the strings on demand all the time, not so much because of the string -> int change.

FWIW, after these changes I can tell that there are only a few places left that use the string representation of types. They almost all use the (successor of) `GetType<T>()`, so they are independent of the actual string chosen. The exception is `ffs_marshal.c` which as a C file cannot employ that mechanism, and hence would benefit from available enum (Though this PR doesn't go there yet). After that, the string representations are only used in debug messages and in the bindings.
E.g. `bpls` output would be affected, and I'm not sure we'd want to go there, but that's another discussion.


germasch:
Besides my usual clang-format troubles, the build apparently fails on Windows with undefined references to the static data members of the `DataType` class. They are there, in `ADIOSTypes.cpp`, and this apparently works as it's supposed to on Linux/Mac. Maybe this is related to special things one has to do to export symbols from DLLs?

williamfgc:
@germasch There is another important benefit of using `std::string` instead of `int` or `class enum` which drove the decision for this particular case: debugging (with serial and parallel debuggers or the old fashion cout) and building tests. This is where most of us (including myself) spend most of our time when creating a MPI-based production library (coding itself is not really time consuming). 

Having `std::string` in a template-primitive-type-based framework helps a lot when debugging, not just at the library level, but when constructing tests. I'd rather see a `float` than `1` when debugging or throwing an exception. We use enum very deep in BP as that's the way types are stored in the buffer. When I see the cost/benefit of using `std::string` as you shown above (it's still not clear to me the overall practical benefit of switching to `int` or `enum`), yeah I'd rather stick with strings. 

In fact, I'd rather switch GetType< T >( ) output to fixed width types. You're right, `unsigned long long int` might not use SSO (in gnu 15 characters), so that's another reason to implement fixed width types as they will all fit in the stack.

eisenhauer:
OMG, this is huge!  All those string comparisons at the core of virtually everything that happens in ADIOS2 gone!

williamfgc:
@eisen I would rather move SST to C++11 to enjoy the benefits of SSO

eisenhauer:
LOL!

williamfgc:
@eisen please, let's keep it professional

germasch:
So I think there are some misconceptions here that SSO solves all the world's problems. SSO is a good thing, but what it addresses is really only avoiding dynamic memory allocation when handling small strings.

## Passing strings

The FAQ says

>C++11: Why are std::string arguments passed sometimes by value and sometimes by reference?
>C++11, provides mechanisms to optimize copying small objects, rather than passing by reference. The latter was always the rule for C++98. When a string is passed by value, it’s assumed that the name will be short, <= 15 characters, most of the time. While passing by reference indicates that the string can be of any size. Check the isocpp guidelines on this topic for more information.

If you look at the [link](http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#f15-prefer-simple-and-conventional-ways-of-passing-information) referenced in there, you will see that it explicitly lists std::string and tells you to pass std::string by reference (always). That really doesn't have anything to do with SSO, or if so only in that SSO makes sizeof(std::string) larger, something like 24 or 32 bytes, which means it's cheaper to pass the 8 byte pointer/reference and take the the hit of the indirect access.

## Comparing strings and comparing integers are nowhere near the same cost

Nor does that really have anything to do with SSO, since the actual strings are accessed via pointer always. (In the case of SSO, the pointer goes into the std::string struct, so it'll be in cache at least).

From https://godbolt.org/z/alDYkd
```
bool compare<int>(int, int):
  cmp edi, esi
  sete al
  ret
bool compare_ref<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > >(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&):
  mov rdx, QWORD PTR [rdi+8]
  xor eax, eax
  cmp rdx, QWORD PTR [rsi+8]
  je .L12
.L9:
  ret
.L12:
  mov eax, 1
  test rdx, rdx
  je .L9
  sub rsp, 8
  mov rsi, QWORD PTR [rsi]
  mov rdi, QWORD PTR [rdi]
  call memcmp
  test eax, eax
  sete al
  add rsp, 8
  ret
```
The machine code generated for comparing two strings has two conditional jumps and potentially a function call, vs essentially a single instruction for integer comparison.

Now, is this going to have measurable impact on adios2 performance? Not likely, except maybe the before-mentioned case of the copy stuff in adiosMemory.inl.

williamfgc:
@germasch I agree with your points and especially your point about impact on adios2.The fact that SSO puts strings in the stack (which the only aspect I care, the others are not measurable for what we do) and allows better debugging is sufficient to benefit the developer/maintainer.

germasch:
And here's my take on debugging:

I agree that enum's can be a bit of a pain in particular for manual debugging -- in fact I hit that when I was trying to figure out the changing ShapeID when writing LocalValue / reading GlobalArray -- printf style debugging will just give you a number. I actually have a suggestion for that:

```
template<typename T>
std::string ToString(T enum_value);
```
That function could be specialized for ShapeID and the various other enum classes, and one could add an ostream operator for them. That would make it easy to get human-readable value (automatic, even, if using `std::cout << my_shape_id;`.

Otherwise, as far as debugging goes, my debugger is quite happy with enum classes.
```
(lldb) p variable
(adios2::core::Variable<unsigned long long> *) $0 = 0x0000000102265868
(lldb) p variable->m_Type
(const adios2::DataType) $1 = UInt64
```
(this is after actually changing DataType to an enum class -- this PR gets you most of the way there).

germasch:
By the way, going all the way to the enum class also gets rid of the static class members which cause DLL symbol troubles for whatever reason.

williamfgc:
@germasch this is why sticking with SSO strings in the C++11 pieces of the code. It is already taking care of those aspects with zero overhead. Keep in mind, several times we can't use modern serial compilers (lldb) in the DOE big machines (our bread and butter) and parallel debuggers can take a while to setup and launch (debug mode) and a cout statement is very common and does the job. I am not against `enums` (you pointed out we use it everywhere else), but for this particular case enums are not buying us much (actually nothing) for what we do/need.

williamfgc:
Closing this as this was a design decision. Switching to enum do not represent any advantage over the current implementation.

chuckatkins:
I'd like to leave this open for the moment and discuss it amongst the dev team on our weekly tcon.  This specific issue came up quite some time ago but nobody had the bandwidth or priority for it so we didn't pursue.  With the bulk of the work already done here, @germasch thank you BTW it's very much appreciated, I think it's worth tabling the discussion here and revisiting it with the team next week.

@germasch we'll revisit this PR in a few days after we talk it over internally.

williamfgc:
@chuckatkins sorry, but I am going to be very vocal about this one. This was settle 2 years ago, I don't want to discuss things already discussed that do not add considerable value to the developer. @germasch appreciate your input, but ultimately I am the one putting up with all the maintenance costs, and we have much bigger problems to solve.

sklasky:
Please do NOT close this request! This seems to be something that we should have someone at least evaluate this. If this is more efficient then I want to know this. Perhaps we need to get Lipeng to do this after he gets done his paper. I need someone on top of this since ecp apps (Kai) are they customers and we must listen to our customers. I don’t have the expertise to say what’s correct but if we can test and Kai’s way works better on summit then I think that I would want this.. so if William is right then some time lost but it’s better to error in the side of the customer and learn ...

PR #1219, opened 2019-02-22
bug fix...

guj:


chuckatkins:
What bug is this fixing?

guj:
@chuckatkins. Oh it is something simple but necessary to fix. Supposed to check with "> 0" but I saw that I did was "> 1". This is needed to add to release branch.  Can I add to it? 

By the way, I saw that on mac, the libraries at build/ kept the rpath, and the libraries in install/ were not.  Is this on purpose?

chuckatkins:
> Oh it is something simple but necessary to fix. Supposed to check with "> 0" but I saw that
> I did was "> 1". 

I can see that :-)   If you wouldn't mind a more descriptive commit message in the future it would be appreciated.  Maybe describe the symptom that the bug fix resolves.

> This is needed to add to release branch. Can I add to it?

We cut the release earlier this week so unfortunately it's too late.

> By the way, I saw that on mac, the libraries at build/ kept the rpath, and the libraries in install/ were not. Is this on purpose?

That is by design.  It's the default behavior of CMake on both Mac and Linux.  There are ways of adjusting that though at the project level that we can explore of necessary.

PR #1222, opened 2019-02-22
A little bit of clean up

germasch:
This one is hopefully not so controversial:
* Use `DataMapEntry` instead of std::pair to make the code more readable.
* do a bit of header cleanup
* remove one left over IOType mapping bit which is not necessary anymoe

### DIFF
--- a/source/adios2/core/IO.cpp
+++ b/source/adios2/core/IO.cpp
@@ -154,8 +154,8 @@ bool IO::RemoveVariable(const std::string &name) noexcept
     if (itVariable != m_Variables.end())
     {
         // first remove the Variable object
-        const std::string type(itVariable->second.first);
-        const unsigned int index(itVariable->second.second);
+        const std::string type(itVariable->second.m_Type);
+        const unsigned int index(itVariable->second.m_Index);
 
         if (type == "compound")
         {
@@ -198,8 +198,8 @@ bool IO::RemoveAttribute(const std::string &name) noexcept
     if (itAttribute != m_Attributes.end())
     {
         // first remove the Variable object
-        const std::string type(itAttribute->second.first);
-        const unsigned int index(itAttribute->second.second);
+        const std::string type(itAttribute->second.m_Type);
+        const unsigned int index(itAttribute->second.m_Index);
 
         if (type.empty())
         {
@@ -304,7 +304,7 @@ IO::GetAvailableAttributes(const std::string &variableName,
             }
         }
 
-        const std::string type(attributePair.second.first);
+        const std::string type(attributePair.second.m_Type);
         attributesInfo[name]["Type"] = type;
 
         if (type == "compound")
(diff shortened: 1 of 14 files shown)


williamfgc:
@germasch what problem are we exactly addressing? ADIOS2 is not expressive enough? It seems to be this is becoming too subjective.

source/adios2/core/IO.cpp
```diff
@@ -198,8 +198,8 @@ bool IO::RemoveAttribute(const std::string &name) noexcept
     if (itAttribute != m_Attributes.end())
     {
         // first remove the Variable object
-        const std::string type(itAttribute->second.first);
-        const unsigned int index(itAttribute->second.second);
+        const std::string type(itAttribute->second.m_Type);
+        const unsigned int index(itAttribute->second.m_Index);
```
germasch: Picking this as an example.

williamfgc: yeah, we are mapping to a local variable called `type` and `index`....seems subjective to me

germasch: But it isn't really a pair, it just happens to contain two values... I agree it is subjective, as is this article: http://maintainablecode.logdown.com/posts/158531-stdpair-considered-harmful, but that's where I'm coming from.

Anyway, should I just redo this PR without the DataMapEntry stuff?

williamfgc:
@germasch I don't want to be rude, but this is not addressing any issue per-se. We don't want PRs that break the current code style, it's there for a reason. The fact that you could read it and understand it tells me it's very clear to the point you can go and modify it. 

That being said, any PR solving a real problem (this is not the case, as I read from your explanation) is more than welcome.

source/adios2/core/IO.cpp
williamfgc: @germasch nice article, but context matters.....in this case you had no problem understanding things when looking at the entire section, not a couple of lines. I'd rather switch GetType < T > to fixed width types so bpls can present types like numpy does (without the non-portable names).

germasch:
@williamfgc I made a new PR (#1224) without the `DataMapEntry` change. I openly admit that that change added very little value, but I had made it because I was in fact confused when looking through the current code. Of course it's easy to see that `it->second.first` is the type, because it's assigned to a variable name `type`. But the kinda two-level data structure that holds Variables and Attributes is not straightforward to the newcomer, and so while I'm sure it's very obvious to you how that data structure is organized, it wasn't to me initially.

Anyway just to say it wasn't meant to be a gratuitous change, I'm generally trying to follow the style and conventions that I see, even when I personally prefer a different style. (E.g., I prefer copy-initializing variables using the `std::string a = b;` syntax over `std::string a(b)`, but I avoid changing those.)

williamfgc:
> But the kinda two-level data structure that holds Variables and Attributes is not straightforward to the newcomer, and so while I'm sure it's very obvious to you how that data structure is organized, it wasn't to me initially.

@germasch please see my internal communication. I want to be very clear: this is not about what's obvious only to me, adios2 serves a community. Decisions were made by consensus. We also work with other newcomers who didn't try to change things their way, but embrace the learning curve. Every single code base out there has (especially using a complex language like C++) has a learning curve, we just don't go changing things that work for others. In addition to this, there is the maintenance cost aspect. Each developer spends a lot of time debugging, testing, deploying, integrating with apps etc. I don't want to disturb any ones efforts. Hope this clarifies things.

PR #1225, opened 2019-02-25
Cleanup: use regular templates in .tcc for Attribute, Variable

germasch:
For some probably historic reason, in the core and cxx11 bindings, Attribute and Variable don't define
their member functions in the .tcc file as usual, but do so in the .cpp file. They don't actually define
functions of the templated class, but rather specializations for every single type, using the usual macro trickery.

While there is need for macro hackery in other parts of the code, there is no need here, so this PR cleans things up to follow the usual guide lines. There are no actual code changes, other than getting rid of some unneeded wrapping (in separate commits).


germasch:
Force pushed an update to fix the clang-format issue. Thanks @williamfgc for previously clarifying which version of clang-format is used here. Unfortunately, macports doesn't have version 3.8, so I've been using 3.7, which seems to mostly do the same thing, except when it doesn't... :(

eisenhauer:
@germasch There's a non-standard homebrew path for clang-format-3.8 that isn't too much trouble.  Instructions are here:
https://github.com/PaddlePaddle/Paddle/issues/7920

williamfgc:
@germasch this was left-over boilerplate from the migration to PIMPL. Thanks!

williamfgc:
@germasch I remember now why I did it this way, from a branch I am working on. Not all Variable (future)  functions will apply to all the types.

germasch:
@williamfgc: C++ let's you specialize classes just fine, no need for macro hackery. I cherry-picked [this](https://github.com/germasch/ADIOS2/commit/a62ecd918ab6a709047b041732bcd8b066839a92) from a branch I have that makes RemoveVariable/Attribute work properly. It replaces `VariableCompound` by `Variable<Compound>` to allow for unified handling of all Variable types.

williamfgc:
@germasch not sure I am following. Macro in Variable.cpp is just a convenience to do specialization, your changes disallowed that as Variable is always associated with all types, this is why I am reverting. BTW, we completely abandoned supporting Compound types, no HPC app we work needs it (there are overheads), and we are very busy with other tasks.

Issue #1226, opened 2019-02-25
c-bindings: adios2_variable_type_string


germasch:
I'm not using the C interface, nor am I planning to ever do that, but I wanted to point out that IMHO `adios2_variable_type_string` should be changed while (or if) it's still possible. (same for attribute, too)

```c
/**
 * Retrieve variable type in string form "char", "unsigned long", etc.
 * @param type output, string form "int", "char"
 * @param size output, type size without '\0'
 * @param variable handler
 * @return adios2_error 0: success, see enum adios2_error for errors
 */
adios2_error adios2_variable_type_string(char *type, size_t *size,
                                         const adios2_variable *variable);

```

In my opinion, C bindings should try to follow the conventions of the C language in as far as possible, but this does not.
* This function cannot be safely used, since one has to pass an allocated char * to it, but one will not know how many characters are needed until after the function returns.
* Okay, so one can uses `char type[100]` and that's probably safe forever (but even in scientific software, buffer overflow bugs should be avoided if possible).
* If one then goes to print the returned string, one will get trailing garbage, because it's not 0-terminated.

Semi-safe use would go like this:
```c
char type[100];
size_t size;
adios2_error rc = adios2_variable_type_string(type, &size, myVar);
assert(rc == 0);
assert(size < 99);
type[size] = 0;
printf("type: %s\n", type);
```
I think chances that a user gets that right are pretty slim. 
It's a C API, so it's not going to be pretty, but my suggestion would be:
```c
adios2_error adios2_variable_type_string(const char **p_type,
                                         const adios2_variable *variable);
```
Use:
```c
const char* type;
adios2_error rc = adios2_variable_type_string(&type, myVar);
assert(rc == 0);
printf("type: %s\n", type);
```
My suggestion would be to return a statically allocated const char*, so if the user wants, they can strdup() it, but in general, one wouldn't need to, and not worry about freeing the string.

Obviously, this is an API change, though at least it will be known at compile time due to the changed function signature. It looks like this function is never used at all nor tested within ADIOS2, but obviously there could be external users. I'd be happy to provide to create a patch if this is considered a useful change.

williamfgc:
[...]

PR #1230, opened 2019-02-26
c-bindings: improvements to API functions returning strings

germasch:
For a more detailed discussion of the underlying issue, see #1226 .

In a nutshell, the current API requires the user to pass in a preallocated buffer, but how large that
buffer had to be is not known until after the call. This PR makes it possible to use those functions safely by calling the API function twice:
* first, with a NULL buffer, so only the size will be returned
* second, preallocate the buffer using the size from step 1, then call the function again.

While safe if used correctly, this interface is unfortunately cumbersome. It doesn't do anything about adding trailing zeros (though that could be added), either.

### DIFF
--- a/bindings/C/c/adios2_c_attribute.cpp
+++ b/bindings/C/c/adios2_c_attribute.cpp
@@ -84,7 +84,7 @@ adios2_error adios2_attribute_type_string(char *type, size_t *size,
 
         const adios2::core::AttributeBase *attributeBase =
             reinterpret_cast<const adios2::core::AttributeBase *>(attribute);
-	return String2CAPI(attributeBase->m_Type, type, size);
+        return String2CAPI(attributeBase->m_Type, type, size);
     }
     catch (...)
     {
--- a/bindings/C/c/adios2_c_attribute.cpp
+++ b/bindings/C/c/adios2_c_attribute.cpp
@@ -25,8 +25,6 @@ adios2_error adios2_attribute_name(char *name, size_t *size,
     {
         adios2::helper::CheckForNullptr(
             attribute, "for attribute, in call to adios2_attribute_name");
-        adios2::helper::CheckForNullptr(
-            name, "for char* char, in call to adios2_attribute_name");
 
         const adios2::core::AttributeBase *attributeBase =
             reinterpret_cast<const adios2::core::AttributeBase *>(attribute);
@@ -80,8 +78,6 @@ adios2_error adios2_attribute_type_string(char *type, size_t *size,
         adios2::helper::CheckForNullptr(
             attribute, "for const adios2_attribute, in call to "
                        "adios2_attribute_type_string");
-        adios2::helper::CheckForNullptr(
-            type, "for char* type, in call to adios2_attribute_type_string");
         adios2::helper::CheckForNullptr(
             size,
             "for size_t* length, in call to adios2_attribute_type_string");
--- a/bindings/C/c/adios2_c_attribute.cpp
+++ b/bindings/C/c/adios2_c_attribute.cpp
@@ -31,9 +31,7 @@ adios2_error adios2_attribute_name(char *name, size_t *size,
         const adios2::core::AttributeBase *attributeBase =
             reinterpret_cast<const adios2::core::AttributeBase *>(attribute);
 
-        *size = attributeBase->m_Name.size();
-        attributeBase->m_Name.copy(name, *size);
-        return adios2_error_none;
+        return String2CAPI(attributeBase->m_Name, name, size);
     }
     catch (...)
     {
@@ -90,9 +88,7 @@ adios2_error adios2_attribute_type_string(char *type, size_t *size,
 
         const adios2::core::AttributeBase *attributeBase =
             reinterpret_cast<const adios2::core::AttributeBase *>(attribute);
-        *size = attributeBase->m_Type.size();
-        attributeBase->m_Type.copy(type, *size);
-        return adios2_error_none;
+	return String2CAPI(attributeBase->m_Type, type, size);
     }
     catch (...)
     {
--- a/bindings/C/c/adios2_c_internal.h
+++ b/bindings/C/c/adios2_c_internal.h
@@ -26,7 +26,7 @@ namespace
 template <int adios2_type>
 struct MapAdios2Type;
 
-inline adios2_error String2CAPI(const std::string &s, char *buf, size_t *size);
+adios2_error String2CAPI(const std::string &s, char *buf, size_t *size);
 
 } // anonymous namespace
 
--- a/bindings/C/c/adios2_c_internal.h
+++ b/bindings/C/c/adios2_c_internal.h
@@ -14,6 +14,8 @@
 
 #include "adios2_c_types.h"
 
+#include <string>
+
 namespace
 {
 /**
--- a/bindings/C/c/adios2_c_internal.h
+++ b/bindings/C/c/adios2_c_internal.h
@@ -24,6 +24,8 @@ namespace
 template <int adios2_type>
 struct MapAdios2Type;
 
+inline adios2_error String2CAPI(const std::string &s, char *buf, size_t *size);
+
 } // anonymous namespace
 
 #include "adios2_c_internal.inl"
--- a/bindings/C/c/adios2_c_internal.inl
+++ b/bindings/C/c/adios2_c_internal.inl
@@ -49,7 +49,7 @@ make_MapAdios2Type(adios2_type_double_complex, std::complex<double>)
     String2CAPI(const std::string &s, char *buf, size_t *size)
 {
     *size = s.size();
-    if (buf)
+    if (buf != nullptr)
     {
         s.copy(buf, *size);
     }
--- a/bindings/C/c/adios2_c_internal.inl
+++ b/bindings/C/c/adios2_c_internal.inl
@@ -43,6 +43,17 @@ make_MapAdios2Type(adios2_type_double_complex, std::complex<double>)
 /* clang-format on */
 #undef make_MapAdios2Type
 
+                                    inline adios2_error
+    String2CAPI(const std::string &s, char *buf, size_t *size)
+{
+    *size = s.size();
+    if (buf)
+    {
+        s.copy(buf, *size);
+    }
+    return adios2_error_none;
+}
+
 } // namespace
 
 #endif /* ADIOS2_BINDINGS_C_C_ADIOS2_C_INTERNAL_INL_ */
--- a/bindings/C/c/adios2_c_internal.inl
+++ b/bindings/C/c/adios2_c_internal.inl
@@ -15,6 +15,8 @@
 #error "Inline file should only be included from its header, never on its own"
 #endif
 
+#include <complex>
+
 namespace
 {
 
--- a/bindings/C/c/adios2_c_variable.cpp
+++ b/bindings/C/c/adios2_c_variable.cpp
@@ -199,12 +199,7 @@ adios2_error adios2_variable_name(char *name, size_t *size,
 
         const adios2::core::VariableBase *variableBase =
             reinterpret_cast<const adios2::core::VariableBase *>(variable);
-        *size = variableBase->m_Name.size();
-        if (name)
-        {
-            variableBase->m_Name.copy(name, *size);
-        }
-        return adios2_error_none;
+        return String2CAPI(variableBase->m_Name, name, size);
     }
     catch (...)
     {
--- a/bindings/C/c/adios2_c_variable.cpp
+++ b/bindings/C/c/adios2_c_variable.cpp
@@ -184,7 +184,7 @@ adios2_error adios2_set_step_selection(adios2_variable *variable,
     }
 }
 
-adios2_error adios2_variable_name(char *name, size_t *size,
+adios2_error adios2_variable_name(const char **name,
                                   const adios2_variable *variable)
 {
     try
@@ -193,11 +193,12 @@ adios2_error adios2_variable_name(char *name, size_t *size,
             variable,
             "for const adios2_variable, in call to adios2_variable_name");
         adios2::helper::CheckForNullptr(
-            size, "for size_t* length, in call to adios2_variable_name");
+            name, "for const char* name, in call to adios2_variable_name");
 
         const adios2::core::VariableBase *variableBase =
             reinterpret_cast<const adios2::core::VariableBase *>(variable);
-        return String2CAPI(variableBase->m_Name, name, size);
+        *name = variableBase->m_Name.c_str();
+        return adios2_error_none;
     }
     catch (...)
     {
--- a/bindings/C/c/adios2_c_variable.cpp
+++ b/bindings/C/c/adios2_c_variable.cpp
@@ -200,7 +200,10 @@ adios2_error adios2_variable_name(char *name, size_t *size,
         const adios2::core::VariableBase *variableBase =
             reinterpret_cast<const adios2::core::VariableBase *>(variable);
         *size = variableBase->m_Name.size();
-        variableBase->m_Name.copy(name, *size);
+        if (name)
+        {
+            variableBase->m_Name.copy(name, *size);
+        }
         return adios2_error_none;
     }
     catch (...)
--- a/bindings/C/c/adios2_c_variable.cpp
+++ b/bindings/C/c/adios2_c_variable.cpp
@@ -257,9 +257,7 @@ adios2_error adios2_variable_type_string(char *type, size_t *size,
 
         const adios2::core::VariableBase *variableBase =
             reinterpret_cast<const adios2::core::VariableBase *>(variable);
-        *size = variableBase->m_Type.size();
-        variableBase->m_Type.copy(type, *size);
-        return adios2_error_none;
+        return String2CAPI(variableBase->m_Type, type, size);
     }
     catch (...)
     {
--- a/bindings/C/c/adios2_c_variable.cpp
+++ b/bindings/C/c/adios2_c_variable.cpp
@@ -184,7 +184,7 @@ adios2_error adios2_set_step_selection(adios2_variable *variable,
     }
 }
 
-adios2_error adios2_variable_name(const char **name,
+adios2_error adios2_variable_name(char *name, size_t *size,
                                   const adios2_variable *variable)
 {
     try
@@ -193,12 +193,11 @@ adios2_error adios2_variable_name(const char **name,
             variable,
             "for const adios2_variable, in call to adios2_variable_name");
         adios2::helper::CheckForNullptr(
-            name, "for const char* name, in call to adios2_variable_name");
+            size, "for size_t* length, in call to adios2_variable_name");
 
         const adios2::core::VariableBase *variableBase =
             reinterpret_cast<const adios2::core::VariableBase *>(variable);
-        *name = variableBase->m_Name.c_str();
-        return adios2_error_none;
+        return String2CAPI(variableBase->m_Name, name, size);
     }
     catch (...)
     {
--- a/bindings/C/c/adios2_c_variable.cpp
+++ b/bindings/C/c/adios2_c_variable.cpp
@@ -193,8 +193,6 @@ adios2_error adios2_variable_name(char *name, size_t *size,
             variable,
             "for const adios2_variable, in call to adios2_variable_name");
         adios2::helper::CheckForNullptr(
-            name, "for char* name, in call to adios2_variable_name");
-        adios2::helper::CheckForNullptr(
             size, "for size_t* length, in call to adios2_variable_name");
 
         const adios2::core::VariableBase *variableBase =
@@ -250,8 +248,6 @@ adios2_error adios2_variable_type_string(char *type, size_t *size,
         adios2::helper::CheckForNullptr(variable,
                                         "for const adios2_variable, in call to "
                                         "adios2_variable_type_string");
-        adios2::helper::CheckForNullptr(
-            type, "for char* type, in call to adios2_variable_type_string");
         adios2::helper::CheckForNullptr(
             size, "for size_t* length, in call to adios2_variable_type_string");
 
--- a/testing/adios2/bindings/C/TestBPWriteTypes.cpp
+++ b/testing/adios2/bindings/C/TestBPWriteTypes.cpp
@@ -116,11 +116,8 @@ TEST_F(BPWriteTypesCC, ADIOS2BPWriteTypes)
         adios2_variable *varR32 = adios2_inquire_variable(ioH, "varR32");
         adios2_variable *varR64 = adios2_inquire_variable(ioH, "varR64");
 
-        size_t size;
-        adios2_variable_name(NULL, &size, varStr);
-        char var_name[size + 1];
-        adios2_variable_name(var_name, &size, varStr);
-        var_name[size] = '\0';
+        const char *var_name;
+        adios2_variable_name(&var_name, varStr);
         EXPECT_EQ(strcmp(var_name, "varStr"), 0);
 
         adios2_engine *engineH =
--- a/testing/adios2/bindings/C/TestBPWriteTypes.cpp
+++ b/testing/adios2/bindings/C/TestBPWriteTypes.cpp
@@ -330,6 +330,7 @@ TEST(ADIOS2_C_API, ReturnedStrings)
 
         const char *var_name;
         adios2_variable_name(&var_name, var);
+        adios2_remove_all_ios(adiosH);
         EXPECT_EQ(strcmp(var_name, "varI8"), 0);
     }
 }
--- a/testing/adios2/bindings/C/TestBPWriteTypes.cpp
+++ b/testing/adios2/bindings/C/TestBPWriteTypes.cpp
@@ -337,26 +337,26 @@ TEST(ADIOS2_C_API, ReturnedStrings)
         // now test the APIs that return strings
         size_t engine_type_size;
         adios2_engine_type(NULL, &engine_type_size, ioH);
-        char engine_type[engine_type_size + 1];
+        char *engine_type = (char *)malloc(engine_type_size + 1);
         adios2_engine_type(engine_type, &engine_type_size, ioH);
         engine_type[engine_type_size] = '\0';
 
         size_t var_name_size;
         adios2_variable_name(NULL, &var_name_size, var);
-        char var_name[var_name_size + 1];
+        char *var_name = (char *)malloc(var_name_size + 1);
         adios2_variable_name(var_name, &var_name_size, var);
         var_name[var_name_size] = '\0';
 
         size_t attr_name_size;
         adios2_attribute_name(NULL, &attr_name_size, attr);
-        char attr_name[attr_name_size + 1];
+        char *attr_name = (char *)malloc(attr_name_size + 1);
         adios2_attribute_name(attr_name, &attr_name_size, attr);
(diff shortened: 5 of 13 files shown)


bindings/C/c/adios2_c_attribute.cpp
```diff
@@ -31,9 +31,7 @@ adios2_error adios2_attribute_name(char *name, size_t *size,
         const adios2::core::AttributeBase *attributeBase =
             reinterpret_cast<const adios2::core::AttributeBase *>(attribute);
 
-        *size = attributeBase->m_Name.size();
-        attributeBase->m_Name.copy(name, *size);
-        return adios2_error_none;
+        return String2CAPI(attributeBase->m_Name, name, size);
```
williamfgc: why not making this part of the helper namespace that returns a void? A more reusable name would be StringToChar, it doesn't have to return an error from enum that is exclusive to the C API. This function is a lot more generic and can be reused in other contexts, please move it to: https://github.com/ornladios/ADIOS2/blob/master/source/adios2/helper/adiosString.h (and in adiosString.cpp)

bindings/C/c/adios2_c_internal.h
```diff
@@ -24,6 +26,8 @@ namespace
 template <int adios2_type>
 struct MapAdios2Type;
 
+inline adios2_error String2CAPI(const std::string &s, char *buf, size_t *size);
+
 } // anonymous namespace
 
 #include "adios2_c_internal.inl"
```
williamfgc: there must be a better way to make this private into a cpp file. This is C++ code which (while hidden) that is publicly expose in C bindings after a `make install`. Not sure will help C developers.

bindings/C/c/adios2_c_internal.inl
```diff
@@ -15,6 +15,8 @@
 #error "Inline file should only be included from its header, never on its own"
 #endif
 
+#include <complex>
```
williamfgc: see comment above, this is a C++ header in a C exposed header

bindings/C/c/adios2_c_variable.cpp
```diff
@@ -200,7 +200,10 @@ adios2_error adios2_variable_name(char *name, size_t *size,
         const adios2::core::VariableBase *variableBase =
             reinterpret_cast<const adios2::core::VariableBase *>(variable);
         *size = variableBase->m_Name.size();
-        variableBase->m_Name.copy(name, *size);
+        if (name)
```
williamfgc: Let's use `if(name != nullptr)` to remove ambiguity

bindings/C/c/adios2_c_internal.h
```diff
@@ -24,6 +24,8 @@ namespace
 template <int adios2_type>
 struct MapAdios2Type;
 
+inline adios2_error String2CAPI(const std::string &s, char *buf, size_t *size);
```
williamfgc: not sure inline means anything. In any case, see comment about moving this function inside helper.

williamfgc:
@germasch I think this is a good start, my two major points:

1. Tests (where the fun begins)
2. Add a function to request size and kill the ambiguity of NULL, adding tests will help you see the context in which the functions are used.

Thanks!

bindings/C/c/adios2_c_attribute.cpp
germasch: I thought about helper:: when I wrote this, but I didn't want to go there, because I don't see any use outside of the C API for this function. Which is also why I named it that way and included
the error code return.
I wrote this tiny helper, but I don't like it, and I actually don't want it to be used anywhere else.

bindings/C/c/adios2_c_internal.h
germasch: This isn't specific to this patch: I introduced the adios2_internal.h header because I needed a place to put stuff that's needed by the interface code (which is written in C++), but not part of the C API. I could have put it somewhere in helper/, but since it's specific to the C bindings, that didn't feel right, so this is where it went.

If you have a better place for a header file that is private to bindings but not part of the interface, let me know. FWIW, the python interface does similar things, though there of course it's obvious that none of the bindings/Python/*.h are part of the Python interface ;(

bindings/C/c/adios2_c_internal.inl
germasch: See above, it's not an exposed header

bindings/C/c/adios2_c_internal.h
germasch: Well, here it may not matter, but I'm pretty sure without the `inline` where the function is defined one will (sometimes) get an undefined symbol. Nevertheless, I wouldn't mind the function being in a .cpp file, but I didn't really want to add yet another file just for it.

bindings/C/c/adios2_c_variable.cpp
germasch: I can do that if that's coding style, but there's no ambiguity in this statement.

germasch:
I definitely take your point about testing, though, in particular because it turns out that this PR is broken... :(

germasch:
So I pushed some changes, which will fail clang-format again :( But that's okay, because it's not really ready, anyway (if ever).

I fixed a bug which shows why one always needs to test stuff (and I didn't :( ). I added (only) one little test, which actually goes to show how hard it is to use this interface, but at least it doesn't change the API.

I then also added a commit ("EXP...") which shows what I think a more convenient interface would look like. The drawback of that is the returned string becomes invalid when the underlying object goes away. Currently, all hell breaks loose if you keep handles to an object which has gone away (even in C++ or Python), so I don't think this makes it much worse, but it is an issue. Of course you can always strdup() the string if you want to own it for good.

There's yet another alternative (but it's also a new API): Pass a preallocated buffer and its length, and get back a possibly truncated 0-terminated string in your buffer. Not a perfect solution either, though.

bindings/C/c/adios2_c_variable.cpp
williamfgc: is `name` a nullptr or a value other than 0?

bindings/C/c/adios2_c_attribute.cpp
williamfgc: that's how all functions in helper started: only one use, over time it helps us maintain code since we have one namespace or reusable generic components. We already have StringToDouble, DimsToString, etc.

testing/adios2/bindings/C/TestBPWriteTypes.cpp
```diff
@@ -116,11 +116,8 @@ TEST_F(BPWriteTypesCC, ADIOS2BPWriteTypes)
         adios2_variable *varR32 = adios2_inquire_variable(ioH, "varR32");
         adios2_variable *varR64 = adios2_inquire_variable(ioH, "varR64");
 
-        size_t size;
-        adios2_variable_name(NULL, &size, varStr);
-        char var_name[size + 1];
-        adios2_variable_name(var_name, &size, varStr);
-        var_name[size] = '\0';
+        const char *var_name;
+        adios2_variable_name(&var_name, varStr);
```
williamfgc: @germasch does this even work? `const char*` means can't change var_name contents, but `adios2_variable_name` is modifying them. This is why I advised to stick with the previous solution.

bindings/C/c/adios2_c_internal.h
williamfgc: @germasch how was this before your PR?

williamfgc:
> I then also added a commit ("EXP...") which shows what I think a more convenient interface would look like.

@germasch Not sure of the point or value of opening an issue and discussing, if you still will go ahead and follow this route. The value of tests is to understand the trade-offs of your interface. See below. This won't make into master as that's exactly what we removed a while ago. 

```
char * name
{
     adios2_init
     adios2_declare_io
     adios2_open
     adios2_variable_name (name, ...
     //adios2_remove_io : another way to invalidate name
}
name is not guaranteed to be valid, adios handler is out of scope, . 
```

bindings/C/c/adios2_c_internal.h
germasch: I've never changed anything about python, my point is that the bindings/python has header files that are used to compile the python interface code (which is written in C++). (And that obviously header files aren't possible part of a python interface).

For the C bindings there used to be only header files that are installed as part of the API, so they were used both as API and they were included in the C++ code that creates the C bindings. Because they were part of the C API, I wasn't able to add some C++ helper code there, which I needed to build the interface. That's why added (and in this PR slightly extended) the not-installed adios_internal.{h,inl}. It's pat of the binding, but not part of the API. An alternative could be to name it `private/adios2_c_helper.h` or something like that.

testing/adios2/bindings/C/TestBPWriteTypes.cpp
germasch: If we didn't want to, by convention, return an error code, one would do
```c
const char *name = adios2_variable_name(varStr);
```
The code above does the very same thing, just passing `name` as a by-reference argument. `const` rules in C/C++ are admittedly somewhat confusing, but this is correct. The `const` goes with the `char`.
```c
const char *name = "abc";
name[0] = 'A'; // error
name = "def"; // that's fine, and that' s what adios2_variable_name does, essentially
```
The const is there so that the C code cannot modify the C string, which still lives in std::string's memory. That's why `.c_str()` returns `const char *`.

germasch:
> > I then also added a commit ("EXP...") which shows what I think a more convenient interface would look like.
> 
> @germasch Not sure of the point or value of opening an issue and discussing, when opinions based on previous experience will be largely dismissed. The value of tests is to understand the trade-offs of your interface. See below. This is exactly what we removed a while ago.

I did say that this PR wasn't in shape for merging -- it's for the sake of discussion (which clearly we're having..). 

> ```
> char * name
> {
>      adios2_init
>      adios2_declare_io
>      adios2_open
>      adios2_variable_name (name, ...
>      //adios2_remove_io : another way to invalidate name
> }
> name is not guaranteed to be valid, adios handler is out of scope, this is why it needs to be a copy.
> ```

So I don't exactly follow the above (most of these don't return strings, they take strings as input.

I know I'm being argumentative here (and if you actually knew me, you'd see that this isn't actually my typical personality.) But apparently, I do have strong opinions on certain aspects of coding, and I can't quite let go ;)

Anyway, I accept that you don't want to return string with limited life time, even though I'd like to have a heard a reason why, after you `RemoveIO` all your adios2_variable* are invalid, but your strings need to remain valid. (Possible argument: There's a hope to fix the other life time issues eventually, so you don't want to introduce new ones.)

I'll drop the last "EXP" (experimental) commit, and I"ll add some more tests.

germasch:
I pushed some update reverting to the all String2CAPI based version, which is at least exactly backward compatible.

I actually put in a test to prove that my "EXP" version does indeed fail if the underlying object goes away (though it only triggers if compiled with address sanitizer). The reverted version is shown to be safe.

I can clean this up to get rid of the "EXP" commit and its revert if you prefer.

bindings/C/c/adios2_c_internal.h
williamfgc: @germasch Not sure when I implied that Python bindings were changed, my point is that those are never accessed by the Python user, unlike headers in C. See my comment about moving it to helper:: as that has been the team's preference with these kind of functionality. There is no need to return `adios2_error`

bindings/C/c/adios2_c_attribute.cpp
williamfgc: > I actually don't want it to be used anywhere else

Why not? We, the team, would prefer it under helper:: as that's how we've been operating for a while for reusability.

williamfgc:
> Anyway, I accept that you don't want to return string with limited life time, even though I'd like to have a heard a reason why,

I thought the pseudo-code was self-explanatory about scope { }. If the adios handler goes away your `name` becomes dangling. Unlike C++'s RAII and objects in which scope and ownership is very clear, in the C API it's the user owning the memory for these strings, and we don't want to mess with it.

bindings/C/c/adios2_c_internal.inl
williamfgc: fair enough

germasch:
> > Anyway, I accept that you don't want to return string with limited life time, even though I'd like to have a heard a reason why,
> 
> I thought the pseudo-code was self-explanatory about scope { }. If the adios handler goes away your `name` becomes dangling. Unlike C++'s RAII and objects in which scope and ownership is very clear, in the C API it's the user owning the memory for these strings, and we don't want to mess with it.

Maybe we do have a real misunderstanding here: End of scope `{ }` in C doesn't call destructors (because the concept doesn't exist), so very little happens. Local variables declared inside a given scope will go away, but they aren't accessible anymore anyway, because they don't exist outside of the scope. You have to explicitly do bad things to cause issues with a variable that go out of scope:
```c
{
   int *p;
  { 
     int c = 1;
     p = &c;
  }
  // don't access *p here!
}
```
When an `adios2_variable *var` goes out of scope, nothing really happens. The memory where the address of the pointer target was stored goes away, but you can't access that anyway, because `var` is out of the scope, so the compiler won't let you access it.  The `Variable` to still exists, the address the pointer used to point to is still valid. A string stored in the `Variable` is still valid. It's not going out of scope that invalidates the string, it's only explicitly calling `adios2_remove_variable()` or similar that causes trouble.

williamfgc:
> A string stored in the `Variable` is still valid. It's not going out of scope that invalidates the string, it's only explicitly calling `adios2_remove_variable()` or similar that causes trouble.

Not the case if the adios handler goes out of scope or `adios2_finalize` gets called, or even worse (which we have seen with apps) if your MPI Communicator passed in `adios2_init` is invalidated. This implies understanding the adios2 internal memory hierarchy as well.

germasch:
Not that the MPI issue is really related, but
> (from [here](https://www.mpich.org/static/docs/v3.2/www3/MPI_Comm_dup.html)) "no library routine should use MPI_COMM_WORLD as the communicator; instead, a **duplicate** of a user-specified communicator should always be used." (emphasis mine)

Not only could the communicator go away, if the library and the application use the same communicator, one might receive messages meant for the other, which is no likely to end well...

williamfgc:
@germasch yeah, not sure how it's related, that's what adios does at init, open and to aggregate substreams. Can you address the points from my review? Thanks.

germasch:
> @germasch yeah, not sure how it's related, that's what adios does at init, open and to aggregate substreams. Can you address the points from my review? Thanks.

Let's not have the MPI discussion here. What other points are there to be addressed? This PR in its current shape let's the app manage the memory, and it's even established by the test. The API in its current shape is backward compatible but hard to use correctly, but at least it's possible now. Mostly, there is just no good solution.

I'll be happy to clean up to commit history to get rid of the excursion into returning an internally managed string, but it's already reverted.

williamfgc:
> What other points are there to be addressed? 

They are not shown on github?

germasch:
> > What other points are there to be addressed?
> They are not shown on github?

Maybe I'm lost.
* Let the application manage the memory: Done
* adios2_c_internal.h breaks the C interface : Not so
* move String2CAPI to helper/: I argued that's it's C API specific and is currently in the same file as other C API specific code. And that it's a cumbersome interface I don't want to encourage to spread.
* add tests: Done
* add separate `_length` functions for getting the length: Well, I'm not doing that, IMO it's awkward enough as it is. FWIW, (somewhat) similar APIs (e.g., HDF5) accept NULL if the user just wants to find out the length.

Nothing prevents you from changing whatever else you want yourself afterwards. The one thing I'd still be open to talk about is 0-termination, but really, I've been doing this for fun, and it isn't so much fun anymore.

williamfgc:
@germasch My suggestions is so the introduced changes can serve better other contributors working and doing maintenance of their own pieces, while working with integration with apps. The impact with apps using the Fortran and Matlab interfaces is also something we haven't evaluated. While you don't want to do that, as you explicitly said, we still have to assign resources if something goes unexpected....and maintaining production software was never fun!

germasch:
Currently the Matlab and Fortran API are on top of the C API, so they have access to these changes in the C API. Are you suggesting it is planned to have those APIs interface directly to the C++ core?

williamfgc:
> Currently the Matlab and Fortran API are on top of the C API, so they have access to these changes in the C API. 

Has that access been tested? Has the Fortran and Matlab APIs been tested with this NULL signature? How we pass NULL from Fortran? How to interface with Fortran modules? How to pass null terminated strings? Is there new code not covered by the tests? Why not moving things to helper when only the return value is unique to C so others won't need to reimplement your function?

> Are you suggesting it is planned to have those APIs interface directly to the C++ core?

No, not at all. It would be duplicating the lines of code in the C bindings.

germasch:
> > Currently the Matlab and Fortran API are on top of the C API, so they have access to these changes in the C API.
> 
> Has that access been tested? Has the Fortran and Matlab APIs been tested with this NULL signature? How we pass NULL from Fortran? How to interface with Fortran modules? How to pass null terminated strings? Is there new code not covered by the tests? Why not moving things to helper when only the return value is unique to C so others won't need to reimplement your function? I also suggested adding a new length function for the above reasons.

This PR implements the minimal solution to this problem, to be backwards compatible. (At least it's the only backward compatible solution I can think of, other than adding new functions to the API). That means that the C API behaves exactly as it used to, except in the case where one is passing NULL for the char*. Presumably no-one is passing NULL right now, since they'd trigger an exception/error. I have added tests for the C API that confirms it still behaves the same as before. Since the Fortran/Matlab API are on top of C API which didn't change behavior, the Fortran / Matlab APIs are unchanged and work as before. I actually did test the Fortran API once (in as far as the automated tests go) once.
I cannot test (or even compile) the Matlab api, but again, it's unchanged.

The Fortran and Matlab API are both written in C, so there's no issue with NULL. They translate in Fortran / Matlab strings, so people using those languages don't have to know about the NULL messiness at all.

Now that the C API can be used safely, the Fortran (and maybe Matlab) API could benefit from that and avoid using hardcoded buffers and maybe a copy. The objective of this PR wasn't to improve the Fortran or Matlab APIs, though, they are left unchanged.

williamfgc:
Yeah, we need tests to proof.

germasch:
[off topic] By the way, having looked at the Fortran interface now, I'll happy declare that fortunately I don't like Fortran very much anyway, so I think I'll manage to stay away from it. But of course not before some comments: (I'm sorry, I guess I don't come across as very nice. But when I looked I saw issues, and I don't have interest in diving deeper, but I figured it's better if it's out there. Feel free to ignore or open an issue for it, whatever you prefer)

* This patch should be applied to fix a minor issue that you missed when reorganizing the C interface to support stdint-based values, otherwise a Fortran application may receive an undefined data type (integer) when reading a file with (unsupported in Fortran) unsigned integer values.

```
diff --git a/bindings/Fortran/modules/adios2_functions_mod.f90 b/bindings/Fortran/modules/adios2_functions_mod.f90
index 88ec90b9e..29e6d4393 100644
--- a/bindings/Fortran/modules/adios2_functions_mod.f90
+++ b/bindings/Fortran/modules/adios2_functions_mod.f90
@@ -46,7 +46,7 @@ contains

         f_type = adios2_type_unknown

-        do i = -1, 11
+        do i = -1, 8
             if (c_type == i) then
                 f_type = c_type
                 exit
```

* It surprises me that you use a 4096+ byte handle for a variable in the Fortran interface (where one would except an 8 byte pointer / id) to support a 4096 character buffer for the variable name, but then when the Fortran application asks for the name, it calls back into C in the first place, using just the 8 byte id and not the buffer at all -- unless I missed something.

germasch:
> Yeah, we need tests to proof the impact.

You mean like you do for other PRs, like this one? #1219 

Is there any doubt that an API that requires you to preallocate a buffer, but doesn't tell you the size until after the call is safe?

You want me do this? (on top of this PR)
```diff
diff --git a/testing/adios2/bindings/C/TestBPWriteTypes.cpp b/testing/adios2/bindings/C/TestBPWriteTypes.cpp
index f8c3d2ad6..6fae1c937 100644
--- a/testing/adios2/bindings/C/TestBPWriteTypes.cpp
+++ b/testing/adios2/bindings/C/TestBPWriteTypes.cpp
@@ -347,8 +347,7 @@ TEST(ADIOS2_C_API, ReturnedStrings)
         adios2_variable_name(var_name, &var_name_size, var);
         var_name[var_name_size] = '\0';

-        size_t attr_name_size;
-        adios2_attribute_name(NULL, &attr_name_size, attr);
+        size_t attr_name_size = 3;
         char *attr_name = (char *)malloc(attr_name_size + 1);
         adios2_attribute_name(attr_name, &attr_name_size, attr);
         attr_name[attr_name_size] = '\0';
```
Yes, just in case you're wondering, if you do guess a size, and the size is smaller than the actual string, it really crashes:
```
[ RUN      ] ADIOS2_C_API.ReturnedStrings
=================================================================
==53902==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x60200002a6b4 at pc 0x000104d9f712 bp 0x7ffeeb0eb250 sp 0x7ffeeb0ea9f0
WRITE of size 7 at 0x60200002a6b4 thread T0
[...]
```
Or are you saying you're not fixing bugs until they get hit in real life by someone having their job killed in the middle of a 10 million core hour run on Titan, because debugging that is just so much more fun?

williamfgc:
@germasch this is great. Please, don't hesitate to report more bugs and potential fixes, it's always appreciated as that's where exactly were we need help. C bindings always have an impact on Fortran and Matlab bindings. The PR you mentioned had tests, and we trust our team.

germasch:
Well, I'm going to dump my current pipeline, but I'm not going to spend this much time arguing anymore.

chuckatkins:
> for example the MPI APIs, not to mention BLAS and LAPACK in which pre-allocation is a must

FWIW all three of these use different API mechanisms w.r.t dynamic data sizes:
* MPI uses `MPI_Probe` to allow you to query the size of the message bevore invoking the corresponding receive call.  This is the approach I believe you're advocating for @williamfgc 
* BLAS algorithms aren't really applicable here as the output sizes are easily known a priori and come directly from the input sizes.
* LAPACK APIs are built around double function calls to determine the necessary size of the work arrays based on the driver's algorithm: The first call passes `LWORK=-1` and returns the size of the work array to use in `WORK(1)` then you can allocate accordingly and call for computation.  This is symantically equivalent to passing in a `NULL` output in C and the approach I believe was taken in this PR @germasch

williamfgc:
@chuckatkins you're spot on. Keep in mind, LAPACK uses a trial-and-error approach from F77 (no dynamic alloc) for block size depending on workspace supplied. From https://www.netlib.org/lapack/lug/node120.html   `If in doubt about how much workspace to supply, users should supply a generous amount (assume a block size of 64, say), and then examine the value of WORK(1) on exit.`

In any case, @sklasky is on my back and I am really busy with our milestones and apps work, in which I should be spending most of my time. So I won't continue this and I'll very more than happy to hand over all of the maintenance work to someone else if this makes it into master.

germasch:
Just to add two more to @chuckatkins's list, that I had looked at when thinking about this:
* MPI_Info_get has another kind of string interface, which takes a buffer and a length and truncates if needed. Safe, easy to use, but the drawback is that there is no way to be sure your string doesn't get truncated.
* H5Aget_name: somewhat of a mix between what MPI_Info_get and the interface in this PR: pass in buffer size, get back possibly truncated string, but also actual size. Can be called with NULL first to get actual size, then again with properly sized buffer.

There is no one perfect solution, in particular not with backwards compatibility, if anyone cares. The functions I mentioned above deal with zero-determination better that my approach. (My approach could be changed in a similar fashion, but that would make it more iffy to use in an optimized Fortran interface -- wouldn't be a show stopper, though).

FWIW, I updated the matlab bindings to take advantage of this PR. It's two simple changes, but I have no way of testing (even compiling) it, so I haven't pushed it here.

For reference:
```diff
commit 978adbf0203802cab52a13429a32361715e310c8
Author: Kai Germaschewski <kai.germaschewski@unh.edu>
Date:   Wed Feb 27 10:08:32 2019 -0500

    MATLAB: use adios2_{variable,attribute}_name() safely

diff --git a/bindings/Matlab/adiosopenc.c b/bindings/Matlab/adiosopenc.c
index fd67860d7..5c45ad800 100644
--- a/bindings/Matlab/adiosopenc.c
+++ b/bindings/Matlab/adiosopenc.c
@@ -236,8 +236,8 @@ void mexFunction(int nlhs, mxArray *plhs[], int nrhs, const mxArray *prhs[])
         adios2_variable *avar = adios_vars[vi];
         /* field NAME */
         size_t namelen;
-        char varname[adios2_string_array_element_max_size];
-
+        adios2_variable_name(NULL, &namelen, avar);
+        char varname[namelen + 1];
         adios2_variable_name(varname, &namelen, avar);
         varname[namelen] = '\0';
         mxSetFieldByNumber(vars, vi, var_field_Name, mxCreateString(varname));
@@ -333,7 +333,8 @@ void mexFunction(int nlhs, mxArray *plhs[], int nrhs, const mxArray *prhs[])
         const adios2_attribute *aa = adios_attrs[ai];
         /* field NAME */
         size_t namelen;
-        char attrname[adios2_string_array_element_max_size];
+        adios2_attribute_name(NULL, &namelen, aa);
+        char attrname[namelen + 1];
         adios2_attribute_name(attrname, &namelen, aa);
         attrname[namelen] = '\0';

```

germasch:
I removed the [WIP] from the title because I believe this is as good as it gets. The changes in here are really minor and backwards compatible:
* The actual changes affect the 6 C API functions that return strings. They now accept NULL as the buffer -- instead of throwing an exception, when called this way, they will return the string size, so that the caller can preallocate an appropriate size buffer. When called with a buffer, the function behaves exactly as before.
* The PR also adds some tests for this functionality.

I'll happily commit to support any problems that appear due to this change.

If this patch gets merged, I can also provide a patch to the matlab bindings, that makes them safe for large strings (and waste less memory for small strings), but I cannot test those. I can also look into updating the Fortran bindings to be safe and more efficient.

eisenhauer:
I don't see an objective reason not to merge this.  It's backwards-compatible and at least allows a safe use of the C API that isn't possible the way things are.

sklasky:
Hi Greg
Are you willing to maintain this? If so then write Norbert, cube the fear is that someone on the ADIOS team has to be responsible. If you guys agree then I’m fine with it. But if it’s only about telling William how to implement and then he has to maintain if, then it will not work out and will cause more problems in the team.

Norbert and I are heading to Amsterdam today.

You should look at https://cyrille.rossant.net/moving-away-hdf5/

<https://cyrille.rossant.net/moving-away-hdf5/>Since I think we need to understand why they don’t like HDF5 and make sure we don’t make the same mistakes.


Thanks


Scott

________________________________

eisenhauer:
Sure.  This patch boils down to adding an "if" statement before a copy operation and a test to see that that works to return the size without doing the copy.  I wouldn't worry about signing up to maintain that.

williamfgc:
@sklasky thanks! @eisenhauer from @germasch : 

> If this patch gets merged, I can also provide a patch to the matlab bindings, that makes them safe for large strings (and waste less memory for small strings), but I cannot test those. I can also look into updating the Fortran bindings to be safe and more efficient.

To provide context: while C bindings have important users they are not as used as Fortran and C++. I'd like to know how to pass a `NULL character` from Fortran to get the size (this is why I was proposing adding a separate function for size to keep Fortran and C bindings 1:1), and I don't even know how to do this in Matlab. That's the full impact of this PR, if someone is willing to maintain it, I'm cool. In any case, please follow @sklasky instructions.

eisenhauer:
Merging as an improvement and a solution for C bindings.  Fortran and MATLAB bindings left for future PRs.

PR #1236, opened 2019-02-27
mpi_comm_dup from the beginning

germasch:
Also came from an unrelated discussion. ADIOS2 MPI_Comm_dup()s its communicators, but potentially too late to protect from the user doing stupid things.

The first commit adds a test that triggers it. The 2nd commit is the simple fix.

### DIFF
--- a/source/adios2/core/ADIOS.cpp
+++ b/source/adios2/core/ADIOS.cpp
@@ -56,7 +56,18 @@ ADIOS::ADIOS(const std::string configFile, MPI_Comm mpiComm,
             " in call to ADIOS constructor\n");
     }
 
-    MPI_Comm_dup(mpiComm, &m_MPIComm);
+    int flag;
+    MPI_Initialized(&flag);
+    if (flag)
+    {
+        MPI_Comm_dup(mpiComm, &m_MPIComm);
+        m_NeedMPICommFree = true;
+    }
+    else
+    {
+        m_MPIComm = mpiComm;
+        m_NeedMPICommFree = false;
+    }
 
     if (!configFile.empty())
     {
@@ -83,6 +94,16 @@ ADIOS::ADIOS(MPI_Comm mpiComm, const bool debugMode,
 ADIOS::ADIOS(const bool debugMode, const std::string hostLanguage)
 : ADIOS("", MPI_COMM_SELF, debugMode, hostLanguage)
 {
+}
+
+ADIOS::~ADIOS()
+{
+    int flag;
+    MPI_Initialized(&flag);
+    if (flag && m_NeedMPICommFree)
+    {
+        MPI_Comm_free(&m_MPIComm);
+    }
 }
 
 IO &ADIOS::DeclareIO(const std::string name)
--- a/source/adios2/core/ADIOS.cpp
+++ b/source/adios2/core/ADIOS.cpp
@@ -47,13 +47,17 @@ namespace core
 
 ADIOS::ADIOS(const std::string configFile, MPI_Comm mpiComm,
              const bool debugMode, const std::string hostLanguage)
-: m_MPIComm(mpiComm), m_ConfigFile(configFile), m_DebugMode(debugMode),
-  m_HostLanguage(hostLanguage)
+: m_ConfigFile(configFile), m_DebugMode(debugMode), m_HostLanguage(hostLanguage)
 {
-    if (m_DebugMode)
+    if (m_DebugMode && mpiComm == MPI_COMM_NULL)
     {
-        CheckMPI();
+        throw std::ios_base::failure(
+            "ERROR: MPI communicator is MPI_COMM_NULL, "
+            " in call to ADIOS constructor\n");
     }
+
+    MPI_Comm_dup(mpiComm, &m_MPIComm);
+
     if (!configFile.empty())
     {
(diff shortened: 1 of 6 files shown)


germasch:
Well, I guess the CI is saved my butt here. So besides forgetting to also `MPI_Comm_free`, the much bigger problem is that if you have a static / global ADIOS2 object, its constructor will call MPI_Comm_dup on MPI_COMM_SELF (by default), and since that happens before MPI is initialized, it's not happy (sometimes -- on my laptop it's fine). 

So what I though is an easy improvement turns into a can of worms that I won't pursue unless someone thinks it's valuable. The only solution I can think of is some hackery that delays the MPI_Comm_dup() until first use, which doesn't sound worth it...

philip-davis:
I like duplicating the comm in the constructor, especially because I hate chasing bugs where the user pulled the comm out from under us. I'm thinking (hoping) there's not too much the user can do to damage MPI_COMM_SELF, so perhaps we could conditionally dup, something like this:

```C
m_MPI_Comm = mpiComm;
if(mpiComm != MPI_COMM_SELF) {
    MPI_Comm_dup(mpiComm, &m_MPIComm);
}
```

That might be a little too kludgy, though. Maybe something like
```C
if(MPI_Initialized()) {
    MPI_Comm_dup(mpiComm, &m_MPIComm);
} else {
    mpiComm = m_MPIComm;
}
```

is a little better. What do you think?

germasch:
I tried (essentially) the second method, which I like. But I hit a new, similar problem, because I also fixed my previous omission of calling `MPI_Comm_free` (only if we dup'd it originally). So the destructor now calls `MPI_Comm_free`, and that goes wrong if it happens after `MPI_Finalize`, and `MPI_Initialized` is no help. And this happens very easily because if you make an `adios2::ADIOS` object in main(), it gets destroyed at the very end of main, ie, after `MPI_Finalize()`.

So I think the real lesson is that a library using MPI really should be initialized only after MPI is up, and be brought down before MPI goes down. But that's probably not realistic to impose onto users.

There area number of not-so-great options:
* Force users to always initialize / shutdown ADIOS while MPI is running (will cause lots of breakage)
* Don't dup MPI_COMM_WORLD, MPI_COMM_SELF, and don't MPI_Comm_free them. That means no trouble for the majority of users. If a custom communicator is used, ADIOS must be initialized after MPI (which will be enforced automatically because you can't make a custom communicator otherwise). In this case, ADIOS must also be brought down before MPI is finalized. Sounds kinda reasonable to me, but will still surprise some users.
* Just leak the dup'd communicator (I don't like this out of principle, though it'd have the least problems).

I seem to have a talent for picking something seemingly simple and opening a can of worms instead. The problem could well be more widespread than this. `MPI_Comm_free()` is also called when an engine is destroyed according to a quick look (on a comm dup'd when it was constructed). If we're properly cleaning up at shutdown time, engines may become unused and `MPI_Comm_free()` their communicators, again likely after MPI_Finalize() in many simple use cases, even those using COMM_WORLD, or COMM_SELF. The fact that this isn't triggering right now makes me wonder whether Engines are actually getting destroyed properly.

I'll push my update, but it's not a solution.

philip-davis:
Thank you for adding that. To the finalize issue, I suppose we could replace the MPI_Initialized check with MPI_Finalized:

```C
ADIOS::~ADIOS()
 {
     int flag;
     MPI_Finalized(&flag);
     if (!flag && m_NeedMPICommFree)
     {
         MPI_Comm_free(&m_MPIComm);
     }
 }
```

We're still leaking the communicator for any object that goes out of scope after `MPI_Finalize()` though, so I see what you mean. An explicit Finalize for the ADIOS method would solve it, but would be a public API change (BTW, I think this is what we do for Engine, doing the `MPI_Comm_free()` in `Engine::Close()`, rather than the destructor.) I guess an explicit call to `ADIOS::~ADIOS()` from the user prior to `MPI_Finalize()` would do the trick, but something feels evil about that. I'm not much of a C++ developer though, so maybe that's fine.

I think my preference would be that the worst thing that can happen is that we leak the communicator. I hate doing that, but practically if we're going out of scope after `MPI_Finalize()` has happened, the application is probably done anyway. Perhaps I am asking for trouble, but hopefully not too much.

In any case, IMO not dup'ing the comm was not the right thing to do, and that is fixed by this is PR so that's a win. Thanks.

germasch:
@philip-davis: `MPI_Finalized` sounds perfect. I had no idea that existed, though in retrospect that wasn't too hard to guess. I had hoped `MPI_Initialized` might go back to false after `MPI_Finalize`, but that's not the case, and it turns out the standard says so.
Let's see whether this one passes the checks.

Trying to force users to call some `adios2_finalize` before `MPI_Finalize` sounds too error-prone, maybe even more so with C++ where destruction is automatic, but at the wrong time, unless one explicitly sets up another scope or something. The justification for `MPI_Finalized` in the standard sounds like it is exactly for this case, ie., a library that needs to know whether MPI is still active.

philip-davis:
Cool. I think this is looking good. Is this ready to be merged from your perspective, or is it still WIP?

germasch:
Thanks. It's ready from my perspective. I removed the [WIP] from the title, so let's see what the powers that be say.

eisenhauer:
Looks good to me...

Issue #1240, opened 2019-02-27
bpls to catch exceptions


williamfgc:
inserting try-catch block on bpls to exit normally (no signal) after an exception is caught

Issue #1239, opened 2019-02-27
Cannot read a BP file


bsmind:
I tried to open a BP file that was able to open with adios 1.13.1 version.
I installed the adios2 by following the instruction in adiosvm repository and passed all tests that comes with adios2 package. 

First, I tried to dump the contents using bpls command that comes with adios2.
And it gave me the following errors:
```
HDF5-DIAG: Error detected in HDF5 (1.8.17) MPI-process 0:
  #000: H5F.c line 604 in H5Fopen(): unable to open file
    major: File accessibilty
    minor: Unable to open file
  #001: H5Fint.c line 1087 in H5F_open(): unable to read superblock
    major: File accessibilty
    minor: Read failed
  #002: H5Fsuper.c line 277 in H5F_super_read(): file signature not found
    major: File accessibilty
    minor: Not an HDF5 file
HDF5-DIAG: Error detected in HDF5 (1.8.17) MPI-process 0:
  #000: H5O.c line 607 in H5Oget_info(): not a location
    major: Invalid arguments to routine
    minor: Inappropriate type
  #001: H5Gloc.c line 253 in H5G_loc(): invalid object ID
    major: Invalid arguments to routine
    minor: Bad value
HDF5-DIAG: Error detected in HDF5 (1.8.17) MPI-process 0:
  #000: H5G.c line 454 in H5Gopen2(): not a location
    major: Invalid arguments to routine
    minor: Inappropriate type
  #001: H5Gloc.c line 253 in H5G_loc(): invalid object ID
    major: Invalid arguments to routine
    minor: Bad value
Error: Could not open this file with any ADIOS2 file reading engines
```
I also tried to open the file in python
```
import adios2
from mpi4py import MPI
comm = MPI.COMM_WORLD
adios = adios2.ADIOS(comm)
ioReader = ioReader.Open("tau-metrics.bp", adios2.Mode.Read)
adios.RemoveAllIOs()
```
and it returns the following error messages:
```
Traceback (most recent call last):
  File "/home/sungsooha/Desktop/CODAR/PerformanceAnalysis/script.py", line 13, in <module>
    reader = ioReader.Open("./data/nwchem/20180801/tau-metrics.bp", adios2.Mode.Read)
ValueError: ERROR: ADIOS2 default BP3 engine doesn't support count statistics
```

Any suggestions?
[tau-metrics.bp.zip](https://github.com/ornladios/ADIOS2/files/2911336/tau-metrics.bp.zip)

[...]

PR #1241, opened 2019-02-27
Enable reading adios1 files with statistics_cnt

williamfgc:
enable reading statistic_cnt from adios1 generated files
bpls to catch exceptions
tested outside CI with external file


williamfgc:
#1239 and #1240

PR #1242, opened 2019-02-27
Fix race condition on writer side

eisenhauer:


eisenhauer:
Well Norbert, I did the commit from your VM, so it looks like you get credit for this!

pnorbert:
This fixes the observed bug in the VM.

Issue #1246, opened 2019-03-02
Open() in python bindings (example code, helloBPReaderHeatMap2D.py)


bsmind:
I cannot run the example code, `./examples/bpReader/helloBPReaderHeatMap2D.py`. 
When I run it, I got the following errors:
```
ibpStream = ioRead.Open('HeatMap2D_py.bp', adios2.Mode.Read. MPI.COMM_SELF)
TypeError: Open(): incompatible function arguments. The following argument types are supported:
     1. (self: adios2.IO, arg0: str, arg1: int) -> adios2::py11::Engine
```

When I looked at the `py11IO.h` file,
it has the following signature,
```
\#ifdef ADIOS2_HAVE_MPI
     Engine Open(const std::string &name, const Mode mode, MPI_Comm comm);
\#endif
```

I also checked the configured header file, `ADIOSConfig.h`, it has
```
\#define ADIOS2_HAVE_MPI
```

Any suggestions?

Thanks,
Sungsoo Ha

germasch:
[Note: I'm not a team member here, just trying to be helpful.]

On your primary problem, I think you have the right idea, i.e, the Python interface you're using appears to be compiled without MPI support. Maybe you have an older version of ADIOS2 / its python bindings lying around that get picked up, rather than the current version, which appears to have picked up MPI correctly according to your `ADIOSConfig.h`? This might help to find out:
```python
>>> import adios2
>>> adios2.__file__
'/Users/kai/build/ADIOS2/build/lib/python3.7/site-packages/adios2.cpython-37m-darwin.so'
```

If you get this to work, you'll run into more trouble, because that example is out of date, and there is another problem in that the version of the `Open()` call with a communicator as last argument isn't properly supported. This makes it work for me: (partially similar to #1233)
```diff
diff --git a/examples/hello/bpReader/helloBPReaderHeatMap2D.py b/examples/hello/bpReader/helloBPReaderHeatMap2D.py
index 2c81b7bc9..8a7b94c77 100644
--- a/examples/hello/bpReader/helloBPReaderHeatMap2D.py
+++ b/examples/hello/bpReader/helloBPReaderHeatMap2D.py
@@ -41,9 +41,8 @@ for i in range(0, Nx):
 adios = adios2.ADIOS(comm)
 ioWrite = adios.DeclareIO("ioWriter")

-varTemperature = ioWrite.DefineVariable("temperature2D", shape,
-                                        start, count, adios2.ConstantDims,
[...]

PR #1247, opened 2019-03-03
WIP: Python fixes

germasch:
This fixes the issues from #1246.
* fix arg order in helloBPReaderHeatMap2D.py
* fix the `IO::Open(..., MPI_Comm)` to actually use the passed comm (it was just using the comm from m_IO)
* add that function to the interface and correctly translate the communicator

I don't think this is the prettiest possible way of doing things, but it works for now (for me, anyway)

### DIFF
--- a/bindings/Python/py11glue.cpp
+++ b/bindings/Python/py11glue.cpp
@@ -366,6 +366,10 @@ PYBIND11_MODULE(adios2, m)
         .def("Open", (adios2::py11::Engine (adios2::py11::IO::*)(
                          const std::string &, const int)) &
                          adios2::py11::IO::Open)
+        .def("Open",
+             (adios2::py11::Engine (adios2::py11::IO::*)(
+                 const std::string &, const int, pybind11::object &comm)) &
+                 adios2::py11::IO::Open)
         .def("AvailableVariables", &adios2::py11::IO::AvailableVariables)
         .def("AvailableAttributes", &adios2::py11::IO::AvailableAttributes)
         .def("FlushAll", &adios2::py11::IO::FlushAll)
--- a/examples/hello/bpReader/helloBPReaderHeatMap2D.py
+++ b/examples/hello/bpReader/helloBPReaderHeatMap2D.py
@@ -41,9 +41,8 @@ for i in range(0, Nx):
 adios = adios2.ADIOS(comm)
 ioWrite = adios.DeclareIO("ioWriter")
 
-varTemperature = ioWrite.DefineVariable("temperature2D", shape,
-                                        start, count, adios2.ConstantDims,
-                                        temperatures)
+varTemperature = ioWrite.DefineVariable("temperature2D", temperatures, shape,
+                                        start, count, adios2.ConstantDims)
 
 obpStream = ioWrite.Open('HeatMap2D_py.bp', adios2.Mode.Write)
 obpStream.Put(varTemperature, temperatures)
--- a/bindings/Python/py11IO.cpp
+++ b/bindings/Python/py11IO.cpp
@@ -13,6 +13,10 @@
 #include "adios2/ADIOSMacros.h"
 #include "adios2/helper/adiosFunctions.h" //GetType<T>
 
+#ifdef ADIOS2_HAVE_MPI
+#include <mpi4py/mpi4py.h>
+#endif
+
 #include "py11types.h"
 
 namespace adios2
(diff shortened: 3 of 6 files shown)


examples/hello/bpReader/helloBPReaderHeatMap2D.py
```diff
@@ -41,9 +41,8 @@
 adios = adios2.ADIOS(comm)
 ioWrite = adios.DeclareIO("ioWriter")
 
-varTemperature = ioWrite.DefineVariable("temperature2D", shape,
-                                        start, count, adios2.ConstantDims,
-                                        temperatures)
+varTemperature = ioWrite.DefineVariable("temperature2D", temperatures, shape,
+                                        start, count, adios2.ConstantDims)
```
williamfgc: Python code in adios2 uses the autopep8 format https://pypi.org/project/autopep8/0.8/  you might run into problems with this. Usage: `autopep8 -i file.py` for in place formatting.

bindings/Python/py11glue.cpp
```diff
@@ -366,6 +366,10 @@ PYBIND11_MODULE(adios2, m)
         .def("Open", (adios2::py11::Engine (adios2::py11::IO::*)(
                          const std::string &, const int)) &
                          adios2::py11::IO::Open)
+        .def("Open",
```
williamfgc: this need to be tested from a Python program, so it can run under CI. Feel free to add tests under testing/adios2/bindings/python calling this function. Only MPI tests are needed.

williamfgc:
@germasch seems like @chuckatkins bump to a more recent PyBind11 version fixes an old issue, I am requesting tests so we can verify in our CI that this is indeed fixed and we can merge.

examples/hello/bpReader/helloBPReaderHeatMap2D.py
germasch: Thanks. I ran autopep8 on it, it's actually happy with it as-is. (Ie, I got lucky)

bindings/Python/py11glue.cpp
germasch: Done (see below).

germasch:
In the latest update, I fixed my usual clang-format issues, and added a test to the CI as follows:

I essentially copied `examples/hello/bpReader/helloBPReaderHeatMap2D.py` to `testing/adios2/bindings/python/TestBPWriteRead2D.py`, which already tests the particular call to `IO::Open`. Then, I automated it (ie, put in some `assert`s that will fail if the result read back doesn't match what's expected), and added it to ctest.

I also did some cleanup on this python test (e.g., actually use 2d numpy arrays). Still right now there are two rather similar python tests, so the original one should probably go away, but it's in a directory collecting examples of how to use the different language bindings, so I didn't want to mess with that.

I'll have another PR on top of this that does some more work to simplify the MPI4PY communicator passing in particular.

germasch:
Alright, so issues with the CI:
* The format check complained about my python, even though I ran autopep8 on the file beforehand (and now did again, and it makes no changes). I fixed this one manually by putting a space between `#` and the comment, but I guess I'm not using autopep8 correctly...?

* codacy complains about me using `assert` in Python.

```
Use of assert detected. The enclosed code will be removed when compiling to optimised byte code.
```
   Well, so isn't that what `assert` is there for? Anyway, `assert` is what pytest uses, so it's kinda the standard way of writing tests in Python. I'd prefer to have Codacy accept, rather than having to invent my own assert.

* one build fails a test:
```
4: Files "/home/adios2/source/examples/heatTransfer/HeatTransfer.Dump.txt" to "/home/adios2/build_fedora-openmpi-ubsan/examples/heatTransfer/HeatTransfer.BPFile.Dump.MxM.txt" are different.
  4/397 Test   #4: HeatTransfer.BPFile.Validate.MxM .............................................***Failed    0.01 sec
```
  I haven't touched this at all, so I feel innocent. Maybe a race condition in the testing?

williamfgc:
@germasch This is all good. Codacy is optional, so I wouldn't mind merging until @chuckatkins can take a look at the `assert` as he has more expertise. About the issue below:

> I haven't touched this at all, so I feel innocent. Maybe a race condition in the testing?

When looking at the logs, I see all the standard output based (bpls) tests failed for all engines, not just that test on that CI target, I'd think is more something in the configuration rather than the tests themselves. Locally, this pattern is reproducible when there is two conflicting installations, but I don't know the details or reproducibility on the CI docker containers or how they clean after themselves. @chuckatkins, what do you think? 

In the meantime, I can always try rerunning the tests again in case some it was some leftovers.

williamfgc:
Actually, there is a minor issue with CircleCI https://status.circleci.com/ right now that won't allow me to rerun tests at the moment. I'll check later.

PR #1249, opened 2019-03-03
pass strings safely to Fortran bindings

germasch:
This PR demonstrates how strings can now be safely passed through to the Fortran API. (Though the implementation is not exactly pretty. But it's that's at least not visible on the interface side.)

#### Background

There are right now 6 function in the C API that "return" strings (ie, copy them into preallocated buffers).
* `adios2_variable_name`
* `adios2_attribute_name`
* `adios2_engine_type`
* `adios2_operator_type`
* `adios2_variable_type_string`
* `adios2_attribute_type_string`

Only the very first one is actually interfaced into Fortran right now, but wasn't tested. 
For the next 3, part of the Fortran API implementation exist, but it is not complete.
For the final 2, there is no trace of them in the Fortran API binding.

[Off-topic here, but I think the last 2 functions should be deleted from the C API. They're not used anywhere, and duplicate functionality which is already there (`adios2_{variable,attribute}_type` returning the adios2_type enum, which is otherwise used throughout the C API.) The only possible value of this functions would be for debugging, but they're really too cumbersome to use to produce quick human readable output.]

#### Changes

This PR does
* not change any APIs
* avoids the fixed size buffer used in interfacing `adios_variable_type`, and avoids one copy. (One copy remain, copying the string into the Fortran `character` string.) It is safe for any size string.
* adds a test for `adios_variable_type` used from Fortran.

I can do the same thing for the other 3 functions, which are currently missing from the Fortran API, if people agree with this approach.

The implementation is not pretty, because two calls have to be made into C to (1) get the string length; then the buffer is allocated and (2) the string is copied. (For what it's worth: Yes, I had two use two C functions to support this interface, but it's internal. Even if the C API had a separate `_length` function, I'd still need those two wrappers, anyway.)

One way to avoid having two layers of glue code (one on the Fortran side, one on the C side) in the Fortran bindings in general would be to use `ISO_C_BINDING`. I believe it's available since Fortran2003. It would simplify the code, though otherwise not really make a difference. BTW, that one provides `C_NULLPTR` or something like that.

### DIFF
--- a/bindings/Fortran/CMakeLists.txt
+++ b/bindings/Fortran/CMakeLists.txt
@@ -35,6 +35,7 @@ set(MODULES
      ${CMAKE_CURRENT_SOURCE_DIR}/modules/adios2_variable_mod.f90
      ${CMAKE_CURRENT_SOURCE_DIR}/modules/adios2_variable_min_mod.f90
      ${CMAKE_CURRENT_SOURCE_DIR}/modules/adios2_variable_max_mod.f90
+     ${CMAKE_CURRENT_SOURCE_DIR}/modules/adios2_operator_mod.f90
 )
 
 add_library(adios2_f ${MODULES} ${F2C})
--- a/bindings/Fortran/f2c/adios2_f2c_attribute.cpp
+++ b/bindings/Fortran/f2c/adios2_f2c_attribute.cpp
@@ -17,6 +17,29 @@
 extern "C" {
 #endif
 
+void FC_GLOBAL(adios2_attribute_name_f2c,
+               ADIOS2_ATTRIBUTE_NAME_F2C)(char *name,
+                                          const adios2_attribute **attribute,
+                                          int *ierr)
+{
+    size_t sizeC;
+    *ierr = static_cast<int>(adios2_attribute_name(name, &sizeC, *attribute));
+}
+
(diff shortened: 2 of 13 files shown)


germasch:
I rebased this PR onto current master and updated it to safely copy strings from C to Fortran for these
4 functions (the other 2 are not used in the Fortran bindings):
* `adios2_variable_name`
* `adios2_attribute_name`
* `adios2_engine_type`
* `adios2_operator_type`

I also added some tests. The Fortran functions by the same name are now safe to call for any string length (they allocate the string they return). However, the last two C functions were also unsafely used in other places, so this patch introduces code like
```
     if (length > 15) stop 'adios2_declare_io: engine_type too long!'
```
in a few (3) places. I don't think that's a good solution, but it's better than silent corruption due to buffer overflows as it is right now in those same places.

So as usual, this PR isn't perfect, but an improvement over what's currently there.

germasch:
The test failures in one circleci build are before-seen validation failures of bpls output, which I'm quite sure aren't related to the Fortran interface work in here.

eisenhauer:
Hmm.  Forcing a rerun of tests.  Thank you for your persistence Kai...

germasch:
@eisenhauer Thanks -- got the checkmark now ;)

PR #1250, opened 2019-03-03
MATLAB: use adios2_{variable,attribute}_name() safely

germasch:
I made this WIP, because I can't test it, so it's not even compile-tested.

Other than that, it's very straightforward, getting rid of the hardcoded 4096 character limit for variable/attribute names and instead using buffers sized for the actual length.

### DIFF
--- a/bindings/Matlab/adiosopenc.c
+++ b/bindings/Matlab/adiosopenc.c
@@ -236,8 +236,8 @@ void mexFunction(int nlhs, mxArray *plhs[], int nrhs, const mxArray *prhs[])
         adios2_variable *avar = adios_vars[vi];
         /* field NAME */
         size_t namelen;
-        char varname[adios2_string_array_element_max_size];
-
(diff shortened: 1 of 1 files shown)


germasch:
I removed the WIP from the title since no-one seems interested / able of actually testing it. I took another look at it, and it's two trivial enough small changes that I'm positive I got it right.

eisenhauer:
Thanks Kai.  It looks like Norbert did the bindings, but he's on travel and hasn't been able to test it.  I've downloaded Matlab, but haven't gotten back to actually doing something with it yet.  One of us will hit this soon...

pnorbert:
I can't test this (on Titan) until I am back.

germasch:
Thanks for being willing to look at this (to both @eisenhauer and @pnorbert). Take your time.

pnorbert:
Well, the current master already breaks the Matlab wrapper, so this will take time.

PR #1251, opened 2019-03-03
Python cleanup

germasch:
Since I'm at it, let's dump on more PR here. This one is on top of #1247, so shouldn't be considered before that one has been worked out.

This PR adds a pybind11 caster to make passing MPI4PY communicators much simpler. It also converts
`ADIOS` back into an actual Python object, rather than a fake-constructor that returns a `py11_ADIOS` object.

### DIFF
--- a/bindings/Python/py11ADIOS.cpp
+++ b/bindings/Python/py11ADIOS.cpp
@@ -18,14 +18,14 @@ namespace py11
 {
 
 #ifdef ADIOS2_HAVE_MPI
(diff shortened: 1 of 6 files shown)


germasch:
I rebased this on top of current master and force pushed. (I also made a minor change, I used the existing `py11types.h` rather than introducing a new `py11mpi.h`.)

Other than my usual clang-format fail, I hope this one will pass...

germasch:
Removed the WIP label because  think this is ready to go.

Issue #1248, opened 2019-03-03
allow assert in python tests?


germasch:
In #1247, I'm hitting an issue where I added some tests that verify that I'm getting the expected result using the `assert` statement in Python. That's how it's commonly done in Python, in fact that's how `pytest` is doing it.

I looked in another python test, and the method currently used there appears to be
```python
if(attrStringData[0] != "one"):
    raise ValueError('attrStringData[0] failed')
```
which will work, but it's more code to write and less explicit when it actually fails, compared to
```python
assert attrStringData[0] == "one"
```
It looks like it's an easy change to have codacy accept `assert` use: https://github.com/heilaaks/snippy/commit/c8bfa51f64f0161572b43a2f726133b148e12c97

williamfgc:
Hey @chuckatkins , can you take a look at this? I'd rather have you touch Codacy based on a recent issue with our accounts. Overall, this Codacy issue is the reason why I moved away with assert on Python, but we use assert heavily in google-test for the C++ and C bindings. It's not a bad thing to enable.

germasch:
I'd be happy to close this, as the PR has been merged and nothing bad happened...

PR #1252, opened 2019-03-04
tell codacy / bandit to accept assert in python

germasch:
This is subject to the outcome of the discussion in #1248 -- me being curious to see whether this actually works to make the otherwise failing `master` branch pass checks.

### DIFF
--- a/bandit.yml
+++ b/bandit.yml
@@ -1,0 +1,2 @@
+# tell codacy / bandit to accept assert in python
+skips: ['B101']


germasch:
I removed the WIP from the title, despite the fact that @chuckatkins hasn't chimed in, because currently all PRs are having their checks fail because of the current use of `assert` in Python. So I think taking this one-liner in the mean time would get the checks to work properly again -- or alternatively I could remove the `assert`s from the Python code for the time being, if that's preferred.

Issue #1243, opened 2019-02-28
Engine::Get with std::vector fails for increasing size in local variable


williamfgc:
Must update dynamically the reallocation when variable SetBlockSelection is used

PR #1256, opened 2019-03-05
Block count bug

williamfgc:
addressing #1243


williamfgc:
Codacy false positive not understanding #ifdef ADIOS2_HAVE_MPI

PR #1257, opened 2019-03-06
rename adios2_engine_type -> adios2_io_engine_type

germasch:
This is an API change (in the C bindings), but I believe `adios2_engine_type` right now is exclusively used by the Fortran bindings and one test (that I had added), so it's very unlikely that any external user is affected.

Its Fortran (f2c) version is already named that way, and it makes more sense,
since the function actually takes an adios2_io, not an adios2_engine.
This will allow to later add a adios2_engine_type function to the C
bindings that actually takes an adios2_engine (if needed). (The CXX11
interface already has Engine::Type(), which does not have an equivalent
 in the C or Fortran bindings).

I also changed the argument order to take the adios2_io first,
since most other adios2_io binding functions do so.

### DIFF
--- a/bindings/C/c/adios2_c_io.cpp
+++ b/bindings/C/c/adios2_c_io.cpp
@@ -693,8 +693,8 @@ adios2_error adios2_flush_all_engines(adios2_io *io)
     }
 }
 
-adios2_error adios2_engine_type(char *engine_type, size_t *size,
-                                const adios2_io *io)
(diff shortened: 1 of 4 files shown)


germasch:
Force pushed with some piece I missed fixed, and I also just kept the argument order the same, so it's pure renaming.

williamfgc:
Closing this PR. Breaks the current (stable) API, this is a hard sell.

PR #1261, opened 2019-03-07
add some tests for C/CXX11 bindings for IO/Engine

germasch:
This tests IO::Open, Engine::Name, Engine::Type, IO::EngineType, using both CXX11 and C bindings.

The C bindings tests are somewhat incomplete, as two CXX11 functions are missing, so I left those out.

In the last two commits (so they can be dropped if needed), I added CXX11 and C API functions for `core::IO::GetEngine`, which allows to get a handle to an existing Engine by name. There's no chance anything gets broken because these are additions, but there is of course a valid question as to whether those are needed. I'm generally not much of a fan for recovering handles by name where one could have just saved the handle, but ADIOS2 generally allows to replace the name for a handle in many places, I suppose a a convenience in particular for people using the Fortran interface. If this gets merged, I'd follow up with a Fortran binding for GetEngine, which is really the main purpose of it all.

FWIW, the XGC code's interface with ADIOS2 (which I don't understand too well yet), maintains its own list of adios2_engines, parallel to the one that's kept in core::IO which arguably is a bit of a hassle.

### DIFF
--- a/testing/adios2/bindings/C/TestBPWriteTypes.cpp
+++ b/testing/adios2/bindings/C/TestBPWriteTypes.cpp
@@ -19,25 +19,29 @@
 
 #include "SmallTestData_c.h"
 
-class BPWriteTypesCC : public ::testing::Test
+class ADIOS2_C_API : public ::testing::Test
 {
 public:
-    BPWriteTypesCC() = default;
-};
+    ADIOS2_C_API()
+    {
+#ifdef ADIOS2_HAVE_MPI
+        adiosH = adios2_init(MPI_COMM_WORLD, adios2_debug_mode_on);
+        MPI_Comm_rank(MPI_COMM_WORLD, &rank);
+        MPI_Comm_size(MPI_COMM_WORLD, &size);
+#else
+        adiosH = adios2_init(adios2_debug_mode_on);
+#endif
+    }
 
-TEST_F(BPWriteTypesCC, ADIOS2BPWriteTypes)
-{
+    ~ADIOS2_C_API() { adios2_finalize(adiosH); }
+
+    adios2_adios *adiosH;
     int rank = 0;
     int size = 1;
+};
 
-#ifdef ADIOS2_HAVE_MPI
-    adios2_adios *adiosH = adios2_init(MPI_COMM_WORLD, adios2_debug_mode_on);
-    MPI_Comm_rank(MPI_COMM_WORLD, &rank);
-    MPI_Comm_size(MPI_COMM_WORLD, &size);
-#else
-    adios2_adios *adiosH = adios2_init(adios2_debug_mode_on);
-#endif
-
+TEST_F(ADIOS2_C_API, ADIOS2BPWriteTypes)
+{
     // write
     {
         // IO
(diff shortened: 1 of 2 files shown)


williamfgc:
> In the last two commits (so they can be dropped if needed), I added CXX11 and C API functions for core::IO::GetEngine, which allows to get a handle to an existing Engine by name.

We had conversations with our C++ application expert users (OpenFoam and PIConGPU) discouraging any kind of access replication of the `adios2::Engine` object/handle. This opens all kinds of problems as Engines interact with system resources (e.g. threads, file, network, RDMA, etc.) and keep a state (open/close, valid, current step), unlike other handlers that only live in memory and use RAII exclusively (e.g. IO, Variable). I'd remove the GetEngine from this PR as not adding it was intended on purpose, more tests are welcome additions, though. 

> FWIW, the XGC code's interface with ADIOS2 (which I don't understand too well yet), maintains its own list of adios2_engines

In Fortran and C either a module or a struct with appropriate handles take care of the adios2 integration. As long as the module/struct is valid the Engine handler is reachable. Also, Engines other  are enable at compile-time along with their dependencies (except for the default). I'd like to talk more about XGC integration and challenges from the application side. Additions/improvements to adios2 are based solely or apps needs.

bindings/CXX11/cxx11/IO.cpp
```diff
@@ -102,6 +102,13 @@ Engine IO::Open(const std::string &name, const Mode mode)
     return Engine(&m_IO->Open(name, mode));
 }
 
+Engine IO::GetEngine(const std::string &name)
```
williamfgc: See my comments in the PR about not having this being intended.

germasch:
> We had conversations with our C++ application expert users (OpenFoam and PIConGPU) discouraging any kind of access replication of the `adios2::Engine` object/handle. This opens all kinds of problems as Engines interact with system resources (e.g. threads, file, network, RDMA, etc.) and keep a state (open/close, valid, current step), unlike other handlers that only live in memory and use RAII exclusively (e.g. IO, Variable). I'd remove the GetEngine from this PR as not adding it was intended on purpose, more tests are welcome additions, though.

Okay, I dropped the last two commits, so this is testing additions only.

> In Fortran and C apps (speaking of only those I interacted with) either a module or a struct with appropriate handles take care of the adios2 integration. As long as the module/struct is valid the Engine handler is reachable. Also, Engines other are enable at compile-time along with their dependencies (except for the default). I'd like to talk more about XGC integration and challenges from the application side. Additions/improvements to adios2 are based solely or apps needs.

XGC keeps handles as `save` Fortran variables, which I think as such is fine. They're created the first time around, and then used on subsequent output. The problem with this is that there is no designated time to close things down, so I suppose that's why all those engines are added to a list and then close is called for all of them in one final IO shutdown function. Not pretty, but I suppose good enough for now (it looks like the IO itself gets leaked). I think ideally, all this cleanup should happen automatically in adios2_finalize(), not sure whether it actually does.

williamfgc:
> Okay, I dropped the last two commits, so this is testing additions only.

Great, let me know when this is ready to merge.

> Not pretty, but I suppose good enough for now (it looks like the IO itself gets leaked). I think ideally, all this cleanup should happen automatically in adios2_finalize(), not sure whether it actually does.

The only object/handler whose memory is owned by the application is `ADIOS/adios2_adios`, all other are dummy references to things that lives inside ADIOS in C++ RAII containers (including IO). As you pointed out, as long as `adios2_finalize` is called adios2 should be the source of leaks. We verify this running valgrind, (MPI implementations add a lot of noise), but we rely on the C++ STL RAII for other resources atuomatic clean up (e.g. transport Close).

PR #1264, opened 2019-03-08
more testing of IO/Engine functionality and a fix for the Fortran bindings

germasch:
This still covers only a small piece of Engine-related IO functionality, though it was enough to find a bug in the Fortran bindings: When returning strings of 0 length from C to Fortran, the old code
would leave the Fortran string in deallocated state, which leads to a crash when checking its content. Fortunately, Fortran allows for 0-size allocated arrays/strings, so the fix is to return just that.

This PR also adds `adios2_engine_name()` to the C and Fortran APIs, as the same functionality exists in the CXX11 API.

### DIFF
--- a/bindings/C/c/adios2_c_engine.cpp
+++ b/bindings/C/c/adios2_c_engine.cpp
@@ -12,6 +12,7 @@
 
 #include "adios2/core/Engine.h"
 #include "adios2/helper/adiosFunctions.h" //GetType<T>
+#include "adios2_c_internal.h"
 
 namespace
 {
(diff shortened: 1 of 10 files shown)


germasch:
Looks like the failures here are just the bpls/validate issue again, no idea why I seem to keep hitting that one...

williamfgc:
@germasch This is great, please rebase your commits (we try to keep commits smalls ~5, unless there is a critical deadline) and force push again, tests will rerun. Thanks!

williamfgc:
@germasch on a second thought, I remember removing the engine_type thing a long time ago in the Fortran bindings in favor of using engine%type as we moved towards using types. This just needs to be updated dynamically, which is probably a bug.

germasch:
I force-pushed this one with fewer commits.

> @germasch on a second thought, I remember removing the engine_type thing a long time ago in the Fortran bindings in favor of using engine%type as we moved towards using types. This just needs to be updated dynamically, which is probably a bug.

This PR is not touching any engine_type API, it does add engine_name, though, which is also missing entirely from C. I guess you can argue that you don't want to have adios2_engine_name in the Fortran API in favor of engine%name, but the C API doesn't have anything right now for it.

williamfgc:
@germasch yes, please remove the Fortran API call.

germasch:
repushed, not adding any API to Fortran, but I'm still adding `adios2_engine_name` to C, because C is otherwise the only API that has no way of getting an engine's name.

PR #1265, opened 2019-03-08
Assorted clean up and small fixes

germasch:
Here's a bunch of little things I've had lie around for a while now. These are mostly independent from reach other, so pieces could be dropped if needed.

It's mostly some cleanup following from the stdint type changes. Maybe the most noteworthy change is moving the `ADIOS2_FOREACH_*_TYPE_1ARG` macros into the CXX11 bindings, as all of the rest of
the code now uses he `_STDTYPE_` macros, so having the `_TYPE_` macro out of the way should prevent any confusion.

### DIFF
--- a/bindings/C/c/adios2_c_attribute.cpp
+++ b/bindings/C/c/adios2_c_attribute.cpp
@@ -38,7 +38,7 @@ adios2_error adios2_attribute_name(char *name, size_t *size,
     }
 }
 
-adios2_error adios2_attribute_type(adios2_type *type,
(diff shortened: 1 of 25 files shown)


williamfgc:
@germasch not sure this adds any value, and create a lot of conflict with current dev branches. Keeping MACRO type in one file is not a bad thing. Also, for future reference let's keep the number of commits small. You can always rebase `git rebase -i ` to squash. We would like to keep master clean, if possible. Are you using an IDE, BTW?

PR #1267, opened 2019-03-08
[COMMENTS?] Fix RemoveVariable/Attribute to really remove the Variable/Attribute

### LINKED ISSUE #1193: RemoveVariable() doesn't fully remove Variable
This is going to be a bit convoluted. So looking at the core::IO code,  thought I saw a bug in how the variable maps are handled. So I tried to actually trigger it, but I couldn't (immediately).

Instead, I found another bug: When the Variable is removed from the `variableMap`, the code actually gets a copy of the entire map, then removes the Variable from the copy, and the original map is unchanged. So that's easy to fix:

```diff
--- a/source/adios2/core/IO.cpp
+++ b/source/adios2/core/IO.cpp
@@ -165,7 +165,7 @@ bool IO::RemoveVariable(const std::string &name) noexcept
 #define declare_type(T)                                                        \
     else if (type == helper::GetType<T>())                                     \
     {                                                                          \
-        auto variableMap = GetVariableMap<T>();                                \
+        auto& variableMap = GetVariableMap<T>();                                \
         variableMap.erase(index);                                              \
         isRemoved = true;                                                      \
     }
```

Now, however, a more fundamental problem with these maps appears, that is, they get into inconsistent state when a variable is removed and then another variable is added. Here is a test:
```cxx
void test2()
{
  using T = int32_t;

  adios2::ADIOS ad("adios.xml", MPI_COMM_WORLD, adios2::DebugON);

  {
    adios2::IO io_writer = ad.DeclareIO("io_writer");
    auto var = io_writer.DefineVariable<T>("var");
    auto var2 = io_writer.DefineVariable<T>("var2");
    auto writer = io_writer.Open("test", adios2::Mode::Write);
    io_writer.RemoveVariable("var");
    auto var3 = io_writer.DefineVariable<T>("var3");
    std::cout << "var2 name " << var2.Name() << std::endl;
    std::cout << "var3 name " << var3.Name() << std::endl;
    
    T test = 99;
    writer.Put(var3, test);
    writer.Close();
  }
}
```

When run, it prints
```
var2 name var2
var3 name var2
```
`var3` didn't really get added, instead it became an alias for `var2`. The problem can be more explicitly seen with this patch:
```diff
diff --git a/source/adios2/core/IO.tcc b/source/adios2/core/IO.tcc
index f595748e..447d03f7 100644
--- a/source/adios2/core/IO.tcc
+++ b/source/adios2/core/IO.tcc
@@ -48,6 +48,8 @@ Variable<T> &IO::DefineVariable(const std::string &name, const Dims &shape,
     auto itVariablePair =
         variableMap.emplace(size, Variable<T>(name, shape, start, count,
                                               constantDims, m_DebugMode));
+    bool emplaceSucceeded = itVariablePair.second;
+    if (!emplaceSucceeded) throw std::runtime_error("emplace failed in IO::DefineVariable");
     m_Variables.emplace(name, std::make_pair(helper::GetType<T>(), size));

     Variable<T> &variable = itVariablePair.first->second;
```
which will make it crash rather than exhibit the incorrect behavior. (Note: There may be a point in always checking the return value from std::map::emplace(), at least in an assert() kind of fashion).

The underlying problem is the one I noted when reading `DefineVariable` in the first place: The code is assuming that variableMap.size() is going to give it a new, unused index, but that is not the case with a map that can shrink. The easiest solution I can think of is to just maintain a per-map index which gets incremented for each Variable that's inserted, but never decremented. That should be fine as long as it doesn't wrap around.

germasch:
This PR fixes #1193, see there for some background.

This PR fixes the issue by keeping a monotonically increasing counter for the index into the variable map, so no two variables can have the same index (unless the counter wraps around, which for an `unsigned int` takes a long time -- but it could be changed to `uint64_t` if one were to worry about it.) Since we now have to keep a counter together with the map, and those are macro generated member variables for each type, I made a custom type for it, `EntityMap`, which is essentially a wrapper around `std::map` that keeps track of the last index used.

It should be noted that while I believe this PR is bug-free, applying it may expose other bugs. Since the existing (buggy) code doesn't actually remove Variables/Attributes, handles remain valid, and so access after remove causes no troubles. Now that objects actually get freed, you better don't access them afterwards.

The PR in its current shape doesn't quite follow the style guide, the extra `class EntityMap` is defined in `core/IO.h` and uses inline definitions. I can easily clean that up, though, if the PR is otherwise acceptable. It includes a test that first works (because the bug is masked by not actually removing variables), then fails after that has been fixed, and finally passes again when using the monotonic and hence unique index.

### DIFF
--- a/source/adios2/ADIOSTypes.h
+++ b/source/adios2/ADIOSTypes.h
@@ -137,6 +137,10 @@ using std::uint64_t;
 using cfloat = std::complex<float>;
 using cdouble = std::complex<double>;
 
+// Compound is never actually defined, it's just used to template with, in
+// particular Variable<Compound>.
+class Compound;
+
 // Limit, using uint64_t to make it portable
 constexpr uint64_t MaxU64 = std::numeric_limits<uint64_t>::max();
 constexpr size_t MaxSizeT = std::numeric_limits<size_t>::max();
--- a/source/adios2/core/ADIOS.cpp
+++ b/source/adios2/core/ADIOS.cpp
@@ -131,8 +131,10 @@ IO &ADIOS::DeclareIO(const std::string name)
         }
     }
 
-    auto ioPair = m_IOs.emplace(
-        name, IO(*this, name, m_MPIComm, false, m_HostLanguage, m_DebugMode));
+    auto ioPair =
+        m_IOs.emplace(std::piecewise_construct, std::forward_as_tuple(name),
+                      std::forward_as_tuple(*this, name, m_MPIComm, false,
+                                            m_HostLanguage, m_DebugMode));
     IO &io = ioPair.first->second;
     io.SetDeclared();
     return io;
--- a/source/adios2/core/IO.cpp
+++ b/source/adios2/core/IO.cpp
@@ -208,7 +208,7 @@ bool IO::RemoveAttribute(const std::string &name) noexcept
 #define declare_type(T)                                                        \
     else if (type == helper::GetType<T>())                                     \
     {                                                                          \
-        auto attributeMap = GetAttributeMap<T>();                              \
+        auto &attributeMap = GetAttributeMap<T>();                             \
         attributeMap.erase(index);                                             \
         isRemoved = true;                                                      \
     }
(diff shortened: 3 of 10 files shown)


williamfgc:
@germasch Creating a wrapper to a C++ container doesn't justify what we are trying to solve. Remove is not a critical function. I wouldn't massage the internals this way.

PR #1270, opened 2019-03-08
use GetType<std::string> consistently

germasch:
Sometimes we use `GetType<std::string>`, sometimes just `"string"`. The code generally uses `GetType<T>()` instead of hardcoded strings for the other types, so this patch consistently does the same for strings.

Also fixes one external use from `helper::GetType` to `GetType`.

### DIFF
--- a/examples/hello/datamanReader/DataManCallbackReceiver.cpp
+++ b/examples/hello/datamanReader/DataManCallbackReceiver.cpp
@@ -16,7 +16,6 @@
 #include <vector>
 
 #include "adios2/ADIOSMacros.h"
-#include "adios2/helper/adiosFunctions.h"
 #include <adios2.h>
 
 // matches Signature2 in ADIOS2
@@ -53,7 +52,7 @@ void UserCallBack(void *data, const std::string &doid, const std::string &var,
     std::cout << "Data : " << std::endl;
 
 #define declare_type(T)                                                        \
-    if (type == adios2::helper::GetType<T>())                                  \
+    if (type == adios2::GetType<T>())                                          \
     {                                                                          \
         for (size_t i = 0; i < dumpsize; ++i)                                  \
         {                                                                      \
--- a/bindings/C/c/adios2_c_attribute.cpp
+++ b/bindings/C/c/adios2_c_attribute.cpp
@@ -152,7 +152,7 @@ adios2_error adios2_attribute_data(void *data, size_t *size,
         {
             // not supported
         }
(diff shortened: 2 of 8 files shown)


examples/hello/datamanReader/DataManCallbackReceiver.cpp
```diff
@@ -53,7 +53,7 @@ void UserCallBack(void *data, const std::string &doid, const std::string &var,
     std::cout << "Data : " << std::endl;
 
 #define declare_type(T)                                                        \
-    if (type == adios2::helper::GetType<T>())                                  \
+    if (type == adios2::GetType<T>())                                          \
```
williamfgc: In this is the case then #include "adios2/helper/adiosFunctions.h" needs to be removed. I'll check with @JasonRuonanWang since he's the one maintaining this.

germasch: I'll be glad to do that.

Also, I forgot to add, if these changes, simple as they are, cause conflicts in current development branches, I'll be glad to hold on to them until those are merged, though I'd have to know.

williamfgc: > cause conflicts in current development branches

For the most part you touched code I maintain, as you go deeper and deal with other engines people funded to develop are responsible for those pieces. I don't want to break their trust so they need to be informed and approve.

JasonRuonanWang: As long as it passes CI, I don't mind removing any header includes.

germasch: I'm perfectly happy with other people reviewing changes that touch their code, in fact  think that's how it should be. I can try to look in the file headers to see who I should try to @mention.

germasch: @JasonRuonanWang other than the header removal, this PR switches `DataManCallbackReceiver.cpp` to use `GetType<T>()` rather than `helper::GetType<T>()`, as external code using the cxx11 bindings should use the former. It won't make any actual difference in that context, but the former handles all C++11 types (e.g, int, long, long long) rather than just stdint base type (e.g. int32_t, int64_t). `helper::GetType<T>()` should be used inside the adios2 implementation, but external code should use `GetType<T>()` as a general rule.

germasch:
re-pushed with the adiosFunctions.h include removed.

PR #1278, opened 2019-03-11
Remove variable and attribute bug

williamfgc:
Addressing #1193


germasch:
FWIW, that's definitely the better fix compared to what I had done.

williamfgc:
@germasch no pb, thanks for catching the bug.

Issue #1132, opened 2019-02-04
NULL engine requested


pnorbert:
Sometimes for performance testing, the easiest way to turn off an output from in an existing application is just to use the NULL engine (in adios 1.x).

williamfgc:
@pnorbert let's talk about it tomorrow, it wasn't clear to me if NULL means 0 buffering and 0 POSIX I/O write, or it's just the latter. Either way, shouldn't be hard to implement, though.

PR #1280, opened 2019-03-11
Adding NullEngine

williamfgc:
Addressing #1132 
Added C++11 bindings null engine

### DIFF
--- a/bindings/C/c/adios2_c_engine.cpp
+++ b/bindings/C/c/adios2_c_engine.cpp
@@ -101,11 +101,6 @@ adios2_error adios2_engine_name(char *name, size_t *size,
         const adios2::core::Engine *engineCpp =
             reinterpret_cast<const adios2::core::Engine *>(engine);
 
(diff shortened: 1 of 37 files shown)


chuckatkins:

germasch:
What's the goal of the "NULL" engine? From #1132, it sounds like it's like `/dev/null` for writing. It looks to me like `SkeletonWriter` basically fits the bill. For reading, I don't think it's clear what one would want.

Independently of that, it might be useful to have a Transport that just sinks everything into a black hole == if one wanted to benchmark, e.g., bp3 serialization without the cost of the actual I/O.

williamfgc:
@germasch yes, that's the goal, switching to NULL should allow to measure adios2 engine overhead in an application. @pnorbert and @jychoi-hpc have used this approach for a long time in adios1.x

williamfgc:
After a discussion with @chuckatkins I am merging this for initial test. If the intention is to only "nullify" the core, we can look for an alternative. This is currently working on the write and read sides to address #1132.

PR #1282, opened 2019-03-12
Docs: Fix Verbatim Markup

ax3l:
I took the time to read the docs and fixed all issues with verbatim markup on the way.

Great docs so far! :)

Please add complex read examples, including "list me all variables" and "list me all attributes" for nested names.


williamfgc:
@ax3l this is great, thanks so much!  BTW, there is the `IO::AvailableVariables` and `IO::AvailableAttributes`  (in high-level Python and C++ APIs it's  available_variables and available_attributes). They return a 2-level std::map (in Python a dictionary) [var_name, [info_key, info_value] ].

Issue #1268, opened 2019-03-08
API Docs/Cheet Sheet: Collective, Blocking and Memory Contracts


ax3l:
Hi,

we are currently sitting in the ADIOS2 tutorial at TU Dresden and as with ADIOS1, the following would be extremely useful:

A cheat sheet (or API documentation) of all public API calls (engine open/close, defines, step begin/end puts) that documents for each of those:
- is it MPI collective or local?
- is it MPI blocking/non-blocking?
- document memory contracts: when is a user allowed or forbidden to modify shared buffers and what will happen if the user does (ignored, overwritten, undefined behavior, etc.)
- does any of the above change with choice of engine?

I personally like a lot about ADIOS1, that most calls on variables are local, non-blocking and not even collective in the MPI sense, but just explicitly documenting this will make starting with ADIOS2 for folks that already know MPI (everyone in HPC) so much easier.

williamfgc:
@ax3l point taken. Thanks for the feedback as usual. In the meantime, have you had the chance to look at https://adios2.readthedocs.io/en/latest/ ?  Would the section on language API bindings (uses doxygen) help? We are always improving the docs based on applications feedback.  Thanks!

ax3l:
Thanks, we enjoy the staging demos very much over here.

> https://adios2.readthedocs.io

Argh, I have overlooked that in the README. Do you want to link it in the repo URL (next to the repo description on GitHub) just to make it more prominent? Will dig in.

williamfgc:
@ax3l good idea. Feel free to provide feedback if the docs are not clear (I will add the cheat sheet on MPI-related calls), all the sources live under `ADIOS2/docs` and we use the `breathe` package to pull doxygen API info from C++ and C headers under `ADIOS2/bindings`. I haven't found a good way to automate Python and Fortran APIs due to their not so friendly nature with doxygen.

germasch:
> * is it MPI collective or local?

Let me second this particular point, as that is something that I'm still not clear on. The only thing I've seen in the docs is

```
Always pass MPI_COMM_SELF if an Engine lives in only one MPI process. Open and Close are collective operations.
```

Which makes sense, but I'm sure that's not all that's collective. In particular, how about `PerformPuts()`? (I'd expect that to be collective) And a simple `Put()`? (I'd expect that to be local for, e.g. BP3, for sure, but I'm much less sure what happens if you're using some M procs -> N files mapping?)

williamfgc:
[...]

PR #1286, opened 2019-03-13
Added docs

williamfgc:
MPI collectives in C++ and C bindings
What ADIOS2 is and isn't in rtd

Addressing #1268  in Doxygen API documentation in readthedocs


williamfgc:
@ax3l does this help #1268

ax3l:
Thanks, that's excellent! :sparkles:

PR #1303, opened 2019-03-18
fix a bug in my test; expose another data corruption bug

germasch:
This is the bug I was actually going for earlier, but I had screwed up my test instead.

### DIFF
--- a/testing/adios2/interface/TestADIOSInterface.cpp
+++ b/testing/adios2/interface/TestADIOSInterface.cpp
@@ -293,16 +293,17 @@ TEST_F(ADIOS2_CXX11_API_Put, MultiBlockPutZeroCopySync2)
     {
         PopulateBlock(myData, b);
     }
+
     std::vector<T> lastBlock(m_Nx / 2);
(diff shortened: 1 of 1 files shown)


testing/adios2/interface/TestADIOSInterface.cpp
```diff
@@ -318,18 +319,19 @@ TEST_F(ADIOS2_CXX11_API_Put, MultiBlockPutZeroCopySync3)
...
 
-    for (int b = 0; b < 1; ++b)
-    {
-        PopulateBlock(myData, b);
-    }
+    // writing block 1 Deferred
+    std::vector<T> lastBlock(m_Nx / 2);
+    std::iota(lastBlock.begin(), lastBlock.end(), T(rank * m_Nx + m_Nx / 2));
+    var.SetSelection(myData.selection(1));
+    engine.Put(var, lastBlock.data(), adios2::Mode::Sync);
```
williamfgc: @germasch At a first glance (I am not running a debugger to confirm this) is that this line might be resizing the buffer thus invalidating the span pointer that is not updated dynamically, calling span.data() after this Put in sync should point to the right location.  Let me know if this works. This is really good for illustration purposes.

PR #1305, opened 2019-03-19
[WIP] prealloc (span) work

germasch:


williamfgc:
Adding a new function breaks the abstraction and makes polymorphism completely unnecessary. The minute your return a span the preallocation is redundant. This was internally decided. Ultimately, you can fork and maintain your version or wrap around the function. If someone is willing to reopen, revisit and back off from its original decision they can as well maintain it. Not only that, but this break re-usability of Get returning a span in streaming (buffer is already pre-allocated). Closing this from my end.

Issue #1320, opened 2019-03-24
sst takes longer than needed buffer from bp3


germasch:
@eisenhauer, I noticed that SstWriter uses `m_Data.m_Buffer.size()` to get the size of the provided buffer. That actually returns the capacity of `m_Data`, though, that is, it's potentially larger than the actual serialized data. It's not likely to have any actual consequences (I don't know if that size is even used at all. If the buffer were copied, it could potentially lead to copying more data than needed, but I don't thing it is ever copied in the first place.) 

Potential fix:

```diff
diff --git a/source/adios2/engine/sst/SstWriter.cpp b/source/adios2/engine/sst/SstWriter.cpp
index b17f7f92..e0e971a2 100644
--- a/source/adios2/engine/sst/SstWriter.cpp
+++ b/source/adios2/engine/sst/SstWriter.cpp
@@ -161,7 +161,7 @@ void SstWriter::EndStep()
         newblock->metadata.DataSize =
             m_BP3Serializer->m_Metadata.m_Buffer.size();
         newblock->metadata.block = m_BP3Serializer->m_Metadata.m_Buffer.data();
-        newblock->data.DataSize = m_BP3Serializer->m_Data.m_Buffer.size();
+        newblock->data.DataSize = m_BP3Serializer->m_Data.m_Position;
         newblock->data.block = m_BP3Serializer->m_Data.m_Buffer.data();
         newblock->serializer = m_BP3Serializer;
         TAU_STOP("Marshaling overhead");
```

That's from how BP3Writer does it:
```cxx
void BP3Writer::WriteData(const bool isFinal, const int transportIndex)
{
    size_t dataSize = m_BP3Serializer.m_Data.m_Position;

...
```

eisenhauer:
@JasonRuonanWang, kicking to you because it's in the BP3 marshaling code...

JasonRuonanWang:
I don't think std::vector<>::size() returns the capacity. It returns the size literally. Instead, std::vector<>::capacity() returns the capacity. 

In any way, at this point, m_Buffer.size() should be exactly the data size that has been used. It should never be over allocated. If it is then I would consider it a bug. So I don't see there is any problem with this.

germasch:
[...]

PR #1325, opened 2019-03-26
Fix size of marshalled BP buffer

eisenhauer:
Refer to issue #1320 .  Thanks @germasch !

### DIFF
--- a/source/adios2/engine/sst/SstWriter.cpp
+++ b/source/adios2/engine/sst/SstWriter.cpp
@@ -158,8 +158,7 @@ void SstWriter::EndStep()
         m_BP3Serializer->AggregateCollectiveMetadata(
             m_MPIComm, m_BP3Serializer->m_Metadata, true);
         BP3DataBlock *newblock = new BP3DataBlock;
-        newblock->metadata.DataSize =
-            m_BP3Serializer->m_Metadata.m_Buffer.size();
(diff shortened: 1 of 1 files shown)


source/adios2/engine/sst/SstWriter.cpp
```diff
@@ -158,8 +158,7 @@ void SstWriter::EndStep()
         m_BP3Serializer->AggregateCollectiveMetadata(
             m_MPIComm, m_BP3Serializer->m_Metadata, true);
         BP3DataBlock *newblock = new BP3DataBlock;
-        newblock->metadata.DataSize =
-            m_BP3Serializer->m_Metadata.m_Buffer.size();
+        newblock->metadata.DataSize = m_BP3Serializer->m_Metadata.m_Position;
         newblock->metadata.block = m_BP3Serializer->m_Metadata.m_Buffer.data();
```
germasch: So this one is actually not the buffer I meant ;) I wasn't talking about the metadata buffer but about the actual data buffer. I haven't looked at the metadata buffer to be sufficiently sure about that one, but from what I remember, I believe that one actually uses m_Buffer as the actual buffer, and doesn't use m_Position.

eisenhauer: The BP3writer uses m_Position for both m_Metadata and m_Data sizes.  So updating both for consistency.  But you might be right about it not being necessary for m_Metadata.

PR #1327, opened 2019-03-26
Some template-related cleanup for Span

germasch:
This is basically all cleanup only. The one API change is that `cxx11::Span` and `core::Span` become top-level classes (ie, not nested within `Variable`). That is for two reasons:
* So that `core::Span` can be forward declared and hence directly used in the pimpl-type `cxx11::Span` without resorting to `reinterpret_cast`.
* So that the `Spans` can be defined / instantiated as a whole class, rather than being all specialized by a macro loop for each type. 

There is no user-visible change, except that `Span` is now also available as `adios2::Span<T>` in addition to `adios2::Variable<T>::Span`, which I'd argue makes life a bit easier. But if you want avoid polluting of the `adios2` namespace, it could be moved to `adios2::detail`.

The patch does remove `Span::rbegin` and `Span::rend` because they were broken (wouldn't compile, but that didn't show because they were never instantiated). They obviously could be added back if one were to properly implement bidirectional or random access operators.

This PR reduces code size (lines of code):
```
 7 files changed, 146 insertions(+), 186 deletions(-)
```

### DIFF
--- a/bindings/CXX11/cxx11/Variable.cpp
+++ b/bindings/CXX11/cxx11/Variable.cpp
@@ -233,59 +233,48 @@ ADIOS2_FOREACH_TYPE_1ARG(declare_template_instantiation)
 #define declare_type(T)                                                        \
                                                                                \
     template <>                                                                \
-    Variable<T>::Span::Span(CoreSpan *coreSpan) : m_Span(coreSpan)             \
+    Variable<T>::Span::Span(core::Span<IOType> *coreSpan)                      \
+    : m_Span(coreSpan)                                                         \
     {                                                                          \
     }                                                                          \
                                                                                \
     template <>                                                                \
     size_t Variable<T>::Span::size() const noexcept                            \
     {                                                                          \
-        const core::Variable<IOType>::Span *coreSpan =                         \
-            reinterpret_cast<const core::Variable<IOType>::Span *>(m_Span);    \
-        return coreSpan->Size();                                               \
+        return m_Span->Size();                                                 \
     }                                                                          \
                                                                                \
     template <>                                                                \
     T *Variable<T>::Span::data() const noexcept                                \
     {                                                                          \
-        const core::Variable<IOType>::Span *coreSpan =                         \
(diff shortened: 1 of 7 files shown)


chuckatkins:

williamfgc:
@germasch I think this is great, truly templated code is always better than macro. I just have two minor requests:

1. Following your own thoughts, let's make `adios2::Span<T>` private to the user, spans can't exist without the notion of a `adios2::Variable<T>`, `adios2::Variable<T>::Span` gives the scope. Other objects can live on the own, `adios2::Span<T>` can't.

2. Reduce the number of commits (`git rebase -i HEAD~n` where n is number of commits back from head). We typically go up to 5 commits per PR, unless we fix CI related stuff.

We can work out rbegin, rend for a later PR (thanks for checking they don't work, I was probably experimenting how they would fit, but never tested). Also, once we get to the read side`cbegin` and `cend` would be very required if we want to provide read-only access to the Variable<T>::Span data.
Great work, thanks!

germasch:
Will do. I think this PR may conflict with the changes in PR #1328, so I'll wait until that one comes to a conclusion, so I can deal with the conflicts at the time.

williamfgc:
@germasch was this ready to merge? I see it has some conflicts that need to be resolved. Thanks!

pnorbert:
Don't merge this until it is tested on Titan if it works as intended.

germasch:
I actually just dusted this off last night. I'll push a rebased version.

pnorbert:
Sorry, I replied to the wrong thread.

williamfgc:
@pnorbert I tested the use of Spans on Titan (I'll do Summit today), works fine.....see PR in adiosvm: https://github.com/pnorbert/adiosvm/pull/27 . This PR is @germasch cleanup of the internal interface which is much better than the current state.

germasch:
I built it on Summit, and the tests ran fine. But obviously it's still possible that some compiler somewhere will be unhappy. However, since this PR only adds some new functionality, and doesn't change existing code, if there is any breakage, it should show at compile time, not cause subtle issues with existing working applications.

PR #1330, opened 2019-03-26
fixed sst-bp bug for local arrays

JasonRuonanWang:


JasonRuonanWang:
@eisenhauer This seems to work with the local arrays in adios2-scratch. Can we add a test for that?

eisenhauer:
Yes, I have a mostly-finished local data test for SST.  Once this gets merged I'll finish that up.  (Ideally, the tests that are used for files should be reusable (and actually be reused) for staging engines, but we're not there yet.)

JasonRuonanWang:
Yeah, I was surprised too. This is the fun part when you try to glue two pieces of code which you are missing a lot of details in either of them. You have to go through a whole lot of hard thinking until you know it's just a single line fix :)

PR #1332, opened 2019-03-27
Added busy command for adios_iotest

philip-davis:
Busy continuously checks the high resolution clock until enough time has elapsed. I think this might be nice to emulate potential CPU interference from the user app on a node.


pnorbert:
Can you add a busy version to the pipe2 example so that the usage of the new command is "documented :-) "?

PR #1338, opened 2019-03-30
Use sane default rpath settings

chuckatkins:


chuckatkins:
@dpugmire @pnorbert @williamfgc this should let you not have to set `LD_LIBRARY_PATH` after installation

PR #1346, opened 2019-04-02
Allow streaming blocking in read mode Python High Level API

williamfgc:


williamfgc:
@pnorbert @JasonRuonanWang killing this PR, the solution is handle inside the engine. Thanks.

PR #1350, opened 2019-04-02
Remove semicolon from macro define

gsjaardema:
The iterator class is defined via the macro `ADIOS2_CLASS_iterator`.  The define ends with a semicolon.  When the symbol is used in source code,  the developer will add a semicolon after the symbol as is done for most code.  This will result in a warning from some compilers about a double semicolon on that line (one from macro, one from line of code)


williamfgc:
@gsjaardema this is great, thanks!

PR #1352, opened 2019-04-03
Print __ for dimensions for global arrays with changing shape

pnorbert:


source/utils/bpls/bpls.cpp
```diff
@@ -758,10 +758,12 @@ int printVariableInfo(core::Engine *fp, core::IO *io,
 
         if (variable->m_ShapeID == ShapeID::GlobalArray)
         {
-            fprintf(outf, "{%zu", variable->m_Shape[0]);
+            Dims d = get_global_array_signature(fp, io, variable);
+            fprintf(outf, "{%s", d[0] > 0 ? std::to_string(d[0]).data() : "__");
```
germasch: Minor comment: Whereas this will likely work just fine with most implementations of std::string, the official way is to use `.c_str()` instead of `.data()`.

chuckatkins: I'll second that. Although i'd also say this should use `fstream` but that's a bigger overhaul.  If you're going to stick with the C I/O stuff for now then `c_str()` is the right way to get `std::string` to C output functions.

germasch: By the way, just in case anyone is interested, the reason for the change that @pnorbert now made is that while `.data()` will typically return the same pointer as `.c_str()`, the return from the former is not guaranteed to be zero-terminated.

Issue #1322, opened 2019-03-24
Put/Span


germasch:
So I'm kinda reluctant to open yet another issue related to spans, nor do I actually much feel like going through this discussion again. Then again, it's still an open issue, so I'll try to summarize where I think we're at.

* `Put(Sync)` and `Put(Deferred) -> PerformPuts()` mirror `MPI_Send` and `MPI_Isend -> MPI_Wait` in terms of when data is ready when data can be touched by whom.

If this is not what ADIOS2's API meant to express, what follows may not apply.

* A way for the app to write directly into ADIOS2's internal buffer can be useful in certain cases, because it avoids a copy of the data (avoids performance penalty) and allocating a second data buffer in the first place (avoids allocating twice as much memory as actually needed.) (#1315)

I think we agree on that. The current Span interface achieves those goals. It comes with caveats (I'm trying to summarize #1290 primarily -- that's a long thread, so the below won't be complete. If you feel I forgot anything important, let me know and I'll add it):
* The memory it returns may not be aligned to usual malloc guarantees (#1314). This may or may not be fixable, and it may or may no be an actual problem for users of the Span interface (but it should be documented at least)
* If the API to return the buffer to be filled to the user is called `Put`, that breaks the expectation that data is ready at the time of `Put`, which is currently satisfied by the two existing `Put(Sync)` and `Put(Deferred)`.
* The buffer returned to the user is subject to change if the underlying adios2 buffer gets realloced. This should be documented. To me, it also means that the choice "Span" is a questionable name for what's returned, because it was meant to imply an analogy to `std::span`, but `std::span` never changes `data()` under you. Anyway, can be handled with documentation.
* The current implementation of this interface uses `Put` to return the buffer, but then doesn't use `Put` at the time the data is ready, so implicitly, the data is ready at `PerformPuts` time. To me, that's not consistent with the existing interface, and also means that there is no API to support potential overlap (the Put(Deferred) -> PerformPuts type) for engines that can use it, like InsituMPI.

So what I suggested in #1290 are two changes:
* rename the Span-returning function to `PutPrealloc` (The actual name is a suggestion that could be discussed (one alternative would be `GetPutBuffer` or something like this). I don't think I've found a perfect name for it, but IMO the current choice of `Put` is already taken to mean something else.
* use `Put(Sync)` and `Put(Deferred)` on prealloced buffers just like on regular buffers.

[...]

PR #1353, opened 2019-04-03
Resolving Put ambiguity

williamfgc:
Separate Variable<T>::Span Put into two
Add single value Put in Span test
This doesn't break or changes current functionality, just fixes a broken
signature
found in #1322


germasch:
So, it's probably obvious, but in my opinion dealing with the ambiguity that is caused by too many `Put` overloads by adding yet another `Put` overload is not exactly the best way to go. (Yes, it technically resolves the ambiguity for the compiler, but one shouldn't forget that humans have to deal with the API, too.)

PR #1356, opened 2019-04-03
fixed a bug in wdm for getting latest available step

JasonRuonanWang:


source/adios2/engine/wdm/WdmReader.cpp
```diff
@@ -135,6 +140,42 @@ StepStatus WdmReader::BeginStepIterator(StepMode stepMode,
...
+
+    size_t maxStep = std::numeric_limits<size_t>::min();
+    for (auto &i : m_MetaDataMap)
+    {
+        if (i.first > maxStep)
+        {
+            maxStep = i.first;
+        }
+    }
+    if (m_CurrentStep > maxStep)
```
germasch: A more elegant (and O(1)) way of getting the maximum key in a map is:
```cxx
if (map.empty()) {
  // FIXME, that may not be what you want, but it's what the code currently does
  maxStep = std::numeric_limits<size_t>::min(); 
} else {
  maxStep = map.rbegin()->first;
}
```

JasonRuonanWang: Thanks for the point. If I remember this is an unordered map. I'm not sure in this case if the rbegin will always give the maximum key

germasch: Sorry, if it's really an unordered_map, my point doesn't hold. I didn't check, I just assumed it was std::map as that's what's used all over the code.

PR #1362, opened 2019-04-07
Clarify build instructions

NAThompson:


williamfgc:
@NAThompson seems the CI got stuck, please do an git amend commit and a force push to trigger it again. Thanks!

NAThompson:
@williamfgc : I intentionally [CI SKIP]'d since this edits no source code.

williamfgc:
@NAThompson thanks for clarifying. My understanding is that we can't merge a PR without rerunning the CI even if does nothing to the code. @chuckatkins can you chime in? Thanks!

chuckatkins:
The CI tests are set up as required right now.  What would be nice though is if we adjust the ci builds themselves to adapt to the diff in the PR and give an early exit for changes that aren't affecting the build.  This is a helpful use case for that, so thank you.  Until we implement that though we can't skip them.

chuckatkins:
I'm also not particularly on board with this change.  I agree that the current documentation is unclear, but I don't think this really helps that and is actually wrong since it implies calling cmake on the source sub-directory instead of the top level build tree.

NAThompson:
Excellent; I thought the instruction to build on the source was weird as well so this can just be fixed rather than an amend commit.

chuckatkins:
Better clarity would be something like this:

1. Clone the repository:

```bash
[~/work]$ git clone https://github.com/ornladios/adios2.git adios2-source
```

2. Create a separate build directory in your work area:

```bash
[~/work]$ mkdir adios2-build && cd adios2-build
```

3. Configure the project with CMake:

```bash
[~/work/adios2-build]$ cmake -DCMAKE_INSTALL_PREFIX=/opt/adios2/2.3.1/gnu/openmpi ../adios2-source
-- The C compiler identification is GNU 7.3.0
-- The CXX compiler identification is GNU 7.3.0
...
```

In general I'd like to encourage people to do "true" out of source builds rather that a build directory inside the source tree.  While either way will work, the total separation of source and build products is a core foundation of CMake.

williamfgc:
I tend to agree with @chuckatkins , I have removed repo source code several times when I meant to remove build directories. @NAThompson , if you don't mind, can you replicate the same info in the readthedocs source for consistency? They are under `ADIOS2/docs/user_guide/source/installation/cmake.rst`
Thanks you both, @chuckatkins and @NAThompson

NAThompson:
@williamfgc : I don't mind; also sorry for DDOS'ing the CI system.

ReadMe.md
```diff
@@ -104,7 +92,7 @@ ADIOS2 build configuration:
 
 -- Configuring done
 -- Generating done
--- Build files have been written to: /home/chuck/adios2/build
+-- Build files have been written to: /home/chuck/ADIOS2/build
```
chuckatkins: `/home/chuck/ADIOS2/build` -> `/home/chuck/adios2_build`

NAThompson: Is the cmake output necessary? It seems reasonable to get rid of it all.

chuckatkins: Necessary, no, helpful, yes.  The documentation shows an abbreviated output to inform users the kind of information they can expect to see.  Namely things like the compiler id in the beginning and option summary and the end let the user know they have a sane configuration.  That being said, I'm fine too if instead of fixing the path in the output you want to just remove that last line "Build files have been written to..." from the output in the docs as it's not particularly informative like the rest of it is.  Up to you.

PR #1364, opened 2019-04-08
Fix build error on xl-16.1.0

chuckatkins:


chuckatkins:
@pnorbert see https://open.cdash.org/viewBuildError.php?buildid=5845493

germasch:
Please note that I added the initializer in 10f1fe164888 for a reason, ie, having that allocation left uninitialized is a bug. So if a compiler is unhappy about it, I'd suggest zeroing that memory in some alternate way.

chuckatkins:
@germasch , tbh I didn't even look at the history, I was just trying to fix the XL builds.  Given that you introduced the build failure in 10f1fe1 then I'm going to go ahead and close this.  Please push an appropriate fix.

Thank

germasch:
Can you try to just replace `{}` by `()`? I don't think I have access to that build in order for me to try myself. If not, I suppose you can use `std::fill` to zero the memory.

Of course, it'd be even better to clean up this code in general. It's been a while since I looked at it, and even then I don't think I completely understood what it's doing. I think it was creating an argv[0], which is usually the executable, in order to make retry_args look like a regular argv array.

PR #1365, opened 2019-04-08
workaround xl-16 compiler bug(?)

germasch:
This attempts an alternate fix to the issue encountered in #1364.

I think that using new style initialization {} was legal in this context,
but I might be wrong. Let's see whether the compiler is happier with the
alternate style.

This isn't actually tested on the compiler that's giving the trouble, so hopefully @chuckatkins can o that.

### DIFF
--- a/source/utils/bpls/bpls.cpp
+++ b/source/utils/bpls/bpls.cpp
@@ -206,7 +206,7 @@ int process_unused_args(adios2sys::CommandLineArguments &arg)
     arg.GetUnusedArguments(&nuargs, &uargs);
 
     std::vector<char *> retry_args;
-    retry_args.push_back(new char[4]{});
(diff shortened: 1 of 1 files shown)


chuckatkins:
LGTM.  Tested on xl-16.1.0 and can confirm it fixes the build error on 16.1.0.  @germasch thanks for the fix; I'll merge it after CI.

chuckatkins:
I believe this worked in xl-13.1.6, was broken in 16.1.0, and seems to work again in xl-16.1.1.  FWIW we've been generally trying to avoid the `{}` constructor syntax (sometimes we're better at it than others) since compiler support seems to have been pretty buggy and unreliable when you get to the closed source compilers (xl, pgi, intel, cray), this case in particular demonstrating that.

PR #1366, opened 2019-04-09
Remove container-overflow error recognized by AddressSanitizer.

NAThompson:


chuckatkins:
lgtm, @NAThompson Thanks for the fix.

tldr:  I believe `resize` here is the appropriate "quick fix" right now, but long term is more complicated.

@pnorbert  I believe the reason for using `reserve` over `resize` was to avoid the cost of pre-initalization, is that correct?  It does violate the contracts in the vector API though, as asan is revealing.  A more appropriate data structure should probably be used to achieve that, which the STL may not provide directly (something like an `std::pair<unique_ptr<int[]>, size_t>` of sorts?)  Using `resize` here though is the appropriate "quick fix" but the underlying issues around the buffer data structures we use internally need to be looked at more deeply since I believe this pattern is in quite a few places (this came up at our all-hands-meeting and we should spend some time off-line as a team on the larger issue).

NAThompson:
@chuckatkins : What are your supported compilers? The easy fix that also elides the expensive OS call to malloc is just to use a stack array. (Do you have enough MPI ranks to cause a stackoverflow? Maybe you do . . . :) ) I did a quick run through the list of compilers on [Godbolt](https://godbolt.org/z/sFpe6M) and the only one I found that didn't support stack arrays was MSVC.

PR #1369, opened 2019-04-10
Fix mpi aggregator state

### LINKED ISSUE #1348: MPIAggregator sending local stack variables via MPI_Isend
See #1326 for more details.

chuckatkins:
This supersedes #1326 and fixes #1348.  The implementation is meant to address several issues:

- It fixes the problem of MPI_Isend using a local stack variable out of scope by using a member variable instead as implemented by @germasch in #1326
- It fixes the problem of state corruption through repeated calls to IExchangeAbsolutePosition before wait by adding a bool check to ensure that only one exchange is active at any given time, thus preventing the situation that causes the state corruption from even occurring.
- It addresses the concern of limiting the exchange to one at a time by moving the implementation into the derived child class, thus allowing alternate aggregations strategies to implement different methods of state tracking and allow simultaneous requests if desired.

This very purposely does not change the API, aggregation algorithm, or any of the MPI calls made.  It only fixes an existing bug and explicitly disallows a pattern that wasn't used anyways by the current implementation.


germasch:
Looks good to me.

PR #1371, opened 2019-04-11
FFS upstream

eisenhauer:
Patches in ffs_conv.c repair a performance bug in FFS with conversion generation turned off.


chuckatkins:
@eisenhauer would you mind doing this from a branch on your fork instead of from the main repo?

eisenhauer:
Ah damn, sorry.   I cloned a clean repo because something weird was happening in the update.sh script, but then I forgot to run the developer script....  Killing and deleting.

PR #1375, opened 2019-04-15
Fix bug in AdvanceStep with timeout on writer failure.

eisenhauer:
Fix bug in AdvanceStep with timeout that caused writer failures not to be recognized.  Add tests for 2x2 case in CommMin and CommPeer.


eisenhauer:
Superceded by #1408 .

PR #1378, opened 2019-04-17
Add a Null2 engine that operates at the core::Engine level

chuckatkins:


germasch:
This looks nice and encapsulated to me ;)

I have one question, just for my education: Why did you make a separate implementation class? This looks pimpl-like to me, but it's not a typical use case for that (at least not to me). Just being curious -- is it some general separate implementation from interface kind of idea?

chuckatkins:
> Why did you make a separate implementation class? This looks pimpl-like to me

Yep, that's exactly what it is.  


> but it's not a typical use case for that (at least not to me). ... is it some general separate implementation from interface kind of idea?

Yes.  The idea is that it keeps the implementation entirely private even with respect to member variables.

williamfgc:
@chuckatkins sounds good to me, should we call it NullWriter? Thanks.

source/adios2/core/IO.cpp
```diff
@@ -576,6 +577,15 @@ Engine &IO::Open(const std::string &name, const Mode mode,
         engine =
             std::make_shared<engine::NullEngine>(*this, name, mode, mpiComm);
     }
+    else if (engineTypeLC == "null2")
```
williamfgc: nullwriter instead of null2?

testing/adios2/engine/SmallTestData.h
```diff
@@ -63,7 +63,7 @@ struct SmallTestData
          std::complex<double>(18.2, 19.2), std::complex<double>(19.2, 20.2)}};
 };
 
-SmallTestData generateNewSmallTestData(SmallTestData in, int step, int rank,
+SmallTestData generateNewSmallTestData(SmallTestData in, size_t step, int rank,
```
williamfgc: I think this is triggering the VS failures. I typically just cast to `int` in the test.

chuckatkins: Yeah, I just didn't have time to look at it last night before I left.  I'll clean it up this morning though.

chuckatkins:
Wrt testing bindings, this is agnostic of the bindings so it shouldn't make any difference.  There's nothing in the bindings that's exercised here that isn't done elsewhere.

testing/adios2/engine/SmallTestData.h
germasch: FWIW, this is the change which is causing the CI failure (MSVC, if I remember).

germasch:
FWIW, I think this would be useful for 2.4.0. The CI failure is easy to fix. I'd agree that "NullWriter" is a nicer name than "Null2".

chuckatkins:
I know, it's an easy fix.  i just got way side tracked.  I'll close the loop on it here shortly and get it merged.

source/adios2/core/IO.cpp
chuckatkins: I've changed it to NullCore

PR #1397, opened 2019-05-07
Use diff instead of cmake -E compare_files for testing

chuckatkins:
@pnorbert @eisenhauer this should help with the CI testing


chuckatkins:
@eisenhauer I believe the validate test failures are due to a hardware change on circleci and how it interacted with the openmpi install.  See https://open.cdash.org/testDetails.php?test=762147510&build=5897119 .  The most recent commit should clear it up.

PR #1436, opened 2019-05-21
[WIP] fixing sst bp compression bug

JasonRuonanWang:


chuckatkins:
@JasonRuonanWang is this problem limited to zfp or is it present on any compressor?  The test code is zfp specific so if the problem is not then the test should be made more generic, but if it is then the tests and source files should at least be named as such.

JasonRuonanWang:
I think this was already fixed in another pull request. Just forgot to close this one.

Issue #1111, opened 2019-01-23
SST: reading when there are zero size blocks


pnorbert:
It seems SST reading does not bring the correct data when some processes write zero size blocks. 
The sick example was seen in EPOCH, where some variables' data is only written from rank 0 but actually all processes define and put() the variable. All non-zero rank writes a count=0 size array. 
This causes no problem at reading with BPFile. But when reading with SST, the data is invalid. 

There is no test or example at the moment to re-create this problem. 
In general, zero-sized blocks are allowed as it regularly happens in particle codes (picongpu, xgc, etc). 

We need tests like this.

JasonRuonanWang:
@pnorbert Did you have this problem when you use FFS or BP?

pnorbert:
I uses default settings. Did not test with both.

eisenhauer:
There is a test for this circumstance now, so I think this is fixed.

PR #1445, opened 2019-05-22
Zero Data Variable Test

eisenhauer:
Add a test that has a global variable which has zero elements contributed by some node.

### DIFF
--- a/testing/adios2/engine/staging-common/CMakeLists.txt
+++ b/testing/adios2/engine/staging-common/CMakeLists.txt
@@ -163,7 +163,9 @@ foreach(test ${SST_TESTS})
     add_common_test(${test} SST)
 endforeach()
 
-set_tests_properties(Staging.2x1ZeroDataVar.CommMin.BP.SST Staging.2x1ZeroDataVar.CommPeer.BP.SST  PROPERTIES WILL_FAIL TRUE)
+if (ADIOS2_HAVE_MPI)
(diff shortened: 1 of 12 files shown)


eisenhauer:
@JasonRuonanWang, this PR adds an SST test in which a single global variable has zero elements contributed by one rank (in this case, all are contributed by another).  The test passes with FFS-based marshaling and fails with BP marshaling.  This is likely what @pnorbert reported in issue #1111 .   I've marked the BP tests as WILL_FAIL in CMake.  Not sure if that will let this PR pass, or if we'll need to fix the bug before it can go in, but maybe this is the test case for committing tests that don't pass...

chuckatkins:
@eisenhauer will this be targeting the release?  If so then please edit the base branch to be `release` and add the `2.4.0` milestone.

eisenhauer:
I'm going to close this and resubmit it reworked after the release.

ax3l:
Oh, so 2.4.0 will not be able to write zero-sized data yet?

williamfgc:
@ax3l this is related to the SST engine which is using the same serializer/deserializer,  if you're talking about files I have a [test](https://github.com/ornladios/ADIOS2/blob/master/testing/adios2/engine/bp/TestBPWriteMultiblockRead.cpp) (without trying compression), look for the Zero keyword in that link. Please use that for your tests and report if there are issues. Thanks!

eisenhauer:
@ax3l OK, zero contribution tests are back in PRs #1622 and  #1624 (simplified version).   They don't do compression, but at present they demonstrate a problem with SST and BP marshaling.  The nature of the problem seems to be that the BP.SST tests are unstable, succeeding or failing based upon a data race created by a bad read request.  Hopefully it can get sorted out soon.

PR #1447, opened 2019-05-23
mpidummy: proper size in snprintf

ax3l:
Fix maximum buffer size and warning
```
  "could not gather data" directive output truncated writing
  22 bytes into a region of size 5 [-Wformat-truncation]
```


eisenhauer:
Thanks Axel!  It looks like this particular bug was copied over from ADIOS1 (where it still exists @pnorbert ...)

ax3l:
Thanks for the hint! Backported in https://github.com/ornladios/ADIOS/pull/198

PR #1450, opened 2019-05-24
append: incredibly ugly but very simple append works (serial, no attrs)

germasch:
This PR shows that appending to a BP3 file isn't all that hard to implement. This is by no means done, it's buggy, limited, and most of all extremely hacky. But it works for one simple test case: Writing 1 variable to a file, then opening that file again with `Mode::Append` and writing a second variable.

Doing this "right" will be quite a bit more work, though not really hard.


germasch:
I will close this for now, as it wasn't meant for merging, anyway. I may go back and do it properly at some point.

PR #1452, opened 2019-05-24
Move ADIOS2 completely to fixed width types

williamfgc:
`adios2::GetType<T>` and `bpls` information now provide fix width types only. Effectively making `bpls` output portable across systems.

```
./bpls ADIOS2BPWriteReadAsStream1D8.bp -lav
File info:
  of variables:  13
  of attributes: 0
  statistics:    Min / Max 

  float complex   cr32     5*{32} = (1.1,i2.1) / (27.1,i28.1)
  double complex  cr64     5*{32} = (11.2,i12.2) / (37.2,i38.2)
  int16_t         i16      3*{32} = -509 / 539
  int32_t         i32      3*{32} = -131069 / 131099
  int64_t         i64      3*{32} = -8589934589 / 8589934619
  int8_t          i8       3*{32} = -5 / 27
  string          iString  scalar = "Testing ADIOS2 String type"
  float           r32      4*{32} = 1.1 / 23.1
  double          r64      4*{32} = 11.2 / 33.2
  uint16_t        u16      2*{32} = 32773 / 32791
  uint32_t        u32      2*{32} = 2147483653 / 2147483671
  uint64_t        u64      2*{32} = 9223372036854775813 / 9223372036854775831
  uint8_t         u8       2*{32} = 133 / 151
```


germasch:
I like this, except that I think there maybe some better choices for some of the types, in particular for the floating point types, ie, "f32"/"f64", and "c32"/"c64", considering this is not a Fortran library. Actually, what I think would be even better to add a couple of characters to make it "float32", "uint64", etc, as that has the least amount of confusion.

And I guess I'm not sure about the "iString"? Is that an Apple thing? ;)

I remember there being some remaining uses of the hardcoded strings that might break now, but maybe you already found them all (in ffs.c).

I think you'll find a lot more breakage in the CI where you'll need to adapt the bpls reference output, which indicates that this change isn't really backward compatible (but I wouldn't complain about it compared to the benefit of portable type names).

germasch:
Forget about my comment on the naming above, I confused the variable name and type name in the bpls output :(

williamfgc:
Thanks @germasch this is possible due to your contribution to move internally to fix width types.

PR #1453, opened 2019-05-28
Fix the creation of fresh attribute indices in DefineAttribute

franzpoeschel:
The array overload of `IO::DefineAttribute( ... )` currently gives out new internal indices for attributes based on the size of `IO::attributeMap`. This results in ambiguous indices as soon as attributes are created after the deletion of other attributes.

Fix this by rewriting the function template according first overload that takes a `const T &value` argument.


williamfgc:
@franzpoeschel thanks! I'll merge the fix (I'd assume this is needed now) and provide a test in a new PR.

williamfgc:
Related to #1193

Issue #1443, opened 2019-05-22
Avoid MPI_Init for Serial I/O when built with MPI


ax3l:
Hi,

we discussed this offline before and we are currently at the point where we would need the feature again. When compiling ADIOS2 with MPI features, it would be wonderful if the serial constructors would still be exposed.

https://github.com/ornladios/ADIOS2/blob/02f246eb328a524eefd00286ce88ed1c0f2172f9/bindings/CXX11/cxx11/ADIOS.h#L67-L80

This would allow users to use a cluster module with MPI support and build small serial tools against it, e.g. for meta-data reads, without the need to call those with `mpirun` (and without `MPI_Init`, ideally).

Is it possible to expose the MPI functionality as a true superset of the serial functionality? Also from a package management point of view, this would make a lot of sense for compatibility.

Regarding the current implementation in [`source/adios2/ADIOSMPI.h`](https://github.com/ornladios/ADIOS2/blob/7b3939beec73029f0fe7e62e54f29e0832e697fa/source/adios2/ADIOSMPI.h), I am a bit afraid we run into usability issues (also seen in practical examples with ADIOS1: https://github.com/ornladios/ADIOS/issues/183). Instead of mocking the MPI APIs, could we potentially just use those as a wrapper and an internal state (MPI-parallel or serial) to switch the implementation at runtime?

williamfgc:
@ax3l a couple of things:

1) in ADIOS2 the way we currently write portable code is using the macro `#ifdef ADIOS2_HAVE_MPI` 

```
#ifdef ADIOS2_HAVE_MPI
    adios2::ADIOS adios(comm);
#else 
    adios2::ADIOS adios;
#endif
```

2) source/adios2/ADIOSMPI.h is completely private, never exposed to the application. In ADIOS1.x mpidummy.h was exposed while in ADIOS2 is private and inside the `namespace adios2`

I don't know how a truly non-conflicting single MPI and Non-MPI library can be deployed. Ideas are welcome or, even better, we can borrow from an existing implementation doing that. Thanks!

williamfgc:
@ax3l a couple of things:

1) in ADIOS2 the way we currently write portable code is using the macro `#ifdef ADIOS2_HAVE_MPI` 

```
#ifdef ADIOS2_HAVE_MPI
    adios2::ADIOS adios(comm);
#else 
    adios2::ADIOS adios;
#endif
```

[...]

PR #1466, opened 2019-05-29
MPICommOnly / mpidummy cleanup; no-MPI_Init support

germasch:
This PR gets rid off `ADIOSMPICommOnly.h`. This probably dates back to the pre-pimpl cxx11 bindings, but isn't useful anymore. Currently, this header gets installed, but only included in the public (Installed) `adios2_c_operator.h`, which doesn't need `MPI_Comm`.

This PR also does some cleanup on mpidummy, including fixing a couple of minor bugs and getting rid of `MPI_Error_string`, which is never used and also doesn't do what the standard says it should be doing.

**ETA:** This PR now also includes a new level of MPI wrapping in `mpiwrap.{h,cpp}`. The functions `SMPI_Comm_rank()` etc behave like the regular MPI counterpart, but they fall back to mpidummy.h in the case where MPI is not initialized, in order to make ADIOS2 more easily usable from serial code even if ADIOS2 itself has been built with MPI.


pnorbert:
Kai, I am working on unifying the parallel and serial builds into one
library. It is much more work than I expected so it will come a bit later.

pnorbert:
Please keep this branch around, I will need to port your fixes into the new
DummyMPI class

germasch:
> Kai, I am working on unifying the parallel and serial builds into one library. It is much more work than I expected so it will come a bit later.

Actually, I have this mostly finished, just needs some cleaning up -- this PR are some preliminary steps. So depending on how much work you've already put into it, I could polish my version and make a PR. I didn't go the class/virtual function route, rather I just explicitly dispatch to the real MPI versus a dummy implementation depending on whether the real MPI has been MPI_Init'd. It could be changed into using derived classes, but I think it's even more code that way. Basically I'm just augmenting the real MPI library to handle one-proc functionality when it's not initialized. I'm also using mpidummy for that, but note that it needs to be compiled with types etc from the real mpi.h in that case, as you've probably found out, too ;)

pnorbert:
Can you share your branch to look at? 

Mine is complicated for sure. Had to change almost every class to use AMPI_Comm instead of MPI_Comm. [https://github.com/pnorbert/ADIOS2/tree/AMPI](https://github.com/pnorbert/ADIOS2/tree/AMPI)

pnorbert:
Does your version work in a serial build when MPI is not available?

germasch:
Yes, it does, in that case it just uses mpidummy. In a parallel build, it uses mpi.h. The MPI functions in that case get dispatched to the regular MPI lib if it's initialized, and to the mpidummy implementation otherwise (so mpidummy provides just the implementation in this case, but not the types / constants, which still come from mpi.h)

The changes throughout the code are replacing `MPI_Comm_rank` with `mpi::MPI_Comm_rank` etc. This doesn't need to be done throughout, necessarily, but for the pieces which are supposed to work when adios2 is compiled with an actual MPI lib, but runs without MPI_Init.

germasch:
I pushed the changes for the MPI+serial support in here, though it's not in shape to be merged. The last two commits are the actual work.

They provide `adios2::mpi::{MPI_Comm_rank,MPI_Comm_size, ...}`. These versions can be called (on 1-proc communicators) even if MPI has not been initialized. In the adios2 w/o MPI case, these functions are just the usual mpidummy versions. If compiled with MPI, these functions wrap the actual MPI library calls, but will fall back to the mpiummy versions if the MPI library hasn't been initialized.

The last commit switches a couple of places to use the `mpi::MPI_*` wrappers instead of the original MPI functions, just enough to make `bpls` work without MPI_Init/MPI_Finalize.

In some other branch, I have some more changes that provide the serial ADIOS constructors even in the compiled-with-MPI case, add some test, etc.

I will say that I'm not entirely happy with this, but it's the least invasive approach I can come up with. In particular, I would have liked to do this in a way that doesn't require changes throughout the code. I don't think that's possible without resorting to other hacks, though. I was hoping I might be able to put the MPI function wrappers into the `adios2::` namespace, so they would get picked up automatically. However, that's not possible because `mpi.h` defines the original functions `extern "C"`, which means they don't honor namespaces and ambiguity ensues.

I have some other ideas, but don't really like those, either.
- `#define MPI_Comm_rank mpi::MPI_Comm_rank` etc in `ADIOSMPI.h`. That would make the change appear to be transparent, though it's kinda fake.
- Play some tricks with dynamic loading of a mpidummy shared library instead of the actual MPI lib when indicated. Sounds rather painful to do, in particular considering that you're supporting a wide variety of systems which I'm sure have their quirks.
- Use the MPI profiling interface to intercept the MPI calls. That's actually quite easy to do, and should be portable if all implementations follow the standard. It's still kind of an abuse of that ability.

Of course one can do a more complete wrapping of MPI, including using virtual base classes, wrapping the data types, etc. I kinda feel that's not the best direction to go, everyone is familiar with MPI and having to learn a different way in adios2 is cumbersome, and 3rd party dependencies won't be covered either way.

Personally, I feel it'd be better to work the other way, ie, use something like this approach to make the adios2 w/o MPI_Init work for what's commonly needed now. For the future, I think it'd be good to avoid needing to use any of those wrappers in the first place. E.g., `Transport` uses `MPI_Comm_rank/size`, but `FilePOSIX` really doesn't care about that.  If there is a transport that actually uses MPI (MPI-I/O?), that could call `MPI_Comm_rank/size` itself, because the presumably requires an actual MPI, or at least I'd say that'd be a reasonable constraint to create. But `FilePOSIX` just shouldn't make any MPI calls, so there's nothing to wrap there.

pnorbert:
Note that this branch in this form does not even compile on Titan. CI VMs do the build fine but have failing tests. 

```
[ 32%] Building CXX object source/adios2/CMakeFiles/adios2.dir/core/Attribute.cpp.o
In file included from /ccs/proj/e2e/pnorbert/ADIOS/ADIOS2/source/adios2/helper/mpiwrap.h:12:0,
                 from /ccs/proj/e2e/pnorbert/ADIOS/ADIOS2/source/adios2/ADIOSMPI.h:13,
                 from /ccs/proj/e2e/pnorbert/ADIOS/ADIOS2/source/adios2/helper/adiosMPIFunctions.h:19,
                 from /ccs/proj/e2e/pnorbert/ADIOS/ADIOS2/source/adios2/helper/adiosFunctions.h:15,
                 from /ccs/proj/e2e/pnorbert/ADIOS/ADIOS2/source/adios2/core/Attribute.cpp:14:
/ccs/proj/e2e/pnorbert/ADIOS/ADIOS2/source/adios2/helper/mpidummy.h:125:16: error: 'MPI_Comm' is not a type
                MPI_Comm comm);
                ^~~~~~~~
/ccs/proj/e2e/pnorbert/ADIOS/ADIOS2/source/adios2/helper/mpidummy.h:128:50: error: 'MPI_Comm' is not a type
                 MPI_Datatype recvtype, int root, MPI_Comm comm);
                                                  ^~~~~~~~
/ccs/proj/e2e/pnorbert/ADIOS/ADIOS2/source/adios2/helper/mpidummy.h:131:19: error: 'MPI_Comm' is not a type
                   MPI_Comm comm);
                   ^~~~~~~~
/ccs/proj/e2e/pnorbert/ADIOS/ADIOS2/source/adios2/helper/mpidummy.h:135:17: error: 'MPI_Comm' is not a type
                 MPI_Comm comm);
                 ^~~~~~~~
/ccs/proj/e2e/pnorbert/ADIOS/ADIOS2/source/adios2/helper/mpidummy.h:138:51: error: 'MPI_Comm' is not a type
                  MPI_Datatype recvtype, int root, MPI_Comm comm);
                                                   ^~~~~~~~
/ccs/proj/e2e/pnorbert/ADIOS/ADIOS2/source/adios2/helper/mpidummy.h:141:14: error: 'MPI_Comm' is not a type
              MPI_Comm comm, MPI_Status *status);
              ^~~~~~~~
/ccs/proj/e2e/pnorbert/ADIOS/ADIOS2/source/adios2/helper/mpidummy.h:143:15: error: 'MPI_Comm' is not a type
               MPI_Comm comm, MPI_Request *request);
               ^~~~~~~~
/ccs/proj/e2e/pnorbert/ADIOS/ADIOS2/source/adios2/helper/mpidummy.h:145:23: error: 'MPI_Comm' is not a type
              int tag, MPI_Comm comm);
                       ^~~~~~~~
/ccs/proj/e2e/pnorbert/ADIOS/ADIOS2/source/adios2/helper/mpidummy.h:147:24: error: 'MPI_Comm' is not a type
               int tag, MPI_Comm comm, MPI_Request *request);
                        ^~~~~~~~
/ccs/proj/e2e/pnorbert/ADIOS/ADIOS2/source/adios2/helper/mpidummy.h:153:28: error: expected ')' before 'comm'
 int MPI_File_open(MPI_Comm comm, const char *filename, int amode, MPI_Info info,
                            ^~~~
/ccs/proj/e2e/pnorbert/ADIOS/ADIOS2/source/adios2/helper/mpidummy.h:153:34: error: expected primary-expression before 'const'
 int MPI_File_open(MPI_Comm comm, const char *filename, int amode, MPI_Info info,
                                  ^~~~~
/ccs/proj/e2e/pnorbert/ADIOS/ADIOS2/source/adios2/helper/mpidummy.h:153:56: error: expected primary-expression before 'int'
 int MPI_File_open(MPI_Comm comm, const char *filename, int amode, MPI_Info info,
                                                        ^~~
/ccs/proj/e2e/pnorbert/ADIOS/ADIOS2/source/adios2/helper/mpidummy.h:153:76: error: expected primary-expression before 'info'
 int MPI_File_open(MPI_Comm comm, const char *filename, int amode, MPI_Info info,
                                                                            ^~~~
/ccs/proj/e2e/pnorbert/ADIOS/ADIOS2/source/adios2/helper/mpidummy.h:154:28: error: expected primary-expression before '*' token
                   MPI_File *fh);
                            ^
/ccs/proj/e2e/pnorbert/ADIOS/ADIOS2/source/adios2/helper/mpidummy.h:154:29: error: 'fh' was not declared in this scope
                   MPI_File *fh);
                             ^~
/ccs/proj/e2e/pnorbert/ADIOS/ADIOS2/source/adios2/helper/mpidummy.h:154:31: error: expression list treated as compound expression in initializer [-fpermissive]
                   MPI_File *fh);
                               ^
/ccs/proj/e2e/pnorbert/ADIOS/ADIOS2/source/adios2/helper/mpidummy.h:162:29: error: expected ')' before 'comm'
 int MPI_Comm_split(MPI_Comm comm, int color, int key, MPI_Comm *comm_out);
                             ^~~~
/ccs/proj/e2e/pnorbert/ADIOS/ADIOS2/source/adios2/helper/mpidummy.h:162:35: error: expected primary-expression before 'int'
 int MPI_Comm_split(MPI_Comm comm, int color, int key, MPI_Comm *comm_out);
                                   ^~~
/ccs/proj/e2e/pnorbert/ADIOS/ADIOS2/source/adios2/helper/mpidummy.h:162:46: error: expected primary-expression before 'int'
 int MPI_Comm_split(MPI_Comm comm, int color, int key, MPI_Comm *comm_out);
                                              ^~~
/ccs/proj/e2e/pnorbert/ADIOS/ADIOS2/source/adios2/helper/mpidummy.h:162:65: error: 'comm_out' was not declared in this scope
 int MPI_Comm_split(MPI_Comm comm, int color, int key, MPI_Comm *comm_out);
                                                                 ^~~~~~~~
/ccs/proj/e2e/pnorbert/ADIOS/ADIOS2/source/adios2/helper/mpidummy.h:162:73: error: expression list treated as compound expression in initializer [-fpermissive]
 int MPI_Comm_split(MPI_Comm comm, int color, int key, MPI_Comm *comm_out);
                                                                         ^
/ccs/proj/e2e/pnorbert/ADIOS/ADIOS2/source/adios2/helper/mpidummy.h:169:60: error: 'MPI_Comm' is not a type
                MPI_Datatype datatype, MPI_Op op, int root, MPI_Comm comm);
                                                            ^~~~~~~~
/ccs/proj/e2e/pnorbert/ADIOS/ADIOS2/source/adios2/helper/mpidummy.h:172:53: error: 'MPI_Comm' is not a type
                   MPI_Datatype datatype, MPI_Op op, MPI_Comm comm);
```

germasch:
I put this up here (as [WIP]) because you asked to take a look at it, not because it's ready (it's clearly not :(). I'll take a look at what's going on on Titan.

germasch:
Well, let's see how the CI goes this time...

I added a fix for the Titan build issue, but I couldn't say I understand why that's needed in the first place. I also rebased and fixed some other bugs. In particular, this change that I made to the dummy `MPI_Gather` broke many things:
```diff
@@ -109,7 +109,7 @@ int MPI_Gather(const void *sendbuf, int sendcnt, MPI_Datatype sendtype,
     size_t nsent = 0, nrecv = 0;
     if (!sendbuf && !recvbuf)
     {
-        return ier;
+        return MPI_ERR_BUFFER;
     }
     if (comm == MPI_COMM_NULL || root)
     {
```
The old behavior returned 0 (success) when either buffer was NULL. I changed the behavior to be the same as what `MPI_Scatter` does, but apparently there is code that relies on passing NULL buffers to MPI_Gather together with 0 counts. So in this latest version, I accept a NULL buffer as long as the corresponding count is 0.

germasch:
There were a couple more failing tests which I fixed with the latest commit (I think -- we'll see).

The failing tests were calling `bpls` on HDF5 files. Since bpls now doesn't call `MPI_Init` (because it's my guinea pig for testing serial use of a parallel adios2 lib), that failed because parallel HDF5 internally `MPI_Comm_dup`s the communicator passed to it. The fix is to not use HDF5 in parallel mode when the communicator encompasses only a single process.

This kinda goes to demonstrate what I was talking about earlier: I think, in the end a cleaner way to solve this problem is to avoid using MPI when it's not needed in the first place, that is, when running on a single proc. This could also be helpful for performance -- for example, `MPI_Bcast`, `MPI_Gather` etc on a single proc end up doing a copy, that could be avoided in the serial case in the first place. The cost, of course, is to have two separate code paths to handle serial vs actually parallel cases. I think it's not a lot of code that would change, though. Wrapping MPI in the way we currently do (both @pnorbert's and my approach do that) is a bit of a band-aid, which can be made to work within the core of adios2, but this only goes so far, external libraries can't be handled this way, and within adios2 it means touching a lot of code and replacing a familiar MPI interface by something new.

There could be some value in wrapping MPI in a more C++-ish fashion, OTOH, the MPI standard tried that a long time ago and eventually through in the towel and deprecated its C++ bindings. Of course, I am biased, but in my approach, the use of MPI is just the same as with real MPI, with the exception of having to put the `mpi::` namespace prefix, so at least it's still the familiar interface, and overhead is small. A potential drawback with what I've done is that I've only updated selected MPI calls, those that are necessary for `bpls` to work. In order to make adios2 work without `MPI_Init` more generally, more calls will need that prefix, and it's not entirely trivial to find all the necessary places. One could, of course, just do a global search and replace, though as I said above, I'd actually be in favor of going the opposite way, that is, avoiding wrapped MPI calls wherever possible instead and aim for getting rid of the band-aid eventually.

pnorbert:
I am aware of that. I could build it on my VM and run gray-scott example
and serial-bpls without any issue, and could also build the serial only
adios library. I just went to Titan to check how it goes and it breaks
before (build) it breaks on the VM (tests).

Your approach is simpler then mine, so if you can complete this, we should
go with your approach.

williamfgc:
@pnorbert @germasch since you're already handling the C++ side, please keep in mind how this would extent to the internal C parts. C doesn't have a notion of `namespace`, not sure if this is required (would any "serial" tool use SST?). The only solution I can think of is the use function pointer in C to map to virtual functions.

germasch:
Good point about handling C. It looks like the only thing affected is SST. I'm not sure how important SST w/o `MPI_Init` is, it'd be nice if someone (@eisenhauer @philip-davis ) could comment on that. A potential use case that I could see is some python analysis script that runs alongside a simulation, connected via SST. In that case, being able to run python without having to deal with MPI would certainly be nice.

I think with my approach, it's not too hard to handle C, I just need to make my wrappers accessible outside of the namespace, using `extern "C"`.  So there could be a `SMPI_Comm_rank` doing the same as `mpi::MPI_Comm_rank`, etc. Though in that case, maybe I should just always use `SMPI_*`, even in C++?

pnorbert:
SST without MPI is the primary case for in situ viz with VisIt. Visit
currently uses both adios installation to build.

germasch:
> SST without MPI is the primary case for in situ viz with VisIt. Visit currently uses both adios installation to build.

So that sounds like a clear reason to include SST support. SST includes its own `sstmpidummy.h`, presumably because it can't use the existing one because it's C++-only. I'm kinda preparing things to support both C and C++ in this PR, but it'll require more changes to `mpidummy.h` to make that includable from C code. My preference would be to move that work to a future PR in order to keep thing here manageable.

germasch:
So this is still in WIP status, but getting close. I'll do a bit of commit squashing, and try to do a bit more meaningful testing. 

I'd be happy to split the PR back in two -- the MPI dummy cleanup that was originally in this PR, and then I could do the new MPI wrappers separately.

For the new MPI wrappers, what I do is to add a bunch of `SMPI_Comm_rank` etc wrappers, and then use those instead of the existing `MPI_Comm_rank` etc in select places.
I think that'll be a good time to merge this work, as it is easy to see that this doesn't change any existing behavior:

- In the no-MPI build, the SMPI wrappers just call the corresponding mpidummy functions, so nothing changes.
- In the with-MPI build, the MPI wrappers call through to the real MPI lib if `MPI_Init` has been called, and the mpidummy version otherwise. Right now, when adios2 is compiled with MPI, `MPI_Init` always has to be called before using adios2, in which case the wrappers forward the call to the MPI library, so nothing changes for existing code.

What's new is that even in the with-MPI build, certain functionality is now available without calling MPI_Init. For example, it is now possible to not call MPI_Init in bpls and have it still work (I've tested that, but it's not included in this PR). I've also made the serial cxx11 ADIOS constructors available even when built with MPI, and that passes some simple testing, but isn't otherwise used.

The next steps will then be to incrementally convert things like bpls, the python interface, etc to take advantage of this.

germasch:
@pnorbert @williamfgc @ax3l: I did the remaining cleanup, so this is ready for review/merging from my side. It doesn't actually change any existing behavior, ie., all the tests / utils are still calling `MPI_Init` as of now (except the one new test I added), in which case the SMPI_* wrappers just call through the real MPI.

It does, however, expose the serial `adios2::ADIOS` constructors always, and BP3/BP4/HDF5 access works w/o needing to call `MPI_init`. There may be more functionality that should be made available w/o initializing MPI, for the most part that should be as easy as changing `MPI_*` to `SMPI_*`.

It passes the CI, and I checked that it builds on summit and titan. I organized the PR  in manageably sized commits which are best looked at one by one. I even added some descriptions to the commit messages ;)

pnorbert:
Please remove the MPI stuff from bpls. It is a sequential tool and the main point of this refactoring is that bpls can run on the login node (of Titan e.g.) where there is no MPI.

pnorbert:
Make sure you remove the lines to call the ADIOS constructor with a communicator in bpls as well.

germasch:
I added the commit that removes MPI_Init/Finalize, that's enough to make this work on Titan in theory (in practice, however, I get an illegal instruction error on Titan, which I don't think is related, but hold on...).

The other use of MPI in bpls is a bit of a bigger change, since bpls currently uses the internal (core::) interface to ADIOS2, not the actual CXX11 bindings. But that shouldn't affect things working (or not) w/o MPI_Init.

pnorbert:
`module swap craype-interlagos craype-istanbul`

will compile a code that can run on the older processors in the login node
as well as on the compute nodes.

Even if it works, this code does not belong to a completely serial code:
```
#ifdef ADIOS2_HAVE_MPI
    core::ADIOS adios(MPI_COMM_SELF, true, "C++");
#else
    core::ADIOS adios(true, "C++");
#endif
```
The idea was that it just simply works with core::ADIOS adios(true, "C++")
with any version of the library.

germasch:
> `module swap craype-interlagos craype-istanbul` will compile a code that can run on the older processors in the login node as well as on the compute nodes.

Thanks, that fixed it.

> Even if it works, this code does not belong to a completely serial code: 

```cxx
#ifdef ADIOS2_HAVE_MPI 
    core::ADIOS adios(MPI_COMM_SELF, true, "C++"); 
#else
   core::ADIOS adios(true, "C++"); 
#endif
```
> The idea was that it just simply works with core::ADIOS adios(true, "C++") with any version of the library.

I agree this doesn't belong into bpls, and I fixed it (I updated the commit to use the serial ctor always and force-pushed). In the bigger picture, though, I think the internals (`core::`) actually should be written to always use communicators, ie., you don't want to have different constructors for `core::IO`, `core::Engine`, etc that deal with the MPI/no-MPI case. I don't think that `bpls` should be using the internal interface in the first place, though. Anyway, not an issue of importance right now.

`bpls` now works on the head node of Titan. There is still a remaining problem in this PR that I just realized: The TestAdiosNoMPI test gets run with `mpiexec -n 4`, and since every instance writes/reads the same `test.bp` file, there are race conditions. If anyone knows how to run that one test without EXEC_WRAPPER, let me know ;)

germasch:
> There is still a remaining problem in this PR that I just realized: The TestAdiosNoMPI test gets run with `mpiexec -n 4`, and since every instance writes/reads the same `test.bp` file, there are race conditions. If anyone knows how to run that one test without EXEC_WRAPPER, let me know ;)

That one was actually kinda trivial to fix, so disregard...

pnorbert:
Next problem: the python binding is still using MPI. All bindings should use the non-mpi constructor when running serial. 

```
...
Writing /ccs/proj/e2e/pnorbert/ADIOS/ADIOS2/build.titan.gnu/install/bin/adios2-config
/ccs/proj/e2e/pnorbert/ADIOS/ADIOS2/build.titan.gnu

pnorbert@titan-ext4:/ccs/proj/e2e/pnorbert/ADIOS/ADIOS2/build.titan.gnu $ export PYTHONPATH=$PYTHONPATH:/ccs/proj/e2e/pnorbert/ADIOS/ADIOS2/build.titan.gnu/install/lib64/python2.7/site-packages
pnorbert@titan-ext4:/ccs/proj/e2e/pnorbert/ADIOS/ADIOS2/build.titan.gnu $ ldmunge /ccs/proj/e2e/pnorbert/ADIOS/ADIOS2/build.titan.gnu/install/lib64

pnorbert@titan-ext4:/ccs/proj/e2e/pnorbert/ADIOS/ADIOS2/build.titan.gnu $ python
Python 2.7.9 (default, Apr 25 2018, 09:46:03) 
[GCC 4.3.4 [gcc-4_3-branch revision 152973]] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> import numpy
>>> import adios2
[Mon Jun  3 13:43:35 2019] [unknown] Fatal error in PMPI_Init_thread: Other MPI error, error stack:
MPIR_Init_thread(537): 
MPID_Init(249).......: channel initialization failed
MPID_Init(638).......:  PMI2 init failed: 1 
Aborted (core dumped)

```

pnorbert:
Also, please don't forget in the parallel build to support both constructors. Rright now this is not the case. On a compute node of Titan this is what I get:

```
pnorbert@titan-login8:/autofs/nccs-svm1_proj/e2e/pnorbert/ADIOS/ADIOS2/build.titan.gnu $ aprun -n 1 python ../tst.py 
Traceback (most recent call last):
  File "../tst.py", line 4, in <module>
    fr = adios2.open("/ccs/home/pnorbert/prism_schema.bp", "r")
TypeError: open(): incompatible function arguments. The following argument types are supported:
    1. (name: unicode, mode: unicode, comm: MPI4PY_Comm, engine_type: unicode=u'BPFile') -> adios2::py11::File
    2. (name: unicode, mode: unicode, comm: MPI4PY_Comm, config_file: unicode, io_in_config_file: unicode) -> adios2::py11::File
```

pnorbert:
You can use my script and file to test your builds on Titan. You need to copy them because you can't list them:
```
/ccs/home/pnorbert/tst.py
/ccs/home/pnorbert/prism_schema.bp
```

This is the output with a serial shared build of ADIOS, on the login node:
```
pnorbert@titan-ext4:/ccs/proj/e2e/pnorbert/ADIOS/ADIOS2/build.titan.gnu-nompi-shared $ python ../tst.py 
File has 0 steps
variable_name: cells
	Type: int32_t
	Min: 0
	Max: 8
	AvailableStepsCount: 1
	Shape: 6, 2
	SingleValue: false


variable_name: points
	Type: double
	Min: 0
	Max: 2
	AvailableStepsCount: 1
	Shape: 9, 3
	SingleValue: false


variable_name: pVar
	Type: double
	Min: 0
	Max: 8
	AvailableStepsCount: 1
	Shape: 9
	SingleValue: false


Variable pVar has 9 elements. Read them now...
pVar = [  0.0  1.0  2.0  3.0  4.0  5.0  6.0  7.0  8.0  ]
```

germasch:
Right, I didn't do anything to the python bindings. They should be converted, too, as I've now done with bpls.

I've been trying to keep the scope limited. This PR is meant to address #1443, which asked for making the serial ADIOS2 constructors available in an MPI build. It does that, and it also makes them actually usable for a subset of ADIOS2 functionality. The original purpose wasn't to make the parallel bpls work on Cray head nodes, or to make the python bindings work without MPI, though those are now things that are possible.

Converting the Python bindings should be straightforward as such, but there are some things to consider, in particular the fact that the Python bindings expose all of ADIOS2, and the existence of a serial constructor may lead users to think that the resulting object is fully usable without MPI, but it is not at this point. In particular, SST has its own separate mpidummy, so the changes in this PR don't affect it, ie., it still won't work without MPI_Init. I'm planning to look into that in the future, but I'm an advocate of an incremental approach. This PR already does touch quite a bit of code (and kinda critical code at that), and achieves a certain set of objectives (cleanup, SMPI wrappers, serial ADIOS ctor, serial bpls). The more gets thrown in with it, the bigger the chance that something will go wrong.

pnorbert:
Okay, fair enough, the python bindings can be added in another PR. I am
just throwing at you more of the work that addresses the original purpose
of this entire refactoring: support Axel's dependence on a single adios
library including python. So this PR can be merged now, but his ticket
should not be considered resolved.

germasch:
You can try this patch (I don't have things set up on Titan to try it). It's entirely untested, but I'd guess it'll likely work, with the constraints mentioned above (ie, only part of ADIOS2 will actually work without MPI_Init, which isn't really a python binding issue, though).

```diff
diff --git a/bindings/Python/py11ADIOS.cpp b/bindings/Python/py11ADIOS.cpp
index 18da959e..afa6147e 100644
--- a/bindings/Python/py11ADIOS.cpp
+++ b/bindings/Python/py11ADIOS.cpp
@@ -29,7 +29,8 @@ ADIOS::ADIOS(MPI4PY_Comm mpiComm, const bool debugMode)
 : ADIOS("", mpiComm, debugMode)
 {
 }
-#else
+#endif
+
 ADIOS::ADIOS(const std::string &configFile, const bool debugMode)
 : m_ADIOS(
       std::make_shared<adios2::core::ADIOS>(configFile, debugMode, "Python"))
@@ -37,7 +38,6 @@ ADIOS::ADIOS(const std::string &configFile, const bool debugMode)
 }

 ADIOS::ADIOS(const bool debugMode) : ADIOS("", debugMode) {}
-#endif

 ADIOS::operator bool() const noexcept { return m_ADIOS ? true : false; }

diff --git a/bindings/Python/py11ADIOS.h b/bindings/Python/py11ADIOS.h
index 0279fe67..9a814baa 100644
--- a/bindings/Python/py11ADIOS.h
+++ b/bindings/Python/py11ADIOS.h
@@ -32,10 +32,9 @@ public:
     ADIOS(const std::string &configFile, MPI4PY_Comm comm,
           const bool debugMode = true);
     ADIOS(MPI4PY_Comm comm, const bool debugMode = true);
-#else
+#endif
     ADIOS(const std::string &configFile, const bool debugMode = true);
     ADIOS(const bool debugMode);
-#endif
     ~ADIOS() = default;

     /** object inspection true: valid object, false: invalid object */
```

germasch:
Thanks for merging it. I looked at your `tst.py`, and I guess it uses the high-level API, that I'm not familiar with -- my patch above for the low-level API won't do anything about it. `tst.py` works on my laptop as-is, though, which actually makes no sense to me. With adios2+mpi, is it supposed to complain about a `adios2.open` signature mismatch, or did it complain about MPI_Init not called?

pnorbert:
The python binding also needs changes in the glue code to allow for those functions. However, doing the same trick as in your diff, leads to compiler error. It seems to me the python binding does not allow for overloading function names. Should we rethink the names of the Open() calls in the python binding?

```
#ifdef ADIOS2_HAVE_MPI

adios2::py11::File Open(const std::string &name, const std::string mode,
                        adios2::py11::MPI4PY_Comm comm,
                        const std::string enginetype)
{
    return adios2::py11::File(name, mode, comm, enginetype);
}

adios2::py11::File OpenConfig(const std::string &name, const std::string mode,
                              adios2::py11::MPI4PY_Comm comm,
                              const std::string &configfile,
                              const std::string ioinconfigfile)
{
    return adios2::py11::File(name, mode, comm, configfile, ioinconfigfile);
}

#endif
adios2::py11::File Open(const std::string &name, const std::string mode,
                        const std::string enginetype)
{
    return adios2::py11::File(name, mode, enginetype);
}

adios2::py11::File OpenConfig(const std::string &name, const std::string mode,
                              const std::string configfile,
                              const std::string ioinconfigfile)
{
    return adios2::py11::File(name, mode, configfile, ioinconfigfile);
}
```

```
$ make
...
[ 62%] Building CXX object bindings/Python/CMakeFiles/adios2py.dir/py11glue.cpp.o
/ccs/proj/e2e/pnorbert/ADIOS/ADIOS2/bindings/Python/py11glue.cpp: In function 'void pybind11_init_adios2(pybind11::module&)':
/ccs/proj/e2e/pnorbert/ADIOS/ADIOS2/bindings/Python/py11glue.cpp:154:50: error: no matching function for call to 'pybind11::module::def(const char [5], <unresolved overloaded function type>, const char [33], pybind11::arg, pybind11::arg, pybind11::arg_v)'
           pybind11::arg("engine_type") = "BPFile");
                                                  ^
In file included from /ccs/proj/e2e/pnorbert/ADIOS/ADIOS2/bindings/Python/py11glue.cpp:11:0:
/ccs/proj/e2e/pnorbert/ADIOS/ADIOS2/thirdparty/pybind11/pybind11/include/pybind11/pybind11.h:812:13: note: candidate: template<class Func, class ... Extra> pybind11::module& pybind11::module::def(const char*, Func&&, const Extra& ...)
     module &def(const char *name_, Func &&f, const Extra& ... extra) {
             ^~~
/ccs/proj/e2e/pnorbert/ADIOS/ADIOS2/thirdparty/pybind11/pybind11/include/pybind11/pybind11.h:812:13: note:   template argument deduction/substitution failed:
/ccs/proj/e2e/pnorbert/ADIOS/ADIOS2/bindings/Python/py11glue.cpp:154:50: note:   couldn't deduce template parameter 'Func'
           pybind11::arg("engine_type") = "BPFile");
```

williamfgc:
@pnorbert those C++ `Open` functions without a comm can be renamed `OpenNoComm` in the glue file. They are not exposed in the Python module, thr user only sees as a single `open` with different signatures and can specify arguments explicitly. That's one of the strengths of Python APIs.

germasch:
Renaming is definitely the easiest way. Or you can do what's done elsewhere in the py11 glue, that is, select the correct overload by a cast:
```diff
diff --git a/bindings/Python/py11ADIOS.h b/bindings/Python/py11ADIOS.h
index 0279fe67..9a814baa 100644
--- a/bindings/Python/py11ADIOS.h
+++ b/bindings/Python/py11ADIOS.h
@@ -32,10 +32,9 @@ public:
     ADIOS(const std::string &configFile, MPI4PY_Comm comm,
           const bool debugMode = true);
     ADIOS(MPI4PY_Comm comm, const bool debugMode = true);
-#else
+#endif
     ADIOS(const std::string &configFile, const bool debugMode = true);
     ADIOS(const bool debugMode);
-#endif
     ~ADIOS() = default;
 
     /** object inspection true: valid object, false: invalid object */
diff --git a/bindings/Python/py11glue.cpp b/bindings/Python/py11glue.cpp
index 9d16e339..89166002 100644
--- a/bindings/Python/py11glue.cpp
+++ b/bindings/Python/py11glue.cpp
@@ -86,7 +86,7 @@ adios2::py11::File OpenConfig(const std::string &name, const std::string mode,
     return adios2::py11::File(name, mode, comm, configfile, ioinconfigfile);
 }
 
-#else
+#endif
 adios2::py11::File Open(const std::string &name, const std::string mode,
                         const std::string enginetype)
 {
@@ -99,7 +99,6 @@ adios2::py11::File OpenConfig(const std::string &name, const std::string mode,
 {
     return adios2::py11::File(name, mode, configfile, ioinconfigfile);
 }
-#endif
 
 PYBIND11_MODULE(adios2, m)
 {
@@ -151,8 +150,13 @@ PYBIND11_MODULE(adios2, m)
         .export_values();
 
 #ifdef ADIOS2_HAVE_MPI
-    m.def("open", &Open, pybind11::arg("name"), pybind11::arg("mode"),
-          pybind11::arg("comm"), pybind11::arg("engine_type") = "BPFile", R"md(
+    m.def("open",
+          (adios2::py11::File(*)(
+              const std::string &name, const std::string mode,
+              adios2::py11::MPI4PY_Comm comm, const std::string enginetype)) &
+              Open,
+          pybind11::arg("name"), pybind11::arg("mode"), pybind11::arg("comm"),
+          pybind11::arg("engine_type") = "BPFile", R"md(
           Simple API MPI open, based on python IO. 
           Allows for passing parameters in source code.
 
@@ -175,9 +179,15 @@ PYBIND11_MODULE(adios2, m)
                   handler to adios File for the simple Python API
     )md");
 
-    m.def("open", &OpenConfig, pybind11::arg("name"), pybind11::arg("mode"),
-          pybind11::arg("comm"), pybind11::arg("config_file"),
-          pybind11::arg("io_in_config_file"), R"md(
+    m.def("open",
+          (adios2::py11::File(*)(
+              const std::string &name, const std::string mode,
+              adios2::py11::MPI4PY_Comm comm, const std::string &configfile,
+              const std::string ioinconfigfile)) &
+              OpenConfig,
+          pybind11::arg("name"), pybind11::arg("mode"), pybind11::arg("comm"),
+          pybind11::arg("config_file"), pybind11::arg("io_in_config_file"),
+          R"md(
           Simple API MPI open, based on python IO. 
           Allows for passing a runtime configuration file in xml format and the
           name of the io element related to the returning File.
@@ -204,16 +214,24 @@ PYBIND11_MODULE(adios2, m)
                   handler to adios File for the simple Python API
     )md");
 
-#else
-    m.def("open", &Open, "High-level API, file object open",
-          pybind11::arg("name"), pybind11::arg("mode"),
-          pybind11::arg("engine_type") = "BPFile");
-
-    m.def("open", &OpenConfig,
-          "High-level API, file object open with a runtime config file",
-          pybind11::arg("name"), pybind11::arg("mode"),
-          pybind11::arg("config_file"), pybind11::arg("io_in_config_file"));
 #endif
+    m.def(
+        "open",
+        (adios2::py11::File(*)(const std::string &name, const std::string mode,
+                               const std::string enginetype)) &
+            Open,
+        "High-level API, file object open", pybind11::arg("name"),
+        pybind11::arg("mode"), pybind11::arg("engine_type") = "BPFile");
+
+    m.def(
+        "open",
+        (adios2::py11::File(*)(const std::string &name, const std::string mode,
+                               const std::string configfile,
+                               const std::string ioinconfigfile)) &
+            OpenConfig,
+        "High-level API, file object open with a runtime config file",
+        pybind11::arg("name"), pybind11::arg("mode"),
+        pybind11::arg("config_file"), pybind11::arg("io_in_config_file"));
 
     pybind11::class_<adios2::py11::ADIOS>(m, "ADIOS")
         // Python 2
@@ -237,7 +255,7 @@ PYBIND11_MODULE(adios2, m)
              "adios2 module starting point, constructs an ADIOS class object",
              pybind11::arg("configFile"), pybind11::arg("comm"),
              pybind11::arg("debugMode") = true)
-#else
+#endif
         .def(pybind11::init<const bool>(),
              "adios2 module starting point "
              "non-MPI, constructs an ADIOS class "
@@ -247,7 +265,6 @@ PYBIND11_MODULE(adios2, m)
              "adios2 module starting point non-MPI, constructs an ADIOS class "
              "object",
              pybind11::arg("configFile"), pybind11::arg("debugMode") = true)
-#endif
         .def("DeclareIO", &adios2::py11::ADIOS::DeclareIO,
              "spawn IO object component returning a IO object with a unique "
              "name, throws an exception if IO with the same name is declared "
```

ax3l:
Thank you so much, that's wonderful! :sparkles: 
Since 2.4.0-rc just came out: will we add these refactorings to the 2.4.0 release? cc @franzpoeschel

williamfgc:
@germasch sounds good. @ax3l yes, this should go into 2.4.0

PR #1473, opened 2019-06-04
Fix python bindings for unified serial and parallel library

pnorbert:
We rely on this trick to work in the type caster:
```
        // Import mpi4py if it does not exist.
        if (!PyMPIComm_Get)
        {
            if (import_mpi4py() < 0)
            {
                throw std::runtime_error(
                        "ERROR: mpi4py not loaded correctly\n"); /* Python 2.X */
            }
        }
```

### DIFF
--- a/bindings/Python/py11glue.cpp
+++ b/bindings/Python/py11glue.cpp
@@ -57,7 +57,7 @@ public:
             if (import_mpi4py() < 0)
             {
                 throw std::runtime_error(
-                        "ERROR: mpi4py not loaded correctly\n"); /* Python 2.X */
+                    "ERROR: mpi4py not loaded correctly\n"); /* Python 2.X */
             }
         }
         // If src is not actually a MPI4PY communicator, the next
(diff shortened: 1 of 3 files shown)


bindings/Python/py11glue.cpp
```diff
@@ -51,10 +51,18 @@ struct type_caster<adios2::py11::MPI4PY_Comm>
...
     {
+        // Import mpi4py if it does not exist.
+        if (!PyMPIComm_Get)
+        {
+            if (import_mpi4py() < 0)
+            {
+                throw std::runtime_error(
+                        "ERROR: mpi4py not loaded correctly\n"); /* Python 2.X */
+            }
+        }
```
germasch: I think I was involved in writing that caster, but I never fully understood the `import_mpi4py` thing.
Why is importing mpi4py still necessary at that point? I would think if the user is passing a communicator from python code, they must already have imported mpi4py themselves, or does that not propagate through into the wrapping layer?

pnorbert:
Apparently import mpi4py at the python level does not make it accessible
from C++. This condition will run in the first invocation (ADIOS
constructor) but will pass in all subsequent invocations. Why, I have no
idea. I took the conditional from the VTK code.

williamfgc:
@pnorbert @germasch thanks, this will largely simplify the [ADIOS2-Jupyter repo](https://github.com/ornladios/ADIOS2-Jupyter) no need to keep two branches for MPI and no-MPI examples

PR #1474, opened 2019-06-04
Rework LatestTimestep as an engine parameter rather than in BeginStep

eisenhauer:


eisenhauer:
This replaces PR #1400.  Note that this accommodates a criticism of #1400, that leaving StepMode::NextAvailable with that name seemed weird, given that there is no other read mode.  So, all occurrences of NextAvailable have been changed to StepMode::Read.  This is a *LOT* of places, because of use in bindings and as the default value for StepMode whenever a timeout is specified.  These changes would likely impact lots of places in the Tutorial as well, at least to the extent that C and fortran bindings were used because you can't rely upon the defaults but will have specified NextAvailable explicitly.

suchyta1:
I have a question. Does this means there's no longer `StepMode::LatestAvailable` when you call `engine.BeginStep()`? And if so, did you make it a parameter you can set with `engine.SetParamater()`, or does a latest available mode just not exist at all any more.

eisenhauer:
Sorry Eric, I missed this comment. StepMode::LatestAvailable is now an engine parameter, AlwaysProvideLatestTimestep, and is interpreted on the reader side at the time of Open().  So essentially LatestAvailable is now an open-time choice and we have better optimizations available for it.

PR #1475, opened 2019-06-05
Fixing mismatched instrumentation in SST layer

khuck:
Fixed missing start/stop calls for two timers in the SST metadata code.  When these are mismatched, it causes TAU to abort when profiling/tracing ADIOS.

### DIFF
--- a/source/adios2/toolkit/sst/cp/cp_writer.c
+++ b/source/adios2/toolkit/sst/cp/cp_writer.c
@@ -1874,6 +1874,7 @@ extern void SstInternalProvideTimestep(
     *Msg = ReturnData->Msg;
     Msg->CohortSize = Stream->CohortSize;
     Msg->Timestep = Timestep;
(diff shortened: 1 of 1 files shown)


eisenhauer:
Generally looks good and easy, but latest commits created a conflict. Rebase and retry?

khuck:
Hmm...I rebased and now it has a bunch more commits.  Hope that's OK.  Let me know if the checks fail...

eisenhauer:
Uh...  Hmm.  I'm a bit worried by the extra commits and what effect those might have.  Maybe the rebase went bad?  Might be better off to cherrypick just the one and try with a new PR?

germasch:
This should help:
```
$ git checkout khuck-patch-2
$ git reset --soft master # assumes master is up-to-date with upstream
$ git commit
```
And then you should have a single commit on top of master, which you can force-push.

khuck:
Thanks @germasch !  That worked.

PR #1479, opened 2019-06-06
Calculate and use absolute step instead of relative step where necessary

pnorbert:
An example where bpls is using data structures directly to do something that is not available in the public API. A refactoring would bring out issues with the public API.

### DIFF
--- a/source/utils/bpls/bpls.cpp
+++ b/source/utils/bpls/bpls.cpp
@@ -1465,7 +1465,7 @@ int readVar(core::Engine *fp, core::IO *io, core::Variable<T> *variable)
     int tidx;                          // 0 or 1 to account for time dimension
     uint64_t nelems;                   // number of elements to read
     // size_t elemsize;                   // size in bytes of one element
-    uint64_t st, ct;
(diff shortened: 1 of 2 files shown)


source/utils/bpls/bpls.cpp
```diff
@@ -2576,18 +2578,43 @@ void print_endline(void)
     nextcol = 0;
 }
 
+template <class T>
+size_t relative_to_absolute_step(core::Variable<T> *variable,
+                                 const size_t relstep)
+{
+    const std::map<size_t, std::vector<size_t>> &indices =
+        variable->m_AvailableStepBlockIndexOffsets;
+    auto itStep = std::next(indices.begin(), 0);
```
germasch: Is there a reason to use `std::next` rather than 
```
auto itStep = indices.begin()
```
directly?

germasch:
I think going towards using the public API / making sure that the public API  makes all necessary data available is a good thing.

FWIW, I have a WIP patch for bpls that also works towards that, but I got stuck because some things are not available from the public API, and others I don't really understand...

PR #1480, opened 2019-06-07
final mpiwrap / mpidummy bits

germasch:
This PR has three changes, which are actually independent of each other:

- use `MPI_COMM_NULL` to keep track of having been invoked via serial API
- remove unused `MPI_File` mpidummy implementation
- minor cleanup in `Transport`.

Some more details are in the commit messages.


williamfgc:
@germasch on a second thought, please disregard my previous comment. These things are from almost 3 years ago and there was a refactoring at some point that moved the uniqueness handling to the `m_Name` given by the TransportManager, in the Transport abstraction. It's OK to remove it. Thanks.

germasch:
Thanks @williamfgc. It's not a big deal either way, whatever you prefer.

germasch:
Changed the status of this PR from Draft to Ready to see whether that'll help with the hanging Travis CI...

chuckatkins:
The "Draft" feature is neat, I've not see that before (it looks like it was just introduced a few months ago).  It seems they're essentially trying to have the idea of "WIP" built in to github with it.

germasch:
Yeah, I like the idea behind the "Draft" PR. Unfortunately, it's missing the feature to switch a PR back into Draft state.

And it does look like draft PRs are in fact not supported by Travis CI at this time :(

germasch:
Alright, so I consider this ready now. The `MPI_COMM_NULL` commit made the code a bit shorter, which  generally take as a sign that things are going the right direction.

There's one workaround in there that I kept:
```cxx
ADIOS::~ADIOS()
{
    // Handle the case where MPI is finalized before the ADIOS destructor is
    // called, which happens, e.g., with global / static ADIOS objects
    int flag;
    MPI_Finalized(&flag);
    if (!flag)
    {
        SMPI_Comm_free(&m_MPIComm);
    }
}
```
I don't really like it, it covers up for applications that create a parallel `adios:ADIOS` instance, passing in a communicator, but then don't destruct until after `MPI_Finalize`. That's arguably just as buggy as doing this with the C API:

```
    [...]
    MPI_Finalize();
    adios2_finalize(ad);
```
Only that in C++ it can be hidden because destructors get called automatically, in particular, destructors for static/global objects get called after the end of `main()`.  The workaround that I kept in there fixes the common case, but this same issue can still happen in more complex scenarios.

I'm not convinced that adios2 should work around this kind of user error. It invites users to get sloppy with this, and in the future, imagine a scenario, where the adios destructor will delete IOs and Engines, and the Engine destructor will trigger a `Close`, which will flush remaining data to disk. Well, if one used MPI aggregation, one is definitely out of luck, since that won't work for sure after `MPI_Finalize`, and adios2 can't work around that case. So it might be better to be strict about it from the beginning, even though that might break some existing applications (should really only be C++ that's affected). Which includes the heatTransfer example, by the way.

ax3l:
>     if (!flag)

Hm, can you maybe store if ADIOS performed the `MPI_Init` and only finalize if that is also the case? I don't think finalizing MPI is a good idea in general. Reasons for that: I might destruct the ADIOS object but keep using an MPI context. On some machines, you are also only allowed to call `MPI_Init` once (e.g. Cray jobs, IIRC).

germasch:
@ax3l, even with the recent patches, ADIOS2 will never call `MPI_Init`, it just doesn't use any MPI calls when used with the serial API. In order, to be able to keep track of whether current functionality is running in serial mode, I slightly abuse `MPI_Comm`. If it's `MPI_COMM_NULL`, adios2 will emulate some MPI functionality, but not actually call into the MPI library. For all other communicators, it uses MPI as usual.

I think having ADIOS2 calling `MPI_Init` itself is not a good idea. First, it's just wrong on a fundamental level, it's `main()`'s job. Second, ADIOS2 doesn't know the right params (`argc`, `argv`, though that's probably not a problem these days). Third, it invites other weird problems, in particular, ADIOS2 calling `MPI_Init`, and then the application calling it at a later time. In particular, with C++, it's quite easy to have constructors run before the start of `main()`, and destructors run after the end of `main()`, creating odd errors. In fact, adios2 being initialized before main() starts, and finalized after it ends happens even in current example code, which is why I kept the workaround above. But I'd rather get rid of it entirely, at which point the logic simply would be:

- the serial API can be used at any time, not interfering with MPI or it's init/finalize at all.
- the parallel API constructor (which takes a communicator) must be called after MPI is initialized, and the object(s) must be destructed before MPI is finalized.

Right now, the piece of code you're talking about partially relaxes the 2nd part of this rule, by working allowing an ADIOS object to be destructed even after MPI_Finalize, but it's essentially a band-aid only, not handling that case in all possible circumstances.

williamfgc:
the "safe" way to C++ libraries that throw exceptions (e.g. STL) is inside a `try/catch` block, that's how C++ examples are presented in adios2, as well as `bpls`. This has been the recommended way for decades, not doing this is basically asking for trouble. ADIOS2 always checks for `std::runtime_error`.

```
MPI_Init(&argc, &argv);
...
try
{
   adios2::ADIOS adios(comm);
  ...
}
catch(std::exception& e)
{
}

MPI_Finalize();
```

germasch:
I fully agree that if one wraps the `ADIOS` object in a try/catch block, it'll certainly be guaranteed that its destructor is called by the end of that block, so calling `MPI_Finalize` after that block means that there won't be any problems with adios2 calling MPI after `MPI_Finalize`. The question is whether we want to disallow the opposite behavior, ie, destructing the `ADIOS` object after `MPI_Finalize`.

This PR, however, doesn't change that behavior, anyway. That is, it still permits destructing `ADIOS` late via the workaround I kept, in order to maintain compatibility. I'd be happy to clean up that final bit (though it would be a backward-incompatible change), but this PR is a prerequisite for that.

ax3l:
> The question is whether we want to disallow the opposite behavior, ie, destructing the ADIOS object after MPI_Finalize.

I think it's quite hard to recover from such a situation, no? ADIOS could still have all kinds of MPI contexts in use...

germasch:
> > The question is whether we want to disallow the opposite behavior, ie, destructing the ADIOS object after MPI_Finalize.
> 
> I think it's quite hard to recover from such a situation, no? ADIOS could still have all kinds of MPI contexts in use...

That's my point. I kept one workaround that recovers from one particular "misuse" of not destructing the `ADIOS` object, that is the case where all of the remaining work is to `MPI_Comm_free` the communicator. The workaround just skips the `MPI_Comm_free` in that instance, in order to avoid a runtime error from the MPI library. But it's a band-aid, because it doesn't handle anything more complex, and as you say, handling the case more generally is basically not possible, in particular if one has to do remaining communication at cleanup time. (An example would be where the destructor would lead to closing an open BP3 file, which would have to use MPI to gather metadata to correctly write the metadata file -- that's just not possible after `MPI_Finalize`). So, in my opinion, we shouldn't have ADIOS2 work around incorrect use by the application at all, not even in the case that we can handle.

Nevertheless, I kept the workaround in there in order to not break existing code, so the PR as it is right now is backwards compatible. I kept the workaround to make it easier to have this PR merged without a discussion on backward-compatibility vs cleanliness, as right now it's safe, so I don't really see any argument against merging it.

I would then make a follow-up PR removing the workaround, which might break existing apps, so for that one there are arguments for/against merging it and the ADIOS2 people have to weigh the merits.

So in terms of this PR, I'll poke @pnorbert and @williamfgc and maybe @chuckatkins @bradking to see whether there is any reason not to take it.

williamfgc:
sounds good to me, but I haven't been involved in all of the details so I'll wait for the others to approve/merge. Thanks!

williamfgc:
@chuckatkins are you OK with this, let's review outstanding PRs before freezing. Thanks!

PR #1483, opened 2019-06-09
Python Module: __version__

ax3l:
The `__version__` attribute is a standardized and recommended attribute for Python modules ([PEP 396](https://www.python.org/dev/peps/pep-0396)). It also comes in very handy to check if one imported a suitable minimum version. These versions
should be strings.
```python
import adios2

print(adios2.__version__)
# '2.4.0'
```

In case we want to add suffixes, such as release candidate identifiers, please be aware of the specifics in [PEP440](https://www.python.org/dev/peps/pep-0440).
Examples:
```
X.YaN    # Alpha release
X.YbN    # Beta release
X.YrcN   # Release Candidate
X.Y      # Final release
X.Y.devN # Development release
```

### DIFF
--- a/bindings/Python/py11glue.cpp
+++ b/bindings/Python/py11glue.cpp
@@ -11,6 +11,7 @@
 #include <pybind11/pybind11.h>
 #include <pybind11/stl.h>
 
+#include <sstream>
 #include <stdexcept>
 
 #include <adios2.h>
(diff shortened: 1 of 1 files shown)


bindings/Python/py11glue.cpp
```diff
@@ -119,6 +120,11 @@ PYBIND11_MODULE(adios2, m)
     m.attr("GlobalValue") = false;
     m.attr("LocalValue") = true;
 
+    std::stringstream versionss;
```
williamfgc: `std::ostringstream` so the intention is clear upfront. Thanks!

germasch:
@ax3l Sometimes, the CI gives random failures for no obvious reasons (I suppose some kind of race). The usual workaround is to amend the last commit to update its timestamp, force-push and keep your fingers crossed ;)

PR #1487, opened 2019-06-11
fixed an unnecessary check in dataman

### LINKED ISSUE #1469: warning/bug? in dataman
I'm getting this warning in DataManWriter:

```
[13/86] Building CXX object source/adios2/CMakeFiles/adios2.dir/engine/dataman/DataManWriter.cpp.o
../source/adios2/engine/dataman/DataManWriter.cpp:157:29: warning: comparison of unsigned expression >= 0 is always true [-Wtautological-compare]
        if (request->size() >= 0)
            ~~~~~~~~~~~~~~~ ^  ~
1 warning generated.
```
I don't know the correct fix is, though. Since the request size is never negative, the current behavior is to always send a reply, even if the request size is 0. If that's the intended behavior, the if statement should just be removed. If the intention was to only reply to messages with a size strictly greater than 0, well, then there's an actual bug.

JasonRuonanWang:
Fixes #1469

### DIFF
--- a/source/adios2/engine/dataman/DataManWriter.cpp
+++ b/source/adios2/engine/dataman/DataManWriter.cpp
@@ -152,7 +152,7 @@ void DataManWriter::MetadataThread(const std::string &address)
         auto request = tpm.ReceiveRequest();
         if (request && request->size() > 0)
         {
-            std::lock_guard lck(m_AggregatedMetadataMutex);
+            std::lock_guard<std::mutex> lck(m_AggregatedMetadataMutex);
             tpm.SendReply(m_AggregatedMetadata);
         }
     }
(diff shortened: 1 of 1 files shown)


source/adios2/engine/dataman/DataManWriter.cpp
```diff
@@ -150,15 +150,10 @@ void DataManWriter::MetadataThread(const std::string &address)
...
         auto request = tpm.ReceiveRequest();
-        if (request == nullptr)
+        if (request && request->size() > 0)
         {
-            continue;
-        }
-        if (request->size() >= 0)
-        {
-            m_AggregatedMetadataMutex.lock();
+            std::lock_guard lck(m_AggregatedMetadataMutex);
```
chuckatkins: Looks like the template argument doesn't resolve by default so you need `std::lock_guard<std::mutex> lck`

PR #1490, opened 2019-06-11
Fix incorrect error message describing CMake options.

NAThompson:


williamfgc:
@NAThompson thanks! This is my bad, the exception and docs are out of sync. This PR solves it.

PR #1494, opened 2019-06-12
Generate the most advanced instructions for the architecture when use…

NAThompson:
…rs do a source build. Add sanitizers for debug builds and keep the frame pointers. Remove unused private variable warnings and mark static functions in headers with definitions as inline to cut down on warnings.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -70,6 +70,14 @@ if(NOT CMAKE_BUILD_TYPE AND NOT CMAKE_CONFIGURATION_TYPES)
   set(CMAKE_BUILD_TYPE "Release" CACHE STRING "Choose the type of build." FORCE)
 endif()
 
+
+if(CMAKE_BUILD_TYPE MATCHES "Release" AND NOT MSVC)
+  set(CMAKE_CXXFLAGS "-g -O3 -march-native -ffast-math -Wfatal-errors")
+elseif(NOT MSVC)
+  set(CMAKE_CXX_FLAGS_DEBUG "-Wall -g -O2 -Wfatal-errors -fsanitize=address -fsanitize=undefined -fno-omit-frame-pointer")
+endif()
+
+
 # Force C++11 and C99
 set(CMAKE_CXX_STANDARD 11)
 set(CMAKE_CXX_STANDARD_REQUIRED True)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -72,9 +72,11 @@ endif()
 
 
 if(CMAKE_BUILD_TYPE MATCHES "Release" AND NOT MSVC)
-  set(CMAKE_CXXFLAGS "-g -O3 -march-native -ffast-math -Wfatal-errors")
-elseif(NOT MSVC)
-  set(CMAKE_CXX_FLAGS_DEBUG "-Wall -g -O2 -Wfatal-errors -fsanitize=address -fsanitize=undefined -fno-omit-frame-pointer")
+  set(CMAKE_CXX_FLAGS "-g -O3 -march-native -ffast-math -Wfatal-errors")
+elseif("${CMAKE_CXX_COMPILER_ID}" STREQUAL "AppleClang" OR
+       "${CMAKE_CXX_COMPILER_ID}" STREQUAL "Clang" OR
+       "${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU")
+  set(CMAKE_CXX_FLAGS_DEBUG "-Wall -g -O0 -Wfatal-errors -fsanitize=address -fsanitize=undefined -fno-omit-frame-pointer")
 endif()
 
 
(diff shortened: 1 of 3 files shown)


CMakeLists.txt
```diff
@@ -70,6 +70,14 @@ if(NOT CMAKE_BUILD_TYPE AND NOT CMAKE_CONFIGURATION_TYPES)
   set(CMAKE_BUILD_TYPE "Release" CACHE STRING "Choose the type of build." FORCE)
 endif()
 
+
+if(CMAKE_BUILD_TYPE MATCHES "Release" AND NOT MSVC)
+  set(CMAKE_CXXFLAGS "-g -O3 -march-native -ffast-math -Wfatal-errors")
+elseif(NOT MSVC)
+  set(CMAKE_CXX_FLAGS_DEBUG "-Wall -g -O2 -Wfatal-errors -fsanitize=address -fsanitize=undefined -fno-omit-frame-pointer")
+endif()
+
```
germasch: I don't think it's a good idea to hardcode compiler flags that are gcc/clang specific. adios2 may be built with xlc, intel, cray, pgi, etc.

I do think the sanitizers are useful for testing / the CI, but shouldn't necessarily be in every debug build. And I prefer `-O0` in my debug builds.

The rest of this patch looks good to me.

NAThompson: Yup, you're exactly right about compiler detection; that's fixed.

As to `-O2`: I wondered how y'all felt about the how long the unit tests took; sometimes `-O0` makes trillions of superfluous stack writes that are pretty frustrating unless you're under a debugger. But if everyone is ok with `-O0` I'm ok with it too.

germasch: I think @chuckatkins should take a look at this. I have no say, anyway, I'm just a contributor. I think even after the change above, you're still setting CMAKE_CXX_FLAGS irregardless of clang/gcc in the Release case.

But I'm also generally not sure about setting CMAKE_CXX_FLAGS in CMakeLists.txt in the first place -- does that mean if I specify those on the command line, they'll be overwritten regardless, or does the command line take precedence?

chuckatkins: This entire block needs to be taken out.  The way you target CPU sets is different on different platforms and the ADIOS project is not in the business of encoding compiler secret sauce.  For example, this is handled in an entirely different mechanism when building on a cray.  CMake has default flags for debug and release across a wide swatch of compilers that's been heavily tested for some time.  While it may not the "super optimal" it certainly gives a safe best guess for optimized configurations; i.e. O3 isn't universally safe, the fastmath and  options aren't available on all the compilers we are supporting, you've dropped the -DNDEBUG flag often used by std library implementations, etc.  If a user wants to specify some specific special sauce flags when they build it then they can certainly do so via the `CFLAGS`, `CXXFLAGS`, and `FFLAGS` environment variables or `-DCMAKE_CXX_FLAGS`, `-DCMAKE_C_FLAGS`, and `-DCMAKE_Fortran_FLAGS` CMake variables (yes, we use all three languages which, while not the point I'm making, would need to be accounted for if this were to go in, which I'm not supportive of).  We do just that in some of our dashboard builds but it's compiler and environment specific and isn't suitable to encode directly into adios.

source/adios2/helper/adiosMemory.inl
```diff
@@ -514,7 +514,7 @@ void Resize(std::vector<T> &vec, const size_t dataSize, const bool debugMode,
 // functions) and copies to the output buffer in blocks. the memory address
 // calculation complexity for copying each block is minimized to O(1), which is
 // independent of the number of dimensions.
-static void NdCopyRecurDFSeqPadding(size_t curDim, const char *&inOvlpBase,
+static inline void NdCopyRecurDFSeqPadding(size_t curDim, const char *&inOvlpBase,
```
chuckatkins: I doubt explicitly marking these inline actually makes any difference since it's non-binding and just a "suggestion" to the compiler anyways.  The likelihood is pretty high that these were already getting inlined by compilers.  Regardless though, it's good for the sake of complete to mark them as such, so I'm supportive of this.

CMakeLists.txt
chuckatkins: This entire block needs to be taken out. The way you target CPU sets is different on different platforms and the ADIOS project is not in the business of encoding compiler secret sauce. For example, this is handled in an entirely different mechanism when building on a cray. CMake has builtin default flags for debug and release across a wide swath of compilers that's been heavily tested for some time. While it may not the "super optimal" it certainly gives a safe best guess for optimized configurations; i.e. O3 isn't universally safe, the fastmath and W options aren't available on all the compilers we are supporting, you've dropped the -DNDEBUG flag often used by std library implementations, etc. If a user wants to specify some specific special sauce flags when they build it then they can certainly do so via the CFLAGS, CXXFLAGS, and FFLAGS environment variables or -DCMAKE_CXX_FLAGS, -DCMAKE_C_FLAGS, and -DCMAKE_Fortran_FLAGS CMake variables (yes, we use all three languages which, while not the point I'm making, would need to be accounted for if this were to go in, which I'm not supportive of). We do just that in some of our dashboard builds but it's compiler and environment specific and isn't suitable to encode directly into adios.

chuckatkins:
@NAThompson if you can drop the CMake changes and limit this to the inline change then I can go ahead and get in merged into our release.

chuckatkins:
I'm going to go ahead and close this for the release since it hasn't seen any additional activity.  If you'd like to open a new one into master after the release please do so.

PR #1496, opened 2019-06-13
support SetMemorySelection on read in BP3

germasch:
Currently, `SetMemorySelection` is entirely ignored on Read/Get. This PR adds support into BP3,
so formats based on BP3 gain the ability to use it. This PR, however, does not try to handle the case of mixed endianness, that is, when `m_ReverseDimensions` is set. **ETA:** If one tries to use `SetMemorySelection` with mixed endianness, the code will catch that case and throw an exception.

Other than added tests, there is only one change to core code, ie., to `BP3Deserializer.tcc`. It should be easy to verify that this change does not break existing code, because in the case where the memory selection is empty (the only currently supported case), the code still calls `ClipContiguousMemory` exactly as before. Behavior is only different in the case where a memory selection has been set, in which case it will use `helper::NdCopy` to handle the copy correctly, as hopefully proven by the bunch of tests that I added.

### DIFF
--- a/bindings/CXX11/cxx11/Variable.h
+++ b/bindings/CXX11/cxx11/Variable.h
@@ -174,7 +174,7 @@ public:
      * Set the local start (offset) point to the memory pointer passed at Put
      * and the memory local dimensions (count). Used for non-contiguous memory
      * writes and reads (e.g. multidimensional ghost-cells).
-     * Currently not working for calls to Get.
+     * Currently Get only works for formats based on BP3.
      * @param memorySelection {memoryStart, memoryCount}
      * <pre>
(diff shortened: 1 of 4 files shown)


germasch:
After some work fixing the tests (not really related to the main purpose of this PR), it now passes the CI. It's a feature I need, and as explained above it only adds this new feature, existing code paths are unchanged. It's ready for review.

williamfgc:
@germasch  thanks! I'd merge it for now, but treat it as experimental in the docs. The reason why is that the NdCopy function hasn't been tested in the big machines.  Recursion depth in that function hasn't been tested at large scales.

PR #1501, opened 2019-06-14
Fix endian reverse in SST+BP

keichi:
This PR fixes a data corruption bug that happens in SST+BP when streaming across machines with different endianness. SstReader bypasses BP3Deserializer if input and output are contiguous memory, but that won't work if the endianness doesn't match.

I've tested the PR on a VM (qemu-ppc64-user) and Fujitsu FX100.


eisenhauer:
Thanks Keichi!  We don't have a test environment for heterogeneous staging, so relying upon you for this!

PR #1502, opened 2019-06-14
Fix BP3Serializer crash on big-endian machines if Endian_Reverse is enabled

keichi:
BP3Serializer assumes little-endian and crashes with the following error on big-endian machines when compiled with Endian_Reverse=ON.

```
terminate called after throwing an instance of 'std::bad_alloc'
  what():  std::bad_alloc
```


williamfgc:
@keichi is the writer and reader both Big Endian? If so, it might be another bug, Endia_Reverse is just to handle interoperability. Turning it off would work as long as the endianness is constant across Writers and Readers.

keichi:
Hi @williamfgc, I have both use cases (BE->BE and BE->LE). This crash happens regardless of the reader endianness as long as Endian_Reverse is enabled. You're right, I can turn Endian_Reverse off for the BE->BE use case, but I don't want to have two separate ADIOS2 builds, one with Endian_Reverse and another without Endian_Reverse.

williamfgc:
@keichi makes sense, thanks for the explanation.

PR #1505, opened 2019-06-14
Unify BP Operations and add BZIP2 and PNG compression

williamfgc:
In addition:
Update README

Disabled DataMan BZIP2 tests @JasonRuonanWang to handle extra metadata for chunks when data > 2Gb


williamfgc:
@NAThompson

source/adios2/operator/compress/CompressBZIP2.cpp
```diff
@@ -0,0 +1,231 @@
...
+ *      Author: William F Godoy godoywf@ornl.gov
+ */
+
+#include "CompressBZIP2.h"
+
+#include <cmath>     //std::ceil
+#include <ios>       //std::ios_base::failure
+#include <stdexcept> //std::invalid_argument
+
+extern "C" {
```
NAThompson: I just took a look at my copy of `bzlib.h`. On line 25, it does its own `extern "C"` work; so ostensibly there's no cause to do our own.

source/adios2/operator/compress/CompressPNG.cpp
```diff
@@ -0,0 +1,220 @@
...
+ *  Created on: Jun 10, 2019
+ *      Author: William F Godoy godoywf@ornl.gov
+ */
+#include "CompressPNG.h"
+
+#include <cstring> // std::memset
+
+extern "C" {
+#include <png.h>
+}
```
NAThompson: My copy of `png.h` has the following comment, line 366:

```cpp
/* Inhibit C++ name-mangling for libpng functions but not for system calls. */
#ifdef __cpluplus
extern "C" {
#endif 
```

So presumably they are doing the right thing here; and there's no need for the `extern "C"` in our code.

source/adios2/operator/compress/CompressPNG.cpp
```diff
@@ -0,0 +1,220 @@
...
+    {"PNG_COLOR_TYPE_GA", PNG_COLOR_TYPE_GA}};
+
+const std::map<std::string, std::set<uint32_t>> CompressPNG::m_BitDepths = {
+    {"PNG_COLOR_TYPE_GRAY", {1, 2, 4, 8, 16}},
+    {"PNG_COLOR_TYPE_PALETTE", {1, 2, 4, 8}},
+    {"PNG_COLOR_TYPE_RGB", {8, 16}},
+    {"PNG_COLOR_TYPE_RGB_ALPHA", {8, 16}},
+    {"PNG_COLOR_TYPE_GRAY_ALPHA", {8, 16}},
+    {"PNG_COLOR_TYPE_RGBA", {8, 16}},
+    {"PNG_COLOR_TYPE_GA", {8, 16}}};
```
NAThompson: It looks like you are doing a very full features exposure of the feature set of `libpng`, but from what I have read, basically all hardware is doing 8 bit bit depths. I haven't yet understood your intended use case, but are the variable bit lengths exposed to the user?

source/adios2/operator/compress/CompressPNG.cpp
```diff
@@ -0,0 +1,220 @@
...
+    const std::size_t ndims = dimensions.size();
+    if (m_DebugMode)
+    {
+        if (ndims != 3 && ndims != 2)
+        {
+            throw std::invalid_argument(
+                "ERROR: image number of dimensions " + std::to_string(ndims) +
+                " is invalid, must be 2 {height,width*bytes_per_pixel} or 3"
+                " {height,width,bytes_per_pixel]} , in call to ADIOS2 PNG "
+                " compression\n");
```
NAThompson: I'm worried about two things here: First, I feel like I'd want to see this error message independently of whether I was in debug mode. Second, the comment states that the meaning of the second argument depends on the presence of the third. I'd be concerned that people will write bugs using this.

source/adios2/operator/compress/CompressPNG.cpp
```diff
@@ -0,0 +1,220 @@
...
+    for (const auto &itParameter : parameters)
+    {
+        const std::string key = itParameter.first;
+        const std::string value = itParameter.second;
+
+        if (key == "compression_level")
+        {
+            compressionLevel = static_cast<int>(helper::StringTo<int32_t>(
+                value, m_DebugMode, "when setting PNG level parameter\n"));
+            if (m_DebugMode)
```
NAThompson: So I still am a bit fuzzy on the philosophy of this library, but I feel like this error message should be given independently of whether `m_DebugMode` is true or false. (You'll have to give me a rundown on the vision for `m_DebugMode` in meatspace.)

source/adios2/operator/compress/CompressPNG.h
```diff
@@ -0,0 +1,87 @@
...
+#include "adios2/core/Operator.h"
+
+namespace adios2
+{
+namespace core
+{
+namespace compress
+{
+
+class CompressPNG : public Operator
```
NAThompson: The thing I worry about with this file is that since png is a lossless format, there's very little scope for high compression ratios. So I worry that users will futz around with this capability without actually generating wins.

Have you seen any wins from playing with the Zlib compress options exposed in `libpng`?

testing/adios2/engine/bp/operations/TestBPWriteReadZfp.cpp
```diff
@@ -411,12 +415,11 @@ void ZfpRate3D(const double rate)
     }
 }
 
-void ZfpRate1DSel(const double rate)
+void ZFPRate1DSel(const std::string rate)
```
NAThompson: Maybe this is bikeshedding, but shouldn't this be a reference?

williamfgc: Using C++11 Small String Optimization https://stackoverflow.com/questions/10315041/meaning-of-acronym-sso-in-the-context-of-stdstring/10319672#10319672

williamfgc:
@NAThompson thanks for your observations. A lot comes down to style, which is "code for the new maintainer", which could be anyone including myself in a few months as I will probably forget all the details. Using things like `extern`, `const`, SSO are more educational material and helps manage the maintainer's time finding things efficiently. Think how long it takes to go into the png and bzip2 to find out they are C libraries that are doing the right thing. This is very helpful as the codebase becomes larger and larger and you're overwhelmed with projects, deadlines, etc.

source/adios2/operator/compress/CompressPNG.h
williamfgc: The serializer in adios2 is just providing memory space. Generating wins is independent of adios2, it really comes down to the application inputs: data characteristics and settings and what the chosen compressor can handle.

source/adios2/operator/compress/CompressPNG.cpp
williamfgc: > (You'll have to give me a rundown on the vision for m_DebugMode in meatspace.)

Sure. DebugMode is for workflow developers, it catches `std::logic_error` (including `std::invalid_argument`) provided by the user. Turning it off is just convenient for static workflows and prevents user input checks already checked during development. On the other hand, `std::runtime_error` is always checked.

testing/adios2/engine/bp/operations/TestBPWriteReadZfp.cpp
NAThompson: Hmm, it doesn't look to me that there is any optimization being performed:

https://godbolt.org/z/edaqCc

The generated asm is identical either way. In any case, the rationale that it documents an expectation of small strings is very reasonable.

cmake/DetectOptions.cmake
```diff
@@ -49,6 +49,16 @@ if(MGARD_FOUND)
   set(ADIOS2_HAVE_MGARD TRUE)
 endif()
 
+# PNG
+if(ADIOS2_USE_PNG STREQUAL AUTO)
+  find_package(PNG 1.6.0)
```
NAThompson: Just one last comment: This broke on VTK-m's CI server; I downgraded it to PNG 1.2.0 and it was ok. (I dunno if you're using any PNG 1.6 features not present in PNG 1.2.)

testing/adios2/engine/bp/operations/TestBPWriteReadZfp.cpp
williamfgc: That's not how SSO works. In the example above the compiler is never aware of string size. it just functions passing by value or reference. SSO means that if the size of the string is known the compiler will [put it in the stack](https://shaharmike.com/cpp/std-string/) passing it by value is the same reason we pass primitives by value.

cmake/DetectOptions.cmake
williamfgc: @NAThompson good call, I think I am....1.6 is pretty old by now and 1.2 ancient

chuckatkins: @williamfgc RHEL7 uses 1.5 and RHEL6 uses 1.2.  If there's not a hard requirement on v1.6 then we should be able to relax it some and just drop the version from the `find_package` call.

williamfgc: @chuckatkins let me double check how much work would it be to downgrade. I am using the v1.6 high-level API for decompression following [their advice on releases](https://libpng.sourceforge.io/index.html) :

```
UPDATE 2019-Apr-14:

    The latest released version is
        libpng-1.6.37 

    [DOWNLOAD from SourceForge]

    For legacy applications,
        libpng-1.5.30
        libpng-1.4.22
        libpng-1.2.59
        libpng-1.0.69 
    are also available, and will be updated only if necessary for security reasons. 
```

chuckatkins: If you are using 1.6 specifically then that's fine.  The only caveat to realize is is that with RHEL7 you can't use the system libpng since it's only 1.5

PR #1506, opened 2019-06-14
Dataspaces Engine

pradsubedi:
This provides a DataSpaces engine for ADIOS2. This engine requires that the dataspaces_server be started before any programs that use the DataSpaces engine and that all servers and clients share the same working directory.

Adds the engine parameter "AppID", so that different client applications can be distinguished by the server.

DataSpaces >=1.8.0 is an external dependency to the DataSpaces engine.

This is a work in progress. Items to do include adding staging tests that run the DataSpaces server, and adding DataSpaces to the CI builds.


philip-davis:
Marked "Experimental" in documentation.

PR #1511, opened 2019-06-17
Bp4 append

lwan86:
Append mode for BP4


pnorbert:
Can you please resolve a simple conflict after merging the changes for LockDefinitions.

PR #1514, opened 2019-06-18
Consistently use BZip2 in CMake code instead of BZIP2

KyleFromKitware:
Other parts of the code use ADIOS_HAVE_BZip2, and this inconsistency
was causing installation failures. Fix the usage to make it consistent.

### DIFF
--- a/docs/user_guide/source/installation/cmake.rst
+++ b/docs/user_guide/source/installation/cmake.rst
@@ -383,7 +383,7 @@ CMake VAR Option               Values                     Description
 ``ADIOS2_USE_Python``          **`AUTO`**/``ON``/OFF      Python >= 2.7 bindings. mpi4py and numpy. Python 3 will be used if Python 2 and 3 are found. If you want a python version not in the path then choose the right pyhton executable by -DPYTHON_EXECUTABLE=... 
 ``ADIOS2_USE_Fortran``         **`AUTO`**/``ON``/OFF      Fortran 90 or above bindings. Must have a Fortran compiler. Default is OFF, must be explicitly set to ON.
 ``ADIOS2_USE_SST``             **`AUTO`**/``ON``/OFF      Simplified Staging Engine (SST) and its dependencies, requires MPI. Can optionally use LibFabric for RDMA transport. Specify the LibFabric install manually with the -DLIBFABRIC_ROOT=... option. 
-``ADIOS2_USE_BZIP2``           **`AUTO`**/``ON``/OFF      `BZIP2 <http://www.bzip.org>`_ compression.              
+``ADIOS2_USE_BZip2``           **`AUTO`**/``ON``/OFF      `BZIP2 <http://www.bzip.org>`_ compression.
 ``ADIOS2_USE_ZFP``             **`AUTO`**/``ON``/OFF      `ZFP <https://github.com/LLNL/zfp>`_ compression (experimental).
 ``ADIOS2_USE_SZ``              **`AUTO`**/``ON``/OFF      `SZ <https://github.com/disheng222/SZ>`_ compression (experimental).
 ``ADIOS2_USE_MGARD``           **`AUTO`**/``ON``/OFF      `MGARD <https://github.com/CODARcode/MGARD>`_ compression (experimental).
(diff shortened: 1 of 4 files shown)


chuckatkins:
So, this should really stay the existing case.  The convention is to have CMake options use UpperCamlCase while when translated to C preprocessor.  For example, in the cmake config files you should see things like:
```cmake
set(ADIOS2_HAVE_Python TRUE)
set(ADIOS2_HAVE BZip2 TRUE)
set(ADIOS2_HAVE_Fortran FALSE)
```
while in `ADIOS2Config.h` you should see corresponding
```c
#define ADIOS2_HAVE_PYTHON 1
#define ADIOS2_HAVE_BZIP2 1
#undef ADIOS2_HAVE_FORTRAN
```

So, if there's a mismatch in option case, that's at least how it's intended to be and should be fixed accordingly.

williamfgc:
@chuckatkins that was me trying to keep the options standard across compression libraries. CMake has [FindPNG](https://cmake.org/cmake/help/v3.15/module/FindPNG.html) as well, so the convention is not clear even within CMake itself, I haven't seen BZip2 in other than CMake (it's bzip2 all over). 

That being said, I have no preference with what is in adios2 as long as it's documented and not confusing to the user. Having BZip2, while having ZFP, PNG, MGARD, SZ everywhere else was doing that. Fortran and FORTRAN have different meanings, so I understand the case in language bindings.

chuckatkins:
> Having BZip2, while having ZFP, PNG, MGARD, SZ

I agree that it's confusing but mixed case is the preferred convention for CMake itself when handling package and target names.  Generally the CMake practice is to try to follow the naming convention used by the package itself in it's documentation.  You can see this in the CMake code base with `FindZLib.cmake`, `FindLibArchive.cmake`, `FindOpenCL.cmake`, etc.  Generally the CMake style will use all uppercase if the entire name is an acronym as in `FindPNG`, `FindCUDA`, etc.  You can see this with our own find modules as well via `FindZeroMQ`, `FindCrayDRC`, and then correspondingly `FindSZ` and `FindMGARD`.  The target names generally match when in mixed case and in the case of all upper case the target names are usually all lower case.

It's certainly a confusing set of heuristics I agree, and maybe not the best for user-facing options, but we do have a reasonable amount of consistency within the ADIOS2 CMake code, albeit not perfect.

williamfgc:
@chuckatkins thanks for the explanation, to add to my own confusion CMake has [FindZLIB](https://cmake.org/cmake/help/v3.15/module/FindZLIB.html) not FindZLib. Anyways, if you agree, I can put back the option as BZip2 and just point at CMake FindModule documentation as-is.

KyleFromKitware:
I have changed the option name and associated variables to `BZip2` per the convention described by @chuckatkins .

docs/user_guide/source/installation/cmake.rst
```diff
@@ -383,7 +383,7 @@ CMake VAR Option               Values                     Description
 ``ADIOS2_USE_Python``          **`AUTO`**/``ON``/OFF      Python >= 2.7 bindings. mpi4py and numpy. Python 3 will be used if Python 2 and 3 are found. If you want a python version not in the path then choose the right pyhton executable by -DPYTHON_EXECUTABLE=... 
 ``ADIOS2_USE_Fortran``         **`AUTO`**/``ON``/OFF      Fortran 90 or above bindings. Must have a Fortran compiler. Default is OFF, must be explicitly set to ON.
 ``ADIOS2_USE_SST``             **`AUTO`**/``ON``/OFF      Simplified Staging Engine (SST) and its dependencies, requires MPI. Can optionally use LibFabric for RDMA transport. Specify the LibFabric install manually with the -DLIBFABRIC_ROOT=... option. 
-``ADIOS2_USE_BZIP2``           **`AUTO`**/``ON``/OFF      `BZIP2 <http://www.bzip.org>`_ compression.              
+``ADIOS2_USE_BZip2``           **`AUTO`**/``ON``/OFF      `BZIP2 <http://www.bzip.org>`_ compression.
```
williamfgc: Thanks for changing the docs!

KyleFromKitware: :)

Issue #1477, opened 2019-06-06
install: internal headers not placed under adios2/


bradking:
As of commit `ec59c81e3d` on `master`, I see a layout in the install tree like this:

```
include/adios2.h
include/adios2_c.h
include/adios2/ADIOS*.h
include/c/*.h
include/cxx11/*.h
```

That means paths to files like `cxx11/Variable.h` do not mention "adios2" in them anywhere.
Instead the layout should be something like

```
include/adios2.h
include/adios2_c.h
include/adios2/ADIOS*.h
include/adios2/c/*.h
include/adios2/cxx11/*.h
```

Cc: @chuckatkins

williamfgc:
@bradking  just curious, what's the advantage for the proposed layout?

bradking:
E.g. if a package manager installs ADIOS2 to the `/usr` prefix then the path `/usr/include/cxx11/Variable.h` does not mention adios at all.

williamfgc:
@bradking thanks, I thought we were following the MPI/openmpi way. At least in my Ubuntu 18 box openmpi headers have the mpi prefix `/usr/include/mpi/` which is a soft link to `/usr/lib/x86_64-linux-gnu/openmpi/include`.   I'd expect cmake to look under `/usr/include/adios2/ ` for public adios2 headers. I have never installed adios2 under /usr....can you verify it adds the `adios2` prefix?

williamfgc:
`ls /usr/include` shows there are many packages doing the above (adding a prefix). I can only speak for my Ubuntu box, haven't used other Linux distros, though.

germasch:
Traditionally, libraries would install their header(s) directly into `/usr/include`, or `/usr/local/include`, basically, `${CMAKE_INSTALL_PREFIX}/include`. That's fine for libraries with one or a few headers, but it clutters thing up and potentially conflicts with other libraries' headers. For example, hdf5 puts `hdf5.h`, `hdf5_hl.h` but also a whole bunch of `H5*.h` headers directly into the `include/` directory. It'd be nicer of those headers were in a subdirectory, but at least by means of the `H5` prefix it's fairly clear that they belong to hdf5 and aren't likely to conflict.

[...]

PR #1515, opened 2019-06-18
Move Fortran modules from include/fortran to include/adios2/fortran

KyleFromKitware:
Fixes: #1477


chuckatkins:
@williamfgc I believe this is fine, but it would be helpful if you could double check that it works still for an external fortran code.

williamfgc:
@chuckatkins sounds good to me!

PR #1517, opened 2019-06-18
Update third-party libraries

KyleFromKitware:
...and disable header and pkgconfig installation


thirdparty/ffs/ffs/ffs/tests/adios2_bug.c
```diff
@@ -5,7 +5,6 @@
 #include <string.h>
 #include <sys/stat.h>
 
-#include "adios2/common/ADIOSConfig.h"
```
KyleFromKitware: Looks like I made a mistake here yesterday... I updated this file in ADIOS even though it's a third-party library. This line created a merge conflict when I updated from upstream. Removing the line should be the correct course of action to resolve the conflict.

chuckatkins: You can also `rm -rf thirdparty/ffs/ffs`, commit, and then re-run the update script and it will pull a fresh copy ignoring the in-tree history.

KyleFromKitware: Done

eisenhauer:
Kyle,
I think the failures you're getting may be due to a latent bug in FFS that I've yet to diagnose and repair.  ADIOS2 is the only thing I've got that triggers it, and it seems to be related to FFS' use of dynamic code generation to speed data transformations.  It has turned out to be hard to track down, so as a temporary patch I had disabled dynamic code generation in FFS in ADIOS2.   However, this PR drops that patch.  Probably the best approach is to re-enable it until I can sort things out.  The patch was to change the default for FFS_CONVERSION_GENERATION_DEFAULT in ffs/CMakeList.txt to 0:

if(FFS_USE_DILL)
....
  set(FFS_CONVERSION_GENERATION_DEFAULT ${DILL_HAS_NATIVE_DCG})
  set(FFS_CONVERSION_GENERATION_DEFAULT 0)       <<<<<< This Commit dropped this line

endif()

chuckatkins:
@eisenhauer How about we just make it a configurable option upstream instead?

```cmake
index ed21732..10e1bde 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -172,7 +172,8 @@ option(FFS_USE_DILL "Enable Dill code generation" ON)
 if(FFS_USE_DILL)
   find_package(dill 2.3.1 REQUIRED)
   set(HAVE_DILL_H ON)
-  set(FFS_CONVERSION_GENERATION_DEFAULT ${DILL_HAS_NATIVE_DCG})
+  option(FFS_CONVERSION_GENERATION_DEFAULT ${DILL_HAS_NATIVE_DCG})
+  mark_as_advanced(FFS_CONVERSION_GENERATION_DEFAULT)
   set(DO_DCG TRUE)
   target_link_libraries(ffs PRIVATE dill::dill)
   list(APPEND _pkg_config_private_reqs "dill >= ${dill_VERSION}")
```

This way it preserves the existing default behavior but lets you override it via `-DFFS_CONVERSION_GENERATION_DEFAULT=OFF`.  That then also allows us to set the option in `thirdparty/ffs/CMakeLists.txt` when we build it.

KyleFromKitware:
> How about we just make it a configurable option upstream instead?

Yes. That is the correct solution. Working on it now.

PR #1519, opened 2019-06-19
Remove LockWriterDefinitions/LockReaderSelections from IO

keichi:
I was seeing weird link-time errors because of this. I guess these are remnants from a previous design. Seems to have been accidentally added in #1478.


eisenhauer:
Hi Keichi.  Yes, these were an artifact of an earlier version of that PR that got left in unnoticed.  Thanks!

PR #1520, opened 2019-06-19
Make the DataInterface engine parameter work in SST RDMA DP

keichi:
@eisenhauer @philip-davis This completes the DataInterface parameter support in RDMA DP that was partially implemented in #1516.


eisenhauer:
Thanks again!

Issue #1518, opened 2019-06-19
Installed header files are broken


keichi:
Installed header files are broken since #1510 was merged:

```
mpic++ -g -std=gnu++11 -O2 -c `/Users/keichi/Projects/research/adios2-install/bin/adios2-config --cxx-flags` -o simulation/IO_adios2.o simulation/IO_adios2.cpp
In file included from simulation/IO_adios2.cpp:16:
/Users/keichi/Projects/research/adios2-install/include/adios2.h:18:10: fatal error: 'cxx11/ADIOS.h' file not found
#include "cxx11/ADIOS.h"
         ^~~~~~~~~~~~~~~
1 error generated.
make: *** [simulation/IO_adios2.o] Error 1
```

Same with CMake builds:
```
In file included from /Users/keichi/Projects/research/adiosvm/Tutorial/gray-scott/plot/render_isosurface.cpp:12:
/Users/keichi/Projects/research/adios2-install/include/adios2.h:18:10: fatal error: 'cxx11/ADIOS.h' file not found
#include "cxx11/ADIOS.h"
         ^~~~~~~~~~~~~~~
1 error generated.
make[2]: *** [CMakeFiles/render_isosurface.dir/plot/render_isosurface.cpp.o] Error 1
make[1]: *** [CMakeFiles/render_isosurface.dir/all] Error 2
make: *** [all] Error 2
```

And C bindings:
```
In file included from test.c:1:
/Users/keichi/Projects/research/adios2-install/include/adios2_c.h:16:10: fatal error: 'c/adios2_c_adios.h' file not found
#include "c/adios2_c_adios.h"
         ^~~~~~~~~~~~~~~~~~~~
1 error generated.
```

keichi:
Confirmed the fix, thanks.

PR #1522, opened 2019-06-19
Fix headers that were broken by #1510

KyleFromKitware:
Fixes: #1518


williamfgc:
@KyleFromKitware is this working after a `make install`?

KyleFromKitware:
@williamfgc Yes

Issue #1166, opened 2019-02-12
add tests after doing a make install on CI


williamfgc:
Not sure if possible, but maybe it is time for external tests. Current master is broken after `make install` as we are not testing outside local ctest.

keichi:
This is just an idea, but maybe we can run CI in [pnorbert/adiosvm](https://github.com/pnorbert/adiosvm) too? This CI would pull the master branch of ADIOS from this repository, build/install ADIOS, and then build the tutorial apps. We can be sure that the tutorial apps are compatible with the latest ADIOS.

williamfgc:
@keichi we have a nightly repo that @chuckatkins kickstarted. We haven't added much, but we should, please see: https://github.com/ornladios/ADIOS2-Testing/tree/master/external/gray-scott . Regression tests would certainly be helpful.

williamfgc:
@keichi  [adios2 cdash](https://open.cdash.org/index.php?project=ADIOS&date=2019-06-19) shows that the master was in fact broken.

keichi:
@williamfgc Thanks, I didn't know this repository existed. It's exactly what I meant. I guess we can add the endianness interoperability tests here in the future?

chuckatkins:
Fixed by #1523

PR #1523, opened 2019-06-19
Add Install tests

KyleFromKitware:
Fixes: #1166


testing/adios2/install/C/main.c
```diff
@@ -0,0 +1,11 @@
+/*
+ * Distributed under the OSI-approved Apache License, Version 2.0.  See
+ * accompanying file Copyright.txt for details.
+ */
+
+#include <adios2_c.h>
+
+int main(void)
+{
+  return 0;
```
chuckatkins: Let's go ahead and add a symbol to validate linking.  It doesn't need to even be working code, so long as it compiles, links, and doesn't segfault.
```C
int main(void)
{
  adios2_finalize(NULL);
  return 0;
}
```
@williamfgc would this snippet execute without segfaulting?

testing/adios2/install/CXX11/CMakeLists.txt
```diff
@@ -0,0 +1,15 @@
+#------------------------------------------------------------------------------#
+# Distributed under the OSI-approved Apache License, Version 2.0.  See
+# accompanying file Copyright.txt for details.
+#------------------------------------------------------------------------------#
+
+cmake_minimum_required(VERSION 3.6)
+project(adios_c_test CXX)
```
chuckatkins: Should `project(adios_c_test CXX)` be `project(adios_cxx11_test CXX)`?

testing/adios2/install/CXX11/main.cxx
```diff
@@ -0,0 +1,11 @@
+/*
+ * Distributed under the OSI-approved Apache License, Version 2.0.  See
+ * accompanying file Copyright.txt for details.
+ */
+
+#include <adios2.h>
+
+int main(void)
+{
```
chuckatkins: Same as with C:
```cxx
int main(int, char **)
{
  adios2::ADIOS adios;
  return 0;
}
```

testing/adios2/install/CXX11/CMakeLists.txt
KyleFromKitware: Indeed, good catch

KyleFromKitware:
> Would it be easier to have the test use ExternalProject? That way you can easily add the manual makefile tests as well and do it in the same way as the cmake based test?

`ctest --build-and-test` has the ability to build a project with your own configure and build steps (or with no configure step.) `ExternalProject` would build the test project at build time, and I would rather build it at test time.

PR #1526, opened 2019-06-20
Bp4 buffer tags

pnorbert:
Add tags to data buffer for easier error checking for the (future) recovery tool. 
Remove unnecessary duplication of dimensions in the metadata (in the data buffer).
Pad the metadata so that the actual data always falls on max_align_t size (in the memory, not in the file).


pnorbert:
Removing this PR for now but still hope to get it into the release

PR #1541, opened 2019-06-24
adios_iotest return code on error

kshitij-v-mehta:
adios_iotest now returns an error code instead of success when an error is encoutered during parsing of cmd line args

### DIFF
--- a/source/utils/adios_iotest/adios_iotest.cpp
+++ b/source/utils/adios_iotest/adios_iotest.cpp
@@ -21,288 +21,297 @@ int main(int argc, char *argv[])
     MPI_Init(&argc, &argv);
 
     Settings settings;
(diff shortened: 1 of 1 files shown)


chuckatkins:
> This is not returning an error code but aborting the executable.

@pnorbert  `MPI_Abort` still returns an error code, it just makes sure all the other MPI processes terminate as well.  I'm okay with having this, but I'll leave it to @pnorbert's discretion.  That being said, the return codes should be different for each error case.  So rather that exiting with `-1` for all errors, I'd rather see it exit with 1, 2, 3, etc. depending on the error.  So either way, change the exit codes passed to `MPI_Abort` or change the exit codes in the `return` statements.

williamfgc:
superseeded by #1585

PR #1543, opened 2019-06-25
Adding dynamic shape to adios_reorganize

williamfgc:


source/adios2/core/VariableBase.cpp
```diff
@@ -17,6 +17,7 @@
 /// \endcond
 
 #include "adios2/core/Engine.h"
+#include "adios2/core/Variable.h"
```
chuckatkins: Parent including the child?

williamfgc: Unfortunately, this is the only way to use `core::VariableBase` with typed functions in utils. I don't see any other way to downcast in member functions, unless we provide "sliced" versions explicitly. This can be reworked later, but it's fixing a current bug in reorganize before the release. It's just timing.

williamfgc: I forgot to mention, this is a request from an ECP project.

PR #1545, opened 2019-06-25
fixed a few problems in ssc tests

JasonRuonanWang:
@chuckatkins Could you help check the cmake stuff and see if there is any smarter ways of doing this? Basically what I needed here is that I need to delete the .ssc file that is generated from previous test runs before running another time. However, cmake does not allow me to run multiple commands in adding a single test. This is a workaround I found on stackoverflow, but it looks quite ugly. Thanks.

### DIFF
--- a/testing/adios2/engine/ssc/CMakeLists.txt
+++ b/testing/adios2/engine/ssc/CMakeLists.txt
@@ -5,26 +5,46 @@
 
 
 if(ADIOS2_HAVE_MPI AND ADIOS2_HAVE_ZeroMQ)
-    add_executable(SscTest SscTest.cpp)
-    target_link_libraries(SscTest adios2 gtest MPI::MPI_C)
-    add_test(NAME SscTest COMMAND "mpirun" "-n" "8" $<TARGET_FILE:SscTest>)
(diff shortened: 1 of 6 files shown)


eisenhauer:
@JasonRuonanWang, it doesn't help here per se, but the solution I've ended up with for cleanup in the staging-common area is to create a directory for each test, cd into it to run and delete it afterwards.  Saves the test framework from having to make naming assumptions about engine contact files, etc.  But I've got run_test.py that has a lot more flexibility that what you get trying to do the same thing in CMake...

JasonRuonanWang:
@eisenhauer That's a great idea. I don't actually understand whether these commands are effective at cmake time or ctest time. But somehow adding these extra rm commands solved my problem. I don't know if it's solved by chance...

eisenhauer:
Well, I think I promised you better docs on how to add tests to staging-common, and I have delivered yet.  I'll try to do that soon.

JasonRuonanWang:
@eisenhauer No rush. I am thinking if we can have a face to face meeting to discuss a series of things including the common-staging tests. Maybe I can come down to Georgia Tech some time.

PR #1547, opened 2019-06-26
Local array read block bug

williamfgc:
Updating count in BP3 and BP4 deserializer
Added test

### DIFF
--- a/testing/adios2/engine/bp/TestBPWriteReadLocalVariables.cpp
+++ b/testing/adios2/engine/bp/TestBPWriteReadLocalVariables.cpp
@@ -1771,7 +1771,7 @@ TEST_F(BPWriteReadLocalVariables, ADIOS2BPWriteReadLocal2DChangeCount)
     const size_t Nx1 = 5;
     // data for block 0 and block 1 per rank
     std::vector<float> data(Nx0 * Nx1);
-    const int32_t startBlock = 3 * mpiRank + 1;
+    const float startBlock = static_cast<float>(3 * mpiRank + 1);
 
     std::iota(data.begin(), data.end(), startBlock);
 
--- a/testing/adios2/engine/bp/TestBPWriteReadLocalVariables.cpp
+++ b/testing/adios2/engine/bp/TestBPWriteReadLocalVariables.cpp
@@ -1792,7 +1792,9 @@ TEST_F(BPWriteReadLocalVariables, ADIOS2BPWriteReadLocal2DChangeCount)
         bpWriter.Put(var_r32, data.data());
         bpWriter.Close();
     }
-
+#ifdef ADIOS2_HAVE_MPI
+    MPI_Barrier(MPI_COMM_WORLD);
+#endif
     {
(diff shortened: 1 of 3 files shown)


testing/adios2/engine/bp/TestBPWriteReadLocalVariables.cpp
```diff
@@ -1754,6 +1754,76 @@ TEST_F(BPWriteReadLocalVariables, ADIOS2BPWriteReadLocal1DSubFile)
...
+        adios2::IO io = adios.DeclareIO("ReaderIO");
+#ifdef ADIOS2_HAVE_MPI
+        adios2::Engine bpReader =
+            io.Open(fname, adios2::Mode::Read, MPI_COMM_SELF);
+#else
+        adios2::Engine bpReader = io.Open(fname, adios2::Mode::Read);
+#endif
+
+        auto var_r32 = io.InquireVariable<float>("r32");
+        std::map<size_t, std::vector<float>> dataIn;
```
chuckatkins: Rather than having to store the data in a map and loop through it after you could do the validation inline for a more compact test:
```c++
std::vector<float> dataIn;
for (size_t b = 0; b < static_cast<size_t>(mpiSize); ++b)
{
    dataIn.clear();
    var_r32.SetBlockSelection(b);
    bpReader.Get(var_r32, dataIn);
    bpReader.PerformGets();

    for (size_t i = 0; i < dataIn[b].size(); ++i)
    {
        EXPECT_EQ(dataIn[i], static_cast<float>(3 * b + 1 + i));
    }
}
```

williamfgc: @chuckatkins true, my goal is to reproduce the conditions that triggered this bug, several Gets in deferred mode (for different block sizes going up and down) over several pieces of data. Thanks for checking.

chuckatkins: Got it.  I often forget what the default mode is, sync or deferred, and I keep thinking it's sync, which is not the case.  Given that, an `std::vector<std::vector<float>>` should be more appropriate than an `std::map<size_t, std::vector<float>>` since your "key" to the map is a zero-based consecutive monotonically increasing integer.

williamfgc: Yeah, that's also from the conditions. The call to Get is creating a new map entry.

chuckatkins: It's the call to `dataIn[b]` that's creating the map entry.  My point is just thaty the map is an unnecessary overhead since you don't need arbitrary comparison functions on key objects since your key range is consecutive integers `[0:mpiSize]`, thus using instead `std::vector<std::vector<float>> dataIn(mpiSize);`

williamfgc: the call to `dataIn[b]` in `Get, which makes it a perfectly valid use case regardless of ordering. I don't think we should spend time on this, should we?

PR #1548, opened 2019-06-26
Use CTest fixtures for parallel test ordering

chuckatkins:


KyleFromKitware:
ADIOS2 has `cmake_minimum_required(VERSION 3.6)`, which does not support test fixtures. We need to have the `DEPENDS` property as well.

chuckatkins:
This is not particularly important so I'll delay further work until after the release.

PR #1554, opened 2019-06-27
bpls: fix some warnings for size_t printf arguments

germasch:
`%PRIu64` is for `uint64_t`, `%zu` is for `size_t`, which on the Mac (and 32-bit
archs) aren't aliases.

[This tiny patch isn't likely to break anything since `%zu` is already used elsewhere in `bpls.cpp`, so I'd consider it safe for the release, though I think these kind of last minute changes are generally better avoided. Anyway, it's certainly not a critical fix, so it could just go into `master` if that's preferred.]

Mentioning @pnorbert .


eisenhauer:
Thank you.  I've been annoyed by these warnings, but not quite annoyed enough to try to fix them...

germasch:
Going to reopen as new PR to see whether I can get the first half of the CI to trigger...

PR #1555, opened 2019-06-27
fix Install test adios_c_test w/MPI

germasch:
Not sure how that passed the CI before, maybe it got exposed by my
PR that switches no-mpi mode per communicator.

Actually, that's probably the explanation, this passed the CI before my last update to the no-mpi unification thing, and the combination of merging those two then failed. But I'd argue it's clearly a bug in the test, fixed here.

### DIFF
--- a/testing/adios2/install/C/main_mpi.c
+++ b/testing/adios2/install/C/main_mpi.c
@@ -7,8 +7,9 @@
 
 #include <stdio.h>
 
-int main(void)
+int main(int argc, char **argv)
 {
+    MPI_Init(&argc, &argv);
     adios2_adios *adios = adios2_init(MPI_COMM_WORLD, adios2_debug_mode_on);
     if (!adios)
     {
@@ -16,5 +17,6 @@ int main(void)
         return 1;
     }
     adios2_finalize(adios);
+    MPI_Finalize();
     return 0;
 }


germasch:
Actually, I meant to make another comment @chuckatkins : When I first tried to reproduce the CI failure with `ctest -VV -R Install.CMake`, it failed because the `Install.Setup` hadn't run, so I had to run that first. I'm not sure that's intended behavior, but I saw you working on this, so I'm mentioning it.

chuckatkins:
@germasch There's other issues besides just MPI_Init with some of the other install tests failing.  @KyleFromKitware is working through them now though.  In the mean time I'm going to disable them so we can get "unstuck".

KyleFromKitware:
@germasch The `Install.Setup` test is indeed supposed to run before the others. The rest of the `Install` tests list `Install.Setup` in their `DEPENDS` properties. I thought about using CTest's [test fixtures](https://cmake.org/cmake/help/v3.15/prop_test/FIXTURES_REQUIRED.html) feature, but CMake 3.6 does not have it... though, now that I look at it, it looks like it can be done in a backwards-compatible way without affecting 3.6 builds.

KyleFromKitware:
@germasch I have just opened #1557, which will fix the MPI issue in addition to another issue which I am still investigating.

chuckatkins:
Superseded by #1557

PR #1557, opened 2019-06-27
Fix broken Install tests

### LINKED ISSUE #1550: Install tests breaks with mpi
I'm not sure why this didn't catch in the CI, but it seems like some of the install tests are failing from not calling MPI_Init inside of the MPI based tests: https://open.cdash.org/testDetails.php?test=778542462&build=5978118

KyleFromKitware:
Fixes #1550


chuckatkins:
Fixes #1557

testing/adios2/install/C/main_mpi.c
```diff
@@ -5,16 +5,28 @@
...
 #include <stdio.h>
 
-int main(void)
+int main(int argc, char **argv)
 {
+    int rank;
+    int size;
+    MPI_Init(&argc, &argv);
+    MPI_Comm_rank(MPI_COMM_WORLD, &rank);
+    MPI_Comm_size(MPI_COMM_WORLD, &size);
```
germasch: Minor comment: no need to get `rank` and `size` if you're not using them.

testing/adios2/install/C/main_mpi.c
```diff
@@ -5,16 +5,28 @@
...
+
     adios2_adios *adios = adios2_init(MPI_COMM_WORLD, adios2_debug_mode_on);
     if (!adios)
     {
         fprintf(stderr, "adios2_init() failed\n");
+        MPI_Finalize();
         return 1;
```
germasch: As a more general comment, as this has been happening in other places, too: There are caveats with calling `MPI_Finalize` on error conditions.

Specifically, if one exits from an MPI program with calling `MPI_Finalize`, at least OpenMPI will recognize this as an error and abort other running processes. But calling `MPI_Finalize` and then exiting with an error code is a regular exit to that process, and other processes will continue running. It's not likely ever a problem here, since nothing much happens other than finalizing things anyway.

But more generally, I think it's not a good habit to get into. If the error occurs only on a subset of all MPI process, the rest of the process will keep running, not knowing that something went wrong. Which makes it quite possible that those processes end up hanging, and rather unlikely they'll behave as intended. So I think as a general rule, one should only call `MPI_Finalize` if it is clear that it will be called collectively (ie, on all processes). But with most errors, it isn't really clear ahead of time (when writing the code) why / when they might get triggered, and so one cannot guarantee that `MPI_Finalize` is called collectively, and hence `MPI_Abort()` is the better choice.

Actually, I just googled, and MPI standard does in fact clarify that `MPI_Finalize` is collective on `MPI_COMM_WORLD`.

germasch:
@KyleFromKitware 

In terms of the error you're probably debugging right now (the CI failure), I think I've seen that before, and I think it's due to not running the test with `mpiexec`:

Running `ctest -VV` on my macbook:
```
560: 1: Test command:/Users/kai/build/ADIOS2/build/Testing/adios2/install/cmake/C/adios_c_test
```

With most installations, MPI programs can just be run directly, but sometimes OpenMPI needs help from `mpiexec` to find its dynamic libs.

PR #1559, opened 2019-06-27
Fix Sst Docs WRT behaviour in no reader case

eisenhauer:


pnorbert:
This should go into the release branch

eisenhauer:
Ugh, yes, it should.

PR #1563, opened 2019-06-28
Test HDF5 with simple tests

eisenhauer:


testing/adios2/engine/staging-common/CMakeLists.txt
```diff
@@ -202,13 +202,32 @@ endif()
...
+#
+if (NOT MSVC)    # not on windows
+    set (HDF5_TESTS ${ALL_SIMPLE_TESTS})
+    # Delayed reader not worth testing on file engines
+    list (FILTER HDF5_TESTS EXCLUDE REGEX "DelayedReader")
+    # Local variables don't work on hdf5 
+    list (FILTER HDF5_TESTS EXCLUDE REGEX ".*Local")
+    # Oddly, some multi-writer tests fail
+    list (FILTER HDF5_TESTS EXCLUDE REGEX "[2345]x")
+    foreach(test ${HDF5_TESTS})
```
chuckatkins: Just for future reference, i'm trying to move us to using the newer syntax of `foreach`:
```cmake
# Old
foreach(var ${var_list_name})
  #do things
endforeach()

# New
foreach(far IN LISTS var_list_name)
  # do stuff
endforeach()
```

It's not worth updating this pr for but it's generally good cmake practice to use the newer syntax.

eisenhauer: I will take note.  (I didn't know there was a newer syntax...)

chuckatkins: In general it's one of the many changes made over the years to try to work on variables directly without having to dereference them.  The newer syntax addresses issues that arise when items in the list might have spaces in them and end up evaluated as separate items.

testing/adios2/engine/staging-common/CMakeLists.txt
```diff
@@ -202,13 +202,32 @@ endif()
...
+        add_common_test(${test} HDF5)
+    endforeach()
+endif(NOT MSVC)
```
chuckatkins: @eisenhauer
```cmake
if (NOT MSVC AND ADIOS2_HAVE_HDF5)
...
endif(NOT MSVC)
```
is always generating a configure warning now from the mismatched arguments to `if` and `endif`.  Please push an update that just removes any arguments to `endif`.  That should eliminate the configure warnings.

PR #1564, opened 2019-06-28
bpls: fix some warnings for size_t printf arguments

germasch:
I'm just redoing #1554, since the CI kept on hanging on that one.

### DIFF
--- a/source/utils/bpls/bpls.cpp
+++ b/source/utils/bpls/bpls.cpp
@@ -1530,12 +1530,12 @@ int readVar(core::Engine *fp, core::IO *io, core::Variable<T> *variable)
     auto shape = variable->Shape(absstep);
     if (verbose > 2)
     {
-        printf("    starting step=%" PRIu64 " absolute step=%" PRIu64
+        printf("    starting step=%" PRIu64 " absolute step=%zu"
                ", dims={",
                stepStart, absstep);
         for (auto dim : shape)
         {
-            printf("%" PRIu64 " ", dim);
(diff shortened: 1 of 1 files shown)


germasch:
I'm not sure what's going on with the CI, though it seems it's working for others, but I've been having the same problem of the checks hanging despite multiple force pushes and moving this into a new PR. I can see on Circle CI that it passed, and  I see a green checkmark next to the commit, but github's status doesn't seem to get it.

At least this is only fixing some warnings (well, bugs with a 32-bit `size_t`, but who has those these days?).

germasch:
Tried adding `[skip ci]` as @chuckatkins was mentioning elsewhere (not that it's really appropriate here), but seems no luck with that either. As I said before, this actually does pass CircleCI (https://circleci.com/gh/germasch/ADIOS2/tree/pr%2Fbpls_warnings), but it won't register correctly with github :(

chuckatkins:
After we get through the release I'm going to look at some other CI options that can get us some better throughput.  The backlog that builds up from the 4 parallel builds that circle limits us to is really becoming a problem.

germasch:
Thanks @chuckatkins . The problem I'm having isn't throughput/the queue limit, though, it is that even after all the jobs go through circleci and pass, they still don't report back to github. You should be able to see that they passed here in ~30 minutes: https://circleci.com/gh/germasch/ADIOS2/tree/pr%2Fbpls_warnings.  

This problem seems specific to me, though, so hopefully it'll just resolve itself eventually...

PR #1567, opened 2019-06-28
Bp4 doclose fix

lwan86:
fixed the issue that data is flushed twice if deferred variables are not empty when calling DoClose() function and a potential rank 0 early return issue in WriteCollectiveMetadataFile() function


chuckatkins:
Replaced by #1568

PR #1568, opened 2019-06-28
Bp4 doclose fix

lwan86:
fixed the issue that data is flushed twice if deferred variables are not empty when calling DoClose() function and a potential rank 0 early return issue in WriteCollectiveMetadataFile() function


chuckatkins:
@williamfgc or @pnorbert should this fix be in the release?

pnorbert:
Oh, damn, this is the second PR and should have gone into the release. I did not even look to check where it goes. Sorry about that.

pnorbert:
I mean second PR for this fix because I closed the first that targeted master.

PR #1572, opened 2019-06-28
Report the availability of RDMA for Staging in CMake summary

eisenhauer:
@philip-davis and @pnorbert .  Just something that came up in the conversation with the BNL guys.  We don’t currently have any obvious output in the CMake “final summary” that we have or have not found the necessary libraries for the RDMA data plane to be active.  Obviously, we won’t know for sure that they can be used until run-time, but assuming that RDMA is a performance-critical build option, it should probably be more obvious that folks have successfully built with it or screwed something up.    That patch simply adds some notice of that, currently placed after all the "Build Options", and simply reporting it "Available" or "Unconfigured".    Norbert, if you concur that this is a good idea, you can just approve it.  Philip, would you please verify on some handy RDMA-enabled platform that this actually does the right thing (I.E. says "Available") and approve if it does?  (I'm leveraging Github instead of Email here.)


eisenhauer:
Actually, @philip-davis , I was able to jump on summitdev where I had libfabric built and tested this.  Seems to work as expected, so no need to verify.

Issue #1553, opened 2019-06-27
row/column major considerations


germasch:
As @pnorbert seems to be working on BP4s on-disk representation (and dumping it), here's a wishlist item that I think would be worth considering:

Currently, ADIOS2 makes the decisions based on row vs col-major layout based on which API is invoked (C++/C: row-major, Fortran: col-major). There are some considerations with that, though. E.g., some (many?) projects switching parts of their code/kernels to C/C++/CUDA, and as that transition continues, they may want to be able to write data in a single app from both languages. e.g., they might want to write Fortran-layout data using the C++ API.

Also, C++'s data layout isn't natively row-major, I'd argue, it's mostly natively non-existant (except for compile-time fixed dimension arrays). Many of the libraries that do multi-d arrays (boost, xtensor, Kokkos) support both layouts, as does Python. So even in a native C++ application, one might use and want to write col-major data. Some apps may be using a mix of row-major and col-major layouts for performance reasons.

So I think there's some work that could be done on the API side, that lets the application specify the data layout they want to use for a given variable. But that's something that can be added over time relatively easily. If supporting mixed layouts is an eventual goal, though, the output format needs to be able to handle it, too, though. I believe BP3 keeps a global layout flag (well, maybe per PG or something), so I think it may not be able to handle mixed layouts well without some kind of compatibility issues, and that's fine. But for BP4, it might be useful to consider this case now, while the binary format is still in flux. It'd probably be easy to add a row-major / col-major flag per Variable. 

[One could also consider a more flexible approach that allows for a general strided approach, e.g., for the case that some array dimensions used in an application has been padded for alignment.]

williamfgc:
@germasch thanks for bringing this up. In adios their no real concept of row-major/column-major as it's all indexed contiguous memory. We reuse these familiar concepts to add another degree of freedom to the `Shape`, `Start`, `Count` per Variable dimension. Basically, adios needs to understand which one is the fastest and the slowest dimension indices.

ADIOS has today two requirements:

- We assume the ordering of dimensions is natural to the language in use and the user is responsible for swapping those accordingly (which somehow they have to way before they use ADIOS2, if let's say going from Fortran to CUDA, see bottom of this message). 

- `bpls` being a C++ reader always presents the data in "row-major" order, basically the last dimension is the fastest/contiguous one, regardless.

[...]

PR #1573, opened 2019-06-29
Exposes IsReverseDims in BlockInfo

williamfgc:
related to request in #1553 from @berkgeveci

### DIFF
--- a/bindings/CXX11/adios2/cxx11/Variable.h
+++ b/bindings/CXX11/adios2/cxx11/Variable.h
@@ -348,6 +348,8 @@ public:
         size_t Step = 0;
         /** reference to internal block data (used by inline Engine) */
         const T *Data() const;
+        /** true: Dims were swapped from column-major, false: not swapped */
+        bool IsReverseDims = false;
 
         // allow Engine to set m_Info
         friend class Engine;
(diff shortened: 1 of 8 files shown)


bindings/CXX11/adios2/cxx11/Variable.h
```diff
@@ -348,6 +348,8 @@ class Variable
         size_t Step = 0;
         /** reference to internal block data (used by inline Engine) */
         const T *Data() const;
+        /** true: Dims were swapped from column-major, false: not swapped */
```
germasch: minor comment: dims may also have been swapped from row-major to col-major, when reading from Fortran.

williamfgc: @germasch this particular struct is only exposed in the C++11 bindings

williamfgc:
@germasch see my comment in #1553, it's not unreasonable to think we might need this info on a per block basis. The block inspection functions in C++ were designed with in-situ viz as one of the use-cases. It's very much needed.

chuckatkins:
@williamfgc please edit the PR to target `master` instead.  If we want to put out a patch release soon after with this then we can.  For now though, given the short time frame, let's keep it out of the release.

williamfgc:
@chuckatkins fine with me

williamfgc:
request in #1553

PR #1577, opened 2019-06-29
fix HDF5Reader build with MSVC

germasch:
Sadly, variable-sized stack arrays are only a feature of C99, but not C++11,
so this patch uses std::vector instead.

This fixes #1485, the second issue there, at least hopefully so -- @ax3l, could you test this one, too, as I don't think the adios2 CI covers this case. (I changed the place identified by the error message from your report, and one more place with the same issue, but there may be more).

@chuckatkins, I haven't checked, but my guess is that the MSVC CI doesn't build ADIOS2 with HDF5, otherwise I would have expected to see this problem earlier. So that may be something for the future todo list.

PR #1575, opened 2019-06-29
fix MSVC/conda build

germasch:
This simply renames `helper::CopyMemory` to `helper::CopyMemoryBlock` to
avoid a name clash with the `CopyMemory` defined in the MSVC build in #1485.

@ax3l, it'd be great if you could check whether this fixes your problem, since it apparently doesn't happen with the MSVC builds that part of the ADIOS2 CI.

PR #1581, opened 2019-06-29
Germasch msvc fixes

pnorbert:
resubmit #1575 and #1577


pnorbert:
I did not check @ax3l's message. Withdrawing this PR

Issue #1485, opened 2019-06-09
Compile Error with MSVC 19.04


ax3l:
Compiler: MSVC 19.0.24241.7
ADIOS2: v2.4.1-rc1

Compiling on Windows with MSVC and CMake+NMake results in an error https://ci.appveyor.com/project/conda-forge/staged-recipes/builds/25153984 of the form
```
%SRC_DIR%\source\adios2/helper/adiosMemory.h(121): warning C4002: too many actual parameters for macro 'RtlCopyMemory'
%SRC_DIR%\source\adios2/helper/adiosMemory.h(114): error C2182: 'memcpy': illegal use of type 'void'
%SRC_DIR%\source\adios2/helper/adiosMemory.h(114): error C2988: unrecognizable template declaration/definition
%SRC_DIR%\source\adios2/helper/adiosMemory.h(114): error C2059: syntax error: 'noexcept'
c:\bld\adios2_1560110677887\work\source\adios2\helper\adiosMemory.inl(260): warning C4002: too many actual parameters for macro 'RtlCopyMemory'
c:\bld\adios2_1560110677887\work\source\adios2\helper\adiosMemory.inl(255): error C2146: syntax error: missing ')' before identifier 'dest'
c:\bld\adios2_1560110677887\work\source\adios2\helper\adiosMemory.inl(255): error C2065: 'dest': undeclared identifier
c:\bld\adios2_1560110677887\work\source\adios2\helper\adiosMemory.inl(255): error C2182: 'memcpy': illegal use of type 'void'
c:\bld\adios2_1560110677887\work\source\adios2\helper\adiosMemory.inl(255): error C2374: 'adios2::helper::memcpy': redefinition; multiple initialization
%SRC_DIR%\source\adios2/helper/adiosMemory.h(114): note: see declaration of 'adios2::helper::memcpy'
c:\bld\adios2_1560110677887\work\source\adios2\helper\adiosMemory.inl(255): error C2988: unrecognizable template declaration/definition
c:\bld\adios2_1560110677887\work\source\adios2\helper\adiosMemory.inl(255): error C2059: syntax error: ','
c:\bld\adios2_1560110677887\work\source\adios2\helper\adiosMemory.inl(255): error C2059: syntax error: ')'
c:\bld\adios2_1560110677887\work\source\adios2\helper\adiosMemory.inl(287): error C2143: syntax error: missing ';' before '{'
[...]

PR #1587, opened 2019-06-30
fix #1577 Pr/fix hdf5 msvc

williamfgc:


chuckatkins:
Fixes #1485

chuckatkins:
@williamfgc why both of these?

williamfgc:
#1485 was stuck, it was just merged

PR #1593, opened 2019-07-01
Python: Fix Windows Output Dir Clash

ax3l:
Avoid a naming conflict when building the python bindings with NMake and MSVC of the form
```
[100%] Linking CXX shared module ..\..\lib\python3.6\site-packages\adios2.cp36-win_amd64.pyd
LINK: command "C:\PROGRA~2\MI0E91~1.0\VC\bin\amd64\link.exe /nologo @CMakeFiles\adios2py.dir\objects1.rsp /out:..\..\lib\python3.6\site-packages\adios2.cp36-win_amd64.pyd /implib:..\..\lib\adios2.lib /pdb:%SRC_DIR%\build\lib\python3.6\site-packages\adios2.pdb /dll /version:0.0 /machine:x64 /INCREMENTAL:NO %PREFIX%\libs\python36.lib -LTCG ..\..\lib\adios2.lib kernel32.lib user32.lib gdi32.lib winspool.lib shell32.lib ole32.lib oleaut32.lib uuid.lib comdlg32.lib advapi32.lib /MANIFEST /MANIFESTFILE:..\..\lib\python3.6\site-packages\adios2.cp36-win_amd64.pyd.manifest" failed (exit code 1149) with the following output:
LINK : fatal error LNK1149: output filename matches input filename '%SRC_DIR%\build\lib\adios2.lib'
NMAKE : fatal error U1077: 'C:\bld\adios2_1561968287896\_build_env\Library\bin\cmake.exe' : return code '0xffffffff'
```

Also fix the proper python installation directory `CMAKE_INSTALL_PYTHONDIR` on windows and make it overwriteable from the command line.


bindings/Python/CMakeLists.txt
```diff
@@ -13,16 +13,29 @@ if(ADIOS2_HAVE_MPI)
...
 )
+set(CMAKE_PYTHON_OUTPUT_DIRECTORY
+  "${ADIOS2_BINARY_DIR}/${CMAKE_INSTALL_PYTHONDIR}"
+  CACHE PATH "Build directory for python modules"
+)
 set_target_properties(adios2py PROPERTIES
   OUTPUT_NAME adios2
-  LIBRARY_OUTPUT_DIRECTORY ${ADIOS2_BINARY_DIR}/${CMAKE_INSTALL_PYTHONDIR}
-  RUNTIME_OUTPUT_DIRECTORY ${ADIOS2_BINARY_DIR}/${CMAKE_INSTALL_PYTHONDIR}
+  ARCHIVE_OUTPUT_DIRECTORY ${CMAKE_PYTHON_OUTPUT_DIRECTORY}
```
ax3l: funnily, this line is needed although one always builds a shared lib for python modules. well.

bindings/Python/CMakeLists.txt
```diff
@@ -13,16 +13,29 @@ if(ADIOS2_HAVE_MPI)
...
+  set(CMAKE_INSTALL_PYTHONDIR_DEFAULT
+    "${CMAKE_INSTALL_LIBDIR}/python${PYTHON_VERSION_MAJOR}.${PYTHON_VERSION_MINOR}/site-packages"
+  )
+endif()
+set(CMAKE_INSTALL_PYTHONDIR "${CMAKE_INSTALL_PYTHONDIR_DEFAULT}"
+  CACHE PATH "Install directory for python modules"
 )
-set(CMAKE_INSTALL_PYTHONDIR "${CMAKE_INSTALL_PYTHONDIR}"
-  CACHE INTERNAL "" FORCE
+set(CMAKE_PYTHON_OUTPUT_DIRECTORY
```
ax3l: WIP: still fighting that one

bindings/Python/CMakeLists.txt
```diff
@@ -13,16 +13,29 @@ if(ADIOS2_HAVE_MPI)
...
+if(WIN32)
+  set(CMAKE_INSTALL_PYTHONDIR_DEFAULT
+    "${CMAKE_INSTALL_LIBDIR}\\site-packages"
+  )
+else()
+  set(CMAKE_INSTALL_PYTHONDIR_DEFAULT
+    "${CMAKE_INSTALL_LIBDIR}/python${PYTHON_VERSION_MAJOR}.${PYTHON_VERSION_MINOR}/site-packages"
+  )
+endif()
+set(CMAKE_INSTALL_PYTHONDIR "${CMAKE_INSTALL_PYTHONDIR_DEFAULT}"
```
ax3l: Fix: Now overwriteable from the command line.

bindings/Python/CMakeLists.txt
```diff
@@ -13,16 +13,29 @@ if(ADIOS2_HAVE_MPI)
...
+  CACHE PATH "Install directory for python modules"
 )
-set(CMAKE_INSTALL_PYTHONDIR "${CMAKE_INSTALL_PYTHONDIR}"
-  CACHE INTERNAL "" FORCE
+set(CMAKE_PYTHON_OUTPUT_DIRECTORY
+  "${ADIOS2_BINARY_DIR}/${CMAKE_INSTALL_PYTHONDIR}"
+  CACHE PATH "Build directory for python modules"
```
chuckatkins: Have you tried this with multi-config generators like Visual studio.  This often doesn't quite work since you need to be able to hole release and debug builds simultaneously in the same build tree.

bindings/Python/CMakeLists.txt
chuckatkins: Rather than hard coding this with a guess you could use introspection:
```cmake
execute_process(
  COMMAND
    ${PYTHON_EXECUTABLE}
      -c [[
import sys
from distutils import sysconfig
sitedir = sysconfig.get_python_lib(plat_specific=True)
if sitedir.startswith(sys.prefix):
    sitedir = sitedir[len(sys.prefix)+1:]

print(sitedir)
]]
  OUTPUT_VARIABLE CMAKE_INSTALL_PYTHONDIR_DEFAULT
)
```

Also be sure to mark the cache variable as advanced.

chuckatkins:
A few things:

- Rather than hard coding this with a guess you could use introspection:

```cmake
execute_process(
  COMMAND
    ${PYTHON_EXECUTABLE}
      -c [[
import sys
from distutils import sysconfig
sitedir = sysconfig.get_python_lib(plat_specific=True)
if sitedir.startswith(sys.prefix):
    sitedir = sitedir[len(sys.prefix)+1:]

print(sitedir)
]]
  OUTPUT_VARIABLE CMAKE_INSTALL_PYTHONDIR_DEFAULT
)
```
- Be sure to mark the cache variable as advanced.
- Does it work with multi-config generators like visual studio which need to keep both release and debug builds in the same tree simultaneously?

ax3l:
> introspection

Uh, that's fancy as well :) Actually, there are only these two, well defined locations inside a given prefix:
https://docs.python.org/3/install/#how-installation-works

But if you consider this more robust we can add this as well.

> Be sure to mark the cache variable as advanced.

thx, done!

> Does it work with multi-config generators like visual studio which need to keep both release and debug builds in the same tree simultaneously?

Yes, it should work well, using it as such with MSVC/VS on AppVeyor as well as with NMake on CF.

chuckatkins:
> > introspection
> 
> Uh, that's fancy as well :) Actually, there are only these two, well defined locations inside a given prefix:
> https://docs.python.org/3/install/#how-installation-works
...
> > But if you consider this more robust we can add this as well.

I don't know if I'd necessarily say robust.  More adaptive for sure, but certainly more complicated as well and prone to cross-compilation issues.  It depends on your system and how "adapted to it" the python install your using is. Introspecting would be the way to ensure it's most likely correct (that's what pybind11 does, but even it's mechanism is a little broken).  Don't bother for now but it's likely a good option if needed.

bindings/Python/CMakeLists.txt
chuckatkins: We shouldn't actually need the ARCHIVE_OUTPUT_DIRECTORY.  Since this is built as a `MODULE` target and not `SHARED` then the only output is a `.dll` without a corresponding `.lib` export library.

ax3l:
Ok, I just squashed the commits and rebased then.

bindings/Python/CMakeLists.txt
ax3l: Yes, I know it's confusing but doesn't work without this also being set on multi-config generators due to a name clash at configure time,

PR #1601, opened 2019-07-06
fix use of "uint"

germasch:
It's not a standard type, so can't be relied on. It gave me trouble in one
config on my Mac (gcc8 + address sanitizer), but also contributed to the
last minute MVSC issues before the 2.4.0 release.

### DIFF
--- a/source/adios2/operator/compress/CompressSZ.cpp
+++ b/source/adios2/operator/compress/CompressSZ.cpp
@@ -279,13 +279,12 @@ size_t CompressSZ::Compress(const void *dataIn, const Dims &dimensions,
     // In C, r[0] is the last dimension. In Fortran, r[0] is the first dimension
     for (int i = 0; i < ndims; i++)
     {
-        uint dsize = dimensions[i];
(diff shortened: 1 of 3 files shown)


source/adios2/operator/compress/CompressSZ.cpp
```diff
@@ -279,7 +279,7 @@ size_t CompressSZ::Compress(const void *dataIn, const Dims &dimensions,
     // In C, r[0] is the last dimension. In Fortran, r[0] is the first dimension
     for (int i = 0; i < ndims; i++)
     {
-        uint dsize = dimensions[i];
+        unsigned int dsize = dimensions[i];
```
pnorbert: both dimensions and r are size_t arrays, so dsize should be too. Or, as the code stands now, this should be a single assignment.

germasch: Well, I was being bug-for-bug compatible ;)

I updated the PR to get rid of `dsize` altogether, which is clearly the better solution.

PR #1614, opened 2019-07-17
Bp4 opentimeout

pnorbert:
…ding within timeout


pnorbert:
wrong branch, no merge

PR #1619, opened 2019-07-22
Fix incorrect error message in SST

keichi:
SST throws incorrect error messages when invalid values are given to CPCommPattern or QueueFullPolicy. Small issue but probably confusing for the user.


eisenhauer:
Thanks Keichi!  Copy and paste errors of course...

PR #1631, opened 2019-07-26
Bump FindMPI to v3.15.1, modified for v3.6 compatibility.

chuckatkins:


chuckatkins:
This fixes a few bugs when trying to use find MPI from dependees

chuckatkins:
Pushed to the wrong remote

PR #1645, opened 2019-07-30
fix reading in MinMaxs array completely from BP4 data

pnorbert:


source/adios2/toolkit/format/bp4/BP4Base.tcc
```diff
@@ -310,12 +310,10 @@ inline void BP4Base::ParseCharacteristics(const std::vector<char> &buffer,
...
-                for (int m = 0; m < M; ++m)
-                {
-                    characteristics.Statistics.MinMaxs[m] =
-                        helper::ReadValue<T>(buffer, position, isLittleEndian);
-                }
+                characteristics.Statistics.MinMaxs.resize(2 * M);
+                helper::CopyFromBuffer<T>(
+                    buffer, position, characteristics.Statistics.MinMaxs.data(),
+                    2 * M);
             }
```
germasch: Looking at this change, is this still going to be endian-safe?

pnorbert: Thanks for the catch. I did not think about endianness.

PR #1646, opened 2019-07-30
Use a 4 part version string

chuckatkins:
This moves towards an automatically updating 4-part version string that
determines the "tweak" version based on the number of commits since the
most recent release.  This allows us to reference version numbers
in-between releases.

For example, `v2.4.0` remains the same but the current master would now automatically become `v2.4.0.168` since it's 168 commits past the release tag.


chuckatkins:
@williamfgc @pnorbert Does this break anything with the BP formatting where it inserts the version number as meta-data into the byte stream headers or footers.

williamfgc:
@chuckatkins since the format itself doesn't change as often as the version tags, I'd keep 3 fields in bp, but that's just me. I'll let @pnorbert decide as he is currently more activate with the development of bp4 and the tooling around it.

chuckatkins:
@williamfgc so the changes in `source/adios2/toolkit/format/bp3/BP3Serializer.cpp` shouldn't cause any issue by adding the extra number?

williamfgc:
@chuckatkins I might have to tests with adios1 bpdump for backwards compatibility. If I see a problem, I'll just make a new PR removing the 4th number.

pnorbert:
I think this will break BP4. The header has 32 bytes for the identity
string and it is already long:
"ADIOS-BP v2.4.0 Index Table   "

pnorbert:
BTW, this code in BP3 was already broken and is more broken now:

    lf_CopyVersionChar(std::to_string(ADIOS2_VERSION_MAJOR), buffer,
position);
    lf_CopyVersionChar(std::to_string(ADIOS2_VERSION_MINOR), buffer,
position);
    lf_CopyVersionChar(std::to_string(ADIOS2_VERSION_PATCH), buffer,
position);
#ifdef ADIOS2_VERSION_TWEAK
    lf_CopyVersionChar(std::to_string(ADIOS2_VERSION_TWEAK), buffer,
position);
#endif
    ++position;

In adios1, this was 4 byte binary data in the footer, a byte for major,
minor and patch, each. Then an empty byte (represented here as ++position).
This code in adios2 will break with any version number >9. Also it is now 5
characters long if TWEAK is 1 character. So yes, bpdump will not work with
this modified BP3 format.

Anyway, I should have stated first: the BP3 format is fixed, no changes are
allowed to it. This tweaking can be discussed for BP4.

chuckatkins:
I can revert the change to bp3 serializer and it will just ignore the tweak component of present.  The main reason for having it is at the build system level so if something depends on a feature in master but not release then it can be expressed as `find_package(adios2 2.4.0.187)`

pnorbert:
Yes, that would be sufficient.

PR #1654, opened 2019-08-02
Fix backwards compatibility logic for FindMPI

chuckatkins:


ax3l:
ok, just bumping the requirement is definitely an option  ^^

PR #1657, opened 2019-08-02
Fixing build error with Endian_Reverse

williamfgc:


chuckatkins:
Is there any reason we need this to be optional instead of just always enabled?

williamfgc:
There are fewer big endian systems, this this is becoming a corner case. So far, only one heterogeneous system we deal with needed this.

williamfgc:
Also, it is not on currently enabled on our CI.

chuckatkins:
I'd be supportive of removing the option and ifdef entirely and just having this be a builtin feature.  The SPARC systems are the only ones we've encountered so far that need it but a few of the newer CPU architectures are likely to use it, i.e. openrisc, maybe the custom Sunway hpc cpus, fpga accelerators, raw sensor data, etc.

@eisenhauer, @pnorbert, your thoughts on just having this be baked in instead of optional?

williamfgc:
@chuckatkins if you have access to those systems it would be great to add at least nightly tests. I don't access to any big endian machine or any of the above architectures. To clarify, this is not needed to run on big endian machines, but when writer and reader have different endianness (that's why I mentioned heterogeneous systems).

PR #1659, opened 2019-08-05
Add Table engine for non-step based extremely large global arrays, and table based data

JasonRuonanWang:
The table engine is a meta-engine that uses BP4 underneath. Before offloading IO operations to the BP4 engine, it merges small puts into larger blocks to reduce the metadata size. The aggregation is done using the ZeroMQ transport.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -118,6 +118,7 @@ adios_option(MPI       "Enable support for MPI" AUTO)
 adios_option(DataMan   "Enable support for DataMan" AUTO)
 adios_option(DataSpaces "Enable support for DATASPACES" AUTO)
 adios_option(SSC       "Enable support for SSC" AUTO)
+adios_option(Table     "Enable support for Table" AUTO)
 adios_option(SST       "Enable support for SST" AUTO)
 adios_option(ZeroMQ    "Enable support for ZeroMQ" AUTO)
 adios_option(HDF5      "Enable support for the HDF5 engine" AUTO)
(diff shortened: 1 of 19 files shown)


CMakeLists.txt
```diff
@@ -118,6 +118,7 @@ adios_option(MPI       "Enable support for MPI" AUTO)
 adios_option(DataMan   "Enable support for DataMan" AUTO)
 adios_option(DataSpaces "Enable support for DATASPACES" AUTO)
 adios_option(SSC       "Enable support for SSC" AUTO)
+adios_option(Table     "Enable support for Table" AUTO)
```
chuckatkins: It looks like the engine isn't actually optional, i.e. there's no way to disable it, so this shouldn't be a user-facing option.

JasonRuonanWang: I just checked and I am actually able to disable it by specifying 
cmake ~/d/a2table -DADIOS2_USE_Table=OFF

JasonRuonanWang: @chuckatkins By the way, how can I quickly check if newer versions of nlohmann::json is already compatible with MSVC or PGI so I can update the library to have these compilers supported for some of my engines that are using nlohmann::json?

chuckatkins: > I just checked and I am actually able to disable it by specifying
> cmake ~/d/a2table -DADIOS2_USE_Table=OFF

I must have misread the logic.  Sorry about that.

cmake/DetectOptions.cmake
```diff
@@ -131,6 +131,17 @@ if(NOT (CMAKE_CXX_COMPILER_ID STREQUAL "PGI") AND NOT MSVC)
     endif()
 endif()
 
+# Table
```
chuckatkins: So, the enable logic here is wrong and doesn't really handle the `ON` case correctly, but I see it's the same logic used in several other places so it's fine to leave it here for this PR.  I'll look at addressing all of them in a separate PR after this.

JasonRuonanWang: @chuckatkins Thanks and please go ahead after I merge. That would be very helpful. I would also like to take a step further and see if there is a way to make the nlohmann::json library work with the PGI compiler and MSVC. So far I have four engines using JSON and we will probably use JSON in the query interface as well. So if this breaks the windows build then it's going to be a blocking issue eventually. I am also looking into https://github.com/Tencent/rapidjson/ and see if we can replace the JSON library with this one, which is said to be faster than the nlohmann implementation. Your comments or ideas would be very helpful.

PR #1662, opened 2019-08-06
Fixes to stabilize KillWriter tests

eisenhauer:
Lengthen kill delay from 2 seconds to 5 seconds.  We've been seeing some CI failures that indicate the reader isn't able to contact the writer.  On a heavily loaded CI machine, 2 seconds might not be enough for writer startup.  Hoping 5 is better.

### DIFF
--- a/source/adios2/engine/sst/SstReader.tcc
+++ b/source/adios2/engine/sst/SstReader.tcc
@@ -124,7 +124,11 @@ void SstReader::ReadVariableBlocks(Variable<T> &variable)
     // wait for all SstRead requests to finish
     for (const auto &i : sstReadHandlers)
     {
-        SstWaitForCompletion(m_Input, i);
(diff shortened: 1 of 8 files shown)


eisenhauer:
OK, this has evolved from an attempt to add some time to the KillWriter tests in the hope that that would help us with CI, to instead be a set of 3 basic changes that hopefully will stabilize the test by closing some holes WRT when failure gets noticed and how it gets handled.  Those fixes boil down to:
- Changing the evpath data plane to correctly return an error code if it is the party to notice a failure.  (This required tweaks to the CP/DP interface as well as changes in evpath_dp.c.)
- Changing the SstReader BP deserialization code so that it noticed if the data fetches failed and raised an exception.
- Changing SstCommonClient so that when it caught an exception out of EndStep() it properly terminated the test and set WriterFailed so that if that was expected, then it was found appropriate.

PR #1668, opened 2019-08-12
python/layout

germasch:
For now, I just want to see what the CI says...

### DIFF
--- a/bindings/Python/py11File.cpp
+++ b/bindings/Python/py11File.cpp
@@ -214,6 +214,12 @@ std::vector<std::string> File::ReadString(const std::string &name,
 
 pybind11::array File::Read(const std::string &name, const size_t blockID)
 {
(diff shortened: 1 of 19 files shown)


williamfgc:
@germasch thanks! This is a great start. I had a branch that was adding support for reading global values, but if this is doing it I'd just work with your PR. Also, would you mind rebasing into a fewer commits, that will help understand communicating the changes so we can weight in. Thanks!

williamfgc:
BTW, I am looking forward to enrich the Python high-level API. You pointed out in one of the commits that there is blockInfo functionality missing and there is also more variable information I'd like to add.

germasch:
For the real PR (#1670), I squashed some of the commits, and included only the python part.

Writing the tests, I certainly have some ideas on improving the high-level API, I may give that a go at some point.

PR #1670, opened 2019-08-12
python hl

germasch:
This is prep work for supporting column-major data in its original order in the python high-level interface, but I think it's useful even by itself.

This PR includes a minor unrelated cleanup removing superfluous semicolons.

It then adds a bunch of tests covering all the variants of the python HL read() functionality, in particular: (1) read whole array / block (2) read a provided selection and (3) read a provided selection and step selection. All 3 overloads are tested against GlobalValue, GlobalArray, LocalValue, LocalArray datasets. Not everything is supported in the current code, so those tests are marked expected failures.

The main part of the PR merges the actual implementation of the various read() overloads into on `DoRead()` function, which takes a Selection and a StepSelection, those being optional:
- An empty `start` means: Start at the origin (0, 0, ...)
- An empty `count` means: Cover whole array (ie, shape) or whole block (ie, count) for GlobalArrays and LocalArrays respectively.
- stepCount = 0 means: Read data from current step.

These changes are internal, ie, the python-facing HL API doesn't change.

This consolidation add support for a few cases, like reading a step selection of GlobalValues. It also makes possible to read a step selection of, say, a whole global array, without having to specify a selection, too:
```python
     read("var', stepStart=1, stepCount=2)
```
Specifying a selection on a LocalArray reads does not work, neither before nor after this PR -- this seems to be an issue of core not supporting it.

This PR squashes (part of) #1668, so if one wants to follow the actual evolution of the code, the former might be easier.

### DIFF
--- a/bindings/Python/py11File.cpp
+++ b/bindings/Python/py11File.cpp
@@ -230,31 +230,14 @@ pybind11::array File::Read(const std::string &name, const Dims &start,
         std::copy(value.begin(), value.end(), pyArray.mutable_data());
         return pyArray;
     }
-#define declare_type(T)                                                        \
-    else if (type == helper::GetType<T>())                                     \
-    {                                                                          \
-        return DoRead<T>(name, start, count, blockID);                         \
-    }
-    ADIOS2_FOREACH_NUMPY_TYPE_1ARG(declare_type)
-#undef declare_type
 
-    throw std::invalid_argument(
-        "ERROR: adios2 file read variable " + name +
-        ", type can't be mapped to a numpy type, in call to read\n");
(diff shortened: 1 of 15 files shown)


germasch:
Oops, meant to close the draft PR instead...

williamfgc:
> Specifying a selection on a LocalArray reads does not work, neither before nor after this PR -- this seems to be an issue of core not supporting it.

Core supports it, here are some [tests using the high-level C++ API](https://github.com/ornladios/ADIOS2/blob/master/testing/adios2/engine/bp/TestBPWriteReadLocalVariablesSelHighLevel.cpp) . We might have to add Python tests. Thanks for the PR looks good.

PR #1672, opened 2019-08-12
Support for BlocksInfo in SST: fix for BP, implementation for FFS, add to test

eisenhauer:
Test to follow SST BP marshaling fix.


JasonRuonanWang:
It would be great to add a test case in which different steps have different block info, to address @chuckatkins 's concern.

eisenhauer:
Agreed.  I need to look to see if anything does that.  ZeroVar test might, but I'm thinking not.

PR #1673, opened 2019-08-12
fixed a bug in BlocksInfo for SST BP marshaling

JasonRuonanWang:


chuckatkins:
Will this always return the block information for the first step now regardless of what step is asked for?  AFAIK that's only valid for the "lock dimensions" case.  Generally on any given step the writer can skip puts on some ranks, vary dimensions and array sizes, etc. which is why we need to inquire the variable again on every step.

JasonRuonanWang:
@chuckatkins No actually I think it returns the current step's blocks info. It's just that in staging cases it always puts the current step's blocks info into index [0] of the data structure. Feel free to verify if you feel unsafe, and let me know if there is any problems.

eisenhauer:
Unfortunately, this includes my quickly thrown-together test, which will fail for the Staging.Modes.* test set.  @JasonRuonanWang , I've incorporated the BP marshaling fix into the other PR, and adjusted it to include a version of the test that should work everywhere.  If you want to kill this PR, I think the other will do.

PR #1677, opened 2019-08-13
python/hl: return single GlobalValue as scalar/0-d array

### LINKED ISSUE #1671: python high-level handling of scalars (GlobalValue)
While working on the python HL API, I noticed this (small) issue: 

When reading a GlobalValue, the API will return a 1-d numpy array with one element, ie., `array([value])`. 

Current behavior:

```python
    def test_GlobalValue1d(self):
        with adios2.open(filename, 'r') as fh:
            for fh_step in fh:
                t = fh_step.current_step()
                val = fh_step.read('global_value')
                self.assertTrue(np.array_equal(val, np.array([global_values[t]])))
```

I'd propose changing this to returning a 0-d array, which is numpy's way of handling scalars. That'll make is possible to just say
```python
self.assertTrue(val == global_values[t])
```
and generally will be less confusing for users of the HL API, as they won't have to unpack the scalar value from the 1-d array first.

germasch:
That's what one should expect for a single value, not a list / 1d array
with a single element. It is a change in behavior, though, though reading
global values was only partially supported previously, anyway.

Fixes #1671

### DIFF
--- a/bindings/Python/py11File.cpp
+++ b/bindings/Python/py11File.cpp
@@ -274,6 +274,11 @@ pybind11::array File::ReadAttribute(const std::string &name,
     {                                                                          \
         core::Attribute<T> *attribute = m_Stream->m_IO->InquireAttribute<T>(   \
             name, variableName, separator);                                    \
+        if (attribute->m_IsSingleValue)                                        \
+        {                                                                      \
+            pybind11::array_t<T> pyArray({});                                  \
+            pyArray.mutable_data()[0] = attribute->m_DataSingleValue;          \
+        }                                                                      \
         pybind11::array_t<T> pyArray(attribute->m_Elements);                   \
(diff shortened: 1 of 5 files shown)


germasch:
So this now passed, after adapting the tests, and as I did that, I figured that reading scalar attributes should be made consistent, too.

That is, adios2 makes a distinction between a scalar (a.k.a. single value) and a length-1 array, and they now get mapped to a numpy scalar (0-d array) ("99") vs a numpy 1-d array ("[99]"), rather than always a 1-d array.

So this PR makes the python HL more consistent overall, but I guess I don't really have strong feelings about it getting merged, because at the same time, it changes existing behavior - and there are many other parts to the HL interface that are not exactly pythonic, anyway.

williamfgc:
@germasch looks good, thanks!

PR #1681, opened 2019-08-14
WIP: return fortran arrays in original order in python high-level interface

### LINKED ISSUE #1665: Reading Fortran data using python interface
When reading data written by a Fortran code using python, indices end up being reversed:

Sample Fortran code (it's based on the included heatmap test)
```fortran
program TestBPWriteReadHeatMap2D
  use mpi
  use adios2

  implicit none

  type(adios2_adios) :: adios
  type(adios2_io) :: ioPut
  type(adios2_engine) :: bpWriter
  type(adios2_variable) :: var

  real(kind=8) :: temperature(0:4, 0:9)

  integer(kind=8), parameter :: ishape(2) = (/5, 10/)
  integer(kind=8), parameter :: istart(2) = (/0, 0/)
  integer :: ierr

  call mpi_init(ierr)
  call adios2_init(adios, MPI_COMM_WORLD, adios2_debug_mode_on, ierr)
  
  call random_number(temperature)
  print*, 'sample temperature(2,8) =', temperature(2, 8)

  call adios2_declare_io(ioPut, adios, 'HeatMapWrite', ierr)
  call adios2_define_variable(var, ioPut, 'temperature', adios2_type_dp, &
                              2, ishape, istart, ishape, &
                              adios2_constant_dims, ierr)
  call adios2_open(bpWriter, ioPut, 'HeatMap2D_f.bp', adios2_mode_write, &
                   ierr)
  call adios2_put(bpWriter, var, temperature, ierr)
  call adios2_close(bpWriter, ierr)

  call adios2_finalize(adios, ierr)
  call mpi_finalize(ierr)

end program TestBPWriteReadHeatMap2D
```
Output:
```sh
[kai@macbook build-fortran (pr/fix-mgard *)]$ bin/test_fortran
 sample temperature(2,8) =  0.13062623044039723
```

Reading the bp file from Python:
```
import adios2
fh = adios2.open("/Users/kai/build/ADIOS2/build-fortran/HeatMap2D_f.bp", "r")
t = fh.read('temperature')
print(t[2, 8])
```
gives
```
---------------------------------------------------------------------------
IndexError                                Traceback (most recent call last)
<ipython-input-10-cf0fcf0017a8> in <module>
      2 fh = adios2.open("/Users/kai/build/ADIOS2/build-fortran/HeatMap2D_f.bp", "r")
      3 t = fh.read('temperature')
----> 4 print(t[2, 8])

IndexError: index 8 is out of bounds for axis 1 with size 5
```

To access the Fortran data element `t(2, 8)` from Python, one has to reverse the index order:
```python
>>> print(t[8, 2])
0.13062623044039723
```
This is related to #1661 , but actually fixable without API additions.

germasch:
So this is what I originally set out to do:

* adds a `m_OriginalLayout` member to `core::Variable`, which defaults to `Unknown`, but BP3 and BP4 engines will set it to the original layout the data was written with it.
* `bpls` behavior is changed slightly (only): If the data was originally written from Fortran, it'll add the note `(transposed)` to the shape, but other than that, it still shows the dimensions reversed / dumps the data transposed. (This could be changed in the future to present the data how it was originally written, if that's considered more useful than current behavior)
* The Fortran HL interface is changed to return arrays in their original index order. Nothing changes for arrays written in row-major, but arrays written as column-major will be read back in their original order. No transposing / rearranging of the data occurs, one just needs to tell numpy what the layout is.

More specifically, the `read` methods gain an additional `order` parameter. `'F'` means ,interpret the data as column-major (so if it was written by Fortran, it'll show up in its original order. If it was originally row-major, this makes it show up transposed. `'C'` means, interpret the data as row-major, which is what current behavior is always. `'K'` means, interpret the data as the layout that it was saved as (if known, default row-major).

If no order parameter is provided, it defaults to `'K'`, which means both data written by C and Fortran will show up in their original order, which is what most people would want. However, it is a change in behavior. The default order, if not provided as an argument, can be easily changed using
```python
import adios2
adios2.file.read_order = 'C'
```
will make restore current behavior. So while this PR changes behavior, it's at least easy to restore the original behavior if anyone turns out to be affected. I chose this way because the default behavior will avoid confusion, at the cost of potentially causing issues to someone relying on the old behavior. One could prioritize the opposite way, ie., keep old behavior by default and everyone who would like to see their Fortran arrays presented in original order would need to explicitly set `read_order` to `'K'`.

Fixes #1665

### DIFF
--- a/bindings/Python/py11File.cpp
+++ b/bindings/Python/py11File.cpp
@@ -70,9 +70,39 @@ size_t File::AddTransport(const std::string type, const Params &parameters)
     return m_Stream->m_IO->AddTransport(type, parameters);
 }
 
-std::map<std::string, adios2::Params> File::AvailableVariables() noexcept
+std::map<std::string, adios2::Params>
+File::AvailableVariables(const std::string order) noexcept
 {
-    return m_Stream->m_IO->GetAvailableVariables();
+    Layout layout = ToLayout(order);
+
+    auto vars = m_Stream->m_IO->GetAvailableVariables();
+    for (auto &desc : vars)
+    {
+        auto type = m_Stream->m_IO->InquireVariableType(desc.first);
+
+        if (false)
+        {
+        }
+#define declare_type(T)                                                        \
+    else if (type == helper::GetType<T>())                                     \
+    {                                                                          \
+        auto var = m_Stream->m_IO->InquireVariable<T>(desc.first);             \
+        if (layout == Layout::Original)                                        \
+        {                                                                      \
+            layout = var->GetOriginalLayout();                                 \
+        }                                                                      \
+        bool reverse_dims = layout == Layout::ColumnMajor;                     \
+        if (reverse_dims)                                                      \
+        {                                                                      \
+            auto shape = var->Shape();                                         \
+            std::reverse(shape.begin(), shape.end());                          \
+            desc.second["Shape"] = helper::VectorToCSV(shape);                 \
+        }                                                                      \
+    }
+        ADIOS2_FOREACH_STDTYPE_1ARG(declare_type)
+#undef declare_type
+    }
+    return vars;
 }
 
 std::map<std::string, adios2::Params> File::AvailableAttributes() noexcept
--- a/bindings/Python/py11File.cpp
+++ b/bindings/Python/py11File.cpp
@@ -25,6 +25,8 @@ namespace adios2
 namespace py11
 {
 
+std::string File::ReadOrder = "K";
+
 File::File(const std::string &name, const std::string mode, MPI_Comm comm,
            const std::string engineType)
 : m_Name(name), m_Mode(mode),
@@ -212,13 +214,15 @@ std::vector<std::string> File::ReadString(const std::string &name,
                                        blockID);
 }
 
-pybind11::array File::Read(const std::string &name, const size_t blockID)
+pybind11::array File::Read(const std::string &name, const size_t blockID,
+                           const std::string &order)
 {
-    return Read(name, {}, {}, blockID);
+    return Read(name, {}, {}, blockID, order);
 }
 
 pybind11::array File::Read(const std::string &name, const Dims &start,
-                           const Dims &count, const size_t blockID)
+                           const Dims &count, const size_t blockID,
+                           const std::string &order)
 {
     const std::string type = m_Stream->m_IO->InquireVariableType(name);
 
@@ -231,12 +235,13 @@ pybind11::array File::Read(const std::string &name, const Dims &start,
         return pyArray;
     }
 
-    return Read(name, start, count, 0, 0, blockID);
+    return Read(name, start, count, 0, 0, blockID, order);
 }
 
 pybind11::array File::Read(const std::string &name, const Dims &start,
                            const Dims &count, const size_t stepStart,
-                           const size_t stepCount, const size_t blockID)
+                           const size_t stepCount, const size_t blockID,
+                           const std::string &order)
 {
     const std::string type = m_Stream->m_IO->InquireVariableType(name);
 
@@ -246,7 +251,8 @@ pybind11::array File::Read(const std::string &name, const Dims &start,
 #define declare_type(T)                                                        \
     else if (type == helper::GetType<T>())                                     \
     {                                                                          \
-        return DoRead<T>(name, start, count, stepStart, stepCount, blockID);   \
+        return DoRead<T>(name, start, count, stepStart, stepCount, blockID,    \
+                         order);                                               \
     }
     ADIOS2_FOREACH_NUMPY_TYPE_1ARG(declare_type)
 #undef declare_type
@@ -352,6 +358,32 @@ adios2::Mode File::ToMode(const std::string mode) const
     }
 
     return modeCpp;
+}
+
+Layout File::ToLayout(std::string order)
+{
+    if (order.empty()) // if not specified, use default
+    {
+        order = ReadOrder;
+    }
+
+    if (order == "C")
+    {
+        return Layout::RowMajor;
+    }
+    else if (order == "F")
+    {
+        return Layout::ColumnMajor;
+    }
+    else if (order == "K")
+    {
+        return Layout::Original;
+    }
+    else
+    {
+        throw std::invalid_argument(
+            "'order' argument to read() must be one of 'C', 'F', 'K'");
+    }
 }
 
 } // end namespace py11
--- a/bindings/Python/py11File.cpp
+++ b/bindings/Python/py11File.cpp
@@ -215,14 +215,14 @@ std::vector<std::string> File::ReadString(const std::string &name,
 }
 
 pybind11::array File::Read(const std::string &name, const size_t blockID,
-                           const std::string &order)
+                           const std::string order)
 {
     return Read(name, {}, {}, blockID, order);
 }
 
 pybind11::array File::Read(const std::string &name, const Dims &start,
                            const Dims &count, const size_t blockID,
-                           const std::string &order)
+                           const std::string order)
 {
     const std::string type = m_Stream->m_IO->InquireVariableType(name);
 
@@ -241,7 +241,7 @@ pybind11::array File::Read(const std::string &name, const Dims &start,
 pybind11::array File::Read(const std::string &name, const Dims &start,
                            const Dims &count, const size_t stepStart,
                            const size_t stepCount, const size_t blockID,
-                           const std::string &order)
+                           const std::string order)
 {
     const std::string type = m_Stream->m_IO->InquireVariableType(name);
 
--- a/bindings/Python/py11File.h
+++ b/bindings/Python/py11File.h
@@ -24,6 +24,8 @@ namespace py11
 class File
 {
 public:
+    static std::string ReadOrder;
+
     const std::string m_Name;
     const std::string m_Mode;
 
@@ -92,14 +94,17 @@ public:
                                         const size_t stepCount,
                                         const size_t blockID = 0);
 
-    pybind11::array Read(const std::string &name, const size_t blockID = 0);
+    pybind11::array Read(const std::string &name, const size_t blockID,
+                         const std::string &order);
 
     pybind11::array Read(const std::string &name, const Dims &start,
-                         const Dims &count, const size_t blockID = 0);
+                         const Dims &count, const size_t blockID,
+                         const std::string &order);
 
     pybind11::array Read(const std::string &name, const Dims &start,
                          const Dims &count, const size_t stepStart,
-                         const size_t stepCount, const size_t blockID = 0);
+                         const size_t stepCount, const size_t blockID,
+                         const std::string &order);
 
     pybind11::array ReadAttribute(const std::string &name,
                                   const std::string &variableName = "",
@@ -121,11 +126,13 @@ public:
 private:
     std::shared_ptr<core::Stream> m_Stream;
     adios2::Mode ToMode(const std::string mode) const;
+    static Layout ToLayout(std::string order);
 
     template <class T>
     pybind11::array DoRead(const std::string &name, const Dims &start,
                            const Dims &count, const size_t stepStart,
-                           const size_t stepCount, const size_t blockID);
+                           const size_t stepCount, const size_t blockID,
+                           const std::string &order);
 };
 
 } // end namespace py11
--- a/bindings/Python/py11File.h
+++ b/bindings/Python/py11File.h
@@ -95,16 +95,16 @@ public:
                                         const size_t blockID = 0);
 
     pybind11::array Read(const std::string &name, const size_t blockID,
-                         const std::string &order);
+                         const std::string order);
 
     pybind11::array Read(const std::string &name, const Dims &start,
                          const Dims &count, const size_t blockID,
-                         const std::string &order);
+                         const std::string order);
 
     pybind11::array Read(const std::string &name, const Dims &start,
                          const Dims &count, const size_t stepStart,
                          const size_t stepCount, const size_t blockID,
-                         const std::string &order);
+                         const std::string order);
 
     pybind11::array ReadAttribute(const std::string &name,
                                   const std::string &variableName = "",
@@ -132,7 +132,7 @@ private:
     pybind11::array DoRead(const std::string &name, const Dims &start,
                            const Dims &count, const size_t stepStart,
                            const size_t stepCount, const size_t blockID,
-                           const std::string &order);
+                           const std::string order);
 };
 
 } // end namespace py11
--- a/bindings/Python/py11File.h
+++ b/bindings/Python/py11File.h
@@ -50,7 +50,8 @@ public:
     size_t AddTransport(const std::string type,
                         const Params &parameters = Params());
 
-    std::map<std::string, adios2::Params> AvailableVariables() noexcept;
+    std::map<std::string, adios2::Params>
+    AvailableVariables(const std::string order) noexcept;
 
     std::map<std::string, adios2::Params> AvailableAttributes() noexcept;
 
--- a/bindings/Python/py11File.tcc
+++ b/bindings/Python/py11File.tcc
@@ -18,11 +18,13 @@ namespace adios2
 namespace py11
 {
 
-static Dims py_strides(const Dims &shape, ssize_t itemsize, bool has_step_dim)
+namespace
+{
+Dims py_strides(const Dims &shape, ssize_t itemsize, bool hasStepDim)
 {
     auto ndim = shape.size();
     Dims strides(ndim, itemsize);
-    if (!has_step_dim)
+    if (!hasStepDim)
     {
         // regular column-major
         for (size_t i = 1; i < ndim; ++i)
@@ -41,12 +43,13 @@ static Dims py_strides(const Dims &shape, ssize_t itemsize, bool has_step_dim)
     }
     return strides;
 }
+} // end anonymous namespace
 
 template <class T>
 pybind11::array File::DoRead(const std::string &name, const Dims &_start,
                              const Dims &_count, const size_t stepStart,
                              const size_t stepCount, const size_t blockID,
-                             const std::string &order)
+                             const std::string order)
 {
     Layout layout = ToLayout(order);
     core::Variable<T> &variable = *m_Stream->m_IO->InquireVariable<T>(name);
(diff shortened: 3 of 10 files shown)


bindings/Python/py11File.tcc
```diff
@@ -60,22 +77,51 @@ pybind11::array File::DoRead(const std::string &name, const Dims &_start,
...
+            shapePy.emplace_back(stepCount);
+        }
     }
-    std::copy(count.begin(), count.end(), std::back_inserter(shapePy));
 
-    pybind11::array_t<T> pyArray(shapePy);
+    pybind11::array_t<T> pyArray;
+    if (layout == Layout::ColumnMajor)
+    {
+        pyArray = pybind11::array_t<T, pybind11::array::f_style>(shapePy);
```
williamfgc: thanks for adding this

williamfgc:
@germasch overall this looks great. Just made a few comments and I'd like to hear from @pnorbert on the changes to bpls.

bindings/Python/py11File.cpp
```diff
@@ -231,12 +235,13 @@ pybind11::array File::Read(const std::string &name, const Dims &start,
...
 
-    return Read(name, start, count, 0, 0, blockID);
+    return Read(name, start, count, 0, 0, blockID, order);
 }
 
 pybind11::array File::Read(const std::string &name, const Dims &start,
                            const Dims &count, const size_t stepStart,
-                           const size_t stepCount, const size_t blockID)
+                           const size_t stepCount, const size_t blockID,
+                           const std::string &order)
```
williamfgc: I'd use `const std::string order` as that's what is used for strings we know are always small.

bindings/Python/py11File.cpp
```diff
@@ -354,5 +360,31 @@ adios2::Mode File::ToMode(const std::string mode) const
     return modeCpp;
 }
 
+Layout File::ToLayout(std::string order)
```
williamfgc: `Layout File::ToLayout( const std::string order) const`

bindings/Python/py11File.h
```diff
@@ -24,6 +24,8 @@ namespace py11
 class File
 {
 public:
+    static std::string ReadOrder;
```
williamfgc: `m_ReadOrder`

bindings/Python/py11File.h
```diff
@@ -92,14 +94,17 @@ class File
...
                                         const size_t blockID = 0);
 
-    pybind11::array Read(const std::string &name, const size_t blockID = 0);
+    pybind11::array Read(const std::string &name, const size_t blockID,
+                         const std::string &order);
 
     pybind11::array Read(const std::string &name, const Dims &start,
-                         const Dims &count, const size_t blockID = 0);
+                         const Dims &count, const size_t blockID,
+                         const std::string &order);
```
williamfgc: `const std::string order` similar in all Read overloads.

bindings/Python/py11File.h
williamfgc: Not sure about the need for `static`, a single workflow can be using multiple instances of File with different data layout order. Making it static is basically a global across all files.

bindings/Python/py11File.h
```diff
@@ -121,11 +126,13 @@ class File
 private:
     std::shared_ptr<core::Stream> m_Stream;
     adios2::Mode ToMode(const std::string mode) const;
+    static Layout ToLayout(std::string order);
```
williamfgc: See my comment about `static` and changing the signature to `Layout File::ToLayout( const std::string order) const`

bindings/Python/py11File.tcc
```diff
@@ -18,13 +18,54 @@ namespace adios2
 namespace py11
 {
 
+static Dims py_strides(const Dims &shape, ssize_t itemsize, bool has_step_dim)
```
williamfgc: let's discuss the need for `static`, also we use camelCase style `const bool hasStepDim`

bindings/Python/py11glue.cpp
```diff
@@ -899,16 +902,17 @@ PYBIND11_MODULE(adios2, m)
...
+        .def(
+            "read",
+            (pybind11::array(adios2::py11::File::*)(
+                const std::string &, const adios2::Dims &, const adios2::Dims &,
+                const size_t, const std::string &order)) &
+                adios2::py11::File::Read,
+            pybind11::return_value_policy::take_ownership,
+            pybind11::arg("name"), pybind11::arg("start") = adios2::Dims(),
+            pybind11::arg("count") = adios2::Dims(),
+            pybind11::arg("block_id") = 0, pybind11::arg("order") = "",
```
williamfgc: Should we make the default `K` explicit? It will help documenting this API.

bindings/Python/py11File.cpp
germasch: It's a static member function, so it can't be const.

 I guess one can argue about whether it should be a const (non-static) member function instead, but since it doesn't use `this`, there's really no point. In fact, it could be made a free function instead, which I thought about, but well, since there aren't any other free functions around there, I figured this keeps it with where it's used.

bindings/Python/py11File.h
germasch: `ReadOrder` is just the default order, and I deliberately wanted it to be "global", so that one could do
```python
import adios2
adios2.File.read_order = 'C'
```
to restore the old behavior. This wouldn't work with a per-instance member variable.

I can prefix it with `m_`, i wasn't sure you also do that for static class variables.

bindings/Python/py11File.tcc
germasch: will do on the camel case. 

This one is static since it's not meant to be used outside of that source file. I could use an anonymous namespace instead to be more C++-y, I guess, if you prefer.

bindings/Python/py11glue.cpp
germasch: Well, I wanted to be the default to be changeable (via `adios2.File.read_order`), so that's why I didn't hardcode it there. If I put `"K"` there, I couldn't tell whether I get `"K"` because the user said `read(..., order='K')`, or just `read()`.

If you think we don't need a configurable default, yes, putting `"K"` there would be cleaner, and getting rid of all the `ReadMode` stuff. But in that case, there'd be no easy way to go back to current behavior, and since this introduces a backward-incompatible change in semantics, I figured there should at least be an easy way to go back to the old behavior.

bindings/Python/py11File.cpp
williamfgc: It can be moved to the `adios2::helper` namespace that contains all the common reusable free functions. See my comment above about `static`.

bindings/Python/py11glue.cpp
williamfgc: @germasch adding K would be cleaner to people used to numpy, other PRs already break the API for column-major generated data, so not a big a deal as long as there is clear documentation for the next release.

bindings/Python/py11File.tcc
williamfgc: @germasch yes, an anonymous namespace is fine, help us make the distinction between static functions as "only used here" and those static member functions that handle static members.

bindings/Python/py11File.h
williamfgc: @germasch not sure I am following. 

```
import adios2
adios2.File.read_order = 'C'

f1 = adios2.open("F.bp")
cm = f1.read(...)
f2 = adios2.open("C.bp")
rm = f2.read(...)
```
would this work? Is there a test for this since it's a new feature?

bindings/Python/py11File.cpp
germasch: I'll move it to `helper` once / if it gains more than one user. It's not generally usable, since it's python-interface specific. But it might get used in the low-level interface eventually,
at which point it makes sense to share it.

germasch:
I added some documentation (good point). I've also added a test to show that `adios2.File.read_mode` works to set the default globally.

I'm now passing the ``order`` string by value, and changed the `static` to an anonymous namespace, and fixed the camelcase variable.

If you would prefer it, I could remove the configurable default for `order`, and just hardcode `order='K'`, ie, `'K'` always being default. It would certainly remove one bit of complexity (which is good), but also the ability to set the default back to the old behavior (always row-major), which personally I wouldn't miss, but someone already using the HL interface might.

williamfgc:
@germasch thanks for addressing all the points. Looks good. 

> If you would prefer it, I could remove the configurable default for order, and just hardcode order='K', ie, 'K' always being default. It would certainly remove one bit of complexity (which is good), but also the ability to set the default back to the old behavior (always row-major), which personally I wouldn't miss, but someone already using the HL interface might.

Let me discuss with more people and find out more, in the meantime we can remove it and think more about it while current users are not affected.

williamfgc:
one minor thing, can we have a separate PR for changes in bpls just so I can merge the changes in Python one this PR is done. Thanks!

germasch:
> one minor thing, can we have a separate PR for changes in bpls (it's a different topic) just so I can merge the changes in Python once this PR is done. Thanks!

I'll drop he bpls changes from this PR.

germasch:
> > If you would prefer it, I could remove the configurable default for order, and just hardcode order='K', ie, 'K' always being default. It would certainly remove one bit of complexity (which is good), but also the ability to set the default back to the old behavior (always row-major), which personally I wouldn't miss, but someone already using the HL interface might.
> 
> Let me discuss with more people and find out more. In the meantime we can remove it and think more about it, while current users are not affected.

The thing is, the way things are right now, people are (potentially) affected. The old behavior is `order='C'`. If I remove the `File.read_order` thing and just hardcode `order='K'` as default argument, the behavior when reading Fortran files will be different (arguably, it will be "correct" now, as the array will show up with its original indexing, but people's definition of "correct" may vary. I'd say, it's definitely less confusing to a new user when indices don't reverse.)

I think that's generally a difficult situation -- one wants to change behavior, because it turns out that some old behavior is not really how it should be. There are basically a few options:
1. Stick with the old behavior, at least by default. The good thing about that is that one doesn't break any existing users. The disadvantage is that the confusing behavior gets carried forward indefinitely, and as more people use the API, the harder it becomes to change.
2. Introduce a new API. Like, `read2`, which changes behavior to how it should have been. That works, but it's ugly (and also kinda confusing).
3. Fix the "buggy" behavior. That's the best solution going forward, only one API, without the confusion. But it breaks code that has to become to rely on the "buggy" behavior.

My suggestion on this (and what the current PR does) is to follow the 3rd option. To limit the breakage (if any), one can put into the release notes:

* adios2.File.read() now honors the original data layout that the data was written with, that is, Fortran arrays will be read in their original indexing order, as opposed to previous behavior, which reverse dimensions. In order to restore original behavior, set `adios2.File.read_mode = 'F'` after importing adios2.

I really dislike option 2.  Option 1 wouldn't be unreasonable, though I suspect it won't really fix the issue that this PR is trying to address, since people don't tend to read documentation in much detail:

Someone doing `x = fh.read('x')` and noticing that the dimension are reversed from their original data is likely to think "Hmmh, looks like adios2 sucks just as bad h5py in that it reverses the dimensions of my array just because I dare using Fortran...", and somehow work around it themselves. Not many people will look through the documentation and notice that if they used `x = fh.read('x', order='C')` things will start behaving as they should.

williamfgc:
@germasch I agree with your points. I just want to hear from other stakeholders before changing current behavior.

germasch:
@williamfgc Sounds good to me. I dropped the bpls related changes from this PR.

NAThompson:
@germasch : Would it be sensible to add a few words about this feature into the `ADIOS2/docs/user_guide` files? This seems like a very useful feature and I don't want it to be buried.

germasch:
> @germasch : Would it be sensible to add a few words about this feature into the `ADIOS2/docs/user_guide` files? This seems like a very useful feature and I don't want it to be buried.

This PR adds a bit of documentation, on the `order` parameter. I didn't feel the need to write too much explaining the new behavior, since it's really what I'd consider the natural behavior in the first place, ie, you access the data in the same index order that you used when writing it.

> Why don't we add the ordering as a metadata parameter in `bp3` and `bp4`? Then we simply interpret the data in the correct order without adding a user parameter.

Well, it's difficult to change existing file formats, though I did suggest to add the information into the metadata in future incarnations. For now the info is on ordering is global per file,  though one probably could improve on that in a backwards compatible way.

> Another question is whether we could do an in-place matrix transpose. Of course we'd need some performance numbers, but assuming higher RAM->CPU bandwidth than Disk->Ram, we should be able to do a matrix transpose on a lower timescale than reading the data. Assuming the performance numbers are reasonable, this seems like a very robust solution to me; no user parameters are required to do the right thing.

I think one can argue either way on this. People here reasonably say they want to focus on performance, and obviously a transpose op may not be cheap, in particular, if it happens on data which is already in memory (e.g., because it's using one of in-memory engines). I don't think it'd necessarily be a bad idea, as long as the user explicitly says "I know what I'm doing". But modern languages like python or c++ (and in some sense even C) have the ability to handle different layouts on the fly, and I'd say that's preferable if available, so that's the low-hanging fruit I'm going for (specifically, I'm thinking from the point of view of my codes, and what I need for them.)

NAThompson:
Ok, so as I understand it:

- The abstraction of `.bp3` as being self-describing is broken, because array ordering is ill-defined.  There is currently no mitigation.
- This PR accepts that the abstraction is broken, but provides mitigation. As such, it is a clear improvement over the status quo, and should be accepted if no other solutions can be provided on a reasonable timescale.

However, let me float an alternative: Fortran writes and reads arrays in C ordering. In this case, we have the context that we are about the perform an expensive disk operation (though the in-memory stores are an important caveat as you mention), and hence the question is whether this operation is expensive relative to the disk ops. I suspect it's not (but haven't written a benchmark yet, so grain of salt), but my tolerance for expensive operations is rather high, because in my view the truly expensive operations are the ones that are performed between your ears when an abstraction leaks.

Now, I don't want to make the perfect the enemy of the good, but the abstraction still leaks: The format is not self-describing. Hence every user is going to hit this same bug, and debug a couple days, and finally find your mitigation-which they'll appreciate! But better that they not hit it in the first place.

Finally, I'm not convinced that the total cost of Fortran performing the transpose is positive. It is amortized over the reads. For instance, your bug report shows you wrote Fortran, and then used Python for reading and subsequent analysis. This seems like a common pattern for many users. Hence if the data is simply written in C ordering, *most* of the operations are done correctly.

So @pnorbert @germasch @williamfgc is this a sensible solution?

germasch:
> * The abstraction of `.bp3` as being self-describing is broken, because array ordering is ill-defined.  There is currently no mitigation.

I don't think the file format is broken for the common case. That is, in my area of science, Fortran is clearly dominant, and always uses column-major. When writing from Fortran, it is correctly recorded that the data is column major.

Adios2 is also doing the right thing when writing row-major data from C/C++. However, there are people (plenty, I suspect, but definitely me), who write column-major data from C/C++, and unless they transpose the data themselves, there is no way to get it onto disk with the correct metadata. That's a separate issue to the one here (it can be addressed to some extent, and I'm planning to do that). I think when writing column-major data from Python, Python will do a transpose for you to row-major, so that's expensive, but what ends up on disk is correctly described.

What this PR addresses is the reading side, in particular, actually using the available information about the storage layout. When using a C/C++ tool like bpls on data written by Fortran, it will present the data with reversed index order, and hence transposed. This PR added a patch to bpls to just at least add a note about the presented data being transposed in this case, but I took that part out as requested.

So this PR only addresses the reading side of the Python HL interface where it uses the original layout information from the metadata to create the Python/numpy array with the original ordering, so that both row-major and column-major array show up in their natural order. On the read side with Python, a transpose is never required, because it supports arbitrary strided layouts. That is also true in C++ if one uses a multi-d array class that supports both data layouts (which I think are most or all of the popular ones). But currently on the C++ side, the information is not available (I'm planning to change that).

The only place where transposes really would be required is if writing both row-major and col-major into the same file, because there is no way to mark individual datasets with separate layouts, so one would have to settle on one or the other and transpose the other data. I'm not sure that's a real life issue, so I don't really have any plans about those. Everything else can be addressed within the current framework, and I'm planning to do that -- this is the first step.

> Finally, I'm not convinced that the total cost of Fortran performing the transpose is positive. It is amortized over the reads. For instance, your bug report shows you wrote Fortran, and then used Python for reading and subsequent analysis. This seems like a common pattern for many users. Hence if the data is simply written in C ordering, _most_ of the operations are done correctly.

Fortran would only have to perform a transpose if reading data that was originally in row-major format (and where the code can't instead just reverse the indices). I suspect that's a very uncommon case. The issue is much more about generic visualization / analysis tools which tend to be written in C/C++/Python, and which will work better if the index order is not reversed when reading column-major data.

> 
> So @pnorbert @germasch @williamfgc is this a sensible solution?

pnorbert:
The enum currently stands both for the actual layout of the variable (so it
may be unknown at the beginning), and for user settings (original tells
adios to represent the array towards the user as it is in the data).

chuckatkins:
@germasch, The functionality and features implemented in here are great.  Given it's significance to the API though it's really something that needs to be implemented within the core and then propagated to all the APIs bindings rather than being implemented just within the python bindings.  While there's currently not complete feature parity across all the APIs, there is for the most part.  A change like this in particular though would need to be available throughout.  Given the disruption of such a change internally and the change to all the public APIs we've decided to push such an implementation to after the end-of-september release.  We do have developer resources on the team to put towards implementing it though but the timeline will be pushed back a bit.

germasch:
Actually, I don't mind dropping this PR for the time being. Maybe not quite for the same reason -- the core functionality needed for the aspect covered here is a very minor addition, but I agree that it leave other aspects to-be-done for now, like HL writing and the low-level Python interface entirely. I wouldn't think having this only available in Python to start is too bad, since Python is the only language that natively supports different layouts. In fact, I think one can learn from it what a modern API for the other bindings might look like.

Anyway, the major reason I agree that this PR may not be the best idea is that it either
- introduces an incompatible change in behavior (with the current default of 'K')
- or, if one were to change the default to 'C', compatibility would be maintained, but it'd also be very unlikely to be of much benefit, except for experts that know to change the default.

Not really directly related, but the python HL API is not very pythonic at all. So there is an opportunity to introduce a more pythonic API and at the same time introduce new, more natural handling of layouts with that.

williamfgc:
> Not really directly related, but the python HL API is not very pythonic at all.

@germasch feel free to propose more pythonic interfaces. Personally, I am not a Python experienced user. Most of the API design came from balancing the input from Python practitioners and what pybind11 v2.0 could support at the time for Python 2 and 3, which feels ancient today.

Issue #1615, opened 2019-07-17
Large number of data races in unit tests


NAThompson:
In preparation for parallel reads, I added the following to the `CMakeLists.txt` and ran the unit tests:

```
if(CMAKE_BUILD_TYPE MATCHES "Debug")
  if(CMAKE_CXX_COMPILER_ID MATCHES "^(AppleClang|Clang|GNU)")
    set(CMAKE_CXX_FLAGS_DEBUG "-Wall -g -O2 -Wfatal-errors -fsanitize=thread -fno-omit-frame-pointer")
  endif()
endif()
```

A very large number of race conditions were detected. A few were in `pthread_create`, which arguably is bad, but not the bailiwick of ADIOS. But a large number of them are originating in ADIOS2. I have attached just one of these here, but many tests are afflicted by these problems.

[race.txt](https://github.com/ornladios/ADIOS2/files/3402662/race.txt)

pnorbert:
This has just started showing up in CI. The parsing of variables is done in multiple threads, and they all create a new variable and **add to the same variable map** using _map_.emplace(). 

./bin/TestBPWriteReadAsStreamADIOS2_Threads
both BP3 and BP4

lwan86:
@pnorbert Could you let me know how to reproduce this bug? I run "TestBPWriteReadAsStreamADIOS2_Threads" several times, it works fine.

williamfgc:
@NAThompson  thanks, we need to start paying more attention to `fsanitize` options. @pnorbert this is probably related to #1125 . I will work on a fix.

chuckatkins:
We currently have ubsan running in the CI for dynamic analysis and we can just as easily add a tsan build.  Once we get a clean tsan build then I can add it to our ci checks.

NAThompson:
@chuckatkins , @williamfgc , @pnorbert : Huge thanks to all of you for taking this seriously. As William notes, you have to work on what's a priority, and being race-clean is mine-but you guys are subjected to the work! So it's huge for me to see PRs getting merged that address this.

PR #1683, opened 2019-08-15
Sanitize

williamfgc:
This is an ongoing effort to sanitize piece of the code. It should be regular practice. Related to #1615

### DIFF
--- a/bindings/C/adios2/c/adios2_c_attribute.cpp
+++ b/bindings/C/adios2/c/adios2_c_attribute.cpp
@@ -169,7 +169,7 @@ adios2_error adios2_attribute_data(void *data, size_t *size,
                 *size = attributeCpp->m_Elements;
                 char **dataT = reinterpret_cast<char **>(data);
 
-                for (auto e = 0; e < *size; ++e)
+                for (size_t e = 0; e < *size; ++e)
                 {
                     attributeCpp->m_DataArray[e].copy(
                         dataT[e], attributeCpp->m_DataArray[e].size());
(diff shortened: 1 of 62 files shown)


williamfgc:
@NAThompson

bindings/C/adios2/c/adios2_c_attribute.cpp
```diff
@@ -169,7 +169,7 @@ adios2_error adios2_attribute_data(void *data, size_t *size,
                 *size = attributeCpp->m_Elements;
                 char **dataT = reinterpret_cast<char **>(data);
 
-                for (auto e = 0; e < *size; ++e)
+                for (size_t e = 0; e < *size; ++e)
```
NAThompson: Just a quick comment: `size` is going to become a worse variable name in the future, as C++17 introduces `std::size` and `using std::size` has gotten into various popular C++ libraries. Whether there will actually be a conflict, I don't know.

source/adios2/toolkit/format/bp4/BP4Deserializer.tcc
```diff
@@ -800,7 +800,12 @@ void BP4Deserializer::DefineVariableInEngineIOPerStep(
                             : header.Path + PathSeparator + header.Name;
 
     core::Variable<T> *variable = nullptr;
-    variable = engine.m_IO.InquireVariable<T>(variableName);
+    {
+        // to prevent conflict with DefineVariable
+        std::lock_guard<std::mutex> lock(m_Mutex);
+        variable = engine.m_IO.InquireVariable<T>(variableName);
```
NAThompson: Great catch on this one.

bindings/C/adios2/c/adios2_c_attribute.cpp
williamfgc: @NAThompson thanks for pointing this out. We need to be in the lookout for new language features.

PR #1687, opened 2019-08-20
Add a reading example to the localArray example

pnorbert:


pnorbert:
Travis is not working for this PR

PR #1690, opened 2019-08-21
fixed a bug in DataMan BeginStep

JasonRuonanWang:


eisenhauer:
Please rebase on master.  Something changed in fedora-ubsan, and what used to work stopped.  My last PR merged a fix.

source/adios2/engine/dataman/DataManReader.cpp
```diff
@@ -101,8 +100,21 @@ StepStatus DataManReader::BeginStep(StepMode stepMode,
                   << m_CurrentStep << std::endl;
     }
 
-    if (m_InitFailed or (m_CurrentStep == m_FinalStep && m_CurrentStep > 0))
+    if (m_InitFailed)
     {
+        if (m_Verbosity >= 5)
+        {
+            std::cout << "DataManReader::BeginStep() returned EndOfStream due to initialization failure" << std::endl;
```
NAThompson: Bikeshedding: Shouldn't this go to `std::cerr` and replace the `std::endl` with a `\n` to avoid the buffer flush?

JasonRuonanWang: Short answer, I don't know. Maybe you are right. Long answer, in principle yes, this is some kind of error. But the only reason I do this here is for debug purpose. If I have some of the information go to std::cout and other go to std::cerr, then it will mess up the sequence of the events over time, and it's much harder for anyone to make sense out of it. For the same reason, I prefer a buffer flush because I want the outputs to reflect the exact time sequence across all processes / threads for every event that happened.

NAThompson: @JasonRuonanWang : That all sounds very sensible to me; sorry for the noise!

source/adios2/engine/dataman/DataManReader.tcc
```diff
@@ -72,12 +72,7 @@ template <typename T>
...
 DataManReader::AllStepsBlocksInfoCommon(const Variable<T> &variable) const
 {
-    std::map<size_t, std::vector<typename Variable<T>::Info>> m;
-    for (const auto &i : m_MetaDataMap)
-    {
-        m[i.first] = BlocksInfoCommon(variable, i.first);
-    }
-    return m;
+    return std::map<size_t, std::vector<typename Variable<T>::Info>>();
 }
```
NAThompson: Shouldn't this function be removed since it now just returns an empty object?

source/adios2/engine/dataman/DataManReader.cpp
```diff
@@ -113,63 +116,46 @@ StepStatus DataManReader::BeginStep(StepMode stepMode,
...
-            auto now_time = std::chrono::system_clock::now();
-            auto duration = std::chrono::duration_cast<std::chrono::seconds>(
-                now_time - start_time);
-            if (duration.count() > timeoutSeconds)
+            if (m_Verbosity >= 5)
             {
-                return StepStatus::EndOfStream;
+                std::cout << "DataManReader::BeginStep() returned EndOfStream, "
+                             "final step is "
+                          << m_FinalStep << std::endl;
```
NAThompson: How do we decide what type of information goes in what verbosity level? (This is just for my own curiosity; I'm not insinuating you've made the wrong choice here . . .)

source/adios2/toolkit/format/dataman/DataManSerializer.cpp
```diff
@@ -957,16 +966,10 @@ size_t DataManSerializer::MinStep()
     }
     return minStep;
 }
+*/
```
NAThompson: Why not just delete this instead of commenting it out?

source/adios2/toolkit/format/dataman/DataManSerializer.cpp
```diff
@@ -1098,6 +1101,86 @@ void DataManSerializer::SetDestination(const std::string &dest)
...
+    for (const auto &i : m_DataManVarMap)
+    {
+        if (latestStep < i.first)
+        {
+            latestStep = i.first;
+        }
+        hasStep = true;
+    }
+
+    if (not hasStep)
```
NAThompson: I just found out that C++ supports `not` as a keyword alias to `!`! (Logically this is unimpeachable, but sounds like it might be better for detecting compiler bugs.. .)

PR #1708, opened 2019-08-29
Initial integration of azure pipelines

chuckatkins:
This is an initial integration of azure pipelines to flush out various
details.  It will co-exist with the CircleCI tooling for the time being
and likely have some code duplication in the scripts as things are shifted.


chuckatkins:
@williamfgc @pnorbert this should give is a much faster turnaround time and better parallelism in builds.  It also allows for better pay-options if needed as well as the ability to run the CI agents on our own compute resources.

PR #1709, opened 2019-08-29
added query component and its CXX binding

guj:


guj:
I plan to have  this as the basic query support  for the sept release.
  
Working on other features  (e,g,   json support, c binding,  python binding,  etc).  I will add  gradually.  

- Junmin

PR #1712, opened 2019-09-03
Improvements for Query API 

guj:
This PR includes:
1. Adding BP4 minmax based query processing
2. Improving xml tags
3. Adding JSON interface
4. Cleaning up query bindings.

### DIFF
--- a/bindings/CXX11/adios2/cxx11/Query.cpp
+++ b/bindings/CXX11/adios2/cxx11/Query.cpp
@@ -10,6 +10,7 @@ QueryWorker::QueryWorker(const std::string &configFile, adios2::Engine &reader)
     m_Worker = std::make_shared<adios2::query::Worker>(*m);
 }
 
+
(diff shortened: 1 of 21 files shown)


chuckatkins:
@guj would you mind editing the PR's title  to something more descriptive than "0829" ?

guj:
Done 
— Junmin

guj:
Thanks for the catch. The README and sample configs in the example/query/ was not committed in the branch. I added now. 

So yes, test_xml is changed to test to reflect the fact it takes JSON file too. I used heatTxf as my data file for my example,  
To run with json query, need 3 arguments: (The bp4io.xml was from the headTxf example bp4 config file)
e.g. bin/queryTest  bp4io.xml your.data.bp4 jsonQuery.

To run with xml, need 2. And just add <adios-query> tag at the end of  the xml file.
 
JsonQuery examples are in example/query/config,  q1.json is a simple query to look at. 

- Junmin

Issue #1699, opened 2019-08-27
Libfabric Dependency Detection


philip-davis:
I am having an issue building on Cori with libfabric support. I believe this comes down to something I am doing wrong in FindLIBFABRIC.cmake, but I can't figure out what. I am seeing this:

```
[ 40%] Linking CXX executable ../../../../../bin/sst_conn_tool
/usr/bin/ld: cannot find -lpkgcfg_lib_PC_LIBFABRIC_atomic-NOTFOUND
/usr/bin/ld: /global/homes/p/pdavis/cori/sw/libfabric/gnu/lib/libfabric.a(src_libfabric_la-fabric.o): in function `fi_ini':
fabric.c:(.text+0xaa2): warning: Using 'dlopen' in statically linked applications requires at runtime the shared libraries from the glibc version used for linking
collect2: error: ld returned 1 exit status
make[2]: *** [source/adios2/toolkit/sst/util/CMakeFiles/sst_conn_tool.dir/build.make:100: bin/sst_conn_tool] Error 1
make[1]: *** [CMakeFiles/Makefile2:2007: source/adios2/toolkit/sst/util/CMakeFiles/sst_conn_tool.dir/all] Error 2
make: *** [Makefile:141: all] Error 2
```

Which looks to me like cmake is inducing a failure because it thought it needed to find the  absolute path libatomic using pkg-config, but couldn't. I see this in `CMakeCache.txt`:

```
//Path to a library.
pkgcfg_lib_PC_LIBFABRIC_alpslli:FILEPATH=/opt/cray/alps/6.6.50-7.0.0.1_3.35__g962f7108.ari/lib64/libalpslli.a

//Path to a library.
pkgcfg_lib_PC_LIBFABRIC_alpsutil:FILEPATH=/opt/cray/alps/6.6.50-7.0.0.1_3.35__g962f7108.ari/lib64/libalpsutil.a

//Path to a library.
pkgcfg_lib_PC_LIBFABRIC_atomic:FILEPATH=pkgcfg_lib_PC_LIBFABRIC_atomic-NOTFOUND

//Path to a library.
pkgcfg_lib_PC_LIBFABRIC_dl:FILEPATH=/usr/lib64/libdl.a

//Path to a library.
pkgcfg_lib_PC_LIBFABRIC_fabric:FILEPATH=/global/homes/p/pdavis/cori/sw/libfabric/gnu/lib/libfabric.a

//Path to a library.
[...]

PR #1713, opened 2019-09-03
Fixes and Enhancements for RDMA on Cori

philip-davis:
Use DRC on Cori, when available. Delay libfabric initialization on the reader until after we have made contact with the writer.

Fix pkg_config_module, which has been failing to properly import the correct link flags for libfabric on some static builds.

Some general cleanup in the RDMA DP code.


philip-davis:
resolves #1699

PR #1720, opened 2019-09-05
Virtual engines: File (the default engine now), FileStream, InSituAna…

pnorbert:
…lysis, InSituVisualization, Code Coupling

File is now BP4 or HDF5 depending on file extension, or BP3 for reading old BP files. 
New section in doc Supported Virtual Engines.


source/adios2/core/IO.cpp
```diff
@@ -467,24 +514,47 @@ Engine &IO::Open(const std::string &name, const Mode mode, MPI_Comm mpiComm)
...
             if (adios2sys::SystemTools::FileIsDirectory(name))
             {
                 engineTypeLC = "bp4";
             }
             else
             {
-                engineTypeLC = "bp3";
+                if (helper::EndsWith(name, ".bp", false))
+                {
+                    engineTypeLC = "bp3";
```
chuckatkins: Should also probably check if it's a directory or not so you can have `foo.bp` as a file use bp3 and as a directory use bp4.

source/adios2/helper/adiosSystem.cpp
```diff
@@ -105,5 +111,47 @@ int ExceptionToError(const std::string &function)
...
+                                                 13,  10, 26, 10};
+            if (tm.GetFileSize(0) >= 8)
+            {
+                char header[8];
+                tm.ReadFile(header, 8, 0);
+                tm.CloseFiles();
+                isHDF5 = !std::memcmp(header, HDF5Header, 8);
+            }
+        }
+        catch (std::exception &e)
```
chuckatkins: Should only be catching exceptions we know about here so we don't swallow and hide ones we can't deal with.

source/adios2/core/IO.cpp
```diff
@@ -81,9 +82,55 @@ IO::IO(ADIOS &adios, const std::string name, const bool inConfigFile,
 {
 }
 
+#define INSERTPARAM(k, v)                                                      \
```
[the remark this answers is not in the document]
pnorbert: thanks, will fix it

source/adios2/helper/adiosSystem.cpp
pnorbert: I don't want to deal with any error here, any error just means I don't consider this to be a HDF5 file. Let the selected engine (BP3) deal with any error later.

source/adios2/core/IO.cpp
```diff
@@ -12,6 +12,7 @@
 #include "IO.tcc"
 
 #include <sstream>
+#include <utility>
```
williamfgc: @pnorbert just a suggestion, when the C++ header has a broad meaning (e.g. algorithm or utility) I put a comment to explain why it was included. `#include <utility> //std::pair`. Helps me a lot when I have to go back and debug C++ headers. Thing like vector, map, set are self-explanatory.

source/adios2/helper/adiosSystem.cpp
```diff
@@ -105,5 +111,47 @@ int ExceptionToError(const std::string &function)
     }
 }
 
+bool IsHDF5File(const std::string &name, helper::Comm &comm,
+                std::vector<Params> &transportsParameters) noexcept
```
williamfgc: `const std::vector<Params> &transportsParameters` ?

source/adios2/helper/adiosSystem.h
```diff
@@ -72,6 +75,9 @@ bool IsZeroIndexed(const std::string hostLanguage) noexcept;
  */
 int ExceptionToError(const std::string &function);
 
+bool IsHDF5File(const std::string &name, helper::Comm &comm,
+                std::vector<Params> &transportsParameters) noexcept;
```
williamfgc: `const std::vector<Params> &transportsParameters` ?

source/adios2/helper/adiosSystem.cpp
chuckatkins: By catching the exception you are dealing with it.  So this is essentially dealing with all errors, and explicitly doing nothing with them.  It should instead catch the exceptions it we know to occur and do nothing with them, letting all other exceptions propagate and not get swallowed so they can be dealt with accordingly by upstream callers.

pnorbert: Okay, I just catch the file system errors arising from the posix transport operations.

PR #1721, opened 2019-09-05
added a query test in testing/.

guj:


guj:
another pull request was used instead

PR #1741, opened 2019-09-15
Add facility for earlier SST info exchange

eisenhauer:


eisenhauer:
@philip-davis Philip, I tagged you just so that you'd see these patches go in.  This is basically the simple idea I outlined via Email.  There's a new (empty) attribute list in initWriter() that you can add things too (only rank 0 matters).  What you put in that list will appear to a new attribute list parameter in initReader() on all reader ranks.

PR #1744, opened 2019-09-16
enabled mpi oversubscription in tests for Table, Ssc and DataMan engines

JasonRuonanWang:


germasch:
Hardcoding `--oversubscribe` is not portable. I believe this has been discussed here before, though I guess not resolved. One option might be to set `OMPI_MCA_rmaps_base_oversubscribe=1` in the (openmpi) test environments. @chuckatkins

chuckatkins:
Oversubscription is already being enable for openmpi in the CI scripts.  See https://github.com/ornladios/ADIOS2/blob/738739045fd44e0b6200c802b8b63263768c8da9/scripts/circle/runOnCircle.sh#L14

chuckatkins:
@JasonRuonanWang what are you trying to address with this?  Perhaps there's a better way.

JasonRuonanWang:
@chuckatkins I was trying to get these tests passed on my laptop. It doesn't work with export OMPI_MCA_rmaps_base_oversubscribe=1, which I don't know why. But if this is an issue then I can close this PR. It's not very important that I have to pass these tests on my laptop.

PR #1750, opened 2019-09-17
BP serializer refactoring

williamfgc:
Introduce BPSerializer base class
Solve diamond problem
inl for public and protected template specialization separation (specialized in inl, generic in tcc)


williamfgc:
Will retarget release branch

Issue #1760, opened 2019-09-18
Current master is broken when configuring with SZ


williamfgc:
the `fortran_add_test_helper` function doesn't handle the BPWriteReadSZ* tests in the operation subdirectory. The following message appears:

```
 Features:
    Library Type: shared
    Build Type:   Debug
    Testing: ON
    Build Options:
      Blosc    : ON
      BZip2    : ON
      ZFP      : OFF
      SZ       : ON
      MGARD    : ON
      PNG      : ON
      MPI      : ON
      DataMan  : OFF
      Table    : OFF
      SSC      : OFF
      SST      : ON
      DataSpaces: OFF
      ZeroMQ   : ON
      HDF5     : ON
      Python   : ON
      Fortran  : ON
      SysVShMem: ON
      Profiling: ON
      Endian_Reverse: OFF
    RDMA Transport for Staging: Unconfigured

-- Configuring done
CMake Error at testing/adios2/bindings/fortran/CMakeLists.txt:9 (add_executable):
  Cannot find source file:

    TestBPWriteReadSZ3D.F90

  Tried extensions .c .C .c++ .cc .cpp .cxx .cu .m .M .mm .h .hh .h++ .hm
  .hpp .hxx .in .txx
Call Stack (most recent call first):
  testing/adios2/bindings/fortran/CMakeLists.txt:76 (fortran_add_test_helper)


CMake Error at testing/adios2/bindings/fortran/CMakeLists.txt:9 (add_executable):
  Cannot find source file:

    TestBPWriteReadSZ2D.F90

  Tried extensions .c .C .c++ .cc .cpp .cxx .cu .m .M .mm .h .hh .h++ .hm
  .hpp .hxx .in .txx
Call Stack (most recent call first):
  testing/adios2/bindings/fortran/CMakeLists.txt:75 (fortran_add_test_helper)


CMake Error at testing/adios2/bindings/fortran/CMakeLists.txt:9 (add_executable):
  No SOURCES given to target: Test.Bindings.Fortran.BPWriteReadSZ3D
Call Stack (most recent call first):
  testing/adios2/bindings/fortran/CMakeLists.txt:76 (fortran_add_test_helper)


CMake Error at testing/adios2/bindings/fortran/CMakeLists.txt:9 (add_executable):
  No SOURCES given to target: Test.Bindings.Fortran.BPWriteReadSZ2D
Call Stack (most recent call first):
[...]

Issue #1742, opened 2019-09-16
Current master is broken when configuring with zfp


williamfgc:
when using `-DADIOS2_USE_ZFP=ON` the following error shows up:

```
-- Verifying Fortran/CXX Compiler Compatibility - Success
CMake Error at testing/examples/heatTransfer/TestBPFileMx1_zfp.cmake:6 (include):
  include could not find load file:

    ADIOS2Functions
Call Stack (most recent call first):
  testing/examples/heatTransfer/CMakeLists.txt:13 (include)


CMake Error at cmake/ADIOSFunctions.cmake:165 (set_property):
  set_property given TEST names that do not exist:

    HeatTransfer.BP3.zfp.Mx1.Validate

Call Stack (most recent call first):
  testing/examples/heatTransfer/TestBPFileMx1_zfp.cmake:33 (SetupTestPipeline)
  testing/examples/heatTransfer/CMakeLists.txt:13 (include)


CMake Error at cmake/ADIOSFunctions.cmake:168 (set_property):
  set_property given TEST names that do not exist:

    HeatTransfer.BP3.zfp.Mx1.Validate

Call Stack (most recent call first):
  testing/examples/heatTransfer/TestBPFileMx1_zfp.cmake:33 (SetupTestPipeline)
  testing/examples/heatTransfer/CMakeLists.txt:13 (include)


CMake Error at cmake/ADIOSFunctions.cmake:171 (set_property):
  set_property given TEST names that do not exist:

    HeatTransfer.BP3.zfp.Mx1.Validate

Call Stack (most recent call first):
  testing/examples/heatTransfer/TestBPFileMx1_zfp.cmake:33 (SetupTestPipeline)
  testing/examples/heatTransfer/CMakeLists.txt:13 (include)


CMake Error at cmake/ADIOSFunctions.cmake:165 (set_property):
  set_property given TEST names that do not exist:

    HeatTransfer.BP4.zfp.Mx1.Validate

Call Stack (most recent call first):
  testing/examples/heatTransfer/TestBPFileMx1_zfp.cmake:61 (SetupTestPipeline)
  testing/examples/heatTransfer/CMakeLists.txt:13 (include)


CMake Error at cmake/ADIOSFunctions.cmake:168 (set_property):
  set_property given TEST names that do not exist:

    HeatTransfer.BP4.zfp.Mx1.Validate

Call Stack (most recent call first):
[...]

PR #1763, opened 2019-09-18
Add compression to CI images

chuckatkins:


chuckatkins:
Fixes #1742 
Fixes #1760

chuckatkins:
@williamfgc I've added ZFP and SZ to the CI images and they should get enabled and tested now.

williamfgc:
@chuckatkins this is great, thanks for fixing these promptly.

PR #1764, opened 2019-09-19
Use auto-built docker images

chuckatkins:


chuckatkins:
@eisenhauer https://open.cdash.org/testDetails.php?test=805856481&build=6118583 looks to be a legit failure, not just a "heisenbug" timeout.

eisenhauer:
Hmm.  The "Cannot bind INET socket" is the cause of the failure here.  That means a bind() operation failed, which may mean too many things running on that host.  So, likely environmental in some way.  (Not that it couldn't maybe be handled better...)

chuckatkins:
> Hmm. The "Cannot bind INET socket" is the cause of the failure here. That means a bind() operation failed, which may mean too many things running on that host.

Right.  It shouldn't be causeing a segfault though.

eisenhauer:
We can make it fail in another way, but fail it will, one way or another.  If we can’t bind sockets, we can’t communicate...

Sent from my iPhone

On Sep 19, 2019, at 9:59 AM, Chuck Atkins <notifications@github.com<mailto:notifications@github.com>> wrote:


Hmm. The "Cannot bind INET socket" is the cause of the failure here. That means a bind() operation failed, which may mean too many things running on that host.

Right. It shouldn't be causeing a segfault though.

eisenhauer:
There are a whole lot of “this should fail better” things that need to be addressed, but I’ll admit that they’ve been given lower priority than functionality and performance.

Sent from my iPhone

On Sep 19, 2019, at 9:59 AM, Chuck Atkins <notifications@github.com<mailto:notifications@github.com>> wrote:


Hmm. The "Cannot bind INET socket" is the cause of the failure here. That means a bind() operation failed, which may mean too many things running on that host.

Right. It shouldn't be causeing a segfault though.

eisenhauer:
I've opened an issue on making SST fail better.  Some of these failures can created relatively easily, but the bind() failure not so much.  So to the extent that we can recreate how you got this, that would be useful.

chuckatkins:
It's near impossible it seems to reliably reproduce these oversubscription->timeout issues.  I think they will essentially disappear as I get all the other tests marked with the "processes" property.

PR #1767, opened 2019-09-20
Added more test macro wrapping

chuckatkins:


chuckatkins:
@pnorbert You can see the HeatTransfer tests running here: https://open.cdash.org/viewTest.php?onlypassed&buildid=6120602

Turns out I had introduced a bug with the `add_subdirectory` order for tests and examples.  It's resolved within this PR.

Issue #1745, opened 2019-09-16
null pointer in MinMax


philip-davis:
I'm doing a large streaming file run with gray-scott coupled to pdf_calc on Theta, and I am sometimes (not always) seeing pdf_calc crash with many of the following strangely-formatted errors:

```
terminate called after throwing an instance of 'terminate called after throwing an instance of 'terminate called after throwing an instance of 'std::invalid_argument'
std::invalid_argument'
std::invalid_argument'
  what():  ERROR: found null pointer in call to Variable<T>::MinMax
```

This is using BP4 with the following configuration (I have combined SST and BP4 parameters for reuse:

```
    <io name="SimulationOutput">
        <engine type="BP4">
            <!-- SST engine parameters -->
            <parameter key="RendezvousReaderCount" value="1"/>
            <parameter key="QueueLimit" value="1"/>
            <parameter key="QueueFullPolicy" value="Block"/>
            <!-- BP4/SST engine parameters -->
            <parameter key="OpenTimeoutSecs" value="900"/>
            <parameter key="BeginStepPollingFrequencySecs" value="1" />
            <parameter key="SubStreams" value="512"/>
        </engine>
    </io>
```

This occurs on the first timestep of the reader. The writer runs to completion. For reference, here is some of the pdf_calc code that precedes the MinMax call:
https://github.com/pnorbert/adiosvm/blob/75bf69b13638f7c67981f43d269d2a19e269da20/Tutorial/gray-scott/analysis/pdf_calc.cpp#L182-L216

philip-davis:
This is on the /projects directory of Theta, so it is a Lustre file system.

JasonRuonanWang:
@philip-davis Is this happening only with BP4 engine? Or SST as well? If it's only BP4 then probably either @pnorbert or @lwan86 is the best person to ask.

JasonRuonanWang:
But even if it's only using BP4 engine, I would still like to ask some further questions just for them to consider. Were you doing staging through files, or purely files? Could you help verify if it happens only in one of the two cases? Or does it happen in both cases?

williamfgc:
@philip-davis thanks for reporting this. If it helps, the error states that the variable is null (false in our API). It'd be good practice to check the variable status after a call to `InquireVariable`, it would make the code safer. Why is not valid is a matter of debugging and getting the stream variable for that step.

philip-davis:
[...]

PR #1772, opened 2019-09-21
WIP: Do not modify steps information in BeginStep if the status is not OK.

pnorbert:
BP4 BeginStep did not handle stepping properly in case of timeout, leading to invalid steps and failing InquireVariable() calls in applications. This PR alone may or may not solve the problem in #1745


pnorbert:
This PR is obsolete as the fix is included by #1733

Issue #1314, opened 2019-03-22
Span.Data() alignment


germasch:
I'm not sure you'll consider that a bug, but Variable::Span<T> will on my Mac return addresses that are aligned by 4 bytes only. While x86_64 can handle misaligned types, I believe in general the requirement is for malloc etc to return addresses aligned by 16 bytes on MacOS, Linux and Win64, so they can potentially be used with SSE2 etc.

williamfgc:
@germasch if this becomes a problem we'll take a look. As far as I know (I haven't tested this), memcpy should be safe, but I don't know how each compiler handles this situation.

germasch:
x86_64 handles misaligned access, except in certain (SSE2) contexts. I suppose Xeon Phi is similar. It's, however, not good for performance -- you might end up having a double sitting on two cache lines, even. memcpy of course is safe, but you're presumably handing the Span to the user so that they'll use it to calculate their output on the fly (ie., they'll actually access it as double, not via memcpy).

From [IBM](https://www.ibm.com/developerworks/library/pa-dalign/index.html):
> On the other hand, modern PowerPC processors lack hardware support for unaligned 64-bit floating-point access. When asked to load an unaligned floating-point number from memory, modern PowerPC processors will throw an exception and have the operating system perform the alignment chores in software. Performing alignment in software is much slower than performing it in hardware.

( I don't know if that applies to Summit, but I'd guess so.)

germasch:
Just to be clear, I don't know if the lack of alignment is an actual issue, either performance wise or even correctness. (As an example,I believe SSE2 `movaps` can cause a segfault or bus error on unaligned data, but if no-one wants to use Span in that fashion, then it's not a problem.)

(I do think that users of Span are likely to use accesses in terms of the actual underlying data type, e.g., double. If one were just using it to memcpy to it, it'd seem much more straightforward to not use Span in the first place and have the memcpy happen in Put(Sync/Deferred). 

I also don't know whether the alignment is easily fixable. If BP3 can handle padding, it might be easy to fix this issue and return properly aligned buffers.

eisenhauer:
[...]

PR #1777, opened 2019-09-23
BP4 span

williamfgc:
Support for zero-copy access to BP4 buffer at write
Added padding for pointer alignment in BP3/BP4 buffers addressing #1314 
Replacing `std::align` with internal function due to [gcc 4.8 bug](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=57350)

### DIFF
--- a/source/adios2/toolkit/format/buffer/heap/BufferSTL.tcc
+++ b/source/adios2/toolkit/format/buffer/heap/BufferSTL.tcc
@@ -23,7 +23,8 @@ namespace format
 template <class T>
 size_t BufferSTL::Align() const noexcept
 {
-    void *currentAddress = m_Buffer.data() + m_Position;
+    void *currentAddress = reinterpret_cast<void *>(
+        const_cast<char *>(m_Buffer.data() + m_Position));
     size_t size = GetAvailableSize();
     std::align(alignof(T), sizeof(T), currentAddress, size);
     return GetAvailableSize() - size;
--- a/source/adios2/toolkit/format/buffer/heap/BufferSTL.tcc
+++ b/source/adios2/toolkit/format/buffer/heap/BufferSTL.tcc
@@ -27,8 +27,8 @@ namespace format
 template <class T>
 size_t BufferSTL::Align() const noexcept
 {
-    // LLVM MIT license
-    // https://github.com/llvm-mirror/libcxx/blob/6952d1478ddd5a1870079d01f1a0e1eea5b09a1a/src/memory.cpp#L217
+    // std::align implementation from llvm libc++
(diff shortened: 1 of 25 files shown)


source/adios2/toolkit/format/buffer/heap/BufferSTL.tcc
```diff
@@ -0,0 +1,51 @@
...
+        if ((size + diff) <= space)
+        {
+            space -= diff;
+        }
+    };
+
+    void *currentAddress = reinterpret_cast<void *>(
+        const_cast<char *>(m_Buffer.data() + m_Position));
+    size_t size = GetAvailableSize();
+#if __GNUC__ > 5
```
chuckatkins: Should this check be `#if (defined __GNUC__) && (__GNUC__ < 5)`?

source/adios2/toolkit/format/buffer/heap/BufferSTL.tcc
```diff
@@ -0,0 +1,51 @@
...
+    auto lf_align = [](const size_t align, const size_t size, void *&ptr,
+                       size_t &space) {
+        const uintptr_t largePtr = reinterpret_cast<uintptr_t>(ptr);
+        const uintptr_t aligned = (largePtr - 1u + align) & -align;
+        const uintptr_t diff = aligned - largePtr;
+        if ((size + diff) <= space)
+        {
+            space -= diff;
+        }
+    };
```
chuckatkins: I'm a bit concerned with the implementation as it seems to be pulled from or at least derived from the implementation in libstdc++, which is gpl or lgpl.  Can we grab the implementation directly from the llvm libc++ as to avoid any license issues?

source/adios2/toolkit/format/buffer/heap/BufferSTL.tcc
williamfgc: @chuckatkins thanks, I am removing this and just disable the VS warning. It's a gcc 4.8 bug.

source/adios2/toolkit/format/buffer/heap/BufferSTL.tcc
williamfgc: will do. Thanks for the reminder on licenses.

williamfgc: Moved to the llvm implementation which is MIT licensed.

chuckatkins: We just need a comment to note attribution.  Something like this should be sufficient:
```
// std::align implementation from llvm libc++
```
Thanks for making the change

PR #1778, opened 2019-09-23
Update EVPath to use epoll() rather than select()

eisenhauer:


chuckatkins:
I think this is reasonably safe to go ahead and have in the release, esp. wrt some of the other changes.

chuckatkins:
@eisenhauer @philip-davis if you're good with it them let's drop the WIP and retarget this for release

eisenhauer:
I think I'm good with it.  Keichi was going to try this with the JAXA code, but I think hasn't managed that yet.  If we merge and he comes back reporting problems, we could quickly change the default back to select...

chuckatkins:
@eisenhauer this will need a different base targeting release as it's currently off master and with pull in stuff we don't want.  Please rebase off release if possible.

chuckatkins:
@eisenhauer I've just reconstructed this off the tip of release.  Please let me know if this is acceptable.

eisenhauer:
Looks good to me...

PR #1779, opened 2019-09-24
added helper::StringToDims

JasonRuonanWang:


source/adios2/helper/adiosString.h
```diff
@@ -120,6 +120,8 @@ void SetParameterValue(const std::string key, const Params &parameters,
  */
 std::string DimsToString(const Dims &dimensions);
 
+Dims StringToDims(const std::string &dimentions);
```
chuckatkins: Typo in argument name in both header and cpp: `dimentions` should be `dimensions`

PR #1781, opened 2019-09-24
Add rapidjson to thirdparty

chuckatkins:
This adds RapidJSON as a header-only third party library.

@JasonRuonanWang you should be able to use it by adding the `RapidJSON` target to the private `target_link_libraries` list of a given target.


JasonRuonanWang:
Awesome! Thanks very much!

chuckatkins:
I'll hold off on merging this until you can at least verify that you can use the target and include the necessary headers.

JasonRuonanWang:
@chuckatkins It worked for me.

PR #1783, opened 2019-09-24
Default to BP3 until read bug is resolved

chuckatkins:


pnorbert:
This PR is not needed as BP4 reading is solved

PR #1785, opened 2019-09-25
ci: Use a 90m timeout for azure pipelines

chuckatkins:


chuckatkins:
@williamfgc this should fix the 60m timeout you encountered the other day

PR #1792, opened 2019-09-28
Revert version bump until the release is tagged

chuckatkins:


chuckatkins:
@williamfgc We want to leave the actual library versioning as is until it's time to tag the release.  The version bump should be the last commit before the release is tagged.

PR #1797, opened 2019-09-30
Add version flags to adios2-config

chuckatkins:


chuckatkins:
@pnorbert is that what you were looking for?

pnorbert:
Yes, thank you for adding this.

PR #1817, opened 2019-10-08
Encapsulate multi-process communication (part 2)

bradking:
Continue work started in #1691 to encapsulate multi-process communication behind a `helper::Comm` API:

* Encapsulate remaining MPI types and constants inside `helper::Comm` and offer equivalents publicly.
* Convert more code from direct MPI use over to `helper::Comm`.  Add APIs to the latter as needed.  * Abstract the `helper::Comm` implementation internally to work with either MPI or a dummy backend.
* Replace MPI with `helper::Comm` in core `ADIOS`, `IO`, and `Stream` interfaces.
* Minimize direct inclusion of `ADIOSMPI.h` to only places that still use MPI directly.

### DIFF
--- a/source/adios2/core/IO.cpp
+++ b/source/adios2/core/IO.cpp
@@ -29,6 +29,7 @@
 #include "adios2/engine/skeleton/SkeletonWriter.h"
 
 #include "adios2/helper/adiosComm.h"
+#include "adios2/helper/adiosCommMPI.h"
 #include "adios2/helper/adiosFunctions.h" //BuildParametersMap
 #include "adios2/toolkit/profiling/taustubs/tautimer.hpp"
 #include <adios2sys/SystemTools.hxx> // FileIsDirectory()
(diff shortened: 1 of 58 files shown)


source/adios2/core/IO.cpp
```diff
@@ -765,7 +763,7 @@ Engine &IO::Open(const std::string &name, const Mode mode, MPI_Comm mpiComm)
 
 Engine &IO::Open(const std::string &name, const Mode mode)
 {
-    return Open(name, mode, m_ADIOS.GetComm().AsMPI());
+    return Open(name, mode, m_ADIOS.GetComm().Duplicate());
```
chuckatkins: Rather than an explicit `Duplicate`, would it instead make sense to pass the comm object by value and have `Duplicate` be implemented in the copy constructor?

bradking: `helper::Comm` is explicitly move-only and has no copy constructor.  This avoids hiding non-trivial operations like `MPI_Comm_dup` behind simple-looking syntax like a (possibly unintentional) copy.

PR #1818, opened 2019-10-09
integration with mgard 0.0.0.2

qliu21:
To integrate MGARD 0.0.0.2


williamfgc:
@qliu21 @pnorbert we also need to add [the MGARD tests](https://github.com/ornladios/ADIOS2/blob/master/testing/adios2/engine/bp/operations/TestBPWriteReadMGARD.cpp) to the CI

qliu21:
OK. They removed the BLOSC related changes in their release, and I plan to add them back to MGARD. I will work on new tests as well.

Issue #1825, opened 2019-10-15
Downstream RPATH overwritten


ax3l:
I am using the `ADIOS2Config.cmake` module for v2.5.0 downstream in openPMD-api.

Somehow, the ADIOS2 `CMakeLists.txt` lines overwrite my *downstream* [`-DCMAKE_INSTALL_RPATH=$ORIGIN`](https://github.com/openPMD/openPMD-api/blob/130f7d3ceef25a1412515987728209ccd0ecb31e/setup.py#L60) setting to read `$ORIGIN/.libsopenpmd_api` instead of `$ORIGIN`.

https://github.com/ornladios/ADIOS2/blob/v2.5.0/CMakeLists.txt#L156-L179

As soon as I am not linking to ADIOS2 anymore, the problem is gone.

Probably we should only set the rpath as CMake property on targets or append to already set, user-provided `CMAKE_INSTALL_RPATH` locations?

ax3l:
the problem is caused downstream by my setuptools scripts.

Reason: `libffi-2009cab3.so.5.0.6` not packaged. Sorry for the noise!

ax3l:
related question: is `libffi` always shared, even if we try to build ADIOS2 as static library?

And if yes, what functionality do I loose if I disable it?

eisenhauer:
@chuckatkins Can say more about the static/shared aspects.  But, libffi is used by dill, which is used by FFS/EVPath for dynamic code generation.  This DCG is used to reduce data transformation overhead in heterogeneous data exchanges, and for remote data transformation/filtering/reduction/subsampling, etc.  But as you've seen in other issues, easy fast support for heterogeneity isn't as much a priority in ADIOS, and there's no ADIOS API for remote data reduction/filtering, so this functionality is largely unused.  libffi comes into the picture when DILL doesn't have a native instruction port.  ffi is then used to create an emulation so EVPath could do still the data reduction, albeit slower than DCG speeds.  libffi *should* only be pulled in on unsupported architectures (x86 is supported).  So, aside from the shared/static nature of its inclusion, I'm curious why it's been pulled in to begin with.  Is this for Conda?  Which platform?

ax3l:
Thanks for the details, this looks interesting for future use!

[...]

PR #1826, opened 2019-10-15
[WIP] CMake RPath: Do not Overwrite User

ax3l:
Do not overwrite user-specified options.

Trying to address #1825 

- [ ] not yet verified this solves the reported problem


ax3l:
does not fix my problem, I do not understand why this pops up as soon as I link against ADIOS2...

Issue #1810, opened 2019-10-06
ADIOS 2.5.0: Static Build Issues Linking Blosc


ax3l:
I am building a static version of ADIOS2 2.5.0 against a static version of C-Blosc 1.15.0 for a PyPI release of openPMD-api with the following CMake flags:

```
-DBUILD_SHARED_LIBS=OFF
-DCMAKE_POSITION_INDEPENDENT_CODE=ON
-DADIOS2_BUILD_EXAMPLES=OFF
-DADIOS2_BUILD_TESTING=OFF
```

with results in the config
```
ADIOS2 build configuration:
  ADIOS Version: 2.5.0
  C++ Compiler : GNU 8.3.1 
    /opt/rh/devtoolset-8/root/usr/bin/c++

  Fortran Compiler : GNU 8.3.1 
    /opt/rh/devtoolset-8/root/usr/bin/gfortran

  Installation prefix: /usr
        bin: bin
        lib: lib64
    include: include
      cmake: lib64/cmake/adios2

  Features:
    Library Type: static (with PIC)
    Build Type:   Release
    Testing: OFF
    Examples: OFF
    Build Options:
      Blosc    : ON
      BZip2    : OFF
      ZFP      : OFF
      SZ       : OFF
      MGARD    : OFF
      PNG      : OFF
      MPI      : OFF
      DataMan  : OFF
      Table    : OFF
      SSC      : OFF
      SST      : ON
      DataSpaces: OFF
      ZeroMQ   : OFF
      HDF5     : ON
      Python   : OFF
      Fortran  : ON
      SysVShMem: ON
      Profiling: OFF
      Endian_Reverse: OFF
    RDMA Transport for Staging: Unconfigured
-- Configuring done
-- Generating done
```

Just or the record, this is the [Docker environment](https://github.com/openPMD/openPMD-api/blob/dev/Dockerfile) I am building in.

ADIOS2 builds the library, but fails in the `adios-config` step with:
```
Configuring adios2-config dummy CMake project
-- The C compiler identification is GNU 8.3.1
-- The CXX compiler identification is GNU 8.3.1
-- Check for working C compiler: /opt/rh/devtoolset-8/root/usr/bin/cc
-- Check for working C compiler: /opt/rh/devtoolset-8/root/usr/bin/cc -- works
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Detecting C compile features
-- Detecting C compile features - done
[...]

PR #1824, opened 2019-10-14
Blosc: Fix Static Builds

ax3l:
Adds search functionality for transitive blosc dependency in static builds.

Fix #1810

- [x] verified this solves the reported problem

PR #1828, opened 2019-10-15
Add a static ci build

chuckatkins:


chuckatkins:
Includes the static blosc fix from @ax3l in #1824

ax3l:
just a note: As a maintainer, you can push to open PRs in ADIOS2 ;)

eisenhauer:
> LGTM.
> Can you check that libffi is build/used static as well? Maybe @eisenhauer knows how to set this, there seam some options in DILL for this.

Chuck has reworked all the EVPath stack CMake stuff to be compatible with how ADIOS2 does things now.  So he's likely a far better authority than me!

chuckatkins:
Fixes #1810

PR #1834, opened 2019-10-16
WIP: More Azure Pipelines migration

chuckatkins:


KyleFromKitware:
@chuckatkins I just pushed a change to pass the compiler on to the install test.

KyleFromKitware:
Hmmm. Still didn't work. I'll need to do some more testing.

PR #1835, opened 2019-10-17
To allow s parameter of mgard to be specified in XML.

qliu21:


qliu21:
@williamfgc Thanks for the help on git merging. Please merge this. Thanks!

PR #1279, opened 2019-03-11
Fix the missing install name dir on osx

HaochengLIU:
This commit allows installed adios2 executables to find its
dependencies properly on osx.

Close #1220

PR #1845, opened 2019-10-30
Pass build configuration in install test

KyleFromKitware:
In order to work with multi-config generators, the install test needs
to know the current working configuration. Pass this information using
the `$<CONFIG>` generator expression.

### DIFF
--- a/testing/install/CMakeLists.txt
+++ b/testing/install/CMakeLists.txt
@@ -12,6 +12,7 @@ function(add_install_cmake_test name)
     "-DTEST_CASE=${name}"
     "-DCMAKE_INSTALL_BINDIR=${CMAKE_INSTALL_BINDIR}"
     "-DCMAKE_INSTALL_CMAKEDIR=${CMAKE_INSTALL_CMAKEDIR}"
+    "-DBUILD_TYPE=$<CONFIG>"
     -P "${CMAKE_CURRENT_SOURCE_DIR}/run_cmake.cmake"
     )
   set_tests_properties(Install.CMake.${name} PROPERTIES
@@ -54,6 +55,7 @@ endfunction()
 add_test(NAME Install.Setup COMMAND "${CMAKE_COMMAND}"
   "-DADIOS2_SOURCE_DIR=${CMAKE_SOURCE_DIR}"
   "-DADIOS2_BINARY_DIR=${CMAKE_BINARY_DIR}"
+  "-DBUILD_TYPE=$<CONFIG>"
   -P "${CMAKE_CURRENT_SOURCE_DIR}/run_install.cmake"
   )
 set_tests_properties(Install.Setup PROPERTIES
--- a/testing/install/run_cmake.cmake
+++ b/testing/install/run_cmake.cmake
@@ -16,6 +16,7 @@ execute_process(COMMAND "${CMAKE_CTEST_COMMAND}"
   --build-options
     "-Dadios2_DIR=${ADIOS2_BINARY_DIR}/testing/install/install/${CMAKE_INSTALL_CMAKEDIR}"
   --test-command "${CMAKE_CTEST_COMMAND}" -V
+  -C "${BUILD_TYPE}"
   RESULT_VARIABLE result
   )
 if(result)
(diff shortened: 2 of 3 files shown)


KyleFromKitware:
> the install tests are still broken on mac for other reasons

Yes, it looks like the install name isn't being set correctly. I'm going to submit a separate PR to fix that.

KyleFromKitware:
It looks like 5f65c14da is what is causing the install name error on macOS:

```cmake
# Some specialized shared library and RPATH handling
if(BUILD_SHARED_LIBS)
  set(CMAKE_INSTALL_RPATH_USE_LINK_PATH TRUE)

  if(APPLE)
    # On Mac OS X, set the dir included as part of the installed library's path:
    if(NOT DEFINED CMAKE_INSTALL_NAME_DIR)
      set(CMAKE_INSTALL_NAME_DIR
        "${CMAKE_INSTALL_PREFIX}/${CMAKE_INSTALL_LIBDIR}")
    endif ()

    # Various homebrew MPI fortran installations break rpath usage on OSX
    if(APPLE AND ADIOS2_HAVE_Fortran AND ADIOS2_HAVE_MPI)
      set(CMAKE_MACOSX_RPATH OFF)
    endif()

  elseif(CMAKE_SYSTEM_NAME MATCHES "Linux")
    # Linux needs some specialized RPATH handling
    string(REGEX REPLACE "[^/]+" ".." relative_base "${CMAKE_INSTALL_LIBDIR}")
    set(CMAKE_INSTALL_RPATH "$ORIGIN/${relative_base}/${CMAKE_INSTALL_LIBDIR}")
  endif()
endif()
```

As best as I can tell, none of this should be necessary, as CMake handles rpath stuff pretty well on its own.

Unfortunately, that commit is a mega-commit consisting of a bunch of squashed commits, and I have no way of tracking down the issue that this code was meant to solve. (I could just remove the code, but I don't want to break whatever prompted it to be added in the first place.)

chuckatkins:
> As best as I can tell, none of this should be necessary, as CMake handles rpath stuff pretty well on its own.
> 
> Unfortunately, that commit is a mega-commit consisting of a bunch of squashed commits, and I have no way of tracking down the issue that this code was meant to solve. (I could just remove the code, but I don't want to break whatever prompted it to be added in the first place.)

If you trace back the history a bit, the commits in question are

- a80afca4 for the rpath handling
  - This is needed explicitly because of how the gfortran dependencies are handled on mac, but I believe it's not related to the issue in question.
- 0c747507 for the INSTALL_NAME_DIR handling
  - @HaochengLIU I don't recall why this was put in.  Do you remember at all?

HaochengLIU:
I remember the code is needed to allows installed adios2 executables to find its
dependencies properly on osx. CMake **cannot** handles rpath stuff pretty well on its own at that time..? Nobert and I hit the same wall years ago. 

If now it's redundant, feel free to remove it.
See details in #1279 .

KyleFromKitware:
OK, I've been able to fix it locally by taking an approach similar to the Linux approach of using `$ORIGIN`:

```cmake
set(CMAKE_INSTALL_RPATH "@loader_path/${relative_base}/${CMAKE_INSTALL_LIBDIR}")
```

However, this is broken by the gfortran logic in a80afca, which sets `MACOSX_RPATH` to `FALSE`. Why does this break with gfortran, and is there an alternative that will allow rpath to work with gfortran?

chuckatkins:
> OK, I've been able to fix it locally by taking an approach similar to the Linux approach of using `$ORIGIN`:
> 
> ```cmake
> set(CMAKE_INSTALL_RPATH "@loader_path/${relative_base}/${CMAKE_INSTALL_LIBDIR}")
> ```
> 
> However, this is broken by the gfortran logic in [a80afca](https://github.com/ornladios/ADIOS2/commit/a80afca421865d49f850261e8d8989bf356eb816), which sets `MACOSX_RPATH` to `FALSE`. Why does this break with gfortran, and is there an alternative that will allow rpath to work with gfortran?

Iirc most of the mpif90 compiler wrappers add a link flag that breaks rpath functionality.

KyleFromKitware:
> Iirc most of the mpif90 compiler wrappers add a link flag that breaks rpath functionality.

Sort of. `mpif90` passes `-Wl,-flat_namespace`, which, in conjunction with [this](http://www.openradar.me/25313838) horrendous bug in Xcode, breaks rpath functionality (it should work in theory.)

I will look for a solution that does not involve rpaths.

chuckatkins:
You could also move the install_name stuff into the conditional that disables rpath, then in the else set it properly.  And then run the install tests on the same conditional.

chuckatkins:
The windows install tests are still failing: https://open.cdash.org/testDetails.php?test=818514303&build=6187477

chuckatkins:
Replaced by #1853

PR #1847, opened 2019-10-31
added common-staging tests back for SSC

JasonRuonanWang:


testing/adios2/engine/staging-common/TestCommonReadAttrs.cpp
```diff
@@ -293,16 +293,19 @@ TEST_F(CommonReadTest, ADIOS2CommonRead1D8)
...
     {
+        if (TimeGapExpected)
+        {
+            EXPECT_TRUE(TimeGapDetected);
+        }
+        else
+        {
+            EXPECT_FALSE(TimeGapDetected);
+        }
         EXPECT_EQ(t, NSteps);
```
eisenhauer: Would you move the "EXPECT_EQ(t, NSteps)" outside the "if (!IgnoreTimeGap)"  to make sure we're checking that we got all the expected steps (even if we're ignoring writer side test delays).

JasonRuonanWang:
Some tests passed everywhere except CI. I don't understand what's wrong. I will close this PR for now.

PR #1849, opened 2019-10-31
Consolidate mpi/nompi bpwriter examples into single sources for each

chuckatkins:


chuckatkins:
@williamfgc I wasn't sure how to adjust the fortran examples since I don't think the `ADIOS2_HAVE_*` preprocesser definitions are exposed through the module.  If you know of a good way to achieve a single source for the fortran example I'd welcome it.

chuckatkins:
From what I've been able to find it seems that if the `.F90` extension is used rather than `.f90` then most Fortran compilers invoke `fpp` by convention.  See updated commit.

williamfgc:
@chuckatkins sounds great.

PR #1876, opened 2019-11-12
Fortran 90 and C adios2_set_selection on local array variables

### LINKED ISSUE #1874: Fortran API + local arrays + set_selection leads to segfault
**Describe the bug**
The Fortran API supports defining local arrays by using the constant `adios2_null_dims` in both shape and start. When changing the size of the local array later with 
`adios2_set_selection(var, 2, adios2_null_dims, (/ nsan, 3 /))`
then the code will segfault during write. 

The problem is that the Fortran function for set_selection does not handle adios2_null_dims and will pass a {-1,0} vector<Dims> to the C++ API, which then will change the variable. Offset from empty vector to this non-empty vector.
`adios2_f2c_variable.cpp:adios2_set_selection_f2c(), line 175.

During writing, in BPSerializer::PutDimensionsRecord() the false branch is executed (believing it's a global array) and BPSerializer.cpp:228 will segfault.

It is not clear to me, which of these functions should be fixed and whether the fix would cause a side effect unknown to me, especially in the serializer.

**To Reproduce**
A test needs to be written. The existing fortran local array test needs to be extended with using adios2_set_selection().

williamfgc:
Fixes #1874

### DIFF
--- a/bindings/C/adios2/c/adios2_c_variable.cpp
+++ b/bindings/C/adios2/c/adios2_c_variable.cpp
@@ -111,15 +111,16 @@ adios2_error adios2_set_selection(adios2_variable *variable, const size_t ndims,
         adios2::helper::CheckForNullptr(variable,
                                         "for adios2_variable, in call to "
                                         "adios2_set_selection");
-        adios2::helper::CheckForNullptr(start, "for start, in call to "
-                                               "adios2_set_selection");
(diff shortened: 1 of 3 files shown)


chuckatkins:
I just updated the text of the message to be "Fixes ISSUE_NUMBER", that way github will automatically close the issue once this is merged.

PR #1881, opened 2019-11-12
Switch to generated target file names for staging-common

### LINKED ISSUE #1843: Staging tests don't work with multiconfig generators
The staging tests are expecting executables to be in bin/Foo whic is not the case for multi-config generators like Visual Studio and XCode.  See https://open.cdash.org/testDetails.php?test=816980556&build=6179972

The test commands likely need to be adapted to use generator expressions like `$<TARGET_FILE:Foo>` instead of assuming `${CMAKE_RUNTIME_OUTPUT_DIRECTORY}/Foo`

eisenhauer:
Fixes #1843

@chuckatkins Should be immaterial otherwise.  Note:  In order to get both the name of the python executable and the generator expressions into run_test.py, the file run_test.in gets passed through both configure_file and file(GENERATE).  If there's a better way to do that, please advise.


chuckatkins:
This looks right.  We don't have any multiconfig generators on CI right now.  I'll pull it into my WIP branch though to make sure it works.

chuckatkins:
Also, I just updated the text of the message to include "Fixes ISSUE_NUMBER", that way github will automatically close the issue once this is merged.

PR #1882, opened 2019-11-13
added exceptions in insitumpi

JasonRuonanWang:
The insitumpi engine will return successfully when there is only writer (or reader) launched. This is a little confusing when the MPI environment is getting complicated, for example, running cheetah workflow on Summit. In this case it's very hard for users to directly tell whether the writers and readers are launched using correct split MPI communicators. So even if they are not launched correctly, both the writers and readers are still going to finish successfully, but of course in this case data is not going to be correct at all. So it's better to throw an exception when there is no writers or no readers launched.

### DIFF
--- a/source/adios2/engine/insitumpi/InSituMPIReader.cpp
+++ b/source/adios2/engine/insitumpi/InSituMPIReader.cpp
@@ -48,7 +48,9 @@ InSituMPIReader::InSituMPIReader(IO &io, const std::string &name,
         insitumpi::AssignPeers(m_ReaderRank, m_ReaderNproc, m_RankAllPeers);
     if (m_RankAllPeers.empty())
     {
-        throw(std::runtime_error("no writers are found"));
+        throw(std::runtime_error(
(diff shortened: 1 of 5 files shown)


chuckatkins:
Is it feasible to add a simple test case for this to ensure the exception gets thrown?

JasonRuonanWang:
@chuckatkins Should be feasible. Let me work on that.

JasonRuonanWang:
@pnorbert Some staging-common tests assume InsituMPI can be executed with only writers or only readers. So if I add exceptions in this case, then those staging-common tests will fail. It may need some work to modify the staging-common testing list. Do you still want to go ahead? If not I can drop this PR.

pnorbert:
I added those tests to make sure they work fine as expected. Now the
expected behavior is that they throw and exception. Can you modify the
tests accordingly? Thanks

JasonRuonanWang:
@chuckatkins This has been stuck since last night. How can I restart it? I don't see a link which I can click or anything.

PR #1900, opened 2019-12-13
Add DP-level reader-side timestep release call

eisenhauer:


philip-davis:
This looks good to me. Not an issue with this PR, but is there a reason to guard the FreeTimestep in SstReleaseStep with a mutex, but not the one in waitForNextMetadata, or is that just an artifact?

eisenhauer:
waitForNextMetadata has a mutex lock that's around more than just the Free call (because wait is traversing more data structures).

philip-davis:
Got it, thanks

PR #1901, opened 2019-12-13
Change some locations where checking DebugMode does more harm than good.

eisenhauer:
DebugMode's best use is to provide a way for users to avoid expensive checks in code thought to be fully-debugged.  It makes less sense when the check to be avoided is cheap, and the consequences are severe (e.g. seg faults or array out-of-bounds access, instead of a nice error).  This cleans up a couple of such uses.

### DIFF
--- a/source/adios2/toolkit/format/bp/bp4/BP4Deserializer.tcc
+++ b/source/adios2/toolkit/format/bp/bp4/BP4Deserializer.tcc
@@ -98,18 +98,14 @@ BP4Deserializer::InitVariableBlockInfo(core::Variable<T> &variable,
         const std::vector<typename core::Variable<T>::Info> blocksInfo =
             BlocksInfo(variable, stepsStart);
 
-        if (m_DebugMode)
+        if (variable.m_BlockID >= blocksInfo.size())
         {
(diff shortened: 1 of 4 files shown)


williamfgc:
@eisenhauer not sure about the motivation for this PR. Is there an app or a measurement so we can conclude these are "cheap"? Also, if they are cheap, why change it?

williamfgc:
The consequence of not using DebugMode is always a seg fault and unsecure code. Again, I would rather fix the countless memory leaks and race conditions in adios2 and dependencies.

source/adios2/toolkit/format/bp/bp4/BP4Deserializer.tcc
```diff
@@ -98,18 +98,14 @@ BP4Deserializer::InitVariableBlockInfo(core::Variable<T> &variable,
         const std::vector<typename core::Variable<T>::Info> blocksInfo =
             BlocksInfo(variable, stepsStart);
 
-        if (m_DebugMode)
+        if (variable.m_BlockID >= blocksInfo.size())
```
williamfgc: I'd rather control all user input check with a single bool. This breaks this internal pattern.

Issue #814, opened 2018-08-22
Build error on Summit


suchyta1:
I'm trying to build the master branch on Summit with the GNU compilers and getting a build error when I do so. I've attached the build log.
[summit-build.log](https://github.com/ornladios/ADIOS2/files/2310468/summit-build.log)

suchyta1:
Versions 2.1.0 and 2.2.0 build OK though.

pnorbert:
Does cmake have a check for complex long doubles? Then we could solve this problem at compile time. Otherwise we need to remove this type.

chuckatkins:
The build is failing with an internal compiler error, which means a compiler bug is hit, not adios.  That being said, we might be able to work around it.  Wrt  long double though, i'd vote to remove it entirely since it's a nonstandard inconsistent type.  Float and double are 32 and 64 bit floating point values respectively.  Long double though is often 96 bit, sometimes 128bit, etc.  That's probably a discussion best left to the dev meetings though.  In the meantime I will see if I can work around the compiler bug.

williamfgc:
https://gcc.gnu.org/wiki/Ieee128PowerPC there is a flag to enable long double on GCC for PowerPC. Not sure if it will solve the issue on Summit

suchyta1:
Thanks, I'll try it out.

suchyta1:
Sorry for the delay, but yes, this fix worked for me.

ax3l:
FWIW, GCC 7.4.0, 8.1.0, 8.1.1, 9.1.0 on Summit do not have the same internal compiler error. So this seams to be fixed in newer versions of GCC.

PR #1907, opened 2019-12-19
Add std::complex<long double> Support 

ax3l:
Implement support for the `std::complex< long double >` type.
We need this downstream in openPMD-api and it's generally a missing type (since we also have a type for `long double`).

Help wanted :)

### To Do

- BP3 engine: BP3 only supported `float` and `double` complex: anything to do?
- [x] test updates (partial)
- update bindings
  - [x] Fortran (also adds missing `long double` type): #1921
  - [ ] Matlab (no fundamental `long double` type?)
  - [x] C (also adds missing `long double` type)
  - [x] Python

As a follow-up, a native ADIOS `bool` type would be great for C++ codes, too.


source/adios2/toolkit/interop/hdf5/HDF5Common.cpp
```diff
@@ -64,6 +64,12 @@ HDF5Common::HDF5Common(const bool debugMode) : m_DebugMode(debugMode)
     H5Tinsert(m_DefH5TypeComplexDouble, "dimg", H5Tget_size(H5T_NATIVE_DOUBLE),
               H5T_NATIVE_DOUBLE);
 
+    m_DefH5TypeComplexLongDouble =
+        H5Tcreate(H5T_COMPOUND, sizeof(std::complex<long double>));
+    H5Tinsert(m_DefH5TypeComplexLongDouble, "dreal", 0, H5T_NATIVE_LDOUBLE);
```
ax3l: note: I would generally vote to rename those labels to `r` and `i` as it is convention by `h5py`
cc @guj

ax3l:
@germasch Can you please take a look if I updated the C bindings completely and properly with the new type? I also added the missing support for `long double` in those.

source/adios2/toolkit/interop/hdf5/HDF5Common.cpp
ax3l: @guj Looks to me like we forget to free those types again in the end, via `H5Tclose`.

germasch:
So FWIW, I was wrong that `long double` is not a standard type -- it is, at least in C. But it's still true that it doesn't have a well defined bit width, so it's not good to use internally because things will break when moving files between systems.

Also, I had thought `long double` support was introduced by this PR, but after looking at it, I now realize it's already there, just not for attributes. In other words, the lack of well-defined bit width is a pre-existing issue, not one introduced by this PR.

ax3l:
The length handling of ADIOS is generally not fixed size, simply by the definition that in C/C++ (`short float`) <= `float` <= `double` <= `long double`.
I am not entirely sure if adios just transparently maps to the platform size when moving data between different data models or if this is a portability issue, but as you said already above this is fully independent of this PR.

In terms of naming, I just used the default size on 64bit linux platforms as assumed in all other places for `float` and `double` as well.

testing/adios2/engine/hdf5/TestHDF5WriteMemorySelectionRead.cpp
```diff
@@ -900,28 +1080,39 @@ void HDF5Steps3D8x2x4(const size_t ghostCells)
...
             EXPECT_EQ(CR64.size(), dataSize);
+            EXPECT_EQ(CR128.size(), dataSize);
 
             EXPECT_EQ(I8.front(), static_cast<int8_t>(step));
             EXPECT_EQ(I16.front(), static_cast<int16_t>(step));
             EXPECT_EQ(I32.front(), static_cast<int32_t>(step));
             EXPECT_EQ(I64.front(), static_cast<int64_t>(step));
             EXPECT_EQ(R32.front(), static_cast<float>(step));
             EXPECT_EQ(R64.front(), static_cast<double>(step));
+            EXPECT_EQ(R128.front(), static_cast<double>(step));
```
ax3l: typo
```suggestion
            EXPECT_EQ(R128.front(), static_cast<long double>(step));
```

germasch:
> The length handling of ADIOS is generally not fixed size, simply by the definition that in C/C++ (`short float`) <= `float` <= `double` <= `long double`.

What happens for integer is that the C++ primitive type (say `unsigned long`) internally gets mapped to u64 (uint64_t, really) or u32, depending on what its actually width on the given system is. So it will be stored with its actual precision. If you store an `unsigned long` on a 64-bit system, it'll be stored with 64 bits. If you open that file on a 32 bit system, you cannot access that variable as an `unsigned long` because that's only 32 bits. You could access it as an `unsigned long long`, though.

For float and double, no such mapping actually happens, though it probably should. The thing is, though, that on all current systems that I'm familiar with, `float` is `float32` and `double` == `float64`, so because of that explicitly mapping them hasn't been needed. However, for `long double` is highly system / compiler dependent, and hence its use is not portable. This is kinda my fault in the sense that back then I fixed the integer types, but not the floating point types.

> I am not entirely sure if adios just transparently maps to the platform size when moving data between different data models or if this is a portability issue, but as you said already above this is fully independent of this PR.

Yes, it's preexisting to this PR, though this PR exposes the problem in additional ways. I'm not saying it's a showstopper, but it'd probably be good to think about what to do with `long double` in the first place, independently of this PR. The options that I see are

1. leaving things as they are, ie, declare use of `long double` to lead to non-portable files (though it worth nothing that reading a file that has `long double` in it on another system can cause buffer overflows, a.k.a, weird crashes and security vulnerabilities)
2. drop support for `long double` altogether (and possible add support for `__float128` if needed by some app.)
3. fix `long double` support, but it looks pretty painful, because on wikipedia it looks like `long double` can be (a) a 64-bit float (b) an 80-bit float stored in 96 bits (c) an 80-bit float stored in 128 bits (d) a "double-double", or (e) an actual 128-bit float, depending on system and compiler (and compiler options). Fun...

Maybe most notably, apparently with MSVC `long double` is just a 64-bit float, most definitely not compatible with the Linux/x86_64 `long double`. I just tried Summit vs Linux/x86_64, and while both have storage size of 128 bits for `long double`, the actual binary representation is different. I suppose using this PR you could try writing a `long double` attribute on Summit and read it on x86_64, and you'll see that you get the wrong answer.

> In terms of variable namings, which are not critical, I just used the default size on 64bit linux platforms as assumed in all other places for `float` and `double` as well.

Unfortunately, even on 64-bit Linux, the binary representation of `long double` (nor the actual precision) is not consistent (ie, on ppc64 != on x86_64), though at least they're all stored in 16 bytes.

ax3l:
> What happens for integer is that the C++ primitive type (say unsigned long) internally gets mapped to u64 (uint64_t, really) or u32, depending on what its actually width on the given system is.

I thought so as well for long, but it's exactly the other way around.
`short`, `int`, `long`, `long long` are the fundamental integer types, even if some of them implement the same size, all others are typedefs.
Ref.: https://en.cppreference.com/w/cpp/language/types

Nevertheless, this is all very off topic for this PR.

germasch:
> I thought so as well for long, but it's exactly the other way around.
> `short`, `int`, `long`, `long long` (+ their unsigned variants) are the fundamental integer types, even if some of them implement the same size, all others are (system-dependent) typedefs.
> Ref.: https://en.cppreference.com/w/cpp/language/types (and actual headers of your stdlib)

I know that -- I guess I wasn't clear. ADIOS2 internally maps the C++ primitive types (short, int, ...) to stdint types. And yes, those stdint types themselves are in fact typedefs for primitive types. The point however, is that it takes the ambiguity away. On x86_64, both `unsigned long` and `unsigned long long` are mapped to `uint64_t`, (and I think `uint64_t` is in fact `unsigned long`, but that might be system dependent and doesn't matter for adios2). The point is that the type is written to disk marked as "u64", no matter what it originally was. And when you open a `Variable<unsigned long>` or a `Variable<unsigned long long>` (which are distinct types)  they both again get mapped internally to `core::Variable<uint64_t>`, so either of those can be used to read the u64 data back.

> Implementing the fundamental types in the user-facing API overloads, as now, is generally the right thing to do. Implementing just the fixed (alias) types leads to problems, namely that some types a user provides will not map overloads.

Yes, even though there was some discussion about that at the time, that's the approach I pushed for, and that got adopted eventually. Obviously, when you support all primitive types, the stdint types, which are aliases for a subset of the primitive types, work just fine as well. Before that change, essentially only the stdint types themselves (and whatever they happen to alias to) were properly supported, and as you say, that lead to problems.

>Assuming the type is known/identical between systems is not good on the file format level, but I don't even know if this is happening here.

That's why it maps to the stdint types internally, which are identical on the filesystem level. The same can't be said of `long double`, though.

source/adios2/toolkit/interop/hdf5/HDF5Common.cpp
guj: I did these in pull request 1913.  not merged yet as I see you made changes in the same  files.  might have  conflict.

ax3l: Thanks, added your commits in #1913 to this PR :)

ax3l:
Ha, the PGI Fortran compiler still does not support quad precision. Welcome to 2020.

source/adios2/toolkit/interop/hdf5/HDF5Common.cpp
```diff
@@ -54,19 +54,32 @@ HDF5Common::HDF5Common(const bool debugMode) : m_DebugMode(debugMode)
...
+    H5Tinsert(m_DefH5TypeComplexLongDouble, "r", 0, H5T_NATIVE_LDOUBLE);
+    H5Tinsert(m_DefH5TypeComplexLongDouble, "i",
+              H5Tget_size(H5T_NATIVE_LDOUBLE), H5T_NATIVE_LDOUBLE);
+
     m_PropertyTxfID = H5Pcreate(H5P_DATASET_XFER);
 }
 
+HDF5Common::~HDF5Common()
+{
+    H5Tclose(m_DefH5TypeComplexLongDouble);
```
ax3l: @guj this seams to write some `H5Tclose(): not a datatype` errors to stderr in some tests.

germasch:
Did you try writing a quad precision attribute using Fortran and reading it back via C++? (Maybe using bpls, if you made that to support `long double`.)

I haven't tried, but I'm pretty sure the number you'll read back in one language isn't the same as the one you've written in the other.

eisenhauer:
> Did you try writing a quad precision attribute using Fortran and reading it back via C++? (Maybe using bpls, if you made that to support `long double`.)
> 
It sure would be nice if we had a lot more of that sort of multi-language/environment test...

ax3l:
This discussion is, as already bespoken above, independent of this PR. Please just open an issue to discuss this so we can stay focused here.

If the public API is fundamental types, an "adios long double" (or "adios long long") one the writer side might map to a "adios double" (or "adios long") on another system. That is fine, but is potentially not the case today since fixed sizes are assumed. Anyway, testable and discuss-able independent of this PR since we already have `long double` (and `long long` et al.).

bindings/Fortran/CMakeLists.txt
```diff
@@ -61,6 +61,33 @@ else()
...
+        set(${success} FALSE PARENT_SCOPE)
+        message(STATUS "Fortran: real(kind=16) NOT supported")
+    endif()
+  endif()
+endfunction()
+
+set(ADIOS2_HAVE_Fortran_REAL16 FALSE)
+_Fortran_has_real16(ADIOS2_HAVE_Fortran_REAL16)
+if(Fortran_HAS_REAL16)
+  target_compile_definitions(adios2_f PUBLIC ADIOS2_HAVE_Fortran_REAL16)
```
ax3l: @chuckatkins I do need some flags here to be able to use `#ifdef` in Fortran source files.

I could not figure out why e.g. `Test.Bindings.Fortran.Remove` can already do this and I cannot here, did I miss a property?

eisenhauer: Oddly, you need to capitalize the postfix...   .F90 instead of .f90.

germasch:
> This discussion is, as already bespoken above, independent of this PR. Please just open an issue to discuss this so we can stay focused here.

I agree that the use of `long double` not being portable between systems is a preexisting issue, and I did't follow up further on the previous discussion.

However, now this PR adds a Fortran interface for `long double` which previously did not exist at all. This new Fortran interface is broken even on the same system, in the sense that if you write a `real*16` using that new Fortran interface, bpls (or anything using the C/C++ API) on the same system will return a different value from what was written. This issue did not previously exist because there was no Fortran interface for `real*16`. Note that I didn't say anything about the additions to the C bindings, which don't have this issue, as at least C and C++ `long double` are compatible. (And the Python bindings are mostly okay, too, though you should use `np.longdouble` not `np.float128` in order to not have it break on Linux-32).

Once you introduce behavior which (in my opinion) is broken, it makes it hard to fix it in the future while maintaining backwards compatibility. It's already happened for `long double` portability across systems, so too late now. But now there'll also be incompatibility between Fortran and C/C++ on the same system, which will be essentially impossible to fix, because what's stored in the file is marked as `long double`, which could mean (on x86_64) 80-bit float if it was written by C/C++, or 128-bit float if it was written by Fortran. 

On x86_64, the fix would be to use `real*10`, not `real*16` in what you've added. Unfortunately, I can't suggest changing the PR accordingly, either, because, e.g., on Power that's not right (I suppose it wouldn't even compile on Power).

You say you want this for OpenPMD. But for OpenPMD, if some Fortran PIC code writes real*16 values, and you read them back in Python or C++ using OpenPMD-API, do you really want it to read back wrong values? 

Anyway, I'm just pointing out what I see as an issue in proliferating an interface that uses an ill-defined datatype without disambiguating it, leading to future headaches if someone ever tries to make handling of > 64-bit floats portable. But I don't make any decisions here, maybe everyone else is fine with having a datatype in adios2 that you can only read back on the same system using the same language.

bindings/Fortran/CMakeLists.txt
chuckatkins: > Oddly, you need to capitalize the postfix... .F90 instead of .f90.

Fortran convention is for `foo.F90` to invoke the compiler with the preprocessor while `foo.f90` gets invoked without it.  While it's not "standardized", it is effectively given that the convention is used by all the compilers we support.

williamfgc:
Related #814 @ax3l thanks for doing the work, let's discuss once everyone is back from vacation. Thanks!

germasch:
I'm glad to see that I'm not the only one who thinks that dropping `long double` support entirely would be an option worth considering (@chuckatkins in #814 said it first).

I'd be surprised to see the x86 (or rather x87) 80-bit extended precision type in use nowadays, as it isn't handled by the MMX/SSE2/whatever units, so there's a question as to whether supporting it properly is worth the effort. On the other hand, there's also an obvious counterargument, which is that adios2 should support all native types of its given language interface, which clearly includes `long double` -- say a code actually uses that type and wants to do checkpointing and so doesn't really care about portability. (I've made the argument for supporting all native types myself a year ago with regard to the integer types).

Supporting it properly would mean to have distinct types in the serializer format for all the things that `long double` might mean, ie., 
* `float80_96` (80-bit stored with 96-bit alignment)
* `float80_128`(80-bit stored with 128-bit alignment)
* `float128` (actual quad precision)
* `float128ibm` (the double-double emulated type)
* `float64` (MSVC's idea of `long double`)
(maybe more?)

The last one actually shouldn't require a distinct type -- in the case of MSVC, writing a `long double` would actually write a `double`, which with MSVC it could be ready back into `long double` or `double`.

One kinda new aspect with these types is that any given system does at most support a subset of them (whereas every relevant system supports all the integer types from 8 to 64 bits). ADIOS2's design is to not do conversions (other than endianness). So if one were to write 80-bit extended precision data on x86, there'd be no way to read it back on a Power system, since it doesn't have a type for it. I'd say that's acceptable -- if you write non-portable types, you get what you asked for. (The alternative of reading it back wrong is worse, IMO).

bindings/Fortran/CMakeLists.txt
ax3l: Thank you so much, today I learned :D

chuckatkins: FWIW, here's a few references to the convention:

- PGI: https://www.pgroup.com/resources/docs/19.10/x86/pgi-user-guide/index.htm#fn-conv-input
- GNU: https://gcc.gnu.org/onlinedocs/gfortran/Preprocessing-and-conditional-compilation.html
- Intel: https://software.intel.com/en-us/fortran-compiler-developer-guide-and-reference-understanding-file-extensions
- XL: https://www.ibm.com/support/knowledgecenter/SSAT4T_16.1.1/com.ibm.xlf1611.lelinux.doc/compiler_ref/cpp.html
- Cray: https://pubs.cray.com/content/S-3901/9.1/cray-fortran-reference-manual/source-preprocessing

Basically they all say to use an uppercase extension for automatic pre-processing.  We could go through the trouble of trying to explicitly add the per-compiler pre-processor flag, which CMake might already know about, but it's much easier to just use the "unofficial standard but convention in practice" method of an uppercase file extension.

ax3l: Cool, yes agreed the ending makes sense for our target compilers. Thanks for the background info!

ax3l: @chuckatkins I can locally compile with GCC-8/GFortran-8 but CI does not agree. Did I oversee something?

ax3l:
Updated C++/C/Python frontend to soley refer to the new (complex) long double types. "Serialization" (aka portability of fundamental types) is not touched by this PR and is an independent task that need to be generally fixed for the BP file format.

I moved the Fortran binding updates for real16 kinds to a follow-up PR in #1921.

germasch:
FWIW, this alleviates my main concern, ie., the Fortran bindings that use `real*16`, which is != `long double` even on the same system are gone from this PR. 

It looks like even though you separated out the Python interface changes into a separate PR #1922, that PR is missing most of the Python-related updates (the python tests in particular), while this PR still has all of the Python changes, so if you want them to be separate, you should still remove them from here. (I didn't look deeply into the Python changes, but I don't think I have fundamental concerns about them -- `np.longdouble` is the same as C/C++ `long double`, ie, consistent on the same system, but not portable, which is what the status of `long double in ADIOS2 is in general.)

ax3l:
Agree.
FWIW, #1922 is only a follow-up to the already merged *complex support as attributes* and is independent of this PR, which adds new types. Yes, additional (already passing) tests are added when we merge this again.

ax3l:
@williamfgc @pnorbert @guj ready for review :)

williamfgc:
@ax3l are you handling the PGI compilation issue in this PR? We saw this on Summit #814 and I want to make sure this PR is not bring it back. Thanks!

ax3l:
I didn't see an issue with PGI but relied on CI for that, which passed.
I thought #814 is about GCC 6.4.0? Cannot see any problem with that locally or in CI.

ax3l:
This seams to be an internal GCC 6.4.0 issue which is fixed on Summit with GCC 7.4.0, 8.1.0, 8.1.1, 9.1.0 - these are the versions I tested now.

williamfgc:
> Ha, the PGI Fortran compiler still does not support quad precision. Welcome to 2020.

@ax3l I was making reference to your comment above wrt PGI. In any case, after reviewing the topic we decided to go for a more robust solution to map currently available non-standard types. @chuckatkins will kickstart the work on the CMake side and we can propagate to BP as in this PR. We can use parts of this PR or build on top.  Thanks for double-checking that there are outstanding issues with gcc <= 6.4.0. My preference is not to merge this PR as-is (breaks `master` compilation with gcc <= 6.4.0 which is still widely used), but keep it open until using mapping the non-standard long double with the corresponding (non-standard) fixed width type is ready.  @chuckatkins please keep us posted. Thanks!

pnorbert:
@ax3l In yesterday's dev meeting the team agreed to do the following (which will be done for the next release)
- Introduce the specific floating point types into the BP format (thanks @germasch  for pointing out a lot of those types)
- CMake will determine at configuration time which types are available for the specific architecture/language/compiler combo
- 'long double' will be a supported non-standard type in the ADIOS C++ interface like 'size_t' is now, and will translate to a specific type in each ADIOS build (so an application may not be portable across machines). 

We thought accepting this PR as is so you can move on, and then do the above as updates but the requirement to use GCC 7 or newer scares us. Can we hold on to this PR for a while, until we have the CMake magic  (@chuckatkins) to get the support for long double and conditional compilation to avoid breaking ADIOS with GCC 5 and 6?

A followup development later will be to implement conversions between types as Operators and add them automatically when application type does not match data type in input.

ax3l:
>> Ha, the PGI Fortran compiler still does not support quad precision. Welcome to 2020.
> I was making reference to your comment above wrt PGI, thanks for checking that's not an issue.

Ah I see, no that's only a Fortran binding issue and has been moved to #1921 and got a CMake feature tests and pre-compiler guards.

ax3l:
Thanks @williamfgc and @pnorbert for the updates!
Your plan sounds good to me, feel free to scavenge this PR as needed.

ax3l:
Just to update this issue: I guess the general goal should be multi-precision float support, especially for half-precision types (bfloat & float16) which already landed in the development branch of e.g. OpenMPI and are actively used in math libraries and apps already. Support for long double is just a very similar problem.

germasch:
It looks to me like this got closed when @chuckatkins pulled in the upstream pybind11, which fixed a pybind11 issue which happened to have the same issue number.

ax3l:
@chuckatkins @pnorbert I forgot what the status on this is. Is there any progress on how to add `std::complex<long double>` to complete the type model (similar to the handling of `long double` in ADIOS2)?

eisenhauer:
@ax3l We're trying to close old, impossible-to-merge-as-is PRs.  It looks like there were action items on this that never actually got addressed.  CMake issues for @chuckatkins, maybe something for @pnorbert...    The comment on half-precision types is likely something we should be thinking about for GPUs too (@anagainaru) .  I think those are unlikely to be addressed prior to our next release, but it's also the case that this PR is applicable to a two-year-old code base and developments like BP5 change the landscape significantly.   Can I suggest closing this PR and maybe creating a summary Issue (if these concerns aren't adequately represented in an issue already)?

PR #1909, opened 2019-12-19
Attributes: Allow Complex Types

ax3l:
This enables support for complex attribute types, just as in ADIOS1.

See request in #1908

- [x] add tests

Issue #1908, opened 2019-12-19
std::complex< T > as attributes


ax3l:
Hi,

is there a specific reason why we do not allow `std::complex< T >` types as attributes in ADIOS2?
Those are primitive types and according to my tests in openPMD-api, using complex types as attributes was possible in ADIOS1.

Can we potentially just update the macros that specialize the overloads to allow complex types for attributes as well?

ax3l:
The [specialization you need for `nlohmann::json`](https://github.com/nlohmann/json/issues/1510#issuecomment-471949812) reads:
```C++
#include <complex>

namespace std
{
    // std::complex handling
    template< class T > void to_json(nlohmann::json &j, const std::complex< T > &p) {
        j = nlohmann::json {p.real(), p.imag()};
    }

    template< class T > void from_json(const nlohmann::json &j, std::complex< T > &p) {
        p.real(j.at(0));
        p.imag(j.at(1));
    }
}
```

ax3l:
Implemented via #1909 and #1922

PR #1922, opened 2020-01-02
Python Attributes: Allow Complex Types

ax3l:
Follow-up to #1909: enables support for complex attribute types, just as in ADIOS1, for Python bindings.

Related to #1908


ax3l:
ping @pnorbert @guj this is a follow-up to #1909 for the `api: Python`.

PR #1927, opened 2020-01-06
Add ASAN build

KyleFromKitware:


KyleFromKitware:
See also: GTkorvo/evpath#30

source/adios2/toolkit/sst/cp/ffs_marshal.c
```diff
@@ -1552,14 +1552,7 @@ extern void SstFFSWriterEndStep(SstStream Stream, size_t Timestep)
...
-    size_t *tmp = MBase->BitField;
-    /*
-     * BitField value is saved away from FMfree_var_rec_elements() so that it
-     * isn't unnecessarily free'd.
-     */
-    MBase->BitField = NULL;
     FMfree_var_rec_elements(Info->MetaFormat, Stream->M);
-    MBase->BitField = tmp;
     FMfree_var_rec_elements(Info->DataFormat, Stream->D);
     memset(Stream->M, 0, Stream->MetadataSize);
```
KyleFromKitware: @eisenhauer The problem is that the pointer is being cleared here. This happened in 1d64c264.

eisenhauer: Ah yes, it's cleared in the memset.  So, I think the preferred fix is to move the MBase->BItField = tmp; assignment to after the memset(Stream->M...).

KyleFromKitware: Understood. I figured you had a good reason for the `memset()`, so I took the conservative approach of `free()`ing it before the `memset()`. Will fix.

KyleFromKitware:
@eisenhauer I'm seeing a memory leak that looks like it should be getting cleaned up by `atexit()`:

```console
323: Direct leak of 16 byte(s) in 1 object(s) allocated from:
323:     #0 0x51fe10 in __interceptor_malloc (/root/build/bin/TestCommonWriteShared+0x51fe10)
323:     #1 0x7febaf5537ff in AddNameToExitList /root/src/source/adios2/toolkit/sst/cp/cp_writer.c:111:37
323:     #2 0x7febaf5534df in writeContactInfoFile /root/src/source/adios2/toolkit/sst/cp/cp_writer.c:156:5
323:     #3 0x7febaf546408 in registerContactInfo /root/src/source/adios2/toolkit/sst/cp/cp_writer.c:182:9
323:     #4 0x7febaf545a05 in SstWriterOpen /root/src/source/adios2/toolkit/sst/cp/cp_writer.c:1218:9
```

You mentioned over in GTkorvo/evpath#30 that you'd disabled some of evpath's `atexit()` handlers, but these look like they're coming from ADIOS itself. Have these been disabled as well, or is there another reason you can think of for the leak?

source/adios2/toolkit/format/bp/bp3/BP3Serializer.tcc
```diff
@@ -370,9 +370,9 @@ void BP3Serializer::PutVariableMetadataInData(
     if (span != nullptr)
     {
         const size_t padLengthPosition = position;
-        uint8_t zero = 0;
+        uint8_t zero[] = {0, 0, 0, 0, 0};
         // skip 1 for paddingLength and 4 for VMD] ending
-        helper::CopyToBuffer(buffer, position, &zero, 5);
+        helper::CopyToBuffer(buffer, position, zero, 5);
```
williamfgc: @KyleFromKitware great catch! Thanks!

eisenhauer:
> @eisenhauer I'm seeing a memory leak that looks like it should be getting cleaned up by `atexit()`:
> 
> ```
> 323: Direct leak of 16 byte(s) in 1 object(s) allocated from:
> 323:     #0 0x51fe10 in __interceptor_malloc (/root/build/bin/TestCommonWriteShared+0x51fe10)
> 323:     #1 0x7febaf5537ff in AddNameToExitList /root/src/source/adios2/toolkit/sst/cp/cp_writer.c:111:37
> 323:     #2 0x7febaf5534df in writeContactInfoFile /root/src/source/adios2/toolkit/sst/cp/cp_writer.c:156:5
> 323:     #3 0x7febaf546408 in registerContactInfo /root/src/source/adios2/toolkit/sst/cp/cp_writer.c:182:9
> 323:     #4 0x7febaf545a05 in SstWriterOpen /root/src/source/adios2/toolkit/sst/cp/cp_writer.c:1218:9
> ```
> 
> You mentioned over in [GTkorvo/evpath#30](https://github.com/GTkorvo/evpath/pull/30) that you'd disabled some of evpath's `atexit()` handlers, but these look like they're coming from ADIOS itself. Have these been disabled as well, or is there another reason you can think of for the leak?

the atexit() isn't disabled, but there's probably a missing free(FileNameList->FileName) in RemoveAllFilesInList().

KyleFromKitware:
@eisenhauer I figured out the problem - `RemoveNameFromExitList` had a faulty algorithm. See my latest patch for the fix.

KyleFromKitware:
Also, it looks like there are certain things that LSAN lets go - see [AddressSanitizerLeakSanitizerDesignDocument#Operation](https://github.com/google/sanitizers/wiki/AddressSanitizerLeakSanitizerDesignDocument#operation) for details. Once I'm finished with ASAN/LSAN, I think we should add a Valgrind build as well, since it is more strict.

eisenhauer:
Glad to get all these fixes in!  I've been hoping for an ASAN/valgrind build for a long while.  Lacking support in CI, the only thing that has received much leak attention is a simple write/read test with default params (BP and CommMin)...

.circleci/config.yml
```diff
@@ -38,9 +38,15 @@ jobs:
     docker:
       - image: ornladios/adios2-auto:fedora-openmpi-ubsan
 
+  "fedora-asan":
+    <<: *defaults
+    docker:
+      - image: kylefromkitware/adios2-auto:fedora-asan
```
KyleFromKitware: This is a temporary image I've been using for development. Could someone build this image and upload it to `ornladios`, or add me to the organization so I can do it myself?

KyleFromKitware:
> Oh, and the atexit() things are mostly freeing things that are reachable by global statics, so if LSAN ignores those, you won't see them until valgrind.

Yeah, Valgrind is my next step, possibly followed by an MPI variant of both ASAN and Valgrind.

KyleFromKitware:
There's one leak left in ASAN's report, and it's a tricky one... it seems that there's a connection being created in `socket_accept_conn()` that occasionally gets an extra reference and never gets freed. It happens non-deterministically. See [this failure](https://open.cdash.org/testDetails.php?test=838749091&build=6291071) for an example.

@eisenhauer I'm a little stuck here. Any thoughts?

eisenhauer:
Try this?

diff --git a/source/adios2/toolkit/sst/cp/cp_writer.c b/source/adios2/toolkit/sst/cp/cp_writer.c
index a3b9a627..66b852a8 100644
--- a/source/adios2/toolkit/sst/cp/cp_writer.c
+++ b/source/adios2/toolkit/sst/cp/cp_writer.c
@@ -644,6 +644,7 @@ static int initWSReader(WS_ReaderInfo reader, int ReaderSize,
                 CMget_conn(reader->ParentStream->CPInfo->cm,
                            reader->Connections[0].ContactList);
 
+            CMConnection_dereference(reader->Connections[0].CMconn);   /* kill pending reference from open */
             if (!reader->Connections[0].CMconn)
             {
                 CP_error(reader->ParentStream, "Connection failed in "
@@ -884,6 +885,7 @@ WS_ReaderInfo WriterParticipateInReaderOpen(SstStream Stream)
                        "Message failed to send to reader in participate in "
                        "reader open\n");
         }
+        CMConnection_dereference(conn);
         free(response.CP_WriterInfo);
         free(response.DP_WriterInfo);
     }

eisenhauer:
Try this?
`diff --git a/source/adios2/toolkit/sst/cp/cp_writer.c b/source/adios2/toolkit/sst/cp/cp_writer.c
index a3b9a627..66b852a8 100644
--- a/source/adios2/toolkit/sst/cp/cp_writer.c
+++ b/source/adios2/toolkit/sst/cp/cp_writer.c
@@ -644,6 +644,7 @@ static int initWSReader(WS_ReaderInfo reader, int ReaderSize,
                 CMget_conn(reader->ParentStream->CPInfo->cm,
                            reader->Connections[0].ContactList);
 
+            CMConnection_dereference(reader->Connections[0].CMconn);   /* kill pending reference from open */
             if (!reader->Connections[0].CMconn)
             {
                 CP_error(reader->ParentStream, "Connection failed in "
@@ -884,6 +885,7 @@ WS_ReaderInfo WriterParticipateInReaderOpen(SstStream Stream)
                        "Message failed to send to reader in participate in "
                        "reader open\n");
         }
+        CMConnection_dereference(conn);
         free(response.CP_WriterInfo);
         free(response.DP_WriterInfo);
     }
`

KyleFromKitware:
@eisenhauer 

That patch caused the following error:

```console
381: ==18879==ERROR: AddressSanitizer: heap-use-after-free on address 0x61200000ff40 at pc 0x7fb67f26f341 bp 0x7ffeb992ba40 sp 0x7ffeb992ba38
381: READ of size 8 at 0x61200000ff40 thread T0
381:     #0 0x7fb67f26f340 in CMConnection_close /root/build/asan/thirdparty/EVPath/EVPath/cm_interface.c:100:22
381:     #1 0x7fb67f687b73 in EvpathDestroyWriter /root/src/source/adios2/toolkit/sst/dp/evpath_dp.c:620:17
381:     #2 0x7fb67f6b7d68 in SstStreamDestroy /root/src/source/adios2/toolkit/sst/cp/cp_common.c:1015:13
381:     #3 0x7fb680fa4d80 in adios2::core::engine::SstWriter::~SstWriter() /root/src/source/adios2/engine/sst/SstWriter.cpp:135:5
```

eisenhauer:
I was afraid of that.  You said it was nondeterministic, and if it was this, then that wouldn't account for the nondeterminism.  Is the issue on the writer side?  I'm still suspicious that this code path (might be involved), but I'd need to dig deeper on a fix.

KyleFromKitware:
> Is the issue on the writer side?

I don't know, I just know that it's an incoming connection. Are connections simplex or duplex?

eisenhauer:
OK, looking at the example failure that you linked earlier.  First, ignore the 3x2, those particular tests are run without MPI without their names changing from the MPI version.  They just end up being 1 writer rank, 1 reader rank.  They are also using the kill_reader test protocol, in which we arbitrarily kill -9 reader processes to make sure they don't mess up the writer when readers go away without warning or proper shutdown.  That whole "without proper shutdown" thing is likely what is giving us the non-determinism.  WRT the question, connections are duplex, and we try to reuse existing connections where possible rather than create multiple between the same processes.  Hence the complexity of the reference count.  This may be a bit ugly to sort out, but I'll take a look at it tomorrow...

chuckatkins:
@eisenhauer fwiw this entire asan build is without MPI i believe.

@kylefromkitware I didn't even think about it before but given that the "kill" tests intentionally force shutdown without cleanup then should those specific tests be excluded from the leak check builds?

KyleFromKitware:
> fwiw this entire asan build is without MPI i believe.

Yes. Once I've taken care of Valgrind I will take a crack at MPI.

> > That whole "without proper shutdown" thing is likely what is giving us the non-determinism.
>
> given that the "kill" tests intentionally force shutdown without cleanup then should those specific tests be excluded from the leak check builds?

Is that what's going on? In that case yes, those tests should absolutely be excluded.

KyleFromKitware:
Actually, if reader processes are being arbitrarily killed, I wonder if the leaks are coming from the reader (in which case we can ignore the leak) or from the writer (in which case it's not properly handling a dropped connection and needs to be fixed.)

KyleFromKitware:
Looks like at least one of the leaks is coming from the writer. From the example failure I linked:

```console
Indirect leak of 1024 byte(s) in 1 object(s) allocated from:
    #0 0x522000 in __interceptor_malloc (/home/adios2/build_fedora-asan/bin/TestCommonServer+0x522000)
    #1 0x7f0fb5b24ce9 in make_tmp_buffer /home/adios2/source/thirdparty/ffs/ffs/ffs/ffs.c:1548:20
    #2 0x7f0fb5b2412d in FFSencode_vector /home/adios2/source/thirdparty/ffs/ffs/ffs/ffs.c:402:5
    #3 0x7f0fb5d32edf in INT_CMwrite_attr /home/adios2/source/thirdparty/EVPath/EVPath/cm.c:3256:12
    #4 0x7f0fb5d30df8 in INT_CMwrite /home/adios2/source/thirdparty/EVPath/EVPath/cm.c:2713:13
    #5 0x7f0fb5d7f829 in CMwrite /home/adios2/build_fedora-asan/thirdparty/EVPath/EVPath/cm_interface.c:613:8
    #6 0x7f0fb62ec308 in SendPeerSetupMsg /home/adios2/source/source/adios2/toolkit/sst/cp/cp_writer.c:469:9
    #7 0x7f0fb62dc9e3 in initWSReader /home/adios2/source/source/adios2/toolkit/sst/cp/cp_writer.c:585:17
    #8 0x7f0fb62da842 in WriterParticipateInReaderOpen /home/adios2/source/source/adios2/toolkit/sst/cp/cp_writer.c:815:21
    #9 0x7f0fb62e42d2 in SstInternalProvideTimestep /home/adios2/source/source/adios2/toolkit/sst/cp/cp_writer.c:2131:18
    #10 0x7f0fb62e8c9c in SstProvideTimestep /home/adios2/source/source/adios2/toolkit/sst/cp/cp_writer.c:2198:5
    #11 0x7f0fb7faa4b5 in adios2::core::engine::SstWriter::EndStep() /home/adios2/source/source/adios2/engine/sst/SstWriter.cpp:280:9
    #12 0x7f0fb80bdb10 in adios2::Engine::EndStep() /home/adios2/source/bindings/CXX11/adios2/cxx11/Engine.cpp:122:15
```

Notice the `SstWriter` frame towards the bottom.

eisenhauer:
Yeah, I'm assuming that any reported leaks are from the writer or from readers that finish naturally (there may be some), 'cause I'm guessing that nobody is around to report leaks from SIGKILL'd procs (and there's be a *lot* more leaks).

KyleFromKitware:
OK, if you're going to take a closer look tomorrow then I will start working on Valgrind in the meantime when I get in tomorrow. If you find a fix, feel free to add it as a commit on this PR (though don't merge yet, I still need to clean up the history a bit and get a properly hosted Docker image.)

KyleFromKitware:
> (and there's be a lot more leaks).

I'm not so sure, actually... if you stop a program at some arbitrary point in time, assuming it's written properly, all allocated memory will still be reachable and thus not be reported by LSAN.

eisenhauer:
So, quick look here.  The control plane isn't calling the dataplane DestroyWriter funcs, ever.  And at least for the EVPath DP, the DestroyWriterPerReader func while extant doesn't pair well with the DestroyWriter func.  In the 1x1 cases that you're testing (non-MPI), we'll be using data preloading, which will mean that the EVPath DP will be calling SendSpeculativePreloadMsgs(), which will call CMget_conn(), which will likely try to use the CMconnection to the reader which was initially created by accept() when it was an incoming connection request, so it'll increment the reference count.  So, while DestroyWriter closes that connection (and decs the ref count), that it doesn't get called probably accounts for your extra ref count and the CMConnection not being free'd.  BUT, I'm not sure why you're not also seeing other EVPath DP data elements showing up as leaks (or why I didn't see them in the limited valgrind runs done before the LSAN CI).  I'm going to proceed with caution here, and I've got a few other things on my plate this morning, so be assured that I'm poking at this, but it's not a one-liner, maybe not quick.  Also it might be wise to pull in @philip-davis because I think the RDMA DP isn't in CI and if we start calling untested Destroy funcs there, we may break master and not find out until we do RDMA runs afterwards.

eisenhauer:
Status:  I've been able to reproduce the CMConnection issue with valgrind applied to the writer.  Will try to track down the circumstances when this happens.

eisenhauer:
I did a PR for CMConnection issue, but apparently to your msan branch, and maybe it's best here.  Perhaps easiest for you to cherrypick it to where you want it?
https://github.com/eisenhauer/ADIOS2/commit/b4185285d7721ad7a4c864272b32f97ad497ab3e

KyleFromKitware:
@eisenhauer I cherry-picked your commit and it fixed the leak in local testing, so I think the actual work on this PR is done. However, it is still blocked by #1935. Once @chuckatkins is finished with that, we can rebase this and merge it.

chuckatkins:
@KyleFromKitware This should be rebased appropriately now.  Once the remaining asan items are cleaned up then we can merge it.

KyleFromKitware:
The latest version of this patch has a connection somewhere that's not being closed with `INT_CMConnection_close()`. This was working fine in the old patch, and seems to be a regression. I don't see anything in the `master` log that sticks out to me. @eisenhauer any idea what might have caused this?

eisenhauer:
Not offhand, but will take a look.

source/adios2/engine/sst/SstWriter.cpp
```diff
@@ -130,7 +130,14 @@ SstWriter::SstWriter(IO &io, const std::string &name, const Mode mode,
     }
 }
 
-SstWriter::~SstWriter() { SstStreamDestroy(m_Output); }
+SstWriter::~SstWriter()
+{
+    SstStreamDestroy(m_Output);
+    if (m_BP3Serializer)
```
chuckatkins: Rather than add an explicit cleanup here, can we just switch `m_BP3Serializer` from a `format::BP3Serializer*` to an `std::unique_ptr<format::BP3Serializer>`?

KyleFromKitware: Will update in the next version.

eisenhauer:
Just a comment for now.  The asan tests that are leaving a CMConnection unfree'd are the Staging.*.Shared* tests.  These tests share the common characteristic that they are opening *TWO* streams between the reader and writer.  In this situation, CM/EVPath will be reusing the underlying bi-directional connections between the processes, multiplexing traffic from each stream over them.  The reference counts should take care of all this, but obviously something isn't going as expected.  I'm still digging.

eisenhauer:
Please try this commit for the CMConnection issue:  https://github.com/eisenhauer/ADIOS2/commit/69bd57078021c7b18182e44d578484ae01040209

KyleFromKitware:
@eisenhauer That worked for me in local testing. Thanks!

chuckatkins:
@KyleFromKitware I just realized you'll probably also want to disable the test repeats for the sanitizer runs:

```diff
diff --git a/scripts/ci/cmake/ci-fedora-asan.cmake b/scripts/ci/cmake/ci-fedora-asan.cmake
index c8346ada..8f1f2375 100644
--- a/scripts/ci/cmake/ci-fedora-asan.cmake
+++ b/scripts/ci/cmake/ci-fedora-asan.cmake
@@ -25,5 +25,6 @@ set(dashboard_track "Analysis")
 set(CTEST_CMAKE_GENERATOR "Unix Makefiles")
 set(CTEST_BUILD_FLAGS "-k -j4")
 
+set(ADIOS_TEST_REPEAT 0)
 list(APPEND CTEST_UPDATE_NOTES_FILES "${CMAKE_CURRENT_LIST_FILE}")
 include(${CMAKE_CURRENT_LIST_DIR}/ci-common.cmake)
diff --git a/scripts/ci/cmake/ci-fedora-ubsan.cmake b/scripts/ci/cmake/ci-fedora-ubsan.cmake
index 02fd1fcf..ee9073b1 100644
--- a/scripts/ci/cmake/ci-fedora-ubsan.cmake
+++ b/scripts/ci/cmake/ci-fedora-ubsan.cmake
@@ -21,5 +21,6 @@ set(dashboard_track "Analysis")
 set(CTEST_CMAKE_GENERATOR "Unix Makefiles")
 set(CTEST_BUILD_FLAGS "-k -j2")
 
+set(ADIOS_TEST_REPEAT 0)
 list(APPEND CTEST_UPDATE_NOTES_FILES "${CMAKE_CURRENT_LIST_FILE}")
 include(${CMAKE_CURRENT_LIST_DIR}/ci-common.cmake)
diff --git a/scripts/dashboard/adios_common.cmake b/scripts/dashboard/adios_common.cmake
index 6b106bf5..aca84a81 100644
--- a/scripts/dashboard/adios_common.cmake
+++ b/scripts/dashboard/adios_common.cmake
@@ -84,8 +84,13 @@ if(NOT dashboard_model)
   set(dashboard_model Experimental)
 endif()
 
-if(CMAKE_VERSION VERSION_GREATER 3.16.20191201)
-  set(CTEST_TEST_ARGS REPEAT "UNTIL_PASS:5")
+if(NOT DEFINED ADIOS_TEST_REPEAT)
+  set(ADIOS_TEST_REPEAT 5)
+endif()
+
+if((CMAKE_VERSION VERSION_GREATER 3.16.20191201 ) AND
+   (ADIOS_TEST_REPEAT GREATER 0))
+  set(CTEST_TEST_ARGS REPEAT "UNTIL_PASS:${ADIOS_TEST_REPEAT}")
 endif()
 
 list(APPEND CTEST_UPDATE_NOTES_FILES "${CMAKE_CURRENT_LIST_FILE}")
```

chuckatkins:
@kylefromkitware I think the failures now are fr me sticking asan in ldpreload.  You can probably take that out.

KyleFromKitware:
That should be everything... once this passes I think we can merge it.

eisenhauer:
Nice!

KyleFromKitware:
Edit: still need to apply @chuckatkins' patch to disable test repetition.

KyleFromKitware:
OK, I've applied it. Now we wait...

KyleFromKitware:
🎉🎉🎉🎉🎉

PR #1928, opened 2020-01-07
Remove engine from IO map at Close

williamfgc:


williamfgc:
@pnorbert please try this to see if deallocation works.

Issue #1891, opened 2019-11-27
Large memory consumption in BufferSTL when using PerformPuts and Flush instead of Begin/EndStep (BP4)


franzpoeschel:
During performance evaluations of the ADIOS2 backend in the [openPMD API](https://github.com/openPMD/openPMD-api), I noticed an unusually large heap memory consumption in non-streaming workflows (in this case the BP4 engine) and traced it back to memory not being freed from the marshalling buffer (`BufferSTL`).

Since openPMD's iterations cannot be easily modeled using the ADIOS2 step concept, this backend only uses steps for streaming engines. For disk-based engines, we use `Engine::PerformPuts` and `Engine::Flush` instead. The documentation for the latter method says:
> Manually flush to underlying transport to guarantee data is moved

This suggests that data should not be present in ADIOS after this call. (?) The figure below shows the memory trace from a small example, writing 30 openPMD iterations from [PIConGPU](https://github.com/ComputationalRadiationPhysics/picongpu) to disk with several flushes per iteration. This memory buildup is not visible when using ADIOS steps.

![Bildschirmfoto vom 2019-11-27 12-01-26](https://user-images.githubusercontent.com/14241876/69718079-af35b100-110d-11ea-85fb-405520e37d41.png)

Did I understand the functionality of `Engine::Flush` correctly? In that case, calling it should free the buffer. If not, is there a suggested alternative to avoid using ADIOS steps without building up heap memory usage in the described way?

williamfgc:
@franzpoeschel first of all, thanks for providing numbers as it always enriches the conversation. The memory growth you're seeing is intended within a single adios2 step, as you reported you are only using one adios2 step. 
A few things: 
1. Are you using Put in sync mode? Usually memory grows as shown in your graph.

2. Flush calls the underlying transports (e.g. POSIX write, fstream write+flush) so it guarantees your data is moved as the docs say. It shouldn't do anything to the memory allocation per-se, but reset the buffer position to 0 for the next Put to avoid growth, if you're not seeing this then it might be a bug.
 
3. Can you try the BP3 engine? BP4 is designed for use-cases with several steps and large metadata (variables, ranks, etc.). You'll probably not see any differences in the allocation, but it would help understand with real numbers.

4. Out of curiosity, what's the reason openPMD steps can't be mapped to adios2 steps? The concept of steps in adios2 is I/O steps or batches of variables, it could be anything (timestep, iteration number, acquired data, etc.)

[...]

PR #1930, opened 2020-01-08
Changes to BP memory management

williamfgc:
Change BP4 memory resize function
Reset Engine smart pointer at Close
Deallocate explicitly at Close (might not be needed)
Might be related to #1891


pnorbert:
After this fix the iotest behaves as expected now

PR #1931, opened 2020-01-09
core: Re-implement IO::Open with an internal engine factory

bradking:
Replace the hard-coded cascading-if table of engine construction with a factory using runtime lookup.  This will be useful for runtime registration of optional engines.

Split registration of IO factory entries.  Register serial engines statically.  Add runtime registration of MPI engines that are guaranteed to be available if the application creates a MPI-backed ADIOS communicator.

The HDF5 engine is special in that it is serial or parallel based on how HDF5 itself was built.  Register it only as one or the other.  Teach the HDF5 engine factory function to create the engine only when using the matching kind of communicator.

The SST engine also supports working with or without MPI.  Leave it registered statically for now.  Later the two variants can be split.

### DIFF
--- a/source/adios2/CMakeLists.txt
+++ b/source/adios2/CMakeLists.txt
@@ -229,6 +229,7 @@ endif()
 
 if(ADIOS2_HAVE_MPI)
   target_sources(adios2 PRIVATE
+    core/IOMPI.cpp
     helper/adiosCommMPI.h  helper/adiosCommMPI.cpp
     engine/insitumpi/InSituMPIWriter.cpp engine/insitumpi/InSituMPIWriter.tcc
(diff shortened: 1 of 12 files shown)


williamfgc:
> This will be useful for runtime registration of optional engines.

@bradking thanks for doing the refactoring. I'm interested in learning more about the above feature and your vision for adding optional engines.

bradking:
@williamfgc this is in preparation to split the libraries such that both serial and MPI-based programs can use the same build of ADIOS2.  The MPI-based engines will then only be available to applications that link the ADIOS2 MPI-specific library.

bradking:
The CI test failures are symptoms of a larger problem with making the HDF5 IO engine available to serial programs when HDF5 is built with MPI.  I'll close this PR while I investigate that further to find a solution.

PR #1932, opened 2020-01-10
Add MSAN build

KyleFromKitware:
This is blocked by #1927.


KyleFromKitware:
Local testing of this build is blood-red. Oh dear...

KyleFromKitware:
@chuckatkins @eisenhauer Looks like to get a working MSAN build, we will need to compile all of ADIOS2's dependencies from source with `-fsanitize=memory`, including `libc++`. See [this link](https://github.com/google/sanitizers/wiki/MemorySanitizerLibcxxHowTo) for details. Before I dig any deeper into this, I'd like to make sure this is a route we want to pursue.

KyleFromKitware:
@eisenhauer Thanks for the explanation. Based on this:

> and covers us for future changes where the reader sends its parameters to the writer too

it sounds to me like it would be better to keep this and scrap #1934, but I will leave it to you to decide.

eisenhauer:
Agreed.  Reverting #1934

KyleFromKitware:
There is a memory error in this build that seems to be triggered by how `long double` is handled. It has a `sizeof()` of 16 (128 bits), but only 80 bits are actually used. The other 48, at the end, remain uninitialized, which results in errors when sending a `long double` over the wire. Here is a minimal example that reproduces the issue:

```c++
#include <iostream>

void print_long_double(const long double& d)
{
  unsigned char* ptr = (unsigned char*) &d;
  for (int i = 0; i < sizeof(d); ++i) {
    if (ptr[i] < 128) {
      std::cout << "less" << std::endl;
    } else {
      std::cout << "more" << std::endl;
    }
  }
}

int main()
{
  long double d;
  memset(&d, 0, sizeof(d));
  d = 42.42;
  print_long_double(d);
  long double d2 = d;
  print_long_double(d2); // Comment this to silence the error
  return 0;
}
```

I'm working on contacting the LLVM team to ask what to do.

scripts/ci/images/fedora-msan/Dockerfile
```diff
@@ -0,0 +1,9 @@
+ARG INSTALL_PREFIX=/opt/msan
+ARG TOOLCHAIN_FILE=toolchain.cmake
+ARG CFLAGS="-fsanitize=memory -fsanitize-memory-track-origins"
```
chuckatkins: Do we need CXXFLAGS set here too?

KyleFromKitware: Might not hurt to add it, but I haven't had an explicit need for it yet. In this context, `CFLAGS` and `CXXFLAGS` are only needed for projects that can't be built with CMake.

chuckatkins:
> create a temporary array of `long double`, `memset()` it to 0, copy the `long double`s into that, and then `memcpy()` that into the _real_ buffer. (Yes, it's a double buffer, but I can't see any way to avoid it without copying the undefined padding bits, 

That should do it.  It's the same way that `memmove` works which is how most of the `std::copy` implementations are implemented.

> unless we hard-code the "real" size of a `long double` to be 80 bits

That's a whole different issue underway elsewhere.  Best leave that one alone here :-)

KyleFromKitware:
>  It's the same way that `memmove` works

I thought `memmove` was typically optimized to not need an intermediate buffer? If the source and destination buffers overlap you can change the copy direction to avoid data corruption.

KyleFromKitware:
Looks like the temporary buffer fixed the ubsan misalignment error.

The only remaining error is in the YAML tests, and it's not a problem with our code. It's a bug of unknown origin somewhere in LLVM. I've already reported it upstream: https://bugs.llvm.org/show_bug.cgi?id=44586

I attempted to work around it by using compiler-rt instead of libgcc, and in the process ran into this issue: https://bugs.llvm.org/show_bug.cgi?id=44762

I'm still waiting to hear back on both issues. Until then, I'll see if I can add a suppression to the problematic tests.

source/adios2/engine/sst/SstParamParser.cpp
```diff
@@ -10,6 +10,8 @@ using namespace adios2::core;
 
 void SstParamParser::ParseParams(IO &io, struct _SstParams &Params)
 {
+    memset(&Params, 0, sizeof(Params));
+
```
chuckatkins: `memset` -> `std::memset`.  We like to use the C++ interface to the C library rather than the C library directly.

source/adios2/helper/adiosMemory.inl
```diff
@@ -74,15 +84,55 @@ void InsertToBuffer(std::vector<char> &buffer, const T *source,
...
 }
 
+template <class T>
+void CopyToBufferSafe(std::vector<char> &buffer, size_t &position,
+                      const T *source, const size_t elements) noexcept
+{
+    // Create a temporary zero-initialized buffer which has the alignment we
+    // need
+    std::unique_ptr<T[]> tmp(new T[elements]);
+    std::memset(tmp.get(), 0, elements * sizeof(T));
```
chuckatkins: Can you drop the `memset` in lieu of the value-initialization constructor `new T[elements]()`?

source/adios2/engine/sst/SstParamParser.cpp
KyleFromKitware: Will update in the next push.

source/adios2/helper/adiosMemory.inl
KyleFromKitware: The problem is that value-initialization might not initialize the other 48 bits. We are attempting to explicitly fill those bits with 0.

KyleFromKitware:
I've rebuilt libc++ in Debug mode, which seems to take care of the YAML issue. It's a horrible hack, but until LLVM responds, I don't have any better ideas.

eisenhauer:
I have a performance concern WRT insisting on initializing the "empty" bytes in a long double array before they are copied to the BP buffer.  This is the code path that is used for application data, so this is potentially gigabytes.  The extra memset() is potentially a very large performance impact, requiring us to pull the destination buffer through the cache an extra time, etc.  IMHO, we should be suppressing the warning about uninitialized memory reads when we're storing 80-bit floats in 128-bit elements, rather that compromising performance to initialize them.  (I don't mind the memset() on things that are known to be small.  We've done that a lot in SST because of the way that moves data.  But I think we have to be much more careful with application data.  We'd like to be moving to a situation where we avoid copies, and this is not be a step in the right direction.)

KyleFromKitware:
@eisenhauer

> IMHO, we should be suppressing the warning about uninitialized memory reads when we're storing 80-bit floats in 128-bit elements, rather that compromising performance to initialize them.

There is a security consideration here. Depending on the use case, sending uninitialized bits over the wire can potentially result in sending sensitive data stored in memory (passwords, PII, etc.) Given the use case for ADIOS, it might not be as big of a deal, but I'd be a lot more concerned if this was for a public-facing server.

Let me know what you think.

chuckatkins:
> I have a performance concern WRT insisting on initializing the "empty" bytes in a long double array before they are copied to the BP buffer. This is the code path that is used for application data, so this is potentially gigabytes. The extra memset() is potentially a very large performance impact, requiring us to pull the destination buffer through the cache an extra time, etc. IMHO, we should be suppressing the warning about uninitialized memory reads when we're storing 80-bit floats in 128-bit elements, rather that compromising performance to initialize them. (I don't mind the memset() on things that are known to be small. We've done that a lot in SST because of the way that moves data. But I think we have to be much more careful with application data. We'd like to be moving to a situation where we avoid copies, and this is not be a step in the right direction.)

I'm inclined to agree with @eisenhauer on this and that that's essentially just "part of the deal" with padded types.  What we're really doing here is copying around user-specified memory so really it's up to the application to provide the properly zero-initialized padding.  For that matter I'm even more included to drop the use of std::copy entirely since we know we're working with raw memory buffers and not complex user-defined types with overloaded assignment operators.
 
> There is a security consideration here. Depending on the use case, sending uninitialized bits over the wire can potentially result in sending sensitive data stored in memory (passwords, PII, etc.) Given the use case for ADIOS, it might not be as big of a deal, but I'd be a lot more concerned if this was for a public-facing server.

Even if it were to used for pii and / passwords (which it's not), it's still operating on user-defined memory.  ADIOS shouldn't be taking the additional responsibility of validating the contents of that memory.  The responsibility is really on the calling application to provide contiguous memory blocks for storage and transmission.

eisenhauer:
I certainly wouldn't claim that you couldn't reconstruct *something* if you got 48 out of every 128 bits of it.  But large data performance is the raison d'etre of ADIOS.  The infosec concerns seem relatively remote compared with the certain large data performance impact.

pnorbert:
+1 for @eisenhauer and performance

chuckatkins:
> ADIOS shouldn't be taking the additional responsibility of validating the contents of that memory. The responsibility is really on the calling application to provide contiguous memory blocks for storage and transmission.

@KyleFromKitware In this vein, it would be worth checking where the msan issue crops up and see if it can be resolved in the test itself by zero-initializing the user-provided memory buffer prior to passing it to adios.

KyleFromKitware:
> @KyleFromKitware In this vein, it would be worth checking where the msan issue crops up and see if it can be resolved in the test itself by zero-initializing the user-provided memory buffer prior to passing it to adios.

Sounds like the best approach to me. I'll take a look.

KyleFromKitware:
Hmmm. The first error I come across when reverting 666b3250 is related to attributes. Unfortunately, `DefineAttribute()` copies the value into an area of memory that is not controlled by the ADIOS consumer, which means we can still get uninitialized bits even when the consumer uses "good practices" as described above.

...and now that I think about it, this is what drove me to modify `CopyToBuffer()` in the first place.

So `memset()`ing gigabytes of user data isn't feasible, but are attributes small enough that they can take the mild performance hit?

eisenhauer:
I'd say yes, that's OK.  Attributes aren't envisioned as being huge...

KyleFromKitware:
Great, thanks!

KyleFromKitware:
Ooh. Hmmm. I didn't think about the fact that `Attribute` might take a non-primitive type like `std::string`, and of course `memset()`ing it yields disastrous results.

chuckatkins:
Maybe rather than specialize on string you could have the two implementations guarded via an enable_if on the is_trivial or is_fundamental type trait?  That allows for future implementations with user defined types to use the generic non-zeroing constructor.

Is that doable without making it a Rube-Goldberg machine?

KyleFromKitware:
> Maybe rather than specialize on string you could have the two implementations guarded via an enable_if on the is_trivial or is_fundamental type trait?

Possibly, but it would need to be moved into `Attribute.tcc`. Right now the macro is explicitly defining the constructors for the list of types.

I wonder if it might actually be better to make something like `adios2::requires_zero_padding<T>` that specializes to `true` for `long double` and defaults to `false` for everything else.

KyleFromKitware:
> I wonder if it might actually be better to make something like `adios2::requires_zero_padding<T>`

The reason I say this is because of the possibility of `std::complex<long double>`, which is not covered by `is_trivial` or `is_fundamental` but still requires zero-padding.

KyleFromKitware:
@chuckatkins @eisenhauer All tests have passed. Anything left here?

source/adios2/core/Attribute.h
```diff
@@ -13,11 +13,23 @@
...
 {
 
+template <class T>
+struct requires_zero_padding : std::false_type
+{
+};
+
+template <>
+struct requires_zero_padding<long double> : std::true_type
+{
```
chuckatkins: I feel like this is a pretty unique case only used by Attribute to address a specific issue.  As such it would be more appropriate to move to the `.cpp` file as an internal helper rather than a public type.  Also, just for convention with the rest of the code base, `requires_zero_padding` should be changed to `RequiresZeroPadding`.

KyleFromKitware: The idea behind putting it in the header was that if there are any non-primitive types that get used in `Attribute` in the future, they can declare `RequiresZeroPadding` in their own header. If you don't foresee that happening then I'll go ahead and move it.

chuckatkins: > The idea behind putting it in the header was that if there are any non-primitive types that get used in Attribute in the future, they can declare RequiresZeroPadding in their own header.  That'd work if it was propagated to the bindings but this is in the `core` which isn't user-facing.  It's the back end implementation for the user-facing `bindings`.  I really think this is specific to the case of `long double`.  There's *lots* of other internal structre that needs to be addressed if we were to support user-defined types, if ever, so I think it's fine to just move it internal.

PR #1935, opened 2020-01-13
Rework CircleCI tooling

chuckatkins:


chuckatkins:
@KyleFromKitware I've consolidated the sanitizer builds into a single container they can all run out of.  Will the `fedora-sanitizers` added here container work for you?

scripts/ci/images/fedora-sanitizers/Dockerfile
```diff
@@ -0,0 +1,50 @@
...
+        -GNinja \
+        -DCMAKE_BUILD_TYPE=Release \
+        -DCMAKE_INSTALL_PREFIX=/opt/llvm/9.0.0/libcxx-msan \
+        -DCMAKE_INSTALL_RPATH='$ORIGIN' \
+        -DLLVM_ENABLE_PROJECTS="libcxx;libcxxabi" \
+        -DLLVM_USE_SANITIZER=MemoryWithOrigins \
+        -DLIBCXX_ENABLE_ABI_LINKER_SCRIPT=OFF \
+        ../../source/llvm && \
+    ninja install-cxxabi install-cxx
+RUN cd ../.. && rm -rf source build
```
KyleFromKitware: This will not remove the build tree cruft, only hide it. Docker images are built in layers, and the cruft will still exist in a lower layer. Instead of `rm`-ing `source` and `build`, it's better to use the `FROM` Docker directive to start a new image, and then `COPY` the needed files into the new image. See [Multi-Stage Builds](https://docs.docker.com/develop/develop-images/multistage-build/) and [my PR](https://github.com/ornladios/ADIOS2/pull/1932/files#diff-63519e5293f296370193560c1789ce55R67) for examples.

KyleFromKitware:
> Will the fedora-sanitizers added here container work for you?

We will need to add a `toolchain.cmake` file to use the custom `libc++` installation, but it looks good so far.

KyleFromKitware:
OK, here's the toolchain file I've come up with:

```cmake
set(CMAKE_C_COMPILER /usr/bin/clang)
set(CMAKE_CXX_COMPILER /usr/bin/clang++)
set(_link_flags "-L /opt/llvm/9.0.0/libcxx-msan/lib -Wl,-rpath,/opt/llvm/9.0.0/libcxx-msan/lib")
set(_compile_flags "-stdlib=libc++ -isystem /opt/llvm/9.0.0/libcxx-msan/include/c++/v1 -fsanitize=memory")
set(CMAKE_EXE_LINKER_FLAGS_INIT ${_link_flags})
set(CMAKE_SHARED_LINKER_FLAGS_INIT ${_link_flags})
set(CMAKE_C_FLAGS_INIT ${_compile_flags})
set(CMAKE_CXX_FLAGS_INIT ${_compile_flags})
```

Please add this somewhere where the MSAN build can read it.

KyleFromKitware:
@chuckatkins mind if I force-push some tweaks to this? I need to add the toolchain file and the custom builds of HDF5 etc. (and it looks like the scripts could do with some cleanup as well.)

scripts/ci/images/fedora-sanitizers/Dockerfile
chuckatkins: Yeah, that was leftover from a previous iteration.  I typically push squashed images anyways so it still ends up removed but I've updated this one to have it all in one step.

chuckatkins:
@kylefromkitware since the asan and msan builds aren't in place yet then let's go ahead and get this merged to update ubsan to use the common infrastructure and then you can rebase your asan and msan branches and update the image accordingly in those.

scripts/ci/images/fedora-sanitizers/Dockerfile
```diff
@@ -0,0 +1,51 @@
...
+    rm -rf source build
+
+# Setup clang with msan-enabled libc++
+#ENV CC=/usr/bin/clang
+#ENV CFLAGS="-fsanitize=memory -fsanitize-memory-track-origins=2 -fno-omit-frame-pointer"
+#ENV CXX=/usr/bin/clang++
+#ENV CXXFLAGS="-fsanitize=memory -fsanitize-memory-track-origins=2 -fno-omit-frame-pointer -stdlib=libc++ -nostdinc++ -I/opt/llvm/9.0.0/libcxx-msan/include/c++/v1 -L/opt/llvm/9.0.0/libcxx-msan/lib64 -Wl,rpath=/opt/llvm/9.0.0/libcxx-msan/lib64"
+
+# Misc cleanup of unneeded files
+RUN rm -rfv /tmp/* /var/cache/dnf
```
KyleFromKitware: Any plans to bring back the `adios2` non-root user?

scripts/ci/images/fedora-sanitizers/Dockerfile
```diff
@@ -0,0 +1,51 @@
...
+        CC=/usr/bin/clang \
+        CXX=/usr/bin/clang++ && \
+    /opt/cmake/bin/cmake \
+        -GNinja \
+        -DCMAKE_BUILD_TYPE=Release \
+        -DCMAKE_INSTALL_PREFIX=/opt/llvm/9.0.0/libcxx-msan \
+        -DCMAKE_INSTALL_RPATH='$ORIGIN' \
+        -DLLVM_ENABLE_PROJECTS="libcxx;libcxxabi" \
+        -DLLVM_USE_SANITIZER=MemoryWithOrigins \
+        -DLIBCXX_ENABLE_ABI_LINKER_SCRIPT=OFF \
```
KyleFromKitware: What's the purpose of this and the rpath adjustment? It gives me the following error when I try to compile a minimal program:

```console
CMake Error at /opt/cmake/share/cmake-3.16/Modules/CMakeTestCXXCompiler.cmake:53 (message):
  The C++ compiler

    "/usr/bin/clang++"

  is not able to compile a simple test program.

  It fails with the following output:

    Change Dir: /root/build/CMakeFiles/CMakeTmp
    
    Run Build Command(s):/usr/local/bin/ninja cmTC_aa33a && [1/2] Building CXX object CMakeFiles/cmTC_aa33a.dir/testCXXCompiler.cxx.o
    clang-8: warning: argument unused during compilation: '-stdlib=libc++' [-Wunused-command-line-argument]
    [2/2] Linking CXX executable cmTC_aa33a
    FAILED: cmTC_aa33a 
    : && /usr/bin/clang++  -nostdinc++ -stdlib=libc++ -isystem /opt/llvm/9.0.0/libcxx-msan/include/c++/v1 -fsanitize=memory  -L /opt/llvm/9.0.0/libcxx-msan/lib -Wl,-rpath,/opt/llvm/9.0.0/libcxx-msan/lib CMakeFiles/cmTC_aa33a.dir/testCXXCompiler.cxx.o  -o cmTC_aa33a   && :
    clang-8: warning: argument unused during compilation: '-nostdinc++' [-Wunused-command-line-argument]
    /usr/bin/ld: /usr/lib64/clang/8.0.0/lib/linux/libclang_rt.msan_cxx-x86_64.a(ubsan_type_hash_itanium.cc.o): undefined reference to symbol '_ZTIN10__cxxabiv121__vmi_class_type_infoE'
    /usr/bin/ld: //opt/llvm/9.0.0/libcxx-msan/lib/libc++abi.so.1: error adding symbols: DSO missing from command line
    clang-8: error: linker command failed with exit code 1 (use -v to see invocation)
```

chuckatkins: msan stuff removed

scripts/ci/images/fedora-sanitizers/Dockerfile
chuckatkins: No.  Setting up the non-root user for the CI jobs to run in was something we did when we first setup container builds as it seemed like a good practice.  Unfortunately it only really seems to work on Circle and breaks GH actions and Azure Pipelins (azure actually creates a non-root user to run in when initializing the container).  Just leaving the users alone makes the image compatible with  any of the CI services.

chuckatkins:
@KyleFromKitware There's some demons in the Ssc engine right now with ubsan / mpi / fedora so I'm dropiong the build until we can resolve it.  I'll merge this as soon as the builds are complete.

PR #1936, opened 2020-01-14
Add documentation for null engine and bp3 option for nvram

williamfgc:


williamfgc:
@ax3l thanks for pointing out the `null` engine is missing in the docs.

ax3l:
@williamfgc offline discussion: please do not forget to link the virtual engines, e.g. the default `File` engine :)

PR #1939, opened 2020-01-20
Add TSAN build

KyleFromKitware:


KyleFromKitware:
Yeah, I was afraid of that. It means we'll have to allocate a duplicate list while holding the lock, and then iterate over *that* with the lock released.

eisenhauer:
> Yeah, I was afraid of that. It means we'll have to allocate a duplicate list while holding the lock, and then iterate over _that_ with the lock released.

It's probably not the only place where we really should be doing that...

KyleFromKitware:
> It's probably not the only place where we really should be doing that...

Yeah, I'm sure my TSAN hunt will reveal more of them.

KyleFromKitware:
The sanitizers just passed, but I'm leaving this as WIP and don't merge because there's an intermittent race condition that I'm trying to track down that seems to happen somewhat rarely.

KyleFromKitware:
Also, this has to be rebased once #1932 is merged.

KyleFromKitware:
So, it looks like the sporadic race condition is in `cp_reader.c`. Unfortunately, it doesn't seem to be clear what functions are responsible for locking in this file, and I think it should be refactored with better locking. I'll take a look at it.

KyleFromKitware:
Alright, I've pushed up what I've done so far for the locking overhaul, but it's nowhere near ready. There's a deadlock that's blocking some of the tests, and TSAN's [deadlock detector](https://github.com/google/sanitizers/wiki/ThreadSanitizerDeadlockDetector) doesn't seem to be working for me. I'm still looking into it.

KyleFromKitware:
All tests have passed, but I would like to get a more thorough review of this, particularly in the SST code.

eisenhauer:
I'll take a look.

eisenhauer:
So, I have some concerns that can maybe be exemplified by this change in cp_writer.c:
```
    static void writeContactInfoFile(const char *Name, SstStream Stream,
                                     attr_list DPAttrs)
    {
        ...
        PTHREAD_MUTEX_UNLOCK(&Stream->DataLock);     //   Added
        AddNameToExitList(AbsoluteFilename);
        PTHREAD_MUTEX_LOCK(&Stream->DataLock);          //  Added
    }
```
Here, a Stream lock is dropped and reacquired around AddNameToExitList() which modifies global data.  The ExitList is the set of contact files that we'll remove if we catch a ^C.  Obviously TSAN ha a problem with that global data when you open multiple Streams and that data is sometimes accessed when holding different locks, a circumstance that TSAN considers an error.  (TSAN would also be worried about the unlocked access in the signal handler, but it doesn't see that in its run, so that is presumably undetected.)  

However dropping the lock around the call makes us no safer.  One of the baseline thread assumptions of ADIOS2 is that only the main program thread makes ADIOS calls.  If that assumption holds, the only the main program thread will end up in SstWriterOpen, which eventually calls AddNameToExitList(), (or SstWriterClose, which calls RemoveNameFromExistList).  So, one thread at a time in this code and it doesn't really matter if it's holding a lock or not.  The upshot is that dropping and reacquiring the lock adds overhead with no gain in safety, and no positive effect other than shutting up TSAN.

Now, I'm not too worried about particular change.  Its one lock cycle per Open()/Close().  Anything that's per file is a lesser worry than per variable, per timestep, etc.  But this is just the first locking change I run across in the cp_writer.c.  I'd like to take the time to get an idea of just how many more lock transactions we're adding...

eisenhauer:
I see significant problems, unfortunately...

It's not always safe to simply drop/reacquire the Stream->DataLock across MPI operations.  In some cases, like this MPI_AllReduce in WriterParticipateInReaderOpen:
````
    PTHREAD_MUTEX_UNLOCK(&Stream->DataLock);
    SMPI_Allreduce(&MyStartingTimestep, &GlobalStartingTimestep, 1, SMPI_LONG,
                   SMPI_MAX, mpiComm);
    PTHREAD_MUTEX_LOCK(&Stream->DataLock);
````
the purpose of the call is to achieve consensus about something across the ranks, and once achieved we'll use that information.  Dropping the DataLock allows the data that that consensus is based upon to change while this rank is in the AllReduce, destroying consensus by allowing a new race condition.  Here, a TimestepRelease message may come in causing a local timestep to be released.  (Then if the consensus decided to use that timestep and the reader tried to read it, we'll die because the data is gone.  If the lock was held, the release message wouldn't be processed until the mutex is unlocked and other state had been updated so that the timestep will remain around.)  While I'm not that familiar with TSAN, I'm guessing that it is poorly equipped to deal with the sort of race condition represented by the incoming message, where locking needs to be less fine-grained to deal with higher-level semantics.  Not to say that we couldn't make something like this work, but at least in this case making simple changes to make TSAN happy by dropping locks and reacquiring them are specifically *not* guaranteed safe and may well introduce new bugs.  This approach of breaking up larger critical sections into smaller ones just isn't something that can be done quite so casually/mechanically.

I apologize for not keeping a closer eye on the commits going on and raising these issues earlier, but I think for the SST code at least, this is going to require substantial changes.  I do think a TSAN build is valuable, and I believe that the things that TSAN finds are indeed things that we should look at, but as it stands I think this PR potentially introduces more actual bugs than it squashes.  @philip-davis, your data plane isn't tested here, but do you concur?  Also, how do you think the DataSpaces engine fare with this?

WRT how to go forward.  I know there's stuff other than SST that is addressed by this build, and it would be good to get it merged, but there's a lot of changes and I think each would have to be looked at carefully if we're not going to introduce new problems.  But let me propose a couple of broad changes in direction.  First, don't release locks across MPI calls.  All of ADIOS2 was designed from the start around the restriction that MPI calls were only to be done by the main program thread.  If it's causing TSAN problems that sometimes we do that with locks held and sometimes not, suppress those warnings.  They don't represent real problems.  Second, don't release locks across the CP_distribute/CP_aggregate calls or things that access CPInfo.  The latter is global data that gets initialized on our first call to create a stream in SST and is deallocated when all references to it are destroyed.  By design, it is read-only in between.  It doesn't matter what locks are held when accessing read-only data, so we can suppress those warnings.  (There is a reference count, but it should be accessed only by the main program thread because of the nature of ADIOS2 design.)  There may be other things in this category too (like the AddNameToExitList noted above), but those are the first things that come to mind from looking at cp_writer.   Generally it's the new UNLOCK/LOCK pairs that have broken up critical sections that cause me the most angst.

Again, sorry to suggest fundamental changes late in this process, but IMHO, it's got to be done...  Also, I'm happy to collaborate more directly in the process rather than just reviewing the outcome.  I'm having to guess about what TSAN didn't like based upon the nature of the changes, which isn't optimal.  We could work this out more jointly, or I can try to sort out SST issues myself...

**Comment by eisenhauer** [comment]
I see significant problems, unfortunately...

It's not always safe to simply drop/reacquire the Stream->DataLock across MPI operations.  In some cases, like this MPI_AllReduce in WriterParticipateInReaderOpen:
````
    PTHREAD_MUTEX_UNLOCK(&Stream->DataLock);
    SMPI_Allreduce(&MyStartingTimestep, &GlobalStartingTimestep, 1, SMPI_LONG,
                   SMPI_MAX, mpiComm);
    PTHREAD_MUTEX_LOCK(&Stream->DataLock);
````
the purpose of the call is to achieve consensus about something across the ranks, and once achieved we'll use that information.  Dropping the DataLock allows the data that that consensus is based upon to change while this rank is in the AllReduce, destroying consensus by allowing a new race condition.  Here, a TimestepRelease message may come in causing a local timestep to be released.  (Then if the consensus decided to use that timestep and the reader tried to read it, we'll die because the data is gone.  If the lock was held, the release message wouldn't be processed until the mutex is unlocked and other state had been updated so that the timestep will remain around.)  While I'm not that familiar with TSAN, I'm guessing that it is poorly equipped to deal with the sort of race condition represented by the incoming message, where locking needs to be less fine-grained to deal with higher-level semantics.  Not to say that we couldn't make something like this work, but at least in this case making simple changes to make TSAN happy by dropping locks and reacquiring them are specifically *not* guaranteed safe and may well introduce new bugs.  This approach of breaking up larger critical sections into smaller ones just isn't something that can be done quite so casually/mechanically.

I apologize for not keeping a closer eye on the commits going on and raising these issues earlier, but I think for the SST code at least, this is going to require substantial changes.  I do think a TSAN build is valuable, and I believe that the things that TSAN finds are indeed things that we should look at, but as it stands I think this PR potentially introduces more actual bugs than it squashes.  @philip-davis, your data plane isn't tested here, but do you concur?  Also, how do you think the DataSpaces engine fare with this?

WRT how to go forward.  I know there's stuff other than SST that is addressed by this build, and it would be good to get it merged, but there's a lot of changes and I think each would have to be looked at carefully if we're not going to introduce new problems.  But let me propose a couple of broad changes in direction.  First, don't release locks across MPI calls.  All of ADIOS2 was designed from the start around the restriction that MPI calls were only to be done by the main program thread.  If it's causing TSAN problems that sometimes we do that with locks held and sometimes not, suppress those warnings.  They don't represent real problems.  Second, don't release locks across the CP_distribute/CP_aggregate calls or things that access CPInfo.  The latter is global data that gets initialized on our first call to create a stream in SST and is deallocated when all references to it are destroyed.  By design, it is read-only in between.  It doesn't matter what locks are held when accessing read-only data, so we can suppress those warnings.  (There is a reference count, but it should be accessed only by the main program thread because of the nature of ADIOS2 design.)  There may be other things in this category too (like the AddNameToExitList noted above), but those are the first things that come to mind from looking at cp_writer.   Generally it's the new UNLOCK/LOCK pairs that have broken up critical sections that cause me the most angst.

Again, sorry to suggest fundamental changes late in this process, but IMHO, it's got to be done...  Also, I'm happy to collaborate more directly in the process rather than just reviewing the outcome.  I'm having to guess about what TSAN didn't like based upon the nature of the changes, which isn't optimal.  We could work this out more jointly, or I can try to sort out SST issues myself...

**Comment by KyleFromKitware** [comment]
@eisenhauer thanks for the feedback. The reason I released and re-acquired the lock in all those places was not due to complaints from TSAN, but because I was under the impression that they might cause a deadlock. I can try taking them out and seeing what happens. And yeah, I realize the SST code is a big change, so it's not surprising that it needs significant rework.

**Comment by eisenhauer** [comment]
One thing I could do that might be helpful is to document our assumptions, the driving concerns and the general plan of the current approach to concurrency in SST.  That's probably a document that should exist out there anyway...

**Comment by eisenhauer** [comment]
Some notes I wrote up on concurrency in SST...

Concurrency issues in SST arise from two primary sources:
1. The interactions between the main program thread and the network handler thread (which asynchronously executes all message handler functions).
2. The asynchronous operation of peers (I.E. readers and writers to which the entity of concern is connected)

Assumptions:
*SST's entry functions (defined in sst.h) are only called by the main program in any rank and are thus serialized.  (This matches the assumptions of ADIOS, which is not thread-safe.)
*The message handler functions (generally CP_*Handler(), registered in cp_common.c) are executed by the network handler thread, so they tend to be serialized as well.  However, some EVPath messaging calls can pause waiting for a remote response.  If this happens inside a network handler function, it is possible that another message will come in before the response that will trigger another handler.  In this case, network handler functions are not strictly serialized.

The Stream is the primary data structure for both the reader and the writer and incoming messages generally identify the local Stream to which they are responsive by carrying along the address of the relevant data structure.  In order to avoid a situation where we might reference free'd memory via this mechanism (perhaps because a message has been delayed), most such structures are never free'd until the messaging system is shut down.  (See LastCallFreeList.)  However, generally we're always concerned about message arrival delays, handlers being invoked at unexpected times causing state to change, etc.

Generally for locking, we're most concerned with locking data structures that might be read/written both by the network handler thread and the main program thread.  Since the Stream is the primary data structure and concurrent streams don't interact per se, we're actually just talking about locking the stream, and we're really mostly concerned about individual fields in the stream structure.  That is, to the extent that there are fields that are only read and written by the main program thread, the serialization assumption above means we don't have to lock for that access.  A similar condition holds for fields which are initialized at the time of stream creation and thereafter are only read.

At this point, of the EVPath functions used in SST, only CMget_conn() and CMinitiate_connection(), have the property that they might pause and a handler might be invoked before they complete.  The other functions can be viewed as local and atomic from this point of view, with the possible exception of CMwrite().  Currently the use of CMwrite() either succeeds immediately with the message queued or sent (if the ENET/RUDP transport is used, or sockets is used and the whole message is buffered), or might potentially block the process (sockets transport with blocking option, which will block the sending thread in write() if either the send is too large to buffer or the remote process hasn't serviced the network to read prior sends).  Generally this shouldn't impact concurrency/locking issues at least in the sense that it won't cause deadlock, but holding a lock across a possibly blocking call might delay the handling of incoming messages, which could affect overall system stability.

Where we've gotten into trouble before:
- most of our sources of deadlock have come from holding locks across EVPath functions like CMget_conn().  If a lock is held across CMget_conn(), the connection initiation protocol *might* wait for a response from the peer, allowing another message to arrive which might try to lock the same lock.  This will deadlock the network handler thread.
- holding locks across calls into the DataPlane functions might possibly have similar results, particularly for the EVPath data plane that also makes calls like CMget_conn().

**Comment by KyleFromKitware** [comment]
@eisenhauer Thanks for the explanation. It sounds like there are three categories of fields in the `SstStream` object:

1. Fields which are only read and written by the main thread. These definitely don't need to be locked.
2. Fields which are read and written by both the main thread and the networking threads. These definitely need to be locked.
3. Fields which are read and written by the main thread but only read (not written) by the networking threads. These may or may not need to be locked - the TSAN documentation calls this [racy/unsafe publication](https://github.com/google/sanitizers/wiki/ThreadSanitizerAboutRaces#racy-publication) and believes it should be guarded due to potential optimization and reordering by the compiler. We'll have to see if TSAN complains about these.

I think the best place to begin would be sifting through all of the `SstStream` fields and putting each one into one of these three categories. Perhaps we could even put them into different structs to make the distinction clear. Could you help with this?

**Comment by eisenhauer** [comment]
Sounds good.  I can help, but likely not until next week.  I’m tied up this morning and afternoon, and the weekend will s spoken for.

**Comment by chuckatkins** [comment]
Currently awaiting review from @eisenhauer

**Comment by eisenhauer** [comment]
WRT the racy/unsafe category above.  At a small scale compiler reordering is of course a worry.  However, I'm thinking more of situations where we know that no network handler can ever access the stream data structure (or the writer-side reader (WSR)) until we actually publish it by sending it in a message, that message gets received, processed, a response sent and that response arrives.  TSAN might not know that, but if the compiler is reordering code to the extent that it moves stuff from after a message exchange to before, then we've got bigger problems.  So if there are TSAN complaints, I'd be much more inclined to educate TSAN with annotations than incur overheads by overlocking (and if there aren't TSAN complaints, so much the better!)

I'm having some conceptual trouble that I think I'll eventually have to remedy by running TSAN myself.  For example, why, in SstReader.cpp, do we need a lock while we're removing all the Attributes in the IO, but we don't need it while we're removing all the Variables?  Those are pretty much treated the same way in SST, and I'm not sure why the differences in semantics would account for the different locking needs.

IMHO, we need to think about splitting up this PR.  I think this is too complex and has changes in too many places to reliably evaluate it as a whole.  Can we do this piece-wise?  I'd propose either:
1. Do a PR on the build-related changes, adding TSAN suppressions so that it runs cleanly.  Then we can try to eliminate those suppressions with future PRs, or
2. Reverse the order and do single-issue PRs, each with a description of the TSAN error, the usage circumstance in which it arises and the proposed fix.  When all the fixes are in, we can add TSAN to CI.

Offhand, 1) would seem to have the added attraction of not letting more TSAN warnings creep in while we're working our way through the existing ones...  Otherwise, I don't have strong feelings either way.

**Comment by eisenhauer** [comment]
With piece-wise goals in mind, I'm looking at the TransportMan parts here and wondering if the patches here are the right fix.  It looks like InitTransports() in BP3/4 calls OpenFilesAsync, which creates an async task to open a set of files.  Presumably this is to push open() times into the background, which is a fine goal as open is often costly.  But, the other semantics here have me wondering 1) if this is the right fix and 2) what the cost is.   WRT 1), there is just the one async task that will exit as soon as the files are open, and generally I think the goal is that they are actually completely open before we operate on anything there.  To that end, code in BP3/4 writer actually waits for the std::future result before trying to write data, etc.   But what about all the other calls in TransportMan?  This patch adds LockTransports() to a lot of them, so there are no locking errors for TSAN, but what that *doesn't* actually accomplish is making sure that that async task has finished before these other calls can continue.  From that point of view, there's a race condition:  If that async task is delayed in starting, these calls can happily grab that lock and go forward.  So if the goal is to make sure that the TransportMan state is sensible, that is, we don't call GetTransportsTypes() or WriteFiles() before the OpenFilesAsync has opened the files, then this patch does not accomplish that.  What it does do, from the point of view of 2) is to introduce a *lot* of (probably) unnecessary locking.  That async task that was introduced to try not to pay the open() cost right away might have been short-lived, but pretty much every call into TransportMan thereafter will pay the price of locking and unlocking a mutex for the duration of program.  This includes Writes, Seeks, Flushes, Closes, Clears, getting file types, etc.  I think we need a fix here where we're not paying extra costs for the duration of the program just because we wanted something during initialization to maybe happen asynchronously.
(Apologies if I've misunderstood what's going on in TransportMan and BP3/4 Writers.)

**Comment by eisenhauer** [comment]
*Probably* the right thing to do with the TransportMan thing is to move the std::future inside the TransportMan object.  OpenFilesAsync wouldn't return std::future, but would instead init a std::future private member.  Where this PR has LockTransports(), it should instead do "if (m_FutureOpenFiles.valid()) { m_FutureOpenFiles.get(); }", and those checks could be deleted from BP3/4Writer.cpp.  The .valid() check is likely just a NULL check, which isn't adding a lot of overhead if engines don't use OpenFilesAsync.   I haven't double-checked to see if this satisfies the BP3/4 situation, but I'd think it would.  Also, I don't know if TSAN is smart enough to be satisfied with this, but I would hope so...

**Comment by KyleFromKitware** [comment]
> it should instead do
>
> ```c++
> if (m_FutureOpenFiles.valid()) {
>   m_FutureOpenFiles.get();
> }
> ```

And what should it do if `!m_FutureOpenFiles.valid()`? Should we `wait()` instead?

**Comment by eisenhauer** [comment]
Actually, get() setting valid to false is great.  If we find valid is false, then either nobody ever created an async task, or they did, and someone else already confirmed that it has completed.  If either of those things are true, we're set.

I have a version I'm playing with here:
https://github.com/eisenhauer/ADIOS2/tree/TransportManTrial
But now that I've rebased I think I have to update CMake.  However, before rebasing this took care of the TransportMan TSAN issues.  (There is an additional problem with ctime() in helper/adiosSystem.cpp, but that's easily remedied by using ctime_r() and a stack buffer instead.)

**Comment by KyleFromKitware** [comment]
I'm looking at the `TransportManTrial` right now, and I like the approach so far. I'll try it in a TSAN build later today.

One thing I'm wondering is if perhaps the `future` should be `mutable` instead of removing the `const` qualifier from the functions that use it. While the functions are technically mutating the state of the class, they are not mutating *data* that we care about.

**Comment by KyleFromKitware** [comment]
> I'd propose either:
>
> 1. Do a PR on the build-related changes, adding TSAN suppressions so that it runs cleanly. Then we can try to eliminate those suppressions with future PRs, or
> 2. Reverse the order and do single-issue PRs, each with a description of the TSAN error, the usage circumstance in which it arises and the proposed fix. When all the fixes are in, we can add TSAN to CI.

I would prefer option 2. If we add suppressions and remove them later on, we are more at risk of forgetting to remove one (unless we are very careful). Option 1 also creates a window where the project pretends to be TSAN compliant but is not actually so, whereas option 2 only does it at the end. The risk of issues creeping in along the way isn't that big of a deal either - it's happened with some of my other PRs, and I was able to deal with it without much of an issue.

**Comment by eisenhauer** [comment]
Changing the future to mutable probably makes sense...  (I'm really here for the distributed system and performance aspects and am largely apathetic about the niceties of particular languages.)

So, lets run with option 2, maybe starting with the TransportMan tweaks.  Shall I fix up my branch for a PR?

**Comment by NAThompson** [comment]
@KyleFromKitware : This is a critical PR. If possible, could you merge the diffs that eliminate race conditions that you've already found, even if the TSAN build is not fully clean?

**Comment by KyleFromKitware** [comment]
If the TSAN build isn't fully clean, then we'll have to make it "not required' until it's fully clean.

**Comment by NAThompson** [comment]
@KyleFromKitware : Right, but perfect is the enemy of the good, and your changes are a huge improvement.

**Comment by KyleFromKitware** [comment]
Indeed... the tricky part is going to be figuring out a way to make a build optional. As far as I can tell, CircleCI does not support "optional" builds, so we'll have to fake it with a `ctest | /bin/true` command.

**Comment by eisenhauer** [comment]
WRT merging what's there, I wouldn't suggest doing that.  From examination we know that some of the changes which split critical sections might make TSAN warnings go away, but introduce new possibilities for race conditions of the sort that TSAN doesn't detect,.  See the comment above about dropping locks around the MPI_AllReduce.  We're going with single-issue PRs that tackle individual races to try to insure we don't do more harm than good.

**Comment by eisenhauer** [comment]
@KyleFromKitware, can we separate and merge the LocalDateTime commit as well?

File: source/adios2/helper/adiosSystem.cpp [hunk 1]
```diff
@@ -48,10 +48,20 @@ bool IsLittleEndian() noexcept
 
 std::string LocalTimeDate() noexcept
 {
+    struct tm now_tm;
+    char buf[30];
```
chuckatkins: I'd rather see this reworked to entirely use `std::chrono` for formatting rather than dropping it to the c functions.

eisenhauer: That works too.  Just observing that this is an independent fix that can go in at any time...

eisenhauer:
Repeating an internal Slack message: To my understanding, the not-TSAN-compliant code involving TransportMan wasn’t actually erroneous. The BP*Writers essentially forked a thread to do something, went off and did something else briefly, waited for that thread, and went about their usual business. TSAN’s problem boiled down to there being no enforcement of this non-erroneous behavior. The “fork” happened inside a not-always-used TransportMan initialization method, the join happened out in engine code.  No other engine called that async method, but if they had, or if the BPWriter changed and if they called another TransportMan method before waiting for the result, something bad _might_ have happened. No locking or other mechanisms protected the simultaneously accessible data in TransportMan, so TSAN rightly screamed “UNSAFE!“. We were able to make TSAN happy by moving the join into TransportMan, but that was not without cost. Rather than a one-time check for thread completion that only happened in BP when that thread-forking-method was used, **_every method call_** in TransportMan now checks to see if there was a thread forked and if it is done yet. So we’ve converted a happens-once check only incurred by the engines that used that async method to a continuous extra overhead that is paid by every user of the class in every method call. Yes, one can argue that it’s “better” for this internal class to enforce its own “safety” on its methods, that this protects against future engine developers who might be less careful, etc. etc..  But we should be clear that in TransportMan while the “safety” gained is abstract, the performance costs are real.

IMHO, there is a limit to the measures we should take to make TSAN happy. I think it’s a great tool, and having it point out questionable situations is fantastic. But I also think we have to look at each TSAN hit carefully and balance costs vs. gains. And here “costs” includes both developer time and execution time. Sorting out exactly what TSAN isn’t happy about, whether it represents a problem that will occur in real life vs. a more abstract issue, and figuring out a code transformation that will make TSAN happy without introducing new bugs or compromising performance is far from trivial and may take effort away from things that actually impact the success of ADIOS. That is, if we’re not performant on Summit, it won’t matter that our code is free of possible race conditions on some future architecture where access to 4-byte integers in memory is somehow not atomic...

source/adios2/helper/adiosSystem.cpp
KyleFromKitware: I used the C API because I couldn't find the formatting function in C++, so I didn't think there was one, but it looks like it was on a different page from where I was looking. Poor documentation... I will fix it.

KyleFromKitware:
I've pushed up the fix for `LocalTimeDate`, but I'll open a separate PR for it.

NAThompson:
@KyleFromKitware : It looks like you've removed a large number of race conditions, but we haven't yet benefited from all this great work. Is there any chance you can cherry pick the race condition fixes that are uncontroversial and build them into separate PRs? I think cb9b45970564f7a88183131678dfdfff1b1682a8 looks good, as does 8861cb45a6ee52fecfb98407da88f7ee0f8736ff and d86cdcb40ee417d065b796719986f25715748959.
 

@eisenhauer : You've made a large number of pretty dramatic improvements to ADIOS recently; do any of them obviate the issues encountered here?

eisenhauer:
Unfortunately, it's hard to separate individual bits from the larger approach.  But I am still kicking this along, just distracted by the other PRs lately.  The reader side looks solid these days, but the writer side is a lot more active asynchronously.

chuckatkins:
@eisenhauer @KyleFromKitware With #2136 now merged can we close this or is it still a placeholder for more work to do?

eisenhauer:
I think close...

chuckatkins:
Closing now that we have a tsan-clean build as of #2136

PR #1940, opened 2020-01-22
WIP: Add TSAN OpenMPI build

KyleFromKitware:


KyleFromKitware:
Looks like OpenMPI has major problems with TSAN (not surprising).

KyleFromKitware:
Closing because it is probably not feasible to use OpenMPI in TSAN. We can always revisit this later.

PR #1942, opened 2020-01-23
enable multi-apps in ssc

shevateng0:
Need the ssc engine to run multiple applications (>2). However, the original program can only communicate between the intra communicators of two applications. I add some codes to allow the commnunication among the intra communicators of multiple applications.


shevateng0:
This branch is messed up. I will close this PR and open a new one.

PR #1943, opened 2020-01-23
enable multi-apps for SSC

shevateng0:
Need the ssc engine to run multiple applications (>2). However, the original program can only communicate between the intra communicators of two applications. I add some codes to allow the commnunication among the intra communicators of multiple applications.

### DIFF
--- a/source/adios2/engine/ssc/SscReader.cpp
+++ b/source/adios2/engine/ssc/SscReader.cpp
@@ -117,6 +117,235 @@ void SscReader::SyncMpiPattern()
         std::cout << "SscReader::SyncMpiPattern, World Rank " << m_WorldRank
                   << ", Reader Rank " << m_ReaderRank << std::endl;
     }
+
+
+    std::vector<int> lrbuf;
+    std::vector<int> grbuf;
+    int m_AppID = 0;
+    int m_AppSize = 0;  //Total number of applications.
+    std::vector<std::vector<int>> m_WriterGlobalMpiInfo;
+    std::vector<std::vector<int>> m_ReaderGlobalMpiInfo;
+
+
+    // Process m_WorldRank == 0 to gather all the local rank m_WriterRank, and find out all the m_WriterRank == 0
+    if (m_WorldRank == 0)
+    {
+        grbuf.resize(m_WorldSize);
+    }
+
+    MPI_Barrier(MPI_COMM_WORLD);
+    MPI_Gather(&m_ReaderRank, 1, MPI_INT, grbuf.data(), 1, MPI_INT, 0, MPI_COMM_WORLD);
+    MPI_Barrier(MPI_COMM_WORLD);
+
+
+    std::vector<int> AppStart;  // m_WorldRank of the local rank 0 process
+    if (m_WorldRank == 0)
+    {
+        for ( int i = 0; i < m_WorldSize; ++i )
+        {
+            if (grbuf[i] == 0)
+            {
+                AppStart.push_back(i);
+            }
+        }
+        m_AppSize = AppStart.size();
+    }
+
+
+    //Each local rank 0 process send their type (0 for writer, 1 for reader) to the world rank 0 process
+    //The AppStart are re-ordered to put all writers ahead of all the readers.
+    std::vector<int> AppType;  // Vector to record the type of the local rank 0 process
+    if (m_ReaderRank == 0) // Send type from each local rank 0 process to the world rank 0 process
+    {
+        if (m_WorldRank == 0) // App_ID
+        {
+            AppType.resize(m_AppSize);
+            for (int i = 0; i < m_AppSize; ++i)
+            {
+                if (i == 0)
+                {
+                    AppType[i] = 1;;
+                }
+                else
+                {
+                    int tmp = 1;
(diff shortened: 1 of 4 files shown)


source/adios2/engine/ssc/SscReader.cpp
```diff
@@ -117,6 +117,235 @@ void SscReader::SyncMpiPattern()
...
+        {
+            ilen = m_ReaderGlobalMpiInfo[i].size();
+        }
+        MPI_Bcast(&ilen, 1, MPI_INT, 0, MPI_COMM_WORLD);
+        m_ReaderGlobalMpiInfo[i].resize(ilen);
+        MPI_Bcast(m_ReaderGlobalMpiInfo[i].data(), ilen, MPI_INT, 0, MPI_COMM_WORLD); // Broadcast readerinfo size
+    }
+
+
+    std::cout << "WorldRank " << m_WorldRank << std::endl;
```
JasonRuonanWang: Could you remove these printing code when it's done?

source/adios2/engine/ssc/SscReader.cpp
```diff
@@ -117,6 +117,235 @@ void SscReader::SyncMpiPattern()
...
+    std::vector<std::vector<int>> m_ReaderGlobalMpiInfo;
+
+
+    // Process m_WorldRank == 0 to gather all the local rank m_WriterRank, and find out all the m_WriterRank == 0
+    if (m_WorldRank == 0)
+    {
+        grbuf.resize(m_WorldSize);
+    }
+
+    MPI_Barrier(MPI_COMM_WORLD);
```
JasonRuonanWang: You don't need to call MPI_Barrier around MPI collective calls.

source/adios2/engine/ssc/SscWriter.cpp
```diff
@@ -34,6 +36,22 @@ SscWriter::SscWriter(IO &io, const std::string &name, const Mode mode,
     m_GlobalWritePattern.resize(m_WriterSize);
     m_GlobalReadPattern.resize(m_ReaderSize);
 
+    /*
```
JasonRuonanWang: Can these be removed?

source/adios2/engine/ssc/SscWriter.cpp
```diff
@@ -120,6 +138,233 @@ void SscWriter::SyncMpiPattern()
         std::cout << "SscWriter::SyncMpiPattern, World Rank " << m_WorldRank
                   << ", Writer Rank " << m_WriterRank << std::endl;
     }
+
+
+    std::vector<int> lrbuf;
+    std::vector<int> grbuf;
+    int m_AppID = 0;
```
JasonRuonanWang: Same as above, these should be in the header file

source/adios2/engine/ssc/SscWriter.cpp
```diff
@@ -120,6 +138,233 @@ void SscWriter::SyncMpiPattern()
...
+    std::vector<std::vector<int>> m_ReaderGlobalMpiInfo;
+
+
+    // Process m_WorldRank == 0 to gather all the local rank m_WriterRank, and find out all the m_WriterRank == 0
+    if (m_WorldRank == 0)
+    {
+        grbuf.resize(m_WorldSize);
+    }
+
+    MPI_Barrier(MPI_COMM_WORLD);
```
JasonRuonanWang: Barriers are not needed

JasonRuonanWang:
@shevateng0 Thanks for the effort. It looks good but I made some minor comments. You will also need to use clang-format on the files you have changed before they can pass the tests.

source/adios2/engine/ssc/SscReader.cpp
shevateng0: I have moved all of the printing codes to the end. They will be printed when m_Verbosity >= 10.

source/adios2/engine/ssc/SscReader.cpp
```diff
@@ -117,6 +117,235 @@ void SscReader::SyncMpiPattern()
         std::cout << "SscReader::SyncMpiPattern, World Rank " << m_WorldRank
                   << ", Reader Rank " << m_ReaderRank << std::endl;
     }
+
+
+    std::vector<int> lrbuf;
+    std::vector<int> grbuf;
+    int m_AppID = 0;
```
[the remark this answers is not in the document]
shevateng0: > These should be put into the header file as class member variables.

I have put them into the header file.

source/adios2/engine/ssc/SscWriter.cpp
shevateng0: > Can these be removed?

All the redundant comments have been removed.

source/adios2/engine/ssc/SscWriter.cpp
shevateng0: > Same as above, these should be in the header file

All the member variables are moved to the header file now.

source/adios2/engine/ssc/SscWriter.cpp
shevateng0: > Barriers are not needed

Thanks. I have deleted those barriers.

PR #1944, opened 2020-01-24
minor tweak of mgard calling as a result of api change.

qliu21:


qliu21:
@williamfgc @pnorbert Any ideas why these tests failed because of HDF5?

CMake Error at /opt/cmake/share/cmake-3.16/Modules/FindPackageHandleStandardArgs.cmake:164 (message):
  Could NOT find HDF5 (missing: HDF5_LIBRARIES HDF5_INCLUDE_DIRS C) (found
  version "")

chuckatkins:
@qliu21 I've just rebased your branch to clear the CI problem.  That being said...

Is there any way for ADIOS to know which CPI needs to be used?  I.e. can we check for something like a version number at configure-time or with a pre-processor definition from the mgard headers so we can know which API needs to be used?

qliu21:
@chuckatkins I haven't done something like that in MGARD but will do this. Most of the recent API changes were due to the initial flawed design.

chuckatkins:
> @chuckatkins I haven't done something like that in MGARD but will do this.

Thanks.  As soon as that's available we will adjust the requirements so ADIOS can ensure that an adequate version is found. Is is find for now until that's available though.

PR #1955, opened 2020-02-04
Add an explicit path for aligned memory.

chuckatkins:
Fix unaligned memory in `CopyToBuffer`

The issue seems to show up sporadically in ubsan.  This should at least address it for this function.

### DIFF
--- a/source/adios2/helper/adiosMemory.inl
+++ b/source/adios2/helper/adiosMemory.inl
@@ -78,8 +78,19 @@ template <class T>
 void CopyToBuffer(std::vector<char> &buffer, size_t &position, const T *source,
                   const size_t elements) noexcept
 {
-    const char *src = reinterpret_cast<const char *>(source);
-    std::copy(src, src + elements * sizeof(T), buffer.begin() + position);
+    // if alignment is appropriate, use type-specific copy
+    // otherwise use raw bytes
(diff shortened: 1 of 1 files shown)


chuckatkins:
@KyleFromKitware I believe this is also related to you ubsan failure see in #1932

williamfgc:
@chuckatkins I think it is great to have this. My input is to have it separate from the current CopyToBuffer function and specialize for `long double`, or the future non-standars types, as suggested by @KyleFromKitware in #1932

KyleFromKitware:
Yes, it is related.

KyleFromKitware:
I'm not sure why ubsan is flagging it as it exists right now. Right now it's copying raw bytes, which should be fine from an ubsan perspective. Am I missing something?

chuckatkins:
So, digging more into a few of the implementations for `std::copy`, I'm not sure this will actually accomplish anything.  In `libc++`, any combination of inputs convertible to the same raw pointer types (`vector<t>::iterator` is convertible to `T*` for instance), eventually ends up with a `memmove`.  So really the way to address the long double issue would be to specialize the function as in #1932 but to only invoke the specialization if the alignment is correct for the type.  If the pointers aren't aligned then there's no real choice but to memcpy / memmove and just copy the padding along with it.

PR #1959, opened 2020-02-11
sst: Re-implement SMPI_ interfaces as C wrappers of Comm encapsulation

bradking:
`adios2::helper::Comm` encapsulates ADIOS multi-process communication, but it is a C++ interface and SST is implemented in C.  Create a C wrapper to use the Comm encapsulation from SST.  This will enable the SST engine to be used by both serial and parallel applications from a single build.


CMakeLists.txt
```diff
@@ -3,7 +3,7 @@
 # accompanying file Copyright.txt for details.
 #------------------------------------------------------------------------------#
 
-cmake_minimum_required(VERSION 3.6)
+cmake_minimum_required(VERSION 3.12)
```
bradking: CMake 3.12 is needed in order to support compiling SST as an object library with its own dependencies and then linking the main adios2 library with the objects.  We should not merge this PR until we decide in general if bumping the requirement to CMake 3.12 is okay

PR #1968, opened 2020-02-14
bindings: Factor out MPI-specific bindings into separate sources

bradking:
Move code conditioned by `ADIOS2_HAVE_MPI` into source files that are built only when that condition is true.  Later this will be useful for compiling the MPI-specific sources in a separate library.


bradking:
This PR has no functional changes but re-organizes code in a way that will help split MPI-specific code into a separate library.  I'm submitting it as a preliminary PR to avoid cluttering the main change with simple content re-organization.  It also avoids lots of `#ifdef ADIOS2_HAVE_MPI` conditions in the sources.

PR #1970, opened 2020-02-17
Allow setting absolute paths for CMAKE_INSTALL_BINDIR

franzpoeschel:
By default, Cmake will set the `CMAKE_INSTALL_BINDIR` variable to `"bin"`, denoting a path within
`CMAKE_INSTALL_PREFIX`. It may also be an absolute path which is currently
not considered in the post-install scripts. This PR allows using absolute paths, too.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -156,8 +156,9 @@ install(FILES ${PROJECT_BINARY_DIR}/source/adios2/common/ADIOSConfig.h
 #------------------------------------------------------------------------------#
 # Some specialized shared library and RPATH handling
 #------------------------------------------------------------------------------#
-set(ADIOS2_RUN_INSTALL_TEST TRUE)
-if(BUILD_SHARED_LIBS)
+option(ADIOS2_RUN_INSTALL_TEST "Enable / disable the install test" TRUE)
+mark_as_advanced(ADIOS2_RUN_INSTALL_TEST)
+if(BUILD_SHARED_LIBS AND ADIOS2_RUN_INSTALL_TEST)
   set(CMAKE_INSTALL_RPATH_USE_LINK_PATH TRUE)
   string(REGEX REPLACE "[^/]+" ".." relative_base "${CMAKE_INSTALL_LIBDIR}")
 
@@ -171,6 +172,7 @@ if(BUILD_SHARED_LIBS)
       else()
         set(CMAKE_INSTALL_NAME_DIR
           "${CMAKE_INSTALL_PREFIX}/${CMAKE_INSTALL_LIBDIR}")
+        message(STATUS "Skipping install tests due to platform specific rpath oddities")
         set(ADIOS2_RUN_INSTALL_TEST FALSE)
       endif()
     else()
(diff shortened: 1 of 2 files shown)


chuckatkins:
What's the use case for this?  Is the intent that `CMAKE_INSTALL_BINDIR` would live outside the specified install prefix?

franzpoeschel:
I am currently experimenting with the [Nix](https://nixos.org/nix/) package manager and ran into this exact issue. So my use case is having an external tool that automatically configures Cmake, using absolute paths. Certainly a rare use case, but it would be handy not having to keep around a patched version of ADIOS2 for this purpose.

chuckatkins:
This will address the "make install" step but the install tests will still be broken since they try to install into a temporary install tree.  Ideally there needs to be an additional set of checks where the install tests are added to skip them if any absolute paths are detected in the install variables.  It's not a particularly awful use case though since it's not a blocker for CI and the issue is really that for a specific configuration outside of a development scenario, a few of the development tests don't work, so I'm okay with not holding this up for that.

franzpoeschel:
Ok, thanks for the head-ups, so I don't run into that problem later on. In that case, it would probably be fair to print a warning if an absolute path is used. Should I add that?

chuckatkins:
> Ok, thanks for the head-ups, so I don't run into that problem later on. In that case, it would probably be fair to print a warning if an absolute path is used. Should I add that?

It should be done at the CMake level, if anywhere, and if you're already going to add the check for absolute paths then you might as well use it to disable the install tests rather than print a warning.  We actually already have a check to disable the install tests for certain problematic rpath situations so the easiest way to address the issue is probably to just make the flag a configurable option:
```cmake
index 02df62b0..94d5a619 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -159,7 +159,8 @@ install(FILES ${PROJECT_BINARY_DIR}/source/adios2/common/ADIOSConfig.h
 # Some specialized shared library and RPATH handling
-set(ADIOS2_RUN_INSTALL_TEST TRUE)
+option(ADIOS2_RUN_INSTALL_TEST "Enable / disable the install test" TRUE)
+mark_as_advanced(ADIOS2_RUN_INSTALL_TEST)
 if(BUILD_SHARED_LIBS)
   set(CMAKE_INSTALL_RPATH_USE_LINK_PATH TRUE)
   string(REGEX REPLACE "[^/]+" ".." relative_base "${CMAKE_INSTALL_LIBDIR}")
@@ -174,6 +175,7 @@ if(BUILD_SHARED_LIBS)
       else()
         set(CMAKE_INSTALL_NAME_DIR
           "${CMAKE_INSTALL_PREFIX}/${CMAKE_INSTALL_LIBDIR}")
+        message(STATUS "Skipping install tests due to platform specific rpath oddities")
         set(ADIOS2_RUN_INSTALL_TEST FALSE)
       endif()
     else()

```

Then you can just configure with `cmake ... -DCMAKE_INSTALL_BINDIR=/foo/bar/bin ... -DADIOS2_RUN_INSTALL_TEST=OFF ... /path/to/source`.  Why don't you go ahead and just add that to the PR here and then we can not worry about the issue in the future if it comes up.

franzpoeschel:
Thank you for your help!

PR #1971, opened 2020-02-17
WIP:  Test reuse of incoming conn for DP

eisenhauer:


eisenhauer:
@jychoi-hpc Would you please try this newest iteration.  It occurred to me that it might be acceptable to have EVPATH DP simply default to re-using the incoming connection if an attempt the attempt to create its own connection failed.  That failure is the main symptom of your peer being behind a NAT, and right now what we do instead is to simply not send messages, which means the program doesn't work anyway.  So this is I think an acceptable fallback that should also kind of transparently handle your situation.  If this works for you, it could be merged immediately.

jychoi-hpc:
Yes, this works. Thank you for the quick fix. Once this merged, I will test again with other features.

eisenhauer:
OK, give it a go...

PR #1974, opened 2020-02-18
table: Enable engine without MPI by cleaning up sub-adios APIs

bradking:
Manage table's sub-adios through the core API.  Avoid using the public C++11 bindings API from the core implementation.

With this we can now access the core ADIOS constructor that accepts the Comm abstraction instead of a raw MPI_Comm.  We can use `Comm::World()` instead of `MPI_COMM_WORLD`.  This allows us to enable the Table engine even when ADIOS is built without MPI.


chuckatkins:
@JasonRuonanWang please take a look

bradking:
Now that we can use the table engine without MPI, its tests are now covered by the sanitizers for the first time.  It looks like some fixes may be needed for that.

PR #1980, opened 2020-02-21
adios_reorganize: do not stay in infinite loop reading an unclosed BP…

pnorbert:
… file if the reading engine is BPFile, BP3, File or the default (picked by empty string). Instead, explain what's going on and finish.


chuckatkins:
It looks like travis is still getting tripped up since the commit hash hasn't changed.  Try just force-pushing a no-op to change the hash and re-cycle the ci: `git commit --no-edit --amend && git push -f`

PR #1981, opened 2020-02-21
added one sided post pull and one sided fence pull modes in ssc

JasonRuonanWang:


JasonRuonanWang:
@chuckatkins Is the clang-format version changed? I got a format error on a file which I didn't touch at all...

diff --git a/source/adios2/core/IOMPI.cpp b/source/adios2/core/IOMPI.cpp

index fabb549..b0296b1 100644

--- a/source/adios2/core/IOMPI.cpp

+++ b/source/adios2/core/IOMPI.cpp

@@ -61,7 +61,6 @@ struct ThrowError

     }

     std::string Err;

 };

-

 }

PR #1984, opened 2020-02-25
TransportMan TSAN changes

eisenhauer:
Change OpenFilesAsync so that it no longer returns a future, but rather the future remains in the class and subsequent calls into the class wait for the async task before they continue.  This preserves the current state of functionality while clearing up TSAN warnings, at a cost of an additional .valid() check on every public member invocation (rather than the one-time cost when those checks were in BP3/BP4 writer).

### DIFF
--- a/source/adios2/toolkit/transportman/TransportMan.cpp
+++ b/source/adios2/toolkit/transportman/TransportMan.cpp
@@ -75,6 +75,7 @@ void TransportMan::OpenFiles(const std::vector<std::string> &fileNames,
                              const std::vector<Params> &parametersVector,
                              const bool profile)
 {
+    WaitForAsync();
     for (size_t i = 0; i < fileNames.size(); ++i)
     {
         const Params &parameters = parametersVector[i];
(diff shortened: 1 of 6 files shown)


source/adios2/toolkit/transportman/TransportMan.cpp
```diff
@@ -421,5 +434,14 @@ void TransportMan::CheckFile(
     }
 }
 
+void TransportMan::WaitForAsync() const
+{
+    // Ensure any prior OpenFilesAsync is complete
+    if (m_FutureOpenFiles.valid())
+    {
+        m_FutureOpenFiles.get();
+    }
```
chuckatkins: Can the functions just call `m_FutureOpenFiles.wait()` instead?

KyleFromKitware: `wait()` also requires `valid()` to be true.

chuckatkins: Then at least `wait` instead of `get`?

eisenhauer: I think consuming the result with get() is the thing to do.  If get() sets valid to false, then it's better for every call after the first to fail on valid() and not have to call wait() too.   Impact for multiple calls to OpenFilesAsync too?

KyleFromKitware:
@eisenhauer Thanks. All tests pass on my local TSAN build (and fail miserably without this change.)

source/adios2/helper/adiosSystem.cpp
```diff
@@ -50,7 +50,6 @@ std::string LocalTimeDate() noexcept
 {
     std::time_t now =
         std::chrono::system_clock::to_time_t(std::chrono::system_clock::now());
-
```
KyleFromKitware: There is still a line removal here. Please squash the commits and make sure `adiosSystem.cpp` is unchanged.

PR #1986, opened 2020-02-26
Fix race condition in LocalTimeDate()

KyleFromKitware:


KyleFromKitware:
Looks like GCC 4.8 does not have `std::put_time()`. We can:

1. Upgrade to GCC >= 5, or
2. Revert back to the C code I originally had

@eisenhauer what would you prefer?

KyleFromKitware:
Hmmm, looks like Intel doesn't have it either. I'll go ahead and switch it back to the C version.

eisenhauer:
I'm going to merge this as a reentrant improvement to what's there.  If we come up with a pure C++ solution later, we can switch to that.

PR #1987, opened 2020-02-26
bindings: Factor out MPI-specific Fortran bindings into separate sources

bradking:
~~Combine the `bindings/Fortran/modules/{mpi,nompi}/*` sources to provide a single definition of the modules providing the `adios2_init` and `adios2_open` generic procedures.  Include the MPI signatures only when ADIOS is built with MPI enabled.~~ EDIT: This part was moved to #1994.

Split the implementations of the serial and MPI variants of the Fortran bindings into separate source files.  For each of the `adios2_init` and `adios2_open` generic procedures, ~~all signatures must remain declared in a single module, so use Fortran submodules~~ (EDIT: generic procedure interfaces can have different signatures added by different modules) put the implementations in separate modules in separate sources.  Later this will be useful for compiling the MPI-specific sources in a separate library.

### DIFF
--- a/bindings/Fortran/modules/adios2_adios_init_mod_serial.f90
+++ b/bindings/Fortran/modules/adios2_adios_init_mod_serial.f90
@@ -1,0 +1,63 @@
+!
+! Distributed under the OSI-approved Apache License, Version 2.0.  See
+!  accompanying file Copyright.txt for details.
+!
+!  adios2_adios_init_mod_serial.f90 : ADIOS2 Fortran bindings for ADIOS
+!                                     class Init functions (serial variants)
+!
+
+module adios2_adios_init_mod_serial
+    use adios2_parameters_mod
+    use adios2_functions_mod
+    implicit none
+
+    interface adios2_init
+        module procedure adios2_init_serial
+        module procedure adios2_init_debug_serial
+        module procedure adios2_init_config_serial
+        module procedure adios2_init_config_debug_serial
+    end interface
+
+contains
+
+    subroutine adios2_init_serial(adios, adios2_debug_mode, ierr)
+        type(adios2_adios), intent(out) :: adios
+        logical, intent(in) :: adios2_debug_mode
+        integer, intent(out) :: ierr
+
+        call adios2_init_config_serial(adios, char(0), adios2_debug_mode, ierr)
+
+    end subroutine
+
+    subroutine adios2_init_debug_serial(adios, ierr)
+        type(adios2_adios), intent(out) :: adios
+        integer, intent(out) :: ierr
+
+        call adios2_init_config_serial(adios, char(0), .true., ierr)
+
+    end subroutine
+
+    subroutine adios2_init_config_serial(adios, config_file, adios2_debug_mode, ierr)
+        type(adios2_adios), intent(out) :: adios
+        character*(*), intent(in) :: config_file
+        logical, intent(in) :: adios2_debug_mode
+        integer, intent(out) :: ierr
+        ! local
+        integer debug_mode
+
+        debug_mode = adios2_LogicalToInt(adios2_debug_mode)
+        call adios2_init_config_serial_f2c(adios%f2c, config_file, debug_mode, ierr)
+        if( ierr == 0 ) adios%valid = .true.
+
+    end subroutine
+
+    subroutine adios2_init_config_debug_serial(adios, config_file, ierr)
+        type(adios2_adios), intent(out) :: adios
+        character*(*), intent(in) :: config_file
+        integer, intent(out) :: ierr
+
+        call adios2_init_config_serial(adios, config_file, .true., ierr)
+
+    end subroutine
+
+end module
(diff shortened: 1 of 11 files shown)


bindings/Fortran/modules/adios2_adios_init_mod_serial.f90
```diff
@@ -2,44 +2,35 @@
...
-        module procedure adios2_init_config_debug
-    end interface
-
 contains
 
-    subroutine adios2_init_nompi(adios, adios2_debug_mode, ierr)
+    module subroutine adios2_init_serial(adios, adios2_debug_mode, ierr)
         type(adios2_adios), intent(out) :: adios
         logical, value, intent(in) :: adios2_debug_mode
         integer, intent(out) :: ierr
```
bradking: The PGI Fortran compiler is unhappy with the use of the `value` attribute on `adios2_debug_mode`.  It complains:

> ```
> PGF90-S-1057-Definition argument name _V_adios2_debug_mode does not match declaration argument name adios2_debug_mode (/.../bindings/Fortran/modules/adios2_adios_init_mod_serial.f90: 16)
>   0 inform,   0 warnings,   1 severes, 0 fatal for adios2_init_serial
> ```

This happens only when it is a `module subroutine`, but making it such is necessary to isolate the implementation in a submodule.

Ideas anyone?

bradking:
gfortran 4.8 is not happy with the `module subroutine` syntax.  According to the [GCC 6 Release Notes](https://gcc.gnu.org/gcc-6/changes.html), Fortran 2008 submodule support was not added until gfortran 6.  We need to decide what to do on gfortran <= 5.  Without submodules I don't think it is possible for `adios2_init` and `adios2_open` to provide both serial and MPI signatures while implementing them in separate libraries.

bindings/Fortran/modules/adios2_adios_init_mod_serial.f90
bradking: I can reproduce this PGI diagnostic in a simple example:

```console
$ cat foo.f90
module foo
  interface
    module subroutine bar(x)
      logical, value, intent(in) :: x
    end subroutine
  end interface
end module

$ cat foo_sub.f90
submodule (foo) sub
contains
  module subroutine bar(x)
    logical, value, intent(in) :: x
  end subroutine
end submodule

$ pgfortran --version
pgfortran 19.10-0 linuxpower target on Linuxpower

$ pgfortran -c foo.f90

$ pgfortran -c foo_sub.f90
PGF90-S-1057-Definition argument name _V_x does not match declaration argument name x (foo_sub.f90: 3)
  0 inform,   0 warnings,   1 severes, 0 fatal for bar
```

germasch: Can't you work around this by not using value here? And then use an interface block for the actual `bind(C)` function locally, and have `bar()` just be a wrapper that calls the C function?

[I haven't looked at this closely, so I might be off base -- but it seems that the problem is the `value`, and that's only useful in `bind(C)` functions, right?]

Either way, that doesn't help with the lack of support of submodules in older gfortran, though.

chuckatkins:
Wrt submodules, is this only an issue with gcc?  Can you check with xl and cray?  We could probably bump the Fortran requirements for gcc to 6 since the Fortran users are pretty much always on an HPC environment with newish compilers.  We'd still want to keep the lower version requirement for c and c++ though.

@pnorbert ?

bradking:
XL Fortran has supported submodules [since version 15.1](https://www.ibm.com/support/knowledgecenter/SSAT4T_15.1.0/com.ibm.xlf151.linux.doc/language_ref/submodules.html) (but not [version 14.1](https://www.ibm.com/support/knowledgecenter/SSAT4T_14.1.0/com.ibm.xlf141.linux.doc/language_ref/modules.html)).

Cray Fortran has supported submodules [since at least version 8.1](https://pubs.cray.com/content/S-3901/CCE%208.1/cray-fortran-reference-manual).  A more direct link is available for [version 8.4](https://pubs.cray.com/content/S-3901/8.4/fortran-reference-manual-s-3901-84/submodules).

pnorbert:
Can't you just add all subroutines into one source file and be done with it? I understand that you may want to add all MPI routines into a separate library later but let's do that if we can build this first.

bradking:
@pnorbert I've split out #1994 with just the changes to combine the mpi/nompi modules.  That should let us get many of the changes merged without solving the problems discussed above yet.

The purpose of this PR is to prepare for the library split.  I've already done similar splits in the C and C++ bindings in an earlier PR.  I'll close this until #1994 is finished.

bindings/Fortran/modules/adios2_adios_init_mod_serial.f90
bradking: I've updated the branch to remove the `VALUE` specifier.  I only had it because it was there before, but I don't think we need it.

bradking:
I've updated the branch here with another approach that avoids using submodules but still puts the serial and MPI symbols in separate `.o` files.

bradking:
Yay, all the compilers were happy with the new approach!

PR #1995, opened 2020-02-28
Add a Destroyed state to streams

eisenhauer:
Hoping to kill a heisen-bug in which a close handler tries to access reader state after the stream is destroyed, but before we completely shut down networking.  I'll run this through CI a few times to see if we trigger anything.


chuckatkins:
@eisenhauer this looks good to me.  Are we okay to merge?

eisenhauer:
I think I'm ready...  Merging...

PR #1996, opened 2020-02-28
Add CMake module to detect float representation

KyleFromKitware:


cmake/CheckTypeRepresentation.cmake
```diff
@@ -0,0 +1,251 @@
...
+    bits. Typically found on x86 architectures. Since x86 is always a little
+    endian architecture, there is no big endian variant of this format.
+
+  ``X86_B80_LE``
+  ``X86_64_B80_LE``
+
+    x86 binary 80-bit floating point number. The ``X86`` variant has 16 bits of
+    padding (96 bits total), and the ``X86_64`` variant has 48 bits of padding
+    (128 bits total). Since x86 is always a little endian architecture, there
+    is no big endian variant of this format.
```
germasch: This is just a nomenclature comment: Maybe it's in the standard somewhere, but I think calling an 80 bit representation "quadruple-precision" is likely to cause misunderstandings, because it really isn't, I think the standard term is "extended precision"., whereas quadruple precision is a representation where all 128-bits are actually used.

And then I'm confused what the difference is between `IEEE754_B128_80_LE` and `X86_64_B80_LE`, since they both appear to refer to 80-bit extended precision padded to 128 bits?

Also, extended precision is part of IEEE754 (in my limited, wikipedia-based understanding, anyway). While they may only be supported on x86{,_64}, I think that naming adds to the confusion even more so since it isn't so clear that the `_64` goes with `X86`.

 My suggestion would be to name them consistently with B<actual-width>, and also encode the actual size directly. (But obviously that's a preference thing):

* `IEEE754_B16_[LB]E` (half)
* `IEEE754_B32_[LB]E` (single)
* `IEEE754_B64_[LB]E` (double)
* `IEEE754_B64_[LB]E` (quad)
* `IEEE754_B80_96_LE` (ext, available on x86)
* `IEEE754_B80_128_LE` (extended, available on x86_64)

I don't mind the IBM name, since that I believe that really is a non-standard representation.

CMakeLists.txt
```diff
@@ -153,6 +153,15 @@ install(FILES ${PROJECT_BINARY_DIR}/source/adios2/common/ADIOSConfig.h
   DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}/adios2/common
 )
 
+#------------------------------------------------------------------------------#
```
chuckatkins: Move to `DetectOptions.cmake` as that's where most of this type of logic is.  I've been trying, sometimes successfully, sometimes not, to keep the top level `CMakeLists.txt` as a "main" and offload most of the feature logic elsewhere.

cmake/CheckTypeRepresentation.cmake
```diff
@@ -0,0 +1,251 @@
...
+
+  ``check_float_type_representation()`` understands the following floating
+  point formats:
+
+  ``IEEE754_B16_LE``
+  ``IEEE754_B16_BE``
+
+    IEEE 754 binary half-precision floating point number. Includes a ``LE`` or
+    ``BE`` suffix, depending on whether the number is little endian or big
+    endian, respectively.
```
chuckatkins: The 16-bit checks should also include the `bfloat16` format

cmake/CheckTypeRepresentation.cmake
chuckatkins: I agree on the documentation / naming being off.  I'd suggest a componentized name structure: `BASETYPE_PRECISION_TAG_ENDIANNESS` with:
- `BASETYPE` being `FLOAT`
- `PRECISION` being `HP`, `SP`, `DP`, `EP` for half, single, double, or extended precision
- `TAG` being format specific
- `ENDIANNESS` being `BE` or `LE`

This would give you something like:
- `FLOAT_HP_B16_BE` : half-precision IEEE754 binary16 big endian
- `FLOAT_HP_B16_LE` : half-precision IEEE754 binary16 little endian
- `FLOAT_HP_BF16_BE` : half-precision bfloat16 big endian
- `FLOAT_HP_BF16_LE` : half-precision bfloat16 little endian
- `FLOAT_SP_B32_BE` :  single-precision IEEE754 binary32 big endian
- `FLOAT_SP_B32_LE` :  single-precision IEEE754 binary32 little endian
- `FLOAT_DP_B64_BE` :  double-precision IEEE754 binary64 big endian
- `FLOAT_DP_B64_LE` :  double-precision IEEE754 binary64 little endian
- `FLOAT_EP_8096_LE` :  extended-precision 80-bit in 96-bit storage, little endian
- `FLOAT_EP_80128_LE` :  extended-precision 80-bit in 128-bit storage, little endian
- `FLOAT_EP_B128_LE` :  extended-precision IEEE754 binary128 little endian
- `FLOAT_EP_B128_BE` :  extended-precision IEEE754 binary128 big endian
- `FLOAT_EP_DD_BE` : extended precision double-double big endian
- `FLOAT_EP_DD_LE` : extended precision double-double little endian

KyleFromKitware: > And then I'm confused what the difference is between `IEEE754_B128_80_LE` and `X86_64_B80_LE`, since they both appear to refer to 80-bit extended precision padded to 128 bits?

`X86_64_B80_LE` is the x86's proprietary extended floating-point format, while `IEEE754_B128_80_LE` is the standard by IEEE. The bit layouts are different. In x86, the padding is at the end, while in IEEE, the unused bits are at the beginning. The x86 format also explicitly includes the first 1 bit in the significand, while in IEEE this is implicit, and thus omitted. The x86 format is *not* an IEEE format.

germasch: Thanks for the explanation. I guess the reason I'm confused is because [wikipedia](en.wikipedia.org/wiki/Extended_precision) says:

>IEEE 754 extended precision formats
>The IEEE 754 floating point standard recommends that implementations provide extended precision formats. The standard specifies the minimum requirements for an extended format but does not specify an encoding.[6] The encoding is the implementor's choice.[7]

>The IA32, x86-64, and Itanium processors support an 80-bit "double extended" extended precision format with a 64-bit significand [...]

So I read this to mean that the x86 80-bit floats are ieee754 compliant, but as I said before, I haven't read the actual standard (nor is wikipedia always right).

I guess that still bears the question, is `IEEE754_B128_80_LE` implemented on any actual hardware?

KyleFromKitware: > is `IEEE754_B128_80_LE` implemented on any actual hardware?

I added this format because I saw it as the output for `__float128` on an x86 compiler. It's almost identical to `IEEE754_B128_LE`, but with the caveat that only 80 bits are actually stored. I'm actually tempted to get rid of `IEEE754_B128_80_LE` entirely and instead output `IEEE754_B128_LE` when that format is found.

cmake/CheckTypeRepresentation.cmake
KyleFromKitware: Do we have any platforms that use `bfloat16`?

cmake/CheckTypeRepresentation.cmake
KyleFromKitware: @chuckatkins I would still like to retain the `IEEE754`, `X86`, and `IBM_EXTENDED` tags in some form in the names. The names you suggested are cryptic and don't say much about how the bits are laid out, only the number of bits. A name like `X86_B80_LE` tells me that it's using the x86 80-bit floating point format, while a name like `IEEE754_B128_LE` tells me that it's using the IEEE 754 128-bit binary format.

KyleFromKitware: Also keep in mind that whatever names we pick now are not fixed - we can change them later. They don't have to be backwards compatible.

cmake/CheckTypeRepresentation.cmake
```diff
@@ -0,0 +1,270 @@
...
+  then ``<var>`` is set to an empty string.
+
+  If ``LANGUAGE`` is set, the specified compiler will be used to perform the
+  check. Acceptable values are ``C`` and ``CXX``.
+
+  ``check_float_type_representation()`` understands the following floating
+  point formats:
+
+  ``FLOAT_HP_IEEE754_LE``
+  ``FLOAT_HP_IEEE754_BE``
```
KyleFromKitware, line 34: I think the new naming scheme should still elude to the fact that it is a binary representation and not a decimal representation. How about:

`FLOAT_HP_B_IEEE754_LE`

Then potential decimal formats in the future could be:

`FLOAT_HP_D_IEEE754_LE`

chuckatkins: Rather than a subset of `FLOAT`, if it's a decimal representation then I'd rather just use a `DECIMAL` prefix instead of `FLOAT` since it really represents a different type with different math and different instructions.  I don't see a broad use case for this with ADIOS though so it sort of falls under the "TODO for feature completeness but probably won't" category.

cmake/CheckTypeRepresentation.cmake
chuckatkins: Not ATM but the way this is set up I think is sufficient to easily extend to add `FLOAT_HP_BFLOAT_{BE,LE}` at a later date.

ax3l:
This PR seems to trigger a regression
```
-- Check representation of float
CMake Error at cmake/CheckTypeRepresentation.cmake:144 (message):
  Could not find type information while trying to determine representation of
  float
  cmake/CheckTypeRepresentation.cmake:226 (__check_type_representation_impl)
  cmake/DetectOptions.cmake:357 (check_float_type_representation)
  CMakeLists.txt:139 (include)
```
in https://github.com/Homebrew/homebrew-core/pull/55556

Compiler: AppleClang 10.0.0.10001145 on macOS 10.13

KyleFromKitware:
> This PR seems to trigger a regression

This functionality isn't currently being used, so I'll just disable it for now.

chuckatkins:
Go ahead and do it as a pr targeting release_26 so we can cut a patch release in a week or two with any of the integration issues that come up if needed.

PR #1999, opened 2020-03-02
cmake: Fix sst object lib dependency propagation with CMake < 3.14

bradking:
CMake < 3.14 forget to propagate private dependencies of object
libraries.  Propagate them the same way CMake >= 3.14 do.


chuckatkins:
Does this still necessitate CMake > 3.12?

bradking:
Now building with CMake 3.12 works.  We still require at least that version.

Issue #2001, opened 2020-03-02
Crash with PGI compilers on POWER9


khuck:
**Describe the bug**
When compiling/testing the gray-scott tutorial example with TAU measurement and PGI compilers, the simulation crashes on exit.  Valgrind shows that ADIOS is deleting the MPI Communicator more than once (during the `adios2::helper::Comm::~Comm` destructor), which proves to be fatal when PGI is the compiler and TAU is preloaded in the environment.  Here's the valgrind output:

```console
==146283== Invalid read of size 1
==146283==    at 0x4723DD4: ompi_comm_invalid (communicator.h:334)
==146283==    by 0x4723BDB: PMPI_Comm_free (pcomm_free.c:51)
==146283==    by 0x744B2C7: adios2::helper::CommImplMPI::~CommImplMPI() (adiosCommMPI.cpp:191)
==146283==    by 0x744B35F: adios2::helper::CommImplMPI::~CommImplMPI() (adiosCommMPI.cpp:194)
==146283==    by 0x6B6AD2B: std::default_delete<adios2::helper::CommImpl>::operator()(adios2::helper::CommImpl*) const (unique_ptr.h:67)
==146283==    by 0x6B6B5C3: std::unique_ptr<adios2::helper::CommImpl, std::default_delete<adios2::helper::CommImpl> >::~unique_ptr() (unique_ptr.h:184)
==146283==    by 0x6B6D19F: adios2::helper::Comm::~Comm() (new_allocator.h:114)
==146283==    by 0x695E887: adios2::core::Engine::~Engine() (Engine.cpp:31)
==146283==    by 0x695E96F: adios2::core::Engine::~Engine() (Engine.cpp:31)
==146283==    by 0x6D8EBB3: adios2::core::engine::BP4Writer::~BP4Writer() (BP4Writer.cpp:41)
==146283==    by 0x6A4394B: void __gnu_cxx::new_allocator<adios2::core::engine::BP4Writer>::destroy<adios2::core::engine::BP4Writer>(adios2::core::engine::BP4Writer*) (new_allocator.h:124)
[...]

PR #2002, opened 2020-03-02
helper: Avoid freeing Comm-held MPI_Comm more than once

khuck:
When various destructors are called, the MPI communicator helper
will clean up by deleting the adios communicator.  Well, this change
prevents that from happening twice.  When compiled with PGI and
executed with TAU profiling, this double-free causes memory to be
corrupted.  This PR fixes issue #2001

### DIFF
--- a/source/adios2/helper/adiosCommMPI.cpp
+++ b/source/adios2/helper/adiosCommMPI.cpp
@@ -198,7 +198,9 @@ void CommImplMPI::Free(const std::string &hint)
     if (m_MPIComm != MPI_COMM_NULL && m_MPIComm != MPI_COMM_WORLD &&
         m_MPIComm != MPI_COMM_SELF)
     {
-        CheckMPIReturn(MPI_Comm_free(&m_MPIComm), hint);
+        MPI_Comm mpiComm = m_MPIComm;
+        m_MPIComm = MPI_COMM_NULL; // prevent freeing a second time
+        CheckMPIReturn(MPI_Comm_free(&mpiComm), hint);
     }
 }
 


chuckatkins:
@bradking ?

bradking:
The openmpi [MPI_Comm_free](https://www.open-mpi.org/doc/v4.0/man3/MPI_Comm_free.3.php) documentation says `The handle is set to MPI_COMM_NULL.`.  Maybe that is not true for MPI in general, so doing it manually as proposed here should be fine.

source/adios2/helper/adiosCommMPI.cpp
```diff
@@ -189,6 +189,7 @@ CommImplMPI::~CommImplMPI()
             m_MPIComm != MPI_COMM_SELF)
         {
             MPI_Comm_free(&m_MPIComm);
+            m_MPIComm = MPI_COMM_NULL; // prevent freeing a second time
```
bradking: The destructor shouldn't run more than once and no other methods of `CommImplMPI` should be called once the destructor finishes.  Therefore this is not necessary here.

The change in `CommImplMPI::Free` should be enough because only that method can reach the corresponding code path more than once.

khuck: I'll test it to see if that's the case...

khuck:
@bradking removed extra assignment in destructor

khuck:
> The openmpi [MPI_Comm_free](https://www.open-mpi.org/doc/v4.0/man3/MPI_Comm_free.3.php) documentation says `The handle is set to MPI_COMM_NULL.`. Maybe that is not true for MPI in general, so doing it manually as proposed here should be fine.

Interesting, because without this fix, it crashes with the openmpi 3.1.3 that comes with the PGI compiler...I guess they haven't read their own documentation/specification.

bradking:
@khuck thanks.  I updated your branch to squash in that fixup and revise the logic to ensure that the assignment occurs even if `CheckMPIReturn` throws an exception.  Please fetch, reset your checkout, and try it out.  No push is needed unless you make further updates after that.

PR #2007, opened 2020-03-03
Move async file open functionality to transports

### LINKED ISSUE #1990: TransportMan mods essentially disable background Opens
So, while PR #1984 does resolve TSAN's concerns, it looks like it does so at the cost of nearly completely eliminating the optimization for which the async was introduced.  Specifically, we're waiting for the async thread at the beginning of every TransportMan call.  BP3/4 call OpenFilesAsync inside InitTransports(), and in each the very next subroutine call is InitBPBuffers(), which calls TransportManager::GetTransportsTypes(), which now waits for the async task so that it can put the transport type in the BP buffers that it's initializing.  The goal of letting substantial other work get done before we wait for the Open() is not fulfilled in any meaningful way.

However, this current arrangement seems unnecessary and I'm hoping @pnorbert can confirm.  GetTransportsTypes returns a vector of strings (because what's not a string?) which match the vector of filenames that the OpenFilesAsync call is opening and which get passed to PetProcessGroupIndex.  Currently these file type values are set only after the files are opened, hence the required wait to get them.  But it seems like the file types should be predictable.  I.E. don't we know the type of the file we're trying to create before we create it?  (Types here are  "File_NULL", "File_POSIX", "File_fstream", "File_stdio", and "WAN_zmq".). So presumably this could be refactored so that we actually achieved some asynchrony here..  Norbert?

eisenhauer:
This implements async open() in the fstream, stdio and posix transports only, and only for the write/create open mode.  @pnorbert , I think this is one of those features that is not exercised in CI at all, and perhaps can't trivially be, lacking a super slow filesystem, so please if there's a way to test other than my own efforts let me know.  (I did add a thread sleep into the asynchronous open threads and it still passed tests, so hopefully OK.)  @KyleFromKitware , this looks good to TSAN on my end, but a check from you would be appreciated too.

Closes #1990

### DIFF
--- a/source/adios2/toolkit/transport/file/FileFStream.cpp
+++ b/source/adios2/toolkit/transport/file/FileFStream.cpp
@@ -23,8 +23,31 @@ FileFStream::FileFStream(helper::Comm const &comm, const bool debugMode)
 {
 }
 
-void FileFStream::Open(const std::string &name, const Mode openMode)
+void FileFStream::WaitForOpen()
 {
+    if (m_IsOpening)
+    {
+        if (m_OpenFuture.valid())
+        {
+            m_OpenFuture.get();
+        }
+        m_IsOpening = false;
+        CheckFile(
+            "couldn't open file " + m_Name +
+            ", check permissions or path existence, in call to POSIX open");
+        m_IsOpen = true;
+    }
+}
+
+void FileFStream::Open(const std::string &name, const Mode openMode,
+                       const bool async)
+{
+    auto lf_AsyncOpenWrite = [&](const std::string &name) -> void {
+        ProfilerStart("open");
+        m_FileStream.open(name, std::fstream::out | std::fstream::binary |
+                                    std::fstream::trunc);
+        ProfilerStop("open");
+    };
     m_Name = name;
     CheckName();
     m_OpenMode = openMode;
@@ -32,10 +55,19 @@ void FileFStream::Open(const std::string &name, const Mode openMode)
     switch (m_OpenMode)
     {
     case (Mode::Write):
-        ProfilerStart("open");
-        m_FileStream.open(name, std::fstream::out | std::fstream::binary |
-                                    std::fstream::trunc);
-        ProfilerStop("open");
+        if (async)
+        {
+            m_IsOpening = true;
+            m_OpenFuture =
+                std::async(std::launch::async, lf_AsyncOpenWrite, name);
+        }
+        else
+        {
+            ProfilerStart("open");
+            m_FileStream.open(name, std::fstream::out | std::fstream::binary |
+                                        std::fstream::trunc);
+            ProfilerStop("open");
+        }
         break;
 
     case (Mode::Append):
@@ -59,9 +91,13 @@ void FileFStream::Open(const std::string &name, const Mode openMode)
                   ", in call to stream open");
     }
 
-    CheckFile("couldn't open file " + m_Name +
-              ", check permissions or path existence, in call to fstream open");
-    m_IsOpen = true;
+    if (!m_IsOpening)
+    {
+        CheckFile(
+            "couldn't open file " + m_Name +
+            ", check permissions or path existence, in call to fstream open");
+        m_IsOpen = true;
+    }
 }
 
 void FileFStream::SetBuffer(char *buffer, size_t size)
@@ -81,6 +117,7 @@ void FileFStream::Write(const char *buffer, size_t size, size_t start)
                   ", in call to fstream write");
     };
 
+    WaitForOpen();
     if (start != MaxSizeT)
     {
         m_FileStream.seekp(start);
@@ -117,6 +154,7 @@ void FileFStream::Read(char *buffer, size_t size, size_t start)
                   ", in call to fstream read");
     };
 
+    WaitForOpen();
     if (start != MaxSizeT)
     {
         m_FileStream.seekg(start);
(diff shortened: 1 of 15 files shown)


source/adios2/toolkit/transport/file/FileFStream.cpp
```diff
@@ -23,19 +23,50 @@ FileFStream::FileFStream(helper::Comm const &comm, const bool debugMode)
...
+
+void FileFStream::Open(const std::string &name, const Mode openMode,
+                       const bool async)
+{
+    auto lf_AsyncOpen = [&](const std::string &name) -> void {
+        ProfilerStart("open");
+        m_FileStream.open(name, std::fstream::out | std::fstream::binary |
+                                    std::fstream::trunc);
+        ProfilerStop("open");
+    };
```
chuckatkins: Should this be setup to work with both writing and reading?
```c++
    auto lf_AsyncOpenWrite = [&](const std::string &name) -> void {
        ProfilerStart("open");
        m_FileStream.open(name, std::fstream::out | std::fstream::binary |
                                    std::fstream::trunc);
        ProfilerStop("open");
    };
    auto lf_AsyncOpenRead = [&](const std::string &name) -> void {
        ProfilerStart("open");
        m_FileStream.open(name, std::fstream::in | std::fstream::binary);
        ProfilerStop("open");
    };
```

eisenhauer: It certainly could be.  However, in the past the use case has been for writers, so async open was only implemented for them.  The use case was that some codes would open output files at init time, but then do a timesteps worth of computation before it needed to write them, so if Open() was going to take a while, doing it async pays off by hiding that delay behind computation.  On the reader side, the analysis codes are maybe less likely to have something to do between opening the file and wanting to get data to work on.  However, unlike the prior versions, pushing this down results in an async task per OS file, rather than one per ADIOS file, so there may be some use in doing open that have substreams in parallel...   I'd defer to Norbert to opine on why this wasn't implemented for readers in the past.

pnorbert: It was simply not required because we don't know any case where the reader would pre-open a file. But it is certainly not a bad idea to support it if it is not expensive for ALL reading

eisenhauer: It wouldn't be expensive in any way, but there is another consequence.  It does move knowledge of whether or not an open() succeeded or failed *out* of the ADIOS-level Open() and into the next operation on that file.  I.E. Open with async will always report success, but the next write(), read(), etc. will raise any exception that the Open might have.  For create-mode writers I think that's probably acceptable as a default (and matches the semantics from before the TSAN mods).  For readers, we probably have to be more careful.  If system-level open() is used as an existence test, then doing it async would bollox that up.

eisenhauer: Actually, I was wrong here.  If we tried to delay reader-side system-level open(), we wouldn't get to to the point where we'd not know if a file existed until after the high-level Open(), because the first thing the BP reader engines do after open is to get the file size and read the metadata.  So if the system-level open() was async, we'd really be waiting on it a few lines later and it would all be relatively pointless.  To get truely async opens, we'd have to restructure the readers a fair bit so that we read the metadata later, maybe on demand at first use.  Not impossible, but likely not worth the trouble without a compelling use case.  For the writers though, after open we're building metadata data structures in memory for the rest of the high-level Open() call and actual file writes or accesses don't occur until later high-level operations happen.    Given this, my suggestion would be to run with this version that makes only the writes async.  This restores prior functionality (async system-level writer open()) in a way that makes TSAN happy.  If we develop a use-case for async high-level reader open(), we can tackle that in a different PR (that would necessarily involve significant changes in BPxReader code as well).  @pnorbert and @chuckatkins, if you concur maybe we can resolve this conversation and merge?

chuckatkins: Sounds good to me.  It makes sense to keep this in the transport anyways.  The only change is request is to rename `lf_AsyncOpen` to `lf_AsyncOpenWrite` just so it's clear that this implementation isn't for both modes.

eisenhauer: I'll rename.  But while I was doing further due diligence, I discovered that while this patch restores the async Open() functionality for the BP3 writer, it does not do so for BP4, because BP4Writer does not completely avoid Write() in its InitTransports() sequence (change related to the new index files in BP4).  My temptation would be to still merge this (after naming change) as it fixes BP3, open an issue on the BP4 async open functionality, and maybe open a 2nd issue on how to test this in CI. If it's important enough to implement, we should test. I've been manually testing by temporarily adding a multi-second delay into the AsyncOpen task, printing "low-level Open done" when it completes and adding a "high-level Open done" when the transport Open completes.  It's working if the "high-level Open" completes before the low-level open.  But I don't commit those delays and output lines obviously.  This is functionality that might be easily broken by changes in the BP engines, can we actually test it in CI?

chuckatkins: >  My temptation would be to still merge this (after naming change) as it fixes BP3, open an issue on the BP4 async open functionality

Agreed

> maybe open a 2nd issue on how to test this in CI.
> ...
> This is functionality that might be easily broken by changes in the BP engines, can we actually test it in CI?

I can't really think of a reliable way.  Any tests that are timing based are inherently unstable on CI given the reduced adn low priority environments they run in.

chuckatkins: lgtm

eisenhauer: Messed up those mods.  Trying again shortly.

eisenhauer: This is merged now, but I have to say that I find it ironic that we can't test functionality designed to help in environments where OS functions might be slow because our CI environment might be slow...

PR #2013, opened 2020-03-03
Prevent hang when throwing out of adios constructor in MPI.

### LINKED ISSUE #1938: Flexible Init with/without config file cannot be added to applications
I want to recover from failure of not finding the xml config and just initialize with default. This results in non-zero ranks hanging in broadcast. 
```
    try {
        ad =  new adios2::ADIOS("adios2_config.xml", world, adios2::DebugON);
    } catch (std::ios_base::failure &e) {
        ad = new adios2::ADIOS(world, adios2::DebugON);
    }
```

When the config file is not present, rank 0 throws an exception, which is caught with the catch statement and then a default init is called. However, all other processes are stuck in a broadcast call for receiving the file content from rank 0 in adiosXML.cpp::ParseConfigXML()

Suggestion: in adiosComm.cpp::Comm::BroadcastFile, the exception should be caught (on rank 0) and then the broadcast should happen with some constant string that indicates error. After that all process should throw the same exception.

NAThompson:
Fixes #1938

### DIFF
--- a/source/adios2/core/ADIOS.cpp
+++ b/source/adios2/core/ADIOS.cpp
@@ -17,6 +17,7 @@
 #include "adios2/core/IO.h"
 #include "adios2/helper/adiosCommDummy.h"
 #include "adios2/helper/adiosFunctions.h" //InquireKey, BroadcastFile
+#include <adios2sys/SystemTools.hxx>
 
 // OPERATORS
 
@@ -65,8 +66,7 @@ ADIOS::ADIOS(const std::string configFile, helper::Comm comm,
 {
     if (!configFile.empty())
     {
-        std::ifstream f(configFile.c_str());
-        if (!f.good())
+        if (!adios2sys::SystemTools::FileExists(configFile))
         {
             throw std::logic_error("Config file " + configFile +
                                    " passed to ADIOS does not exist.");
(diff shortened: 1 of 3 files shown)


source/adios2/core/ADIOS.cpp
```diff
@@ -64,6 +65,12 @@ ADIOS::ADIOS(const std::string configFile, helper::Comm comm,
 {
     if (!configFile.empty())
     {
+        std::ifstream f(configFile.c_str());
```
chuckatkins: `adios2sys::SystemTools::FileExists` from `#include <adios2sys/SystemTools.hxx>`

testing/adios2/engine/bp/TestBPNoXMLRecovery.cpp
```diff
@@ -0,0 +1,44 @@
...
+#endif
+
+int main(int argc, char *argv[])
+{
+    int rank = 0;
+#ifdef ADIOS2_HAVE_MPI
+    int size = 1;
+    MPI_Init(&argc, &argv);
+    MPI_Comm_rank(MPI_COMM_WORLD, &rank);
+    MPI_Comm_size(MPI_COMM_WORLD, &size);
```
chuckatkins: size isn't actually being used here and can be removed

PR #2016, opened 2020-03-03
Inline engine fixes

caitlinross:
There's some clean up and stuff to the inline engine, along with these two fixes:
- All variables now have their block info cleared out when starting to write a new step, instead of only the variables that were read on the previous step.
- Reader now reports that it is on the same step as the Writer, since for the inline engine, it doesn't make sense for them to be on different steps.

### DIFF
--- a/source/adios2/core/Engine.tcc
+++ b/source/adios2/core/Engine.tcc
@@ -160,19 +160,13 @@ template <class T>
 typename Variable<T>::Info *Engine::Get(Variable<T> &variable,
                                         const Mode launch)
 {
-    if (m_DebugMode)
-    {
-        // CommonChecks<T>(variable, nullptr, {{Mode::Read}}, "in call to Get");
-    }
-
+    typename Variable<T>::Info *info = nullptr;
     switch (launch)
     {
     case Mode::Deferred:
-        // TODO different? Should use DoGetDeferred?
-        return DoGetBlockSync(variable);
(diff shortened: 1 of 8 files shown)


caitlinross:
@chuckatkins

chuckatkins:
>     * All variables now have their block info cleared out when starting to write a new step, instead of only the variables that were read on the previous step.

Great!

> * Reader now reports that it is on the same step as the Writer, since for the inline engine, it doesn't make sense for them to be on different steps.

I'm not so sure that's the case since everything written on a given step shouldn't be available in the reader until *after* end step.

pnorbert:
For an inline engine I would expect to see:

Writer: step 1
Reader: step 1
Writer: step 2
Reader: step 2
...

Technically, the reader is running after the writer's EndStep, so they are
not at the "same step at the same time".

chuckatkins:
There's a bunch of issues with the inline engine that didn't really become clear until it started getting worked on.  This first pr addresses a squirrely bug with variables showing up in steps when they shouldn't.  There will be some follow up prs to 1) ensure the engine "contract" is enforced, i.e. making sure variables are only available after endstep is complete, 2) prevent beginstep on a reader until endstep is complete on the writer, and similarly prevent beginstep on the writer until all readers have finished their previous endstep 3) add appropriate synchronization primitives to let readers and writers live on separate threads.

PR #2019, opened 2020-03-03
Add MpiHandshake to enable XGC-COUPLER-GENE communication pattern

JasonRuonanWang:


source/adios2/helper/adiosMpiHandshake.h
```diff
@@ -0,0 +1,60 @@
...
+ * adiosMpiHandshake.h
+ */
+
+#ifndef ADIOS2_HELPER_ADIOSMPIHANDSHAKE_H_
+#define ADIOS2_HELPER_ADIOSMPIHANDSHAKE_H_
+
+#include "adios2/common/ADIOSConfig.h"
+#ifndef ADIOS2_HAVE_MPI
+#error "Do not include adiosCommMPI.h without ADIOS2_HAVE_MPI."
+#endif
```
bradking: Please update the file name in the `#error` line to match this header.

JasonRuonanWang: Done

source/adios2/helper/adiosMpiHandshake.cpp
```diff
@@ -0,0 +1,286 @@
...
+    auto start_time = std::chrono::system_clock::now();
+    while (!Check(filename))
+    {
+        auto now_time = std::chrono::system_clock::now();
+        auto duration = std::chrono::duration_cast<std::chrono::seconds>(
+            now_time - start_time);
+        if (duration.count() > timeoutSeconds)
+        {
+            throw(std::runtime_error("Mpi handshake timeout"));
+        }
```
chuckatkins: There should be some sort of short sleep here to ensure the process doesn't go into a blocking spin wait.  Without the sleep then  you run the risk in an oversubscription scenario of one process completely consuming the CPU code and not allowing others to proceed.  Something like `std::this_thread::sleep_for(100ms);` should be sufficient.

JasonRuonanWang: Done

PR #2022, opened 2020-03-05
bindings: Convert Fortran MPI- and non-MPI implementations to submodules

bradking:
Convert the separate modules implementing Fortran `adios2_init` and `adios2_open` generic procedure signatures for MPI and non-MPI variants into submodules.  This allows the primary modules to be built first and the implementation submodules built later in separate libraries.

In order to support compilers that do not support Fortran submodules, use preprocessor conditions to optionally build without submodules as before.

### DIFF
--- a/bindings/Fortran/CMakeLists.txt
+++ b/bindings/Fortran/CMakeLists.txt
@@ -22,12 +22,12 @@ add_library(adios2_f
   modules/adios2_parameters_mod.f90
   modules/adios2_adios_mod.f90
   modules/adios2_adios_init_mod.F90
-  modules/adios2_adios_init_mod_serial.f90
+  modules/adios2_adios_init_mod_serial.F90
   modules/adios2_attribute_mod.f90
   modules/adios2_attribute_data_mod.f90
   modules/adios2_io_mod.f90
   modules/adios2_io_open_mod.F90
-  modules/adios2_io_open_mod_serial.f90
+  modules/adios2_io_open_mod_serial.F90
   modules/adios2_io_define_variable_mod.f90
   modules/adios2_io_define_attribute_mod.f90
   modules/adios2_engine_mod.f90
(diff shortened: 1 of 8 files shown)


bradking:
For reference, this follows up after #1994 and #1987.  It turns out that in order to put the MPI and non-MPI implementations in separate libraries we do need to use submodules.  Otherwise the (non-submodule) module dependencies are circular between the two libraries.  This PR switches to submodules where supported but otherwise falls back to the previous approach.

bradking:
In the CI results we can see that the `try_compile` for `ADIOS2_HAVE_FORTRAN_SUBMODULES` passed on most of the builds.  It failed on gfortran 4.8 as expected, and the build still worked by falling back to the plain module approach.

bindings/Fortran/CMakeLists.txt
```diff
@@ -8,6 +8,32 @@ include(FortranCInterface)
...
+    integer, intent(in) :: x
+  end subroutine
+  module subroutine bar_real(x)
+    real, intent(in) :: x
+  end subroutine
+end submodule
+program main
+end program
+]] ADIOS2_HAVE_FORTRAN_SUBMODULES SRC_EXT F90)
+
```
chuckatkins: I'd like this to be factored out into `cmake/ADIOSFunctions.cmake`.  That's essentially where we keep our "library" of custom cmake functions used for the build.

bindings/Fortran/CMakeLists.txt
```diff
@@ -49,15 +52,18 @@ target_include_directories(adios2_f
 
 target_link_libraries(adios2_f PRIVATE adios2)
 
+if(ADIOS2_HAVE_FORTRAN_SUBMODULES)
+  target_compile_definitions(adios2_f PRIVATE "$<$<COMPILE_LANGUAGE:Fortran>:ADIOS2_HAVE_FORTRAN_SUBMODULES>")
+endif()
 
 if(ADIOS2_HAVE_MPI)
+  target_compile_definitions(adios2_f PRIVATE "$<$<COMPILE_LANGUAGE:Fortran>:ADIOS2_HAVE_MPI_F>")
```
chuckatkins: Is there a scenario where these files would ever be built with `ADIOS2_HAVE_MPI_F` not defined?

bradking: This is needed to tell the *other* sources that MPI parts will be present.

bradking: These two non-MPI sources need to know whether the MPI sources will be included:

* `bindings/Fortran/modules/adios2_adios_init_mod.F90`
* `bindings/Fortran/modules/adios2_io_open_mod.F90`

PR #2025, opened 2020-03-08
Fix unaligned memory access in SSC engine

keichi:
@JasonRuonanWang I was getting a bus error due to unaligned memory access on the JAXA system (it has strict alignment requirements). This PR removes the unaligned accesses.


JasonRuonanWang:
Thanks. This will hopefully fix some errors I saw from the common staging tests too, which I have just promised to bring back to SSC.

PR #2032, opened 2020-03-10
Fix bugs for Fortran column major conversion in SSC

JasonRuonanWang:

PR #2037, opened 2020-03-12
Add a little bit more of a description of tests in staging-common

eisenhauer:


eisenhauer:
The fortran install test failure of #2032 has returned...  @chuckatkins , @bradking ?  Thoughts?

bradking:
Strange.  I don't think anything in #2032 was done to resolve the `Install.Make.Fortran` failure it saw.  Maybe there is a race or other spurious condition causing it.  Unfortunately the only error message visible in the test output is `collect2: error: ld returned 1 exit status`.  It has no indication of why the link failed.  I'm not able to reproduce it locally.

eisenhauer:
I did try a force-push to see if it might be a race, but it didn't go away.   Also seeing it on PR #2041 .

bradking:
In other PRs I can see that `el7-gnu8-openmpi-ohpc-static` is taking significantly longer than the other builds.  `Install.Make.Fortran` needs to link a large binary due to using static libraries.  Since the only error we see is `ld returned 1 exit status` I wonder if `ld` is being terminated early by something external.

eisenhauer:
I have a vague recollection of seeing some disk space errors somewhere, which I now can’t remember...  This is late in the build, possible that’s the problem?

bradking:
I've opened #2044 to host the CI discussion because it is unrelated to this PR.

chuckatkins:
Once #2046 is merged you can rebase the failure away.

PR #2041, opened 2020-03-16
Revamp locking macros on SST reader side

eisenhauer:
No changes in locking here, just modifying the macros so that the Locked boolean is stream-local and we're a bit more uniform.  Doing piece-wise movement towards making TSAN happy to make it easier to evaluate each change.


bradking:
This is hitting #2044 in CI.

chuckatkins:
Once #2046 is merged you can rebase the static test failure away

Issue #2053, opened 2020-03-19
SubStreams parameter causes SST and other engines to work badly


pnorbert:
**Describe the bug**
The BP engines have an engine parameter SubStreams. If left in the settings accidentally when using SST or other engines using BP serialization, they will not work properly on the reading side. It will run but the data will be corrupt on the reading side. 

**To Reproduce**
We see this behavior with nwchem sorter or gray-scott. 

**Additional context**
The problem lies in the core that allows parameters to be set that does not work for an engine but is (silently) used by subcomponents (BP serializer in this case).

eisenhauer:
I was just looking at BPBase where this stuff is pulled out of the parameter list passed to Init.  So I think the question is how do we handle this?  Should we've got knowledge isolated in BP (what parameters it knows) but where there is compatibility issues with things elsewhere (e.g. SST).  SST sees the params first, and it could theoretically sanitize it, removing known incompatible parameter settings.  Or it could reach into BP*Serializer afterwards to set things to something safe.  But we could be asking for trouble when someone adds new BP parameters and doesn't know to possibly add it to the parameter sanitizers in other engines.  Maybe have something like a "RestoreSafeDefaults" method in BPBase that takes an engine name and unsets BP parameters as appropriate?  At least then we've got the BP parameter operations all in one place...  Other thoughts?

eisenhauer:
Closed by #2054

PR #2054, opened 2020-03-19
For specific engines, only utilize certain BP parameters

eisenhauer:
Add an EngineType parameter to BPBase::Init().  If the engine type is specified, use it to load only an "approved" subset of the specified BP parameters, leaving others at their defaults

This is one possible way to address #2053.  It doesn't give warnings about ignored parameters (generally we don't), or anything else, but it is a way to help make sure that engines that use BP don't get surprised by unsupported or new parameters that change the way BP behaves.


chuckatkins:
Looks fine to me.

pnorbert:
I approve this but the first task I want to give to Dmitry is to work on the parameter handling in a separate class/structure so that a central place has all the parameters and a utility can even print the supported options for each engine.

eisenhauer:
I like that plan for Dmitry...

PR #2055, opened 2020-03-20
Debug mode changes

eisenhauer:
Simple tweaks to a few files using DebugMode.  The idea here is that if it's worth raising an exception in DebugMode, it's worth raising it unconditionally.  None of the changes here affect anything where the actual checks inside an "if(DebugMode)" conditional are more expensive than a compare or two.  It does remove some checks which had the effect of making semantics different in DebugMode than in normal mode.  (Specifically, normal mode would allow multiple variables with the same name to be defined without error, but with unclear consequences.  Now multiple definition is disallowed always.)

(This is just for a few files in core for now, but other uses could benefit from examination.)

### DIFF
--- a/source/adios2/core/IO.tcc
+++ b/source/adios2/core/IO.tcc
@@ -33,7 +33,7 @@ Variable<T> &IO::DefineVariable(const std::string &name, const Dims &shape,
                                 const bool constantDims)
 {
     TAU_SCOPED_TIMER("IO::DefineVariable");
-    if (m_DebugMode)
+
     {
         auto itVariable = m_Variables.find(name);
         if (!IsEnd(itVariable, m_Variables))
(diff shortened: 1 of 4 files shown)


source/adios2/core/IO.tcc
```diff
@@ -103,15 +103,12 @@ Attribute<T> &IO::DefineAttribute(const std::string &name, const T &value,
                                   const std::string separator)
 {
     TAU_SCOPED_TIMER("IO::DefineAttribute");
-    if (m_DebugMode)
+    if (!variableName.empty() && InquireVariableType(variableName).empty())
```
pnorbert: Well, this is not free, but it's good to raise an exception anyway

pnorbert: I mean, this is just some cost when defining attributes which might happen thousands of times but it will not affect IO performance. So I agree with removing the test here.

eisenhauer: Agreed.

PR #2063, opened 2020-03-24
Fix warnings visible on CI builds

bradking:
Perform minor cleanup and refactoring to fix warnings that have appeared during CI (but not blocked merging) for a while.


bradking:
The goal of this PR is to clear up warnings during CI, so please do not merge it if any remain.

chuckatkins:
With this in place the I believe we can try to add back in the "warnings as errors" flags in the appropriate `scripts/ci/cmake/` ctests scripts that it's been disabled in.

bradking:
> try to add back in the "warnings as errors" flags 

I added a commit to do that.

scripts/ci/cmake/ci-suse-pgi.cmake
```diff
@@ -8,9 +8,9 @@ env_module(load pgi)
 set(ENV{CC}  pgcc)
 set(ENV{CXX} pgc++)
 set(ENV{FC}  pgfortran)
-#set(ENV{CFLAGS} -Werror)
+set(ENV{CFLAGS} -Werror)
 set(ENV{CXXFLAGS} --brief_diagnostics)
```
bradking: @chuckatkins the PGI C compiler does not understand `-Werror`.  Do you know the correct options to use here for C and C++?

chuckatkins: AFAIK it doesn't have such an option.

bradking:
Yay, no warnings and `/WX` is now enforced on MSVC!

PR #2064, opened 2020-03-24
Added Engine::BlocksInfo call to Python bindings

### LINKED ISSUE #2023: Add Engine::BlocksInfo to the python bindings
Engine::BlocksInfo() seems to be missing from the python bindings. This is essential for extracting metadata information to build skel I/O models. 

This seems like it is a minor addition, but requires some knowledge of the pybind tool.

isosc:
resolves #2023

### DIFF
--- a/bindings/Python/py11Engine.cpp
+++ b/bindings/Python/py11Engine.cpp
@@ -281,39 +281,40 @@ void Engine::LockReaderSelections() const
     m_Engine->LockReaderSelections();
 }
 
-std::vector<std::map<std::string,std::string>> Engine::BlocksInfo(std::string &var_name, const size_t step) const
+std::vector<std::map<std::string, std::string>>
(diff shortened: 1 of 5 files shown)


chuckatkins:
@bradking based on how this is failing the static build: `/__w/ADIOS2/ADIOS2/helper/TestHelperStrings.cpp:175:1: fatal error: error writing to /tmp/ccbfUwF7.s: No space left on device`, it looks like this is the likely cause of the previous linker failures we were seeing.  I'll see if I can address it by moving the tmpdir to the local disk instead of /tmp.

@isosc I'll let you know when the fix is merged and you can rebase accordingly.

isosc:
Thanks @chuckatkins

chuckatkins:
Rebased on master to bypass static failure

PR #2069, opened 2020-03-25
SST Reader-side TSAN mods

eisenhauer:
Currently these mods ignore the attributes issue (updated by both the main and network thread due to discard/install skipped model).   They also ignore writer-side TSAN issues.  But, run clean for me with the following TSAN suppression file:
  race:*RemoveAllAttributes*
  race:*EvpathReadRequestHandler*
  race:*EvpathDestroyWriter*
  race:*cp_writer.c*
  race:libadios2_evpath.so
Tracking down and fixing those suppressions will be the subject of future PRs.

Change-wise, this adds some comments on routines WRT what thread will be calling them and what their locking needs are.  Generally critical sections are broadened to include state shared by the threads.  Locks are preferentially acquired at entry to interface routines, rather than internally.

### DIFF
--- a/source/adios2/toolkit/sst/cp/cp_reader.c
+++ b/source/adios2/toolkit/sst/cp/cp_reader.c
@@ -21,27 +21,34 @@
 #include "cp_internal.h"
 
 #define gettid() pthread_self()
+#define MUTEX_DEBUG
 #ifdef MUTEX_DEBUG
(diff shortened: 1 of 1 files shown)


eisenhauer:
Whoops @bradking , I meant to tag @KyleFromKitware ...

Issue #2000, opened 2020-03-02
Starting with ADIOS2


aruhela:
What is the recommended way to start with ADIOS2? 

Is there available any Video tutorial to get an overview and basic usage of it ?

pnorbert:
Hi, we don't have a video tutorial but here is the PDF of our latest tutorial. 
[https://users.nccs.gov/~pnorbert/adios-tutorial-2.5.0.pdf](https://users.nccs.gov/~pnorbert/adios-tutorial-2.5.0.pdf)

aruhela:
Thanks for sharing the tutorial.

ax3l:
Maybe we can add the link to the latest tutorial in the manual. That's the first entry point for new users.

caitlinross:
Closing since we now have [self-directed tutorial available on read the docs](https://adios2.readthedocs.io/en/latest/tutorials/overview.html)

PR #2080, opened 2020-03-30
Korteweg de-Vries equation example

NAThompson:
This should help with #2000.

A couple things:

- I should probably add a "step attribute" for the time.
- I will add graphing ability once the Python binding problems are hammered out.

### DIFF
--- a/examples/KortewegdeVries/CMakeLists.txt
+++ b/examples/KortewegdeVries/CMakeLists.txt
@@ -3,16 +3,6 @@
 # accompanying file Copyright.txt for details.
 #------------------------------------------------------------------------------#
 
-add_subdirectory(basics)
-add_subdirectory(hello)
-add_subdirectory(query)
-add_subdirectory(useCases)
-add_subdirectory(KortewegdeVries)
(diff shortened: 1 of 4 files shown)


pnorbert:
Shouldn't we have this rather in [https://github.com/ornladios/ADIOS2-Examples](https://github.com/ornladios/ADIOS2-Examples)?

NAThompson:
@pnorbert : That's fine; either way. But the fact that I had no clue that the ADIOS2-examples repo existed, along with the existence of an `examples` directory here, does not indicate that many other people will find it. It also appears less well-maintained, so I fear it will be susceptible to bitrot.

williamfgc:
@NAThompson I can help you migrating to the ADIOS2-Examples repo (in my spare time). Reality is, ADIOS2-Examples is easy to maintain as long as ADIOS2 doesn't break its API. As for unit tests, they should be under testing, there is nothing run/verified under examples so it got pretty outdated over time.

NAThompson:
@williamfgc : Ok, cool. I'll open a PR in the examples project.

PR #2084, opened 2020-03-31
Fixing inline engine contracts

caitlinross:


docs/user_guide/source/engines/inline.rst
```diff
@@ -6,21 +6,50 @@ The Inline engine provides in-process communication between the writer and reade
...
+For successful operation, the writer will perform a step, then the reader will perform a step in the same process. Data is decomposed between processes, and the writer can write its portion of the data like other ADIOS engines. When the reader starts its step, the only data it has available is that written by the writer in its process. To select this data in ADIOS, use a block selection. The reader then can retrieve whatever data was written by the writer.
 
-Please refer to the examples or tests for the typical access pattern. Note, however, that the ``Get()`` method:
+.. note::
+ The inline engine does not support Sync mode for writing. In addition, since the inline engine does not do any data copy, the writer should avoid changing the data contents before the reader has read the data.
```
NAThompson, line 24: Can `adios2::mode != adios2::mode::Sync` be enforced in code? (Sorry if that's already enforced, but I didn't see it if it is.)

docs/user_guide/source/engines/inline.rst
```diff
@@ -6,21 +6,50 @@ The Inline engine provides in-process communication between the writer and reade
...
+     inlineReader.Get(var, info);
+ }
+ inlineReader.EndStep();
+
+ // do application analysis -
+ // use info.Data() to get the pointer for each element in blocksInfo
+
+ // After any desired analysis is finished, writer can now reuse data pointer
+
+Note, however, that the ``Get()`` method:
```
NAThompson, line 52: When building these docs I get:

```
/Users/4nt/canonical_adios/docs/user_guide/source/engines/dataspaces.rst:55: WARNING: Unexpected indentation.
source/engines/inline.rst:28: WARNING: Error in "code-block" directive:
maximum 1 argument(s) allowed, 6 supplied.
```

NAThompson: I fixed it with this diff

```
diff --git a/docs/user_guide/source/engines/inline.rst b/docs/user_guide/source/engines/inline.rst
index e4761151..4b86a93d 100644
--- a/docs/user_guide/source/engines/inline.rst
+++ b/docs/user_guide/source/engines/inline.rst
@@ -26,34 +26,35 @@ For successful operation, the writer will perform a step, then the reader will p
 Typical access pattern:
 
 .. code-block:: c++
- // ... Application data generation
 
- inlineWriter.BeginStep();
- inlineWriter.Put(var, data); // always use deferred mode
- inlineWriter.EndStep();
- // Unlike other engines, data should not be reused at this point,
- // though ADIOS cannot enforce this.
- // Must wait until reader is finished using the data.
+    // ... Application data generation
 
- inlineReader.BeginStep();
- auto blocksInfo = inlineReader.BlocksInfo(var, step);
- for (auto& info : blocksInfo)
- {
-     var.SetBlockSelection(info.BlockID);
-     inlineReader.Get(var, info);
- }
- inlineReader.EndStep();
+    inlineWriter.BeginStep();
+    inlineWriter.Put(var, data); // always use deferred mode
+    inlineWriter.EndStep();
+    // Unlike other engines, data should not be reused at this point,
+    // though ADIOS cannot enforce this.
+    // Must wait until reader is finished using the data.
 
- // do application analysis -
- // use info.Data() to get the pointer for each element in blocksInfo
+    inlineReader.BeginStep();
+    auto blocksInfo = inlineReader.BlocksInfo(var, step);
+    for (auto& info : blocksInfo)
+    {
+        var.SetBlockSelection(info.BlockID);
+        inlineReader.Get(var, info);
+    }
+    inlineReader.EndStep();
 
- // After any desired analysis is finished, writer can now reuse data pointer
+    // do application analysis -
+    // use info.Data() to get the pointer for each element in blocksInfo
+
+    // After any desired analysis is finished, writer can now reuse data pointer
 
 Note, however, that the ``Get()`` method:
 
 .. code-block:: c++
 
- void Engine::Get<T>(                                       \
+    void Engine::Get<T>(                                       \
         Variable<T>, typename Variable<T>::Info & info, const Mode);
 
 is preliminary, and may change in the future.
```

testing/adios2/engine/inline/TestInlineWriteRead.cpp
```diff
@@ -610,6 +616,248 @@ TEST_F(InlineWriteRead, InlineWriteRead2D2x4)
...
+
+            // Make a 1D selection to describe the local dimensions of the
+            // variable we write and its offsets in the global spaces
+            adios2::Box<adios2::Dims> sel({mpiRank * Nx}, {Nx});
+
+            EXPECT_THROW(wvar_iString.SetSelection(sel), std::invalid_argument);
+            wvar_i32.SetSelection(sel);
+
+            inlineWriter.Put(wvar_iString, testData.S1);
+            EXPECT_THROW(inlineWriter.Put(wvar_i32, testData.I32.data(),
```
NAThompson, line 817: And there is the unit test for the throw; sry!

docs/user_guide/source/engines/inline.rst
NAThompson: Resolved by unit test I didn't notice..

NAThompson:
@caitlinross : Great work; do you feel it's ready to merge?

(Note: I didn't run it under tsan/msan for review; I figured CI is now getting that. . .)

docs/user_guide/source/engines/inline.rst
caitlinross: So I think it's enforced as much as it can be. In the case of variables that aren't single values, an error is thrown if a PutSync is performed. In the case of single values, they're by default treated as sync, even if the user set the mode to deferred, right? So it seems like if I throw an error on those, then it will fail even when the user has actually selected deferred for single value variables.

caitlinross: Oh somehow your last comment about the unit test didn't show up until after I added my comment

docs/user_guide/source/engines/inline.rst
caitlinross: Fixed. I think - I just tried to build the docs locally, but I think I'm having an issue with some of the python packages.

caitlinross:
@chuckatkins Will you check over this?

chuckatkins:
This looks good.  Rebasing on master will get rid of the static failure

caitlinross:
@chuckatkins Done and looks like all the tests are good now.

caitlinross:
Actually don't merge yet @chuckatkins. I'm updating ADIS for this and I think I noticed a change I should make.

caitlinross:
Okay now it should be good as long as I didn't break any tests. I made one change in the InlineReader constructor. Previously it was trying to access the InlineWriter, but only used it to output some info when verbosity is set to 5. I don't think there's any reason to enforce that the writer has to be opened before the reader (they just both have to be opened before either calls BeginStep/etc. And this is noted in the documentation now). And allowing them to be opened in either order makes it simpler to set things up in ADIS.

NAThompson:
Looks like everything is ready to be merged; @caitlinross, is this correct?

caitlinross:
Yeah, I just don't have authorization to merge it.

chuckatkins:
Lgtm

eisenhauer:
done.

chuckatkins:
@caitlinross Also nice job on adding the error conditions in the test cases

PR #2085, opened 2020-03-31
Refactor SSC in preparation for supporting local arrays and local values

JasonRuonanWang:


JasonRuonanWang:
> Wouldn't work here better using m_ShapeID?
> if (variable->m_ShapeID == ShapeID::LocalValue)
> ...

Oh okay. I had an impression that there was such a thing, but didn't find it in the Variable class. I will have a look.

PR #2086, opened 2020-04-01
Added support to handle HDF5 API interface change

guj:
The newly released HDF5 1.12 has a slight API change. Detecting the HDF5 version and use the right call signature accordingly.


source/adios2/CMakeLists.txt
```diff
@@ -269,6 +269,10 @@ if(ADIOS2_HAVE_HDF5)
 
   foreach(lib IN LISTS adios2_hdf5_objs)
     target_link_libraries(${lib} PRIVATE ${HDF5_C_LIBRARIES})
+    message(STATUS HDF5_VERSION)
+    if(HDF5_VERSION VERSION_LESS_EQUAL 1.11.0)
+      target_compile_definitions(${lib} PRIVATE H5_USE_110_API)
+    endif()
```
germasch, line 275: You should probably remove the status message, assuming it was for debugging.

But this is also not the correct way of handling this -- that macro is something that hdf5 1.12 understands -- if it is defined, it'll make the API compatible to 1.10. That is,
```
 if(HDF5_VERSION VERSION_GREATER_EQUAL 1.12.0)
      target_compile_definitions(${lib} PRIVATE H5_USE_110_API)
  endif()
```

by itself should fix the issue.

source/adios2/toolkit/interop/hdf5/HDF5Common.cpp
```diff
@@ -1260,7 +1260,11 @@ void HDF5Common::ReadAttrToIO(core::IO &io)
     hsize_t numAttrs;
     // herr_t ret = H5Gget_num_objs(m_FileId, &numObj);
     H5O_info_t oinfo;
+#ifdef H5_USE_110_API
     herr_t ret = H5Oget_info(m_FileId, &oinfo);
+#else
+    herr_t ret = H5Oget_info(m_FileId, &oinfo, H5O_INFO_ALL);
+#endif
```
germasch, line 1267: Alternatively, you can adapt the code to support the new API, as you're doing here, but then you should not use this macro, and fact you don't need a new macro at all, you can just use
```cxx
#if H5_VERSION_GE(1,12,0)
  herr_t ret = H5Oget_info(m_FileId, &oinfo, H5O_INFO_ALL);
#else
  herr_t ret = H5Oget_info(m_FileId, &oinfo);
#endif
```

guj:
@NAThompson Please checkout again and see if it works for you. Thanks.

source/adios2/CMakeLists.txt
guj: if (version >= 1.12)
   define(H5_USE_110_API)? 

I think you may not mean it.

source/adios2/toolkit/interop/hdf5/HDF5Common.cpp
guj: yes this would also work.

source/adios2/CMakeLists.txt
guj: I removed the status message.  Good catch, thanks!

I tried with both hdf5. 1.10 and 1.12. My original ifdef works.

source/adios2/toolkit/interop/hdf5/HDF5Common.cpp
NAThompson: I think @germasch has the right fix here. I think HDF5's suggestion for vandalizing the CMakeLists file to do backward compatibility is pretty weird, esp. since a source code fix is required anyway.

source/adios2/CMakeLists.txt
germasch: > if (version >= 1.12)
> define(H5_USE_110_API)?
> 
> I think you may not mean it.

Yes, I did mean it. H5_USE_110_API is a macro support by hdf5 1.12 that activates compatibility with 1.10, so then you don't have to change your code.

@chuckatkins ?

source/adios2/toolkit/interop/hdf5/HDF5Common.cpp
germasch: My point is that one can choose either way: define the compatibility macro in CMakeLists.txt, in that case one doesn't need to change the source code at all.

Or, don't define the compatibility macro, but change the source code as shown above.  That means making the source code use the 1.12 API. The previous way is telling hdf5 1.12 to still provide the 1.10-compatibile API, so the source codes doesn't need to change.

germasch:
Just to be clear, I don't mean to say that this fix doesn't work -- it does. But it uses a macro that has been given a certain meaning by hdf5 very differently from how it was intended.

source/adios2/toolkit/interop/hdf5/HDF5Common.cpp
chuckatkins: > HDF5's suggestion for vandalizing the CMakeLists file to do backward compatibility is pretty weird

FWIW the `CMakelist.txt` change was my suggestion, not theirs.

guj: @chuckatkins 
Oh that clarifies things.  I will change it back to the usual H5_VERSION comparison when I merge the h5vol changes.

chuckatkins:
@germasch is right, this is definitely not the right way to use the preprocessor definition.  Rather that informing the user, it actually sets up macros in `hdf5.h` to provide backwards compatibility interfaces.  If `H5_USE_110_API` is defined for `>= 1.11` then the user code doesn't have to change.  It could be done either in the `CMakeLists.txt` as `-DH5_USE_110_API` or in the source as `#define H5_USE_110_API` before `hdf5.h` is included.  However, using the version comparison macro instead allows the code to adapt when the newer API is available.  The change should go in now though rather than waiting until the vol is implemented since it's really orthogonal to that.

source/adios2/toolkit/interop/hdf5/HDF5Common.cpp
NAThompson: @chuckatkins : You were forced into that path since you're providing a library which depends on HDF5. I think the HDF5 guys had envisioned that the workaround would be used by end users only.

guj:
@chuckatkins 
OK I did a pull request to simply using the good old H5_VERSION instead of flags from CMakeList.txt.
However / Azure Pipelines (macos macos1015_xcode111_make) fails at ffs for a generated file cod_node.c
Error message see below.  I have reran it a few times.  Same thing.  Any idea? Any thing changed since last night? 
 
---------------------
[ 17%] Building C object thirdparty/ffs/ffs/CMakeFiles/ffs.dir/cod_node.c.o
cod/cod.y:2274:51: error: use of undeclared identifier 'cod_yytext'; did you mean 'cod_yylex'?
    yycontext->error_func(yycontext->client_data, yytext);
                                                  ^~~~~~
                                                  cod_yylex
cod/cod.y:138:16: note: expanded from macro 'yytext'
#define yytext cod_yytext

guj:
All passed. Thanks @chuckatkins , although not sure how you made it work.

PR #2087, opened 2020-04-01
removed comment

guj:


NAThompson:
@guj: Since you did a nice source code fix, is the CMakeLists.txt change necessary?

Wouldn't simply checking the `H5_VERS_INFO` string be sufficient?

PR #2094, opened 2020-04-02
WIP: Burstbuffer

pnorbert:
Just checking which compilers are happy with my code


chuckatkins:
@pnorbert you can get rid of the `el7-gnu8-openmpi-ohpc-static` failure by rebasing on current master, I fixed it in #2092.  w.r.t. the windows failures, `unistd.h` is only available on `*nix`.  Can you use the `cstdio` functions instead?

chuckatkins:
> w.r.t. the windows failures, `unistd.h` is only available on `*nix`. Can you use the `cstdio` functions instead?

You can also use proper C++ fstreams and put them into unbuffered mode: https://gcc.gnu.org/onlinedocs/libstdc++/manual/streambufs.html#io.streambuf.buffering

> there are times when writing to a file should be like writing to standard error; no buffering should be done because the data needs to appear quickly (a prime example is a log file for security-related information). The way to do this is just to turn off the buffering before any I/O operations at all have been done (note that opening counts as an I/O operation):
> 
> ```cxx
> std::ofstream    os;
> std::ifstream    is;
> 
> os.rdbuf()->pubsetbuf(0,0);
> is.rdbuf()->pubsetbuf(0,0);
> ```

pnorbert:
@Chuck Atkins <chuck.atkins@kitware.com>
Thanks for the links. Is there any study showing that unistd and unbuffered
streambuf/stdio behaves the same way (on Unix)? I don't find any.

I don't think burst buffer is a thing to support on Windows but I
definitely don't want to sacrifice performance on the big machines because
of them.

germasch:
ADIOS2 already has an abstraction for POSIX/stdio/fstream I/O by way of its file transports. Maybe using that is an option to consider?

chuckatkins:
@pnorbert See #2104.  It adds a `<parameter key="Buffered" value="false"/>` parameter to disable buffered I/O.  With that it should be easy enough to compare performance numbers.

pnorbert:
Removing this PR from now until I fix the IO routines

PR #2104, opened 2020-04-05
Enable "buffered" parameter for file transports

chuckatkins:
This allows the `fstream` and `stdio` transports to be configured for unbufferd I/O by setting the buffer to `null`.  See  https://gcc.gnu.org/onlinedocs/libstdc++/manual/streambufs.html#io.streambuf.buffering

The defaults remain unchanged but this now allows an extra transport parameter to disable buffered i/o.


williamfgc:
@chuckatkins Looks great! Nice added portable functionality!

chuckatkins:
Just adding some tests, then I'll update...

PR #2105, opened 2020-04-07
added Value in SSC BlockInfo

JasonRuonanWang:


JasonRuonanWang:
@pnorbert 

I have added Value in SSC BlocksInfo in this PR, please have a try with your application and see if it works. 

There is a problem with this implementation. For the initial step, if you call BlocksInfo before getting any actual variable data, then it wouldn't work. This is because all variables including single values in SSC are put into data rather than metadata, and at this particular moment, before you actually read any data in the first step, data is not transferred to the reader side yet. 

I would have implemented it so that the BlocksInfo function itself can call GetSync() to actually read data, which would for sure solve the problem, but currently BlocksInfo() is defined as a const function, which does not allow me to change any class members. This makes it impossible to actually read anything in BlocksInfo. 

So, either we ask users to always call BlocksInfo() after at least one GetSync or GetDeferred, or we make BlocksInfo a non-const function.

PR #2106, opened 2020-04-07
Fix python install dir when only the interpreter is found

chuckatkins:


chuckatkins:
@ax3l this should fix your ci error in openpmd

ax3l:
Thank you, test passes again :)

PR #2108, opened 2020-04-07
Remove examples as examples are moving to the ADIOS2-Examples repo.

NAThompson:


williamfgc:
Not sure if we need to remove these. They are built optionally, never installed, and they are more like "hello world" examples that help devs and users browsing on github. `ADIOS2-Examples` is for tutorials. They serve a different purpose.

chuckatkins:
Keeping these examples servers a variety of purposes, one of which is for CI (the heat transfer example provides complex use-case testing that the unit tests can't cover).  I do think we could probably reduce many of the ones that are there as they are largely redundant keeping the examples in the repo as a general matter is very useful.  The separate examples repo is better for the more complex mini-app type examples that we would use in the tutorials.

PR #2109, opened 2020-04-07
Added a write/read BP test for a 10D array. Just because DCA++ has a …

pnorbert:
…test like that.


williamfgc:
I'm glad they are not Fortran

PR #2115, opened 2020-04-08
Kill DebugMode spec in constructors

eisenhauer:


williamfgc:
I can certainly use a debug mode flag for avoiding checks on some of the large metadata sets I'm working with. Also, breaking the API would make it really hard to the ecosystem maintenance (distros). Just wanted to know if @pnorbert @sklasky @chuckatkins decided this is the path forward. Thanks!

eisenhauer:
There are no API changes here, and yes, there is consensus.

pnorbert:
@williamfgc We could not see any measurable difference with/without the checks, even up to 100k variables. The actual operations overweight the checks in every case. So we decided that the flag is unnecessary and Greg took on eliminating it from more than 300 function signatures. 

This cleanup does not address that we may add global level bool flags to ADIOS that can be used everywhere in the code consistently. Like Verbose. But it seems unlikely that we need a separate debug mode for sanity checks.

The public API does not change with this PR, the debug flag will stay in at least deprecated public functions.

williamfgc:
@pnorbert thanks, so the spec is not killed as the title suggested, just removed internally (which makes sense if this was measured, this was a all-or-nothing kind of approach anyways). It's good to keep the possibility of removing checks (basically keeping an optional debugMode) we might find "expensive" along the way for apps. I'm seeing very different I/O patterns recently in which data reads are not necessarily the bottleneck while exploring adios2 usage for performance. Will keep you posted.

pnorbert:
@williamfgc On that note, I just saw in a  simple test of one-block write/read on a single process (though a 10D array) that the read is much slower than by a competing product. I am curios what you will find in the read to be inefficient.

williamfgc:
@pnorbert yes, read-performance is the domain I'm looking into. We might need adjustments in the adios2 readers for large metadata cases (tracking 10D arrays sounds fun) after profiling. Some debugging flag would help remove unnecessary checks (after all it's a feature intended for performance). The competing product might have optimized that read pattern a while ago. Thanks!

PR #2118, opened 2020-04-08
Burstbuffer support using fstream

pnorbert:


pnorbert:
@chuckatkins @williamfgc Would you please review this code carefully? I am not used to ifstreams/ofstreams, so particular attention is needed for how IO exceptions are handled here. 

Rough idea: FileDrainer is a class to 1) operate a queue of tasks 2) to implement read/write I/O. FileDrainerSingleThread runs one thread which processes the queue and copies blocks of data from input file to output file, always at the end of the files; and sometimes write a block of memory to output file (metadata index, profiling.json, 1-byte activate flag at offset 38). BP4Writer controls this thread.

source/adios2/toolkit/burstbuffer/FileDrainer.cpp
```diff
@@ -0,0 +1,279 @@
...
+    else
+    {
+        OutputFile f = std::make_shared<std::ofstream>();
+        m_OutputFileMap.emplace(path, f);
+        Open(f, path, append);
+        return f;
+    }
+}
+
+void FileDrainer::Open(InputFile f, const std::string &path)
```
williamfgc, line 159: Since `InputFile` is a `shared_ptr` this add to the ownership counter. Ownership is not needed so passing `InputFile &`  should be fine.

source/adios2/toolkit/burstbuffer/FileDrainer.cpp
```diff
@@ -0,0 +1,279 @@
...
+        f->open(path, std::ios::out | std::ios::app | std::ios::binary);
+    }
+    else
+    {
+        f->rdbuf()->pubsetbuf(0, 0);
+        f->open(path, std::ios::out | std::ios::trunc | std::ios::binary);
+    }
+}
+
+void FileDrainer::Close(InputFile f) { f->close(); }
```
williamfgc, line 181: same here`InputFile&`.

source/adios2/toolkit/burstbuffer/FileDrainer.cpp
```diff
@@ -0,0 +1,279 @@
...
+void FileDrainer::Close(OutputFile f) { f->close(); }
+
+bool FileDrainer::Good(InputFile f) { return (f->good()); }
```
williamfgc, line 184: same here `InputFile& f`

source/adios2/toolkit/burstbuffer/FileDrainer.h
```diff
@@ -0,0 +1,149 @@
...
+    size_t fromOffset;
+    size_t toOffset;
+    std::vector<char> dataToWrite; // memory to write with Write operation
+
+    FileDrainOperation(DrainOperation op, const std::string &fromFileName,
+                       const std::string &toFileName, size_t countBytes,
+                       size_t fromOffset, size_t toOffset, const void *data);
+};
+
+typedef std::map<std::string, std::shared_ptr<std::ifstream>> InputFileMap;
```
williamfgc, line 61: Not sure if `std::shared_ptr` is needed at all. `ifstream` and `ofstream` can exist (as `false`) without being opened. I'd stick with objects as possible to prevent managing ownership.

source/adios2/toolkit/burstbuffer/FileDrainer.h
```diff
@@ -0,0 +1,149 @@
...
+
+    /** Join the thread. Main thread will block until thread terminates */
+    virtual void Join() = 0;
+
+    /** turn on verbosity. set rank to differentiate between the output of
+     * processes */
+    void SetVerbose(int verboseLevel, int rank);
+
+protected:
+    std::queue<FileDrainOperation> operations;
```
williamfgc, line 105: Nice!

testing/utils/iotest/burstbuffer-BP4-nsub1.xml
```diff
@@ -0,0 +1,20 @@
+<?xml version="1.0"?>
+<adios-config>
+
+    <!--===========================================
+           Configuration for io_T1 group
+        ==========================================-->
+
+    <io name="io_T1">
```
williamfgc, line 8: I'd try adding also YAML config files tests.

docs/user_guide/source/engines/bp4.rst
```diff
@@ -66,7 +68,15 @@ This engine allows the user to fine tune the buffering operations through the fo
 
 12. **StatsBlockSize**: Calculate Min/Max for a given size of each process output. Default is one Min/Max per writer. More fine-grained min/max can be useful for querying the data. 
 
-13. **Node-Local**: For distributed file system. Every writer process must make sure the .bp/ directory is created on the local file system. Required for using local disk/SSD/NVMe in a cluster.
```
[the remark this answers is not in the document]
pnorbert: Yes, both NodeLocal and Node-Local works as parameters. I thought it was just a mistake though to have a dash in the name.

source/adios2/toolkit/burstbuffer/FileDrainer.cpp
```diff
@@ -0,0 +1,279 @@
...
+    FileDrainOperation operation(op, fromFileName, toFileName, countBytes,
+                                 fromOffset, toOffset, data);
+    std::lock_guard<std::mutex> lockGuard(operationsMutex);
+    operations.push(operation);
+}
+
+void FileDrainer::AddOperationSeekEnd(const std::string &toFileName)
+{
+    std::string emptyStr;
+    AddOperation(DrainOperation::SeekEnd, emptyStr, toFileName, 0, 0, 0);
```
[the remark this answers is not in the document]
pnorbert: I experienced compiler error with "" in the argument list. Attributed to the fact that I was expecting a const string&. Maybe I was wrong and moved on with this solution

source/adios2/toolkit/burstbuffer/FileDrainer.cpp
```diff
@@ -0,0 +1,279 @@
...
+}
+
+/*FileDrainOperation::FileDrainOperation(std::string &toFileName,
+                                       size_t countBytes, size_t toOffset, )
+: op(DrainOperation::Write), fromFileName(""), toFileName(toFileName),
+  countBytes(countBytes), append(false), fromOffset(0), toOffset(toOffset){
+
+};*/
+
+FileDrainer::FileDrainer() {}
```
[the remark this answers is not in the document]
pnorbert: There was something there but at the end now this is empty. Will remove it.

source/adios2/toolkit/burstbuffer/FileDrainer.h
pnorbert: I started with that first. Then the XL compiler complained that I was trying to use a deleted copy constructor (of ofstream). I did not know how to solve it, so moved to using shared ptr. 

I could not figure out how to emplace a new ofstream object into the map and then return a reference back to the caller that xlc++ would accept.

testing/utils/iotest/burstbuffer-BP4-nsub1.xml
pnorbert: habits just don't want to die...

source/adios2/toolkit/burstbuffer/FileDrainer.cpp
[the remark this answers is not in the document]
williamfgc: Interesting, `const std::string&` should be fine with "" or any temporary , another option is `std::string("")`. But this is so minor, shouldn't matter.

source/adios2/toolkit/burstbuffer/FileDrainer.h
williamfgc: A comment would be helpful since it's very valuable info on the XL compilers (we'll probably forget). Also, `fstream` doesn't have a copy constructor (only move), so maybe `std::unique_ptr` is more suitable?

williamfgc: Unless this is accessed concurrently, then nevermind since we can't pass `std::unique_ptr &` around. I'd keep `std::shared_ptr` and pass `std::shared_ptr&` as much as possible.

source/adios2/toolkit/burstbuffer/FileDrainer.cpp
pnorbert: Why don't you suggest the same for the Seek() and Good() functions? They are in the protected area while Open/Close is in the private area. Does it make any difference?

williamfgc: Sorry, missed those.  But yes, passing by reference (non-owning manner) should be used as much as possible.

source/adios2/toolkit/burstbuffer/FileDrainer.h
NAThompson: If you pass the `std::shared_ptr` by reference, the reference count is not incremented. Is that what is intended?

source/adios2/toolkit/burstbuffer/FileDrainer.cpp
```diff
@@ -0,0 +1,279 @@
...
+void FileDrainer::Open(InputFile f, const std::string &path)
+{
+
+    f->rdbuf()->pubsetbuf(0, 0);
+    f->open(path, std::ios::in);
+}
+
+void FileDrainer::Open(OutputFile f, const std::string &path, bool append)
+{
+
```
williamfgc, line 168: move `f->rdbuf()->pubsetbuf(0, 0);` here

source/adios2/toolkit/burstbuffer/FileDrainer.cpp
```diff
@@ -0,0 +1,279 @@
...
+            }
+        }
+        buffer += readSize;
+        count -= readSize;
+        totalRead += readSize;
+    }
+    return std::pair<size_t, double>(totalRead, totalSlept);
+}
+
+size_t FileDrainer::Write(OutputFile f, size_t count, const char *buffer,
```
williamfgc, line 256: `const size_t count`  is more clear as `count` is not changing in the function. A local non-const copy to pass it `f->write` would have to be inserted. Unlike POSIX's write fstream is not meant to run in a loop.

source/adios2/toolkit/burstbuffer/FileDrainer.cpp
```diff
@@ -0,0 +1,277 @@
...
+void FileDrainer::CloseAll()
+{
+    for (auto it = m_OutputFileMap.begin(); it != m_OutputFileMap.end(); ++it)
+    {
+        // if (it->second->good())
+        //{
+        Close(it->second);
+        //}
+        m_OutputFileMap.erase(it);
+    }
```
bradking, line 185: @pnorbert the test failures on Windows are due to invalidating the loop iterator here (and in the input loop below).  The `.erase()` removes the current iterator so the `++it` in end of the `for()` tries to increment an invalid iterator.  In a Debug build the MSVC iterator debugging catches it.  Otherwise this likely strays off and doesn't actually close or flush all of the files, hence the test failure.

I fixed it locally with

```patch
diff --git a/source/adios2/toolkit/burstbuffer/FileDrainer.cpp b/source/adios2/toolkit/burstbuffer/FileDrainer.cpp
index 72b0b85e..97eb793a 100644
--- a/source/adios2/toolkit/burstbuffer/FileDrainer.cpp
+++ b/source/adios2/toolkit/burstbuffer/FileDrainer.cpp
@@ -181,16 +181,16 @@ void FileDrainer::CloseAll()
         //{
         Close(it->second);
         //}
-        m_OutputFileMap.erase(it);
     }
+    m_OutputFileMap.clear();
     for (auto it = m_InputFileMap.begin(); it != m_InputFileMap.end(); ++it)
     {
         // if (it->second->good())
         //{
         Close(it->second);
         //}
-        m_InputFileMap.erase(it);
     }
+    m_InputFileMap.clear();
 }

 void FileDrainer::Seek(InputFile &f, size_t offset, const std::string &path)
```

With that the `Utils.IOTest.BurstBuffer.1to1.BP4.Write` test passes.

pnorbert: Thanks for pointing out this mistake. After fixing this, the test passes on Windows only because the draining of metadata succeeds and the followup tests with bpls only read the metadata. The data, however, is not drained at all. 

```
Drain 0: Copy from bb\burstbuffer-32KB.bp\data.0 -> burstbuffer-32KB.bp\data.0 33025 bytes
ADIOS THREAD ERROR: FileDrainer couldn't read from file bb\burstbuffer-32KB.bp\data.0 offset = -1 count = 32927 bytes but only 0. File size at time of reading = -1
: iostream stream error
Drain 0: Copy from bb\burstbuffer-32KB.bp\md.0 -> burstbuffer-32KB.bp\md.0 277 bytes
Drain 0: Write to file burstbuffer-32KB.bp\md.idx 64 bytes of data from memory (no seek)
    App 1: Max write time = 0.0009631
    App 1: Min write time = 0.0009631
App 1 Step 3:
Drain 0: Copy from bb\burstbuffer-32KB.bp\data.0 -> burstbuffer-32KB.bp\data.0 32961 bytes  -- Skip because of previous error
Drain 0: Copy from bb\burstbuffer-32KB.bp\md.0 -> burstbuffer-32KB.bp\md.0 213 bytes
```

The current code skips further operations on any file where an error has happened, basically ruining the draining effort. What I am trying to figure out on Windows that how to recover from EOF status and read successfully later.  Resetting the flags with ifstream.clear() does not help, not even closing the file and opening a new one and starting from scratch. I wrote some standalone test where it works, just not here in this FileDrainer thread.

pnorbert: The strange thing is that I can seek to the end-10 and read the last 10 bytes successfully from the file, but the actual read from offset 0 for the entire size of block (33025 bytes) fails. 

-- Read from bb\burstbuffer-32KB.bp\data.0 start = 0 count = 33025 fileSize = 33025 iostate = 0
-- Test read of 10 bytes from end resulted in iostate = 0 gcount = 10
-- Read reached EOF for bb\burstbuffer-32KB.bp\data.0. read = 98 bytes  iostate = 3

pnorbert: @bradking Thank you for your suggestive question about binary mode. I forgot to add that flag for the read side. This was causing the problem I was seeing.

PR #2121, opened 2020-04-09
Add rendezvous reader count in dataman

JasonRuonanWang:


JasonRuonanWang:
@rkube @jychoi-hpc 

This should fix the problem you saw in the KSTAR-NERSC workflow where a lot of the initial steps are dropped. Now the default behavior is that the writer only starts writing when there is a reader connected. 

If you need other configurations you can set the engine parameter RendezvousReaderCount to a different number, or take a look at the doc at https://github.com/ornladios/ADIOS2/blob/master/docs/user_guide/source/engines/dataman.rst after I merge this PR.

PR #2126, opened 2020-04-10
ppc64le emulation in CI

chuckatkins:


eisenhauer:
Nice!

chuckatkins:
Don't get too excited just yet.  I'm able to build and run the containers on my local machine after registering the `qemu-ppc64le-static` cpu emulation binaries with the kernels binfmf_misc but I'm trying to find out if any of the ci services have it.  So far GitHub actions is a no.

chuckatkins:
@eisenhauer Okay, so this is pretty darn cool.  It seems to be working: https://open.cdash.org/buildSummary.php?buildid=6458866

The trick will really be to determine a reduced build configuration that's still useful since it's so much slower.

eisenhauer:
Something is better than nothing!

NAThompson:
@chuckatkins : This really is cool. So hard to do.

I wonder if, wrt to the expense, you might simply do a compile and not bother running the unit tests. It's still a positive change, relative to what currently exists.

chuckatkins:
@NAThompson.  That's essentially what I've done. The compile takes a little over an hour so I limited the tests to just a few of the "interface" tests which are very short.  My goal was to really be able to get xlc coverage in ci since we've encountered some known issues with it.

williamfgc:
Yeah, this is pretty neat. XL is very different. Thanks @chuckatkins !

KyleFromKitware:
+1 on the last commit (workaround for bad code generation.) Just tested it locally and all tests are passing.

chuckatkins:
@KyleFromKitware After much trial and error, mostly getting the right entrypoint setup and figuring out I needed a statically compiled Node.js to override the one provided by GitHub Actions, I'm able to get this working as a regular container job.  Thanks for the help out-of-band on the suggestions for adjusting the docker image.

Issue #2127, opened 2020-04-10
Deprecated DebugMode param in ADIOS2 init


eisenhauer:
DebugMode has been deprecated (because it was not found to have an effect on performance) and has been removed from ADIOS2 internals, leaving all sanity checks active.  However, this now-inoperable parameter still exists in the various language bindings, either in ADIOS object constructor or init() functions.  For each binding we need to choose a strategy and timeline for transitioning them out (or decide to leave them if it's viewed as too disruptive).

williamfgc:
May I suggest keeping it for future use at the savings of not breaking API for the ecosystem maintainers and users e.g. @ax3l and myself? This is a common performance feature introduced in databases and I found it very useful in the scientific workflows I worked in my past experience. While it might not be the bottleneck in current HPC write cases, it is common to avoid multiple checks in metadata extraction of complex workflows. That being said, I'd like to use adios2 and keep contributing for different I/O problems, but removing features and making the API unstable (as of v2.5.0) in every version makes it very difficult to justify its use for the long-term as a production ready technology, CC other stakeholders: @chuckatkins @pnorbert @sklasky @dpugmire

williamfgc:
Hope this helps the discussion: I ran some benchmarks (kudos to @NAThompson for introducing me to google-benchmark) for a proxy app that resembles data I/O patterns we are trying to tackle (code is below). **Bottomline:** Debug mode can squeeze 2-3% for write by just switching a `bool` flag, I still need to look at read. Squeezing every second is important for the app I'm working with, and this option in adios2 could help.

Running these on my local Mac. These are far from stress tests (1-5M vars-1-5M attributes).
adios2 v2.5.0 `debug mode = true`:
```
Benchmark                       Time             CPU   Iterations
-----------------------------------------------------------------
BM_adios2_write/100000 31966208858 ns   31921618000 ns            1
BM_adios2_write/200000 66933369580 ns   66847440000 ns            1
BM_adios2_write/300000 99180746067 ns   98890918000 ns            1
BM_adios2_write/400000 144998962734 ns   143423604000 ns            1
BM_adios2_write/500000 192458612287 ns   187584078000 ns            1
```

adios2 v2.5.0 `debug mode = false`
```
Benchmark                       Time             CPU   Iterations
-----------------------------------------------------------------
BM_adios2_write/100000 30867953721 ns   30791826000 ns            1
BM_adios2_write/200000 64943908377 ns   64832113000 ns            1
[...]

PR #2132, opened 2020-04-13
bindings: Remove debug_mode flag from C adios2_init_{serial,mpi}

bradking:
The `adios2_init{,_config}_{serial,mpi}` constructor functions were added by #1946.  They have not yet been included in a release, so we can change the signatures.  Drop the `debug_mode` argument.

~~Create undocumented/private `..._with_debug_mode` variants of the constructor functions to implement the legacy `adios2_init` and `adios2_init_config` constructors (whose args vary based on use of MPI).~~ EDIT: Define the legacy `adios2_init` and `adios2_init_config` constructors (whose args vary based on use of MPI) as function-like preprocessor macros that drop the `debug_mode` argument.  This preserves compatibility with callers of the original signatures.

Issue: #2127

### DIFF
--- a/bindings/C/adios2/c/adios2_c_adios.h
+++ b/bindings/C/adios2/c/adios2_c_adios.h
@@ -22,19 +22,17 @@ extern "C" {
 #endif
 
 #if ADIOS2_USE_MPI
-#define adios2_init adios2_init_mpi
-#define adios2_init_config adios2_init_config_mpi
+#define adios2_init(comm, debug_mode) adios2_init_mpi(comm)
+#define adios2_init_config(config_file, comm, debug_mode)                      \
+    adios2_init_config_mpi(config_file, comm)
 
 /**
  * Starting point for MPI apps. Creates an ADIOS handler.
  * MPI collective and it calls MPI_Comm_dup
  * @param comm defines domain scope from application
- * @param debug_mode true: extra user-input debugging information, false:
- * run without checking user-input (stable workflows)
  * @return success: handler, failure: NULL
  */
-adios2_adios *adios2_init_mpi(MPI_Comm comm,
-                              const adios2_debug_mode debug_mode);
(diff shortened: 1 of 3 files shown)


chuckatkins:
Since these are new interfaces and we're moving to deprecate "debug more` anyways tehn we can drop `_debug_mode` from the new interfaces entirely.

bindings/C/adios2/c/adios2_c_adios.h
```diff
@@ -22,19 +22,22 @@ extern "C" {
 #endif
 
 #if ADIOS2_USE_MPI
-#define adios2_init adios2_init_mpi
-#define adios2_init_config adios2_init_config_mpi
+#define adios2_init adios2_init_mpi_with_debug_mode
+#define adios2_init_config adios2_init_config_mpi_with_debug_mode
```
chuckatkins, line 26: Can we drop the `_with_debug_mode` symbol entirely?  Something like:
```c++
#define adios2_init(comm, debug_mode) adios2_init_mpi(comm)
#define adios2_init_config(config, comm, debug_mode) adios2_init_config_mpi(config, comm)
```
That  way we wouldn't even need the separate function implementation.  What do you think?

bradking: Ah, right.  My original patch was from before we removed the `debug_mode` functionality.

chuckatkins: @eisenhauer doing so would render the `debug_mode` parameter in the "deprecated" interface to be a noop.  Would that be acceptable or should we still preserve the functionality?

eisenhauer: Works for me....

bradking: Since #2125 the `debug_mode` parameter has been a no-op already.

PR #2136, opened 2020-04-14
TSAN build

eisenhauer:
Check to see if my clean local TSAN runs also happen in CI.


eisenhauer:
This PR is just the "add TSAN build" commit cherry-picked from @KyleFromKitware 's prior PR.  I think master branch is TSAN clean now.   (I did discover that gcc TSAN doesn't have the code attributes for disabling TSAN in a subroutine, so it won't be clean for that without a suppressions file.  But it seems clean for clang.).

NAThompson:
@eisenhauer : Awesome work!

Issue #2137, opened 2020-04-14
ADIOS 2.5.0: core/Stream fails on i686


ax3l:
**Describe the bug**
ADIOS 2.5.0 does not build on Linux `i686` (docker image: `quay.io/pypa/manylinux2010_i686`) with GCC 7.3.1.

The build error is:
```
In file included from /project/ADIOS2-2.5.0/source/adios2/core/Stream.cpp:12:0:
/project/ADIOS2-2.5.0/source/adios2/core/Stream.tcc: In instantiation of ‘void adios2::core::Stream::Write(const string&, const T&, bool, bool) [with T = std::basic_string<char>; std::string = std::basic_string<char>]’:
/project/ADIOS2-2.5.0/source/adios2/core/Stream.cpp:196:1:   required from here
/project/ADIOS2-2.5.0/source/adios2/core/Stream.tcc:128:9: error: narrowing conversion of ‘18446744073709551613’ from ‘long long unsigned int’ to ‘unsigned int’ inside { } [-Wnarrowing]
         Write(name, &datumLocal, {adios2::LocalValueDim}, {}, {}, vParams(),
         ^~~~~
/project/ADIOS2-2.5.0/source/adios2/core/Stream.tcc: In instantiation of ‘void adios2::core::Stream::Write(const string&, const T&, bool, bool) [with T = signed char; std::string = std::basic_string<char>]’:
/project/ADIOS2-2.5.0/source/adios2/core/Stream.cpp:196:1:   required from here
/project/ADIOS2-2.5.0/source/adios2/core/Stream.tcc:128:9: error: narrowing conversion of ‘18446744073709551613’ from ‘long long unsigned int’ to ‘unsigned int’ inside { } [-Wnarrowing]
/project/ADIOS2-2.5.0/source/adios2/core/Stream.tcc: In instantiation of ‘void adios2::core::Stream::Write(const string&, const T&, bool, bool) [with T = short int; std::string = std::basic_string<char>]’:
/project/ADIOS2-2.5.0/source/adios2/core/Stream.cpp:196:1:   required from here
/project/ADIOS2-2.5.0/source/adios2/core/Stream.tcc:128:9: error: narrowing conversion of ‘18446744073709551613’ from ‘long long unsigned int’ to ‘unsigned int’ inside { } [-Wnarrowing]
[...]

PR #2138, opened 2020-04-15
Cast to size_t due to error in aarch64 gcc 7.3.1

williamfgc:
Address #2137 

@ax3l does this work?

### DIFF
--- a/source/adios2/core/Stream.tcc
+++ b/source/adios2/core/Stream.tcc
@@ -125,8 +125,8 @@ void Stream::Write(const std::string &name, const T &datum, const bool isLocal,
     const T datumLocal = datum;
     if (isLocal)
     {
-        Write(name, &datumLocal, {adios2::LocalValueDim}, {}, {}, vParams(),
-              endStep);
+        Write(name, &datumLocal, {static_cast<size_t>(adios2::LocalValueDim)},
+              {}, {}, vParams(), endStep);
     }
     else
     {
--- a/source/adios2/core/VariableBase.cpp
+++ b/source/adios2/core/VariableBase.cpp
@@ -368,7 +368,7 @@ void VariableBase::InitShapeType()
         }
         else
         {
-            if (m_Shape != Dims{LocalValueDim})
+            if (m_Shape != Dims{static_cast<size_t>(LocalValueDim)})
             {
                 throw std::invalid_argument(
                     "ERROR: LocalValue string variable " + m_Name +


germasch:
Wouldn't the correct fix be to make `LocalValueDim` (and friends) to be `size_t`? (And then use `std::numeric_limits<size_t>::max() - 2` for its value.)

I guess I would argue it'd actually be better to not use the system-dependent `size_t` for dimensions at all, but instead use `uint64_t`. But that's a pretty invasive change.

williamfgc:
@germasch that's possible, but since `size_t` should be 64-bit in targetted platforms it is the same. adios2 never supported 32-bit systems properly. 32-bit systems wouldn't be able to write/read large data (> 2 Gb) anyways. More systems (e g. Linux Mint) are phasing out support for 32-bit these days.

pnorbert:
The requirements are that the LocalValueDim 64-bit constant is properly saved in the BP format where it is stored on 64-bit, and then that it is properly interpreted on any system reading the file. The application is supposed to use the "adios2::LocalValueDim" constant and never rely on its value. So casting internally would be fine.

williamfgc:
@pnorbert that's correct. That would require adding support to handle the 32-bit limit value for size_t. I don't think it's there, along with CI on 32-bit systems, can be an enhancement.

germasch:
> @germasch that's possible, but since `size_t` should be 64-bit in targetted platforms it is the same. adios2 never supported 32-bit systems properly. 32-bit systems wouldn't be able to write/read large data (> 2 Gb) anyways. More systems (e g. Linux Mint) are phasing out support for 32-bit these days.

I think that supports my argument that the dims should be `uint64_t`, rather than `size_t`, doesn't it? (But I still hold that that's kinda of big invasive change, though if on all supported system sizeof(size_t) == 8, it shouldn't break the API, at least.)

BTW, Linux-32 can write files larger than 4GB without trouble, but I agree that's not a target architecture that needs to be supported.

However, the fact that @ax3l gets that warning leads me to think that `sizeof(size_t) < sizeof(uint64_t)` on aarch64, because otherwise I don't see how that's a narrowing conversion. I don't know whether that's true, though, it'd seem odd for a 64-bit architecture to have a 32-bit size_t. Maybe @ax3l can print `sizeof(size_t)`?

williamfgc:
> I think that supports my argument that the dims should be uint64_t, rather than size_t, doesn't it?

@germasch this was a very early discussion, the trade-off is the amount of casting you'd need when interacting with the `size_t`-based STL containers. Ultimately, for practical purposes `size_t` was adopted which is `uint64_t` in all existing business. 

> BTW, Linux-32 can write files larger than 4GB without trouble

That's correct, we run POSIX write in a loop for the above reason. I was referring to the memory allocation limits in 32-bit, not file sizes, any writer/reader is limited by this. 

> However, the fact that @ax3l gets that warning leads me to think that sizeof(size_t) < sizeof(uint64_t) on aarch64

Yes, that's the interesting part. My understanding is that aarch64 size_t is 64-bit, waiting on @ax3l to see if this patch makes any difference.

ax3l:
Thank you for the PR!

Double-checking my report again, this error did indeed occur on `manylinux2010_i686` which I forgot to de-select properly.

I'll check if this fixes the build issue on `i686` (and one could add this to CI, just to make sure the four common data models in C++ compile: https://en.cppreference.com/w/cpp/language/types#Data_models

ax3l:
@williamfgc there is another occurrence:
```
/project/ADIOS2-2.5.0/source/adios2/toolkit/format/bp/BPBase.h:171:33: warning: large integer implicitly truncated to unsigned type [-Woverflow]
         size_t StatsBlockSize = 1125899906842624;
                                 ^~~~~~~~~~~~~~~~
[ 55%] Building CXX object source/adios2/CMakeFiles/adios2.dir/core/Operator.cpp.o
[ 55%] Building CXX object source/adios2/CMakeFiles/adios2.dir/core/Stream.cpp.o
[ 56%] Building CXX object source/adios2/CMakeFiles/adios2.dir/core/Variable.cpp.o
[ 56%] Building CXX object source/adios2/CMakeFiles/adios2.dir/core/VariableBase.cpp.o
/project/ADIOS2-2.5.0/source/adios2/core/VariableBase.cpp: In member function ‘void adios2::core::VariableBase::InitShapeType()’:
/project/ADIOS2-2.5.0/source/adios2/core/VariableBase.cpp:321:46: error: narrowing conversion of ‘18446744073709551613’ from ‘long long unsigned int’ to ‘unsigned int’ inside { } [-Wnarrowing]
             if (m_Shape != Dims{LocalValueDim})
```

https://travis-ci.com/github/openPMD/openPMD-api/jobs/319746575

ax3l:
Excellent, that's it! :)

germasch:
So considering that we now know that we're talking about an actual 32-bit system, we know that `size_t` is 32 bits.  Considering that

```
constexpr uint64_t LocalValueDim = MaxU64 - 2;
```

when one casts to `size_t`, one cuts of all the leading `1` bits. When this eventually gets written to disk, as @pnorbert says, it'll be expanded back to to `uint64_t`, I don't think it'll equal `LocalValueDim`, because all the leading bits are now 0?

Besides, I'd guess if you build adios2 with testing enabled on a 32-bit system, you'll get a whole bunch more of these errors, because it's `uint64_t` but the code attempts to put it into `size_t` dimension.

I personally don't think it's particularly important to support 32-bit systems. But if you do, I think it'd be good to add a CI image for that and check that things actually work.

williamfgc:
@germasch yes, TDD (test driven development) is the way to go. Feel free to open a ticket before this PR is merged and the conversation gets lost. I'd check with @pnorbert to see if it's in the roadmap. From my side we won't need 32-bit support anytime soon.

PR #2150, opened 2020-04-21
Fix CI warnings on PGI

bradking:
These have crept into `master` since PGI has no `-Werror`.


bradking:
There is one warning left in `source/adios2/toolkit/interop/hdf5/HDF5Common.cpp` due to using runtime `if` with a compile-time constant.

Issue #2144, opened 2020-04-20
Double free corruption on ppc64le with XL


chuckatkins:
See https://open.cdash.org/testDetails.php?test=872605402&build=6476675

KyleFromKitware:
Confirmed in my own testing. Investigating now.

KyleFromKitware:
For my testing, I've been using the `Interface.ADIOSDefineAttributeTest.GetAttribute` test. Here's what I've found so far.

In this test, the failure is being caused by a second or invalid call to `free()` inside the `std::string` destructor for `m_TestData.S1array[0]`, inside `TEST_F(ADIOSDefineAttributeTest, GetAttribute)`. What's interesting is that when I put the following at the end of the test function:

```c++
    std::string& asdf = m_TestData.S1array[0];
    printf("%p %p %p %i\n", &asdf, &asdf[0], "one", sizeof(std::string));
```

I don't get the double-free/corruption error. It seems that referencing the string's `[0]` in a mutable fashion mutates the state of the string somehow, bringing it out of some ugly state. However, changing it to:

```c++
    const std::string& asdf = m_TestData.S1array[0];
    printf("%p %p %p %i\n", &asdf, &asdf[0], "one", sizeof(std::string));
```

causes the error to reappear, since the string is not being mutated.

KyleFromKitware:
Unfortunately, I have not been able to find debuginfo for `libstdc++`, and have not been able to get a closer look at what's going on inside `std::string`.

chuckatkins:
So the working theory is something wonky about how XL is doing `std::string` handling, which would in turn be pervasive across the rest of adios given the extent of string handling that's done.  Seems entirely plausible.

Do you think you could try to recreate a micro-example that reproduces the issue outside of adios so we can determine if it's a compiler bug or perhaps just some subtle undefined behavior we're relying on?

KyleFromKitware:
That is going to be my next step. I took a brief crack at it yesterday but didn't have any luck. I'll try doing some things a little differently.

KyleFromKitware:
OK, I have a minimal working example:

```c++
#include <array>
#include <string>

struct TestStruct
{
  std::array<std::string, 1> arr = {{"one"}};
};

int main(void)
{
  TestStruct s;
  return 0;
}
```

I will see if I can find some help in debugging this.

KyleFromKitware:
Hmmm. I'm having a really hard time finding a place to report a bug to IBM.

[...]

PR #2151, opened 2020-04-21
Work around double-free/corruption error in XL

KyleFromKitware:
Fixes: #2144

### DIFF
--- a/testing/adios2/engine/SmallTestData.h
+++ b/testing/adios2/engine/SmallTestData.h
@@ -20,7 +20,7 @@
 struct SmallTestData
 {
     std::string S1 = "Testing ADIOS2 String type";
-    std::array<std::string, 1> S1array = {{"one"}};
+    std::array<std::string, 1> S1array;
     std::array<std::string, 3> S3 = {{"one", "two", "three"}};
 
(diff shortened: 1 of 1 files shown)


chuckatkins:
This doesn't completely fix the issue, several tests still hit the double-free.  Are there other places this needs to be made?

KyleFromKitware:
>  Are there other places this needs to be made?

I think so. Investigating now.

KyleFromKitware:
After encountering another, harder-to-debug issue (`corrupted double-linked list`), I decided to back up and see if something is going wrong at a more fundamental level. What I found is that the compiler is linking to both `libibmc++.so.1` and `libstdc++.so.6`. This seems very suspicious to me... seems like a ripe opportunity for conflicting memory management code to step on each other's toes. I'm doing some more investigation now.

chuckatkins:
If you want to debug it in a privileged environment you can use the `ornladios/adios2:power8-el7-xl` docker image I've set up.  You just need to first setup the in-kernel emulation via binfmt-misc: `docker run --rm --privileged multiarch/qemu-user-static:register --reset`.  You should then be able to use a docker container with the XL compiler in an environment you have root on.

KyleFromKitware:
Closing in favor of the fix in #2126.

PR #2155, opened 2020-04-22
toolkit: link taustubs to 'dl' library explicitly

bradking:
The `tautimer.cpp` source file uses `dlopen` and friends.  Link explicitly to the library providing those symbols.  Previously the symbols were only available when the library was used by a consumer that happened to link the 'dl' library or equivalent.

### DIFF
--- a/source/adios2/CMakeLists.txt
+++ b/source/adios2/CMakeLists.txt
@@ -13,6 +13,8 @@ target_include_directories(taustubs PRIVATE
       $<BUILD_INTERFACE:${PROJECT_BINARY_DIR}/source>
 )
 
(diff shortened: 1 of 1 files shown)


bradking:
I noticed this while working on the modern bindings+core+mpi library split, but didn't include it in previous PRs because it wasn't needed.  Now that those are done I'm polishing this up.

pnorbert:
Wait wait. Wasn't this the intention of @khuck? Is this fix going to cause
trouble?

chuckatkins:
Whoops...

khuck:
I think this is fine, it just adds -dl to the link

PR #2157, opened 2020-04-22
Optimize SST/BP Reader

eisenhauer:
In the existing formulation, ReadVariableBlocks() is a per-variable call.  It requests all writer-side data necessary to fulfill the local read (possibly multiple requests), waits for those requests to complete and then fills the read buffers for that variable.  If latency is long on those requests, this incurs a lot of overhead because multiple variables are handled only serially, while SST can handle many concurrent data requests.  This refactoring breaks ReadVariableBlocks() into two routines, ReadVariableBlocksRequests() and ReadVariableBlocksFill(), with a step in between that waits for all the read requests to complete.  The object is to maximize concurrency in the read requests by issuing the read requests for all pending variables (not just one) and then waiting on all those requests at once.

### DIFF
--- a/source/adios2/engine/sst/SstReader.cpp
+++ b/source/adios2/engine/sst/SstReader.cpp
@@ -528,6 +528,10 @@ void SstReader::PerformGets()
     }
     else if (m_WriterMarshalMethod == SstMarshalBP)
     {
+        std::vector<void *> sstReadHandlers;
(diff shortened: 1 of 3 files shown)


jychoi-hpc:
@jychoi-hpc I am just putting myself to be notified when this is merged.

PR #2164, opened 2020-04-24
Add Findenet module

KyleFromKitware:


chuckatkins:
@eisen can we even use the upstream enet?

KyleFromKitware:
@chuckatkins you mean @eisenhauer? :)

eisenhauer:
Not without mods, no.  There is no CMake in upstream (at least not the regular one, though there are several), and I had to add at least one small subroutine to expose the internal FD.

chuckatkins:
I don't think we can.  And with the mirror in gtkorvo we added a CMake build with package configs, so I don't think we need the find module

KyleFromKitware:
> There is no CMake in upstream

Right, the whole point of this module is to find an enet installation *without* CMake. That's not what's blocking us here. It sounds like the blocker is that @eisenhauer has made fundamental changes to enet that have not been incorporated into upstream. Unfortunately, this means that a package that goes into Debian proper will not be able to use enet. (I encountered exactly the same issue with VTK a few years ago - we made some modifications to libharu that never got into upstream, and thus never got into Debian, and so I couldn't use libharu in the Debian package.)

chuckatkins:
Could it be packaged as adios2-enet?

KyleFromKitware:
> Could it be packaged as adios2-enet?

I'll ask.

chuckatkins:
@eisenhauer  So I assume this should be closed then since we can't use the enet this would find anyways?

eisenhauer:
Probably...  Unless this is useful to find the GTkorvo version packaged separately...

chuckatkins:
Nope, since we'd just use the cmake package config.

Issue #2143, opened 2020-04-18
Option to link to external dependencies instead of thirdparty copies


williamfgc:
**Why is this feature important?**
In preparation to package for major Linux distros repos (Debian and Fedora). Third party copies are ~discouraged~ prohibited if already part of the distribution system, see [Fedora](https://fedoraproject.org/wiki/Bundled_Libraries?rd=Packaging:Bundled_Libraries) and [Debian](https://wiki.debian.org/UpstreamGuide#No_inclusion_of_third_party_code) docs on the subject.  

**What is the potential impact of this feature in the community?**
Having Debian and Fedora distributions would ease the integration of adios2 in projects (my current ones included) due to widespread distribution in other distros (e.g. Ubuntu, Centos). 

**Describe the solution you'd like and potential required effort**
Extend `ADIOS2_USE_EXTERNAL_*` to cover the dependencies of interest, if external dependency not found it might fallback to thirdparty copy inside ADIOS2. I'll be happy to do a PR, might need a new CI container. 

**Additional context**
Current `ADIOS2/thirdparty` list (empty stays as-is):

EVPath  
GTest:  Not needed for installation
KWSys: Can Kitware folks @chuckatkins @bradking @KyleFromKitware provide context?
RapidJSON: target
atl
dill
enet: target
ffs
mingw-w64: not needed
nlohmann_json: target
pugixml: target
pybind11: target
yaml-cpp: target

Note: the targeted libraries seem to have a stable API, so making these tasks easier to maintain.
Input is welcome. CC: @ax3l @pnorbert

ax3l:
Thanks, same is highly encouraged on conda-forge and spack.

bradking:
> KWSys: Can Kitware folks provide context?

KWSys is designed specifically to be configured by host projects into a unique namespace.  There are no external installations.  It is always deployed as part of a host project, and each such deployment is effectively a distinct library configured (both in name and features) for that host project.  The upstream KWSys repository is separate only to de-duplicate source code and maintenance.

chuckatkins:
W.r.t. KWSys, it is internal-only by design.  See https://github.com/ornladios/ADIOS2/issues/2143#issuecomment-616560299

[...]

PR #2171, opened 2020-04-27
Remove RapidJSON for upcoming release packaging

williamfgc:
This removes the currently not used RapidJSON modified copy in `thirdparty`. Since it's not used, it will simplify upstream packaging efforts (e.g. Debian policy discussed in #2143 ).

The current implementation would be parked in this [branch](https://github.com/williamfgc/ADIOS2/tree/with-thirdparty-rapidjson) if required.

CC: @ax3l


chuckatkins:
Please hold off on this until we talk through it on the Tuesday dev call.

williamfgc:
@chuckatkins sounds good, feel free to ping @ax3l and myself since we (mostly him) are putting a lot of effort on packaging for easy deployment to our customers.

williamfgc:
@chuckatkins thanks, I'd be happy to put it back after the release.

Issue #2089, opened 2020-04-01
yaml-cpp import


suchyta1:
Is it possible to reference ADIOS' internal yaml-cpp build in another project's CMake setup? For example, I'm working on the EFFIS workflow tool, which also uses yaml-cpp. The EFFIS install works OK if I use yaml-cpp version 0.6.3, which is the version ADIOS currently uses, but causes a segfault if EFFIS uses a different version of yaml-cpp. (yaml-cpp is nominally a PRIVATE link in EFFIS, but I see the same thing with PUBLIC or INTERFACE.)
 
Since EFFIS needs ADIOS, I was just going to pick up on what ADIOS had built if I could. It doesn’t seem to be built for this though, with how it’s only part of the BUILD_INTERFACE, and how it’s effectively private, i.e. without the yaml-cpp headers in the ADIOS header directory. I tried a hackish use of linking adios2::yaml-cpp in the EFFIS cmake, but that doesn't build (and I had to use the yaml-cpp header from the yaml-cpp repo, since it wasn’t in the ADIOS build output.)

chuckatkins:
The way it's currently implemented, `yaml-cpp` isn't even an actual library file, it's source files are directly included in `libadios`, which also means we can't consume an externally provided one either.  We could transition it to a proper third-party library where we could then build our internal copy or consume an external one, but that wouldn't really address what you want.  Regardless of how it's implemented, adios's internal copies of third party libraries are explicitly for internal use only and not intended to be provided for direct consumption by anything else.

The best way to deal with what you want would be to allow adios to use either an internal or external version and then build both EFFIS and ADIOS with the same external yaml-cpp.

suchyta1:
OK, that makes sense. Thanks for the info. I may attempt a PR to allow an external link, but for now I'll just use the matching version.

williamfgc:
@suchyta1 the part that is interesting is the seg fault caused by the internal/private implementation. The internal yaml-cpp symbols (or any thirdparty in adios2) shouldn't be reachable, not even globally in the adios2 library. As an example of a similar situation, adios2 uses a fixed version of pugixml different from VTK's pugixml. Did you try running `ldd` on your dynamic executable/libraries? Are you using the same compiler for adios2 and effis/yaml-cpp? Does Effis work without adios2 using yaml-cpp other than 0.6.3?

suchyta1:
[...]

PR #2172, opened 2020-04-27
Enable External yaml-cpp compilation

williamfgc:
yaml-cpp has been pretty stable and ships with `cmake` support.
Works on vanilla systems: 

- Ubuntu 18.04 from synaptic 
- macOS Mojave 10.14.6 from Homebrew 2.2.13

Can be added to CI. 
Related to #2143 and #2089
CC: @ax3l

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -127,6 +127,7 @@ adios_option(Fortran   "Enable support for Fortran bindings" AUTO)
 adios_option(SysVShMem "Enable support for SysV Shared Memory IPC on *NIX" AUTO)
 adios_option(Profiling "Enable support for profiling" AUTO)
 adios_option(Endian_Reverse "Enable support for Little/Big Endian Interoperability" AUTO)
+adios_option(EXTERNAL_yaml-cpp "Enable support for using a external yaml-cpp library for packaging" AUTO)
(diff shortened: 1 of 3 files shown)


CMakeLists.txt
```diff
@@ -127,6 +127,7 @@ adios_option(Fortran   "Enable support for Fortran bindings" AUTO)
 adios_option(SysVShMem "Enable support for SysV Shared Memory IPC on *NIX" AUTO)
 adios_option(Profiling "Enable support for profiling" AUTO)
 adios_option(Endian_Reverse "Enable support for Little/Big Endian Interoperability" AUTO)
+adios_option(EXTERNAL_yaml-cpp "Enable support for using a external yaml-cpp library for packaging" AUTO)
```
ax3l, line 130: I think this might belong in `thirdparty/CMakeLists.txt`

williamfgc: @ax3l that's a good question. At first I tried that path, but my lack of cmake knowledge showed up. I couldn't reuse the target from `find_package` coming from the thirdparty directory, it was just going away. I thought cmake target have global scope, maybe that has changed or I am missing something. In any case, we are going with @KyleFromKitware  PR #2175, so I am closing this.

williamfgc:
Closing, replaced with #2175

PR #2178, opened 2020-04-28
Add option to use external GTest

KyleFromKitware:


williamfgc:
@KyleFromKitware thanks for the effort. From the packaging side, `gtest` is not needed at installation, only during build. But I'd double-check with @chuckatkins if this PR would be needed for other purposes. Thanks!

KyleFromKitware:
> From the packaging side, gtest is not needed at installation, only during build.

Yes, but if we want to run the tests during the Debian build, we will need the external GTest to do so.

As I see it, there is very little cost in externalizing GTest here.

KyleFromKitware:
Closing in favor of #2184.

PR #2181, opened 2020-04-28
Add option to use external RapidJSON

KyleFromKitware:


williamfgc:
@KyleFromKitware rapidjson is currently not being used at all. @chuckatkins was going to reach out see PR #2171

KyleFromKitware:
Per @chuckatkins' comment on #2143:

> It's currently not being used but was added as a transition path away from nlohnann_json both due to performance reasons and issues with compilers we need to support for which workarounds were in conflict with upstream development policies.

This leads me to believe that RapidJSON is the way of the future, and should thus be externalized. I will do the same for nlohnann_json since it is still currently being used.

williamfgc:
Correct, for future releases. I was narrowing it down for the upcoming release. This can be easily reinstated when it is actually used in future releases

KyleFromKitware:
Closing in favor of #2184.

PR #2174, opened 2020-04-27
Add option to use external pugixml

KyleFromKitware:

PR #2183, opened 2020-04-29
Refactor helper and query to keep pugixml private

williamfgc:
Put common XML functionality in helper adiosXMLUtil.h
Refactor parsers in Query and adios2XML 
Forward declare pugixml object to keep the dependency private, allowing static external dependencies (e.g. home brew to be linked)

Related to #2143 and #2174


williamfgc:
@chuckatkins  Maybe this a longer term effort to wrap around `pugixml` completely and/or introduce classes in helper? The intention for now is to be able to reuse the `pugixml` interface objects internally given the current code structure. Currently, helper doesn't use classes and I didn't want to be invasive given that there is an upcoming release. This could be a good topic to address after the release.

chuckatkins:
It's already a heavy refactor so even without switching to classes the implementation of the functions can still be moved into the cpp file keeping just the function prototypes in the header.  That wouldn't really change any of the implementation structure and it would keep the pugixml data structures out of header.

williamfgc:
> It's already a heavy refactor 

@chuckatkins I am modifying the implementation, not the class structure of the query (maybe a separate PR to think this through?, I'd ask @guj). This PR is addressing a separate issue towards making `pugixml` private.

williamfgc:
BTW, I see an issue with a SST test reported by one of the memory sanitizers see [here]( https://circleci.com/gh/ornladios/ADIOS2/35629?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-checks-link)

williamfgc:
> It's already a heavy refactor so even without switching to classes the implementation of the functions can still be moved into the cpp file keeping just the function prototypes in the header.

@chuckatkins I'm actually a bit confused about the requested changes after reviewing the code. Consumers of adiosXMLUtils need to interact with pugixml objects. Thus they are declared with forward declarations making it already a private implementation (essentially prototypes as they are unknown types). I don't know how to further hide this without a private implementation class. An example would be helpful.

chuckatkins:
I believe this does it: https://github.com/chuckatkins/ADIOS2/commit/e6d9e32d332c3fe500093811cd9ef97fd0d9c272

It keeps all the code in the same place with a minimal amount of changes but just makes the few pieces directly interacting with pugi as private helper functions.  With no pugixml data structures, actual or forward declared, in any public interface.  Essentially just addressing the the singular issue at hand and pushing a more significant refactoring of the query and xml code to after the release.

KyleFromKitware:
I still don't entirely understand why this change is needed. The files that reference pugixml are all internal - they are not installed and do not export anything pugi-related. Is there something I'm missing?

williamfgc:
@chuckatkins correct, and that part hasn't been changed in its structure. This PR is offloading the common pugixml functions to a helper, that's all. Before this PR, there were copies of the same functions, and one header was including `pugixml.hpp`. That's the scope essentially.

@KyleFromKitware linking with a static external library shows an issue with header inclusion if using an external pugixml. Homebrew, another packaging target see what @ax3l added [here](https://formulae.brew.sh/formula/adios2), distributes a static `pugixml` on macOS Mojave . This PR makes the implementation completely private. Sorry, it I wasn't clearly communicated in my first comment or in #2184

chuckatkins:
It's not ideal but as long as it's out of the public interface for now I suppose it's fine.  We'll look at re-working the class structure and code organization post release.

chuckatkins:
Can we drop the pugixml dependency from the `examples/query/queryTest` executable now too?

guj:
Yes,  you can comment it out in the CMakeFile
And I will deal with it when I have time in a week or two. 

- Junmin

PR #2184, opened 2020-04-29
Add options to use external versions of dependencies

KyleFromKitware:


chuckatkins:
@KyleFromKitware wrt to the ci build, rather than installing the dependencies in every build, let's build an image that has all the external dependencies and use a cache file that explicitly turns all of the options on (skipping of course the ones that Debian won't have packages for like ZFP, SZ, and MGARD).  See https://github.com/ornladios/ADIOS2/pull/2193/files#diff-c1a4ff950e9dc8ac61d3e38e1f1c2d43 .  We should also put the build on GitHub Actions instead of CircleCI since there's still extra slack in the 20 available simultaneous builds whereas circle is already saturated at 4.

source/adios2/CMakeLists.txt
```diff
@@ -117,7 +117,7 @@ target_include_directories(adios2_core
     $<BUILD_INTERFACE:${ADIOS2_SOURCE_DIR}/source>
     $<BUILD_INTERFACE:${ADIOS2_BINARY_DIR}/source>
 )
-target_link_libraries(adios2_core PRIVATE adios2sys_interface pugixml_private taustubs yaml-cpp)
+target_link_libraries(adios2_core PRIVATE adios2sys_interface adios2::thirdparty::pugixml_impl taustubs adios2::thirdparty::yaml-cpp)
```
williamfgc, line 120: If it helps, after #2183 we don't need to distinguish the internal pugixml with `_impl` they are both (thirdparty and external) private dependencies.

KyleFromKitware:
> rather than installing the dependencies in every build

This is deliberate. Debian Sid is a moving target, and packages are always being updated. I want to make sure that ADIOS always builds correctly against the latest versions of packages in Debian.

thirdparty/CMakeLists.txt
```diff
@@ -20,19 +20,65 @@ endfunction()
 
 include(CMakeDependentOption)
 
+set(ADIOS2_USE_EXTERNAL_DEPENDENCIES OFF CACHE BOOL
+  "Use externally supplied dependencies")
+set(ADIOS2_USE_EXTERNAL_GTKORVO_STACK "${ADIOS2_USE_EXTERNAL_DEPENDENCIES}" CACHE BOOL
```
chuckatkins, line 25: AFAICT EVPath is the only actual dependency we're using and all the other pieces, i.e. atl, ffs, dill, enet, are dependencies of EVPath.  @eisenhauer is that right?

If that's the case then what does `ADIOS2_USE_EXTERNAL_GTKORVO_STACK` give you that `ADIOS2_USE_EXTERNAL_EVPATH` doesn't?

scripts/ci/cmake/ci-debian-sid.cmake
```diff
@@ -0,0 +1,13 @@
+# Client maintainer: chuck.atkins@kitware.com
+
+set(dashboard_cache "
+ADIOS2_USE_EXTERNAL_DEPENDENCIES=ON
+ADIOS2_USE_EXTERNAL_GTKORVO_STACK=OFF
```
chuckatkins, line 5: This should still turn on the explicit features, otherwise a failure in auto-detection would successfully configure and build a configuration we don't expect.

scripts/ci/setup/ci-debian-sid.sh
```diff
@@ -0,0 +1,26 @@
...
+  cmake \
+  ninja-build \
+  g++ \
+  gfortran \
+  pkg-config \
+  libpugixml-dev \
+  libyaml-cpp-dev \
+  pybind11-dev \
+  libgtest-dev \
+  nlohmann-json3-dev
```
chuckatkins, line 26: We'll also want to explicitly add bzip2, c-blosc, hdf5, and zeromq.

scripts/ci/gh-actions/linux-setup.sh
```diff
@@ -1,3 +1,18 @@
 #!/bin/bash
 
+set -e
+
+export CI_ROOT_DIR="${GITHUB_WORKSPACE}/.."
+export CI_SOURCE_DIR="${GITHUB_WORKSPACE}"
+
+SETUP_SCRIPT=${CI_SOURCE_DIR}/scripts/ci/setup/ci-${GITHUB_JOB}.sh
```
chuckatkins, line 8: `GITHUB_JOB` -> `GH_YML_JOBNAME`

thirdparty/CMakeLists.txt
KyleFromKitware: It controls the default value for `ADIOS2_USE_EXTERNAL_<gtkorvo-dependency>`. It is possible, though admittedly unlikely, to use the built-in EVPath while using an external atl/ffs/etc.

I suppose I could take this option out and just make the default `OFF` for all of them. Then anyone who wants to use an external version has to explicitly set it to `ON`, which most likely just means setting `ADIOS2_USE_EXTERNAL_EVPATH` to `ON`.

chuckatkins: Isn't it already setup that way but with `cmake_dependent_option`? The only thing I see this really controlling is a shortcut to using an internal EVPath with external dependencies, which is really a one off use case that seldom, if ever, people would use.  And even if they did they can still add the extra options for it.

thirdparty/CMakeLists.txt
```diff
@@ -20,19 +20,65 @@ endfunction()
...
+  "Use an externally supplied GTest library")
+set(ADIOS2_USE_EXTERNAL_PUGIXML "${ADIOS2_USE_EXTERNAL_DEPENDENCIES}" CACHE BOOL
+  "Use an externally supplied pugixml library")
+set(ADIOS2_USE_EXTERNAL_PYBIND11 "${ADIOS2_USE_EXTERNAL_DEPENDENCIES}" CACHE BOOL
+  "Use an externally supplied pybind11 library")
+set(ADIOS2_USE_EXTERNAL_NLOHMANN_JSON "${ADIOS2_USE_EXTERNAL_DEPENDENCIES}" CACHE BOOL
+  "Use an externally supplied nlohmann_json library")
+set(ADIOS2_USE_EXTERNAL_YAMLCPP "${ADIOS2_USE_EXTERNAL_DEPENDENCIES}" CACHE BOOL
+  "Use an externally supplied yaml-cpp library")
+
```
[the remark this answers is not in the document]
KyleFromKitware: I suppose they could. This isn't really any different - just a stylistic difference.

KyleFromKitware:
Should we also do a non-MPI build on Debian, or is it fine to just do MPI?

williamfgc:
@KyleFromKitware that's a very good question. The closest I see is `HDF5` packages:

- libhdf5-serial-dev : dummy transition to libhdf5-dev
- libhdf5-openmpi-dev
- libhdf5-mpich-dev

I think adios2 would follow the same pattern to make our life easier (I wonder about `libadios2-dev`, we might not have one)
- libadios2-serial-dev
- libadios2-openmpi-dev
- libadios2-mpich-dev

pnorbert:
The difference is that libadios2-openmpi-dev or libadios2-mpich-dev could
be used as the serial library, kinda making libadios2-serial-dev
unnecessary. But how do you make users know that?

williamfgc:
@pnorbert yes, in principle this is going to be different from the current pattern in Debian. Maybe `libadios2-serial-dev` should always be installed? I'd have to dig deeper with Debian configurations and learn more about @bradking target separation, though.

chuckatkins:
> The difference is that libadios2-openmpi-dev or libadios2-mpich-dev could be used as the serial library, kinda making libadios2-serial-dev unnecessary.

@bradking and I talked about this last week and it's an important distinction to make.  While an MPI build does provide serial libraries it wouldn't take the place of an exclusively serial build.  The serial libraries provided with the MPI build are inherently tied to the MPI build so it makes sense to keep a separate `-serial` package.

thirdparty/CMakeLists.txt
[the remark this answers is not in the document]
chuckatkins: Please do as it's what the rest of the options are already using.

.github/workflows/build-and-test.yml
```diff
@@ -38,6 +39,8 @@ jobs:
           container: ornladios/adios2:ci-suse-pgi
         - jobname: suse-pgi-openmpi
           container: ornladios/adios2:ci-suse-pgi-openmpi
+        - jobname: debian-sid-openmpi
+          container: debian:sid
 
     steps:
     - uses: actions/checkout@v2
```
chuckatkins, line 46: It looks like we need to move the checkout step to after the `Setup` step since the action isn't able to find an installed git in the debian build so the dashboard submissions aren't showing up.

chuckatkins:
This is looking great!  Coming along nicely.

chuckatkins:
Re packaging: the serial libs provided by an MPI build should not be siphoned off into a serial package, they need to remain two separate builds.  If anything it's due to the hdf5 support and that we can't support both serial and parallel versions of it in the same build.

.github/workflows/build-and-test.yml
KyleFromKitware: Can you run the checkout after the setup? Doesn't the script need to be there (checked out) in order to run it? I'll try it and see what happens.

KyleFromKitware: Yeah, that failed miserably. Not surprising.

I guess I'll have to make a minimal image that has `git` initially installed just to bootstrap the process. The rest of the dependencies will still be updated on the fly.

chuckatkins: I think using the upstream image like you have now is good so maybe you can add an initial first step to just install git and make it conditional on not using an ornladios image and matches "debian"

KyleFromKitware: Good point. I'll try that.

KyleFromKitware: OK, looks like that worked.

KyleFromKitware:
> The closest I see is `HDF5` packages:

Yikes. The [`debian/rules`](https://salsa.debian.org/debian/hdf5/-/blob/unstable/debian/rules) for building this is not pretty.

KyleFromKitware:
OK, I'll go ahead and add a serial Debian build as well.

KyleFromKitware:
All of the [test failures](https://open.cdash.org/testDetails.php?test=878833954&build=6506727) for the Debian build say:

```console
terminate called after throwing an instance of 'std::runtime_error'
  what():  binding zmq socket failed
```

Anyone seen this before?

eisenhauer:
> All of the [test failures](https://open.cdash.org/testDetails.php?test=878833954&build=6506727) for the Debian build say:
> 
> ```
> terminate called after throwing an instance of 'std::runtime_error'
>   what():  binding zmq socket failed
> ```
> 
> Anyone seen this before?

@JasonRuonanWang ?

JasonRuonanWang:
> > All of the [test failures](https://open.cdash.org/testDetails.php?test=878833954&build=6506727) for the Debian build say:
> > ```
> > terminate called after throwing an instance of 'std::runtime_error'
> >   what():  binding zmq socket failed
> > ```
> > 
> > 
> > Anyone seen this before?
> 
> @JasonRuonanWang ?

I have never seen this before. Did anyone change any ZMQ related stuff recently?

eisenhauer:
> > > All of the [test failures](https://open.cdash.org/testDetails.php?test=878833954&build=6506727) for the Debian build say:
> > > ```
> > > terminate called after throwing an instance of 'std::runtime_error'
> > >   what():  binding zmq socket failed
> > > ```
> > > 
> > > 
> > > Anyone seen this before?
> > 
> > 
> > @JasonRuonanWang ?
> 
> I have never seen this before. Did anyone change any ZMQ related stuff recently?

i think the key here is that the debian build is new, so the environment might be causing ZMQ problems, or need tweaks...

KyleFromKitware:
> i think the key here is that the debian build is new, so the environment might be causing ZMQ problems, or need tweaks...

Indeed. No other builds are failing, and I can't reproduce it locally, so I think it's some combination of Debian + the GitHub environment. I'm trying to debug it right now.

KyleFromKitware:
So, I have now been able to reproduce the issue locally. It looks like running the tests in parallel triggers the error, while running them serially does not. I'm tempted to just run the tests serially and be done with it, but this still doesn't explain why all the other builds run in parallel just fine. I'm going to do some more investigation.

eisenhauer:
@JasonRuonanWang Might this be due to ZMQ using a fixed port by default (causing conflicts with multiple instances)?  If so, just forcing them to run serially might indeed be a simple solution...

chuckatkins:
It makes sense.  A socked bind failure could easily occur if two processes are trying to use the same tcp port.  All of our other MPI builds run their tests serially anyways.

JasonRuonanWang:
Yes that's very possible. Let me take a look? Which test was failing? Or all of them?

KyleFromKitware:
>  All of our other MPI builds run their tests serially anyways.

In that case it's a no-brainer. I've just updated the MPI test to run serially. That should fix it.

KyleFromKitware:
> Which test was failing? Or all of them?

The `Engine.DataMan.DataManEngineTest` tests fail when running in parallel due to the serial and MPI tests using the same ports. Not all of them fail at once, but at least one of them fails every time you run in parallel.

JasonRuonanWang:
> > Which test was failing? Or all of them?
> 
> The `Engine.DataMan.DataManEngineTest` tests fail when running in parallel due to the serial and MPI tests using the same ports. Not all of them fail at once, but at least one of them fails every time you run in parallel.

Do you want me to merge a fix to the master and then you rebase? Or do you want me to make changes in this branch? Or you can just feel free to change the port numbers in all dataman tests to make them more scattered. I am OK anyway.

KyleFromKitware:
> Do you want me to merge a fix to the master and then you rebase?

I think we can open a separate PR to fix the port numbers independent of this. All the other MPI builds run serially, so those can be updated at the same time as the port numbers are fixed.

KyleFromKitware:
I re-ran the tests due to a spurious failure in one of the builds. Everything looks good so far.

JasonRuonanWang:
I have merged the PR which changes the port numbers. Please rebase and let me know if there is still problems.

KyleFromKitware:
OK, I've rebased and made the Debian MPI build parallel again.

KyleFromKitware:
@JasonRuonanWang that still didn't work. I'm still seeing bind errors.

I'm going to make the MPI tests serial. We can change it in a follow-up PR.

JasonRuonanWang:
> @JasonRuonanWang that still didn't work. I'm still seeing bind errors.
> 
> I'm going to make the MPI tests serial. We can change it in a follow-up PR.

How many MPI processes do you use for each job?

KyleFromKitware:
> How many MPI processes do you use for each job?

@eisenhauer @chuckatkins ?

eisenhauer:
> > How many MPI processes do you use for each job?
> 
> @eisenhauer @chuckatkins ?

Well, for the staging-common tests, the reader and writer sizes are encoded into the test name.   The biggest is 3x5 (or 5x3), but they vary.

KyleFromKitware:
I have added an mpich build.

KyleFromKitware:
> I have added an mpich build.

...and it promptly fell flat on its face.

KyleFromKitware:
> ...and it promptly fell flat on its face.

Looks like there's no `mpi4py` for mpich. So I've disabled Python in the mpich build.

KyleFromKitware:
Any insight on [these](https://open.cdash.org/viewTest.php?onlyfailed&buildid=6513009) test failures? They all say:

```console
Fatal error in PMPI_Rget: Wrong synchronization of RMA calls , error stack:
PMPI_Rget(178): MPI_Rget(origin_addr=0x55a3525e2170, origin_count=847, MPI_CHAR, target_rank=0, target_disp=0, target_count=847, MPI_CHAR, win=0xa0000000, request=0x55a3525e06a0) failed
MPID_Rget(93).: Wrong synchronization of RMA calls
```

eisenhauer:
> Any insight on [these](https://open.cdash.org/viewTest.php?onlyfailed&buildid=6513009) test failures? They all say:
Sorry, no clue...

JasonRuonanWang:
> Any insight on [these](https://open.cdash.org/viewTest.php?onlyfailed&buildid=6513009) test failures? They all say:
> 
> ```
> Fatal error in PMPI_Rget: Wrong synchronization of RMA calls , error stack:
> PMPI_Rget(178): MPI_Rget(origin_addr=0x55a3525e2170, origin_count=847, MPI_CHAR, target_rank=0, target_disp=0, target_count=847, MPI_CHAR, win=0xa0000000, request=0x55a3525e06a0) failed
> MPID_Rget(93).: Wrong synchronization of RMA calls
> ```

These seem to be the one-sided MPI calls in SSC. I have never tried MPICH.

chuckatkins:
> > How many MPI processes do you use for each job?
> 
> @eisenhauer @chuckatkins ?

I think the most we use is 8.  I've had reliability problems with mpich and oversubscription in the past so that may be behind some of the test failures.  Check if there's any environment variables or mpiexec arguments that need to be added to better handle the oversubscription of MPI processes to available slots with the mpich build.  There's quite a few we set for openmpi to workaround a variety of issues.

KyleFromKitware:
Maybe it would be easier if we hold off on mpich for now. I'm not well-versed enough in MPI to debug this issue, and fixing it in this PR is scope creep. We can add an mpich build and fix the issues in a follow-up PR.

JasonRuonanWang:
> Maybe it would be easier if we hold off on mpich for now. I'm not well-versed enough in MPI to debug this issue, and fixing it in this PR is scope creep. We can add an mpich build and fix the issues in a follow-up PR.

Sounds good. The main use cases of SSC all rely on OpenMPI. So we are not very concerned whether it works with MPICH for now. You can also turn off SSC tests in the MPICH build too.

PR #1829, opened 2019-10-15
Use one-sided MPI in SSC engine

JasonRuonanWang:

PR #2186, opened 2020-04-29
helper: Cleanup adiosNetwork

bradking:
Remove unused `Handshake{Reader,Writer}` from `adiosNetwork`.  These have been unused since #1829.  Also, `adiosMpiHandshake` is now available.

Update when `adiosNetwork` provides `AvailableIpAddresses`.  Since #1659 the function is needed by the Table engine.  Since #1329 the function is not directly used by anything else.


bradking:
@JasonRuonanWang thanks.  I was just removing this because it looked unused.  I'm happy to leave it in.  However, we should at least update the conditions in which it is provided.  Currently if one turns off dataman and ssc but turns on table then it won't compile.

Is there any reason to make these functions conditional on other packages?  We could just turn them on all the time.  They are not too big.

JasonRuonanWang:
If I remember, the conditions set was because some headers this function uses is not available on Windows. Ultimately, DataMan will use all functions in this file, and Table will use AvailableIpAddresses() only.

bradking:
Okay, closing in favor of #2187

PR #2187, opened 2020-04-29
helper: Always provide adiosNetwork functions on UNIX platforms

bradking:
Since #1659 the `AvailableIpAddresses` function is needed by the Table engine but its conditions only make it available for dataman or ssc.  Drop those conditions so it is available in general.

For now we still need to leave these out on Windows.  The `AvailableIpAddresses` function in particular needs a conditional alternative implementation on Windows.

### DIFF
--- a/source/adios2/helper/adiosNetwork.cpp
+++ b/source/adios2/helper/adiosNetwork.cpp
@@ -13,7 +13,6 @@
 #include "adios2/toolkit/transport/file/FileFStream.h"
 
 #ifndef _WIN32
-#if defined(ADIOS2_HAVE_DATAMAN) || defined(ADIOS2_HAVE_SSC)
 
 #include <iostream>
 #include <thread>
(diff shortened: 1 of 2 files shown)


JasonRuonanWang:
@bradking Thanks for the help!

chuckatkins:
w.r.t the pgi errors, we currently disable anything using the json library due to compiler issues.

bradking:
@chuckatkins thanks.  That PGI error may be why these conditions were there in the first place.  Shall we just add "or Table" to the conditions instead?

chuckatkins:
They were.  basically Jason's engines use json and the json library we use has both pgi and xl issues so we disable those engines, and thus json, for those compilers.

JasonRuonanWang:
Hopefully we can get rid of these headache issues by completely moving to RapidJson after the release.

bradking:
Thanks.  Closing in favor of #2188.

PR #2188, opened 2020-04-29
helper: Enable adiosNetwork functions for Table engine

bradking:
Since  #1659 the `AvailableIpAddresses` function is needed by the Table engine but its conditions only make it available for dataman or ssc.  Update the conditions to enable the functions for Table and drop them for SSC (which does not use them).  Also add a comment explaining why the conditions are needed in the first place.


bradking:
@JasonRuonanWang thanks.  I updated the conditions to add Table and drop SSC.

PR #2197, opened 2020-05-03
Added IO.GetVariableNames() function which returns a vector of name-t…

### LINKED ISSUE #2196: Faster version of available_variables() in python bindings.
As @khuck & co reported, available_variables() is a costly call. 
Need a faster way to get the list of variables in python.

A followup PR is introducing 'get_variable_names()' that returns a list of pairs (name, type). The low-level python and the original C++ API call is IO.GetVariableNames() that returns a vector of pairs.

pnorbert:
…ype string pairs. It is much faster than the current AvailableVariables() function. Added a python high-level function 'get_variable_names()' which returns a list of string-pairs (name,type). Added python test. Fixes #2196

### DIFF
--- a/bindings/CXX11/adios2/cxx11/IO.cpp
+++ b/bindings/CXX11/adios2/cxx11/IO.cpp
@@ -154,6 +154,13 @@ std::string IO::EngineType() const
     return m_IO->m_EngineType;
 }
 
+std::vector<std::pair<std::string, std::string>> IO::GetVariableNames() const
+    noexcept
+{
+    helper::CheckForNullptr(m_IO, "in call to IO::GetVariablesDataMap");
+    return m_IO->GetVariableNames();
+}
+
 // PRIVATE
 IO::IO(core::IO *io) : m_IO(io) {}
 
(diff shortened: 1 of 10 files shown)


williamfgc:
@pnorbert thanks, I think this is great! Also, I'm thinking the current `available_variables` could be further generalized by passing a list of keys to only extract certain info. Something like: `vars_info = fstep.available_variables(keys=[type, dims, ...])` while an empty list would the default returning everything that way current behavior and API are not broken. But if this what 90% of people would do, it's fine to have a separate function.

pnorbert:
This new function is ready to merge but it does not serve too much purpose for the TAU use case after figuring out that the problem was the use of a debug build. Also @williamfgc's comment is valid, maybe we should rework the high-level python function instead of having yet another function. The C++/low-level python extra function GetVariableNames though is fine I think.

williamfgc:
@pnorbert let me work on overloading `available_variables` . When are you freezing for `v2.6.0`?

pnorbert:
Thanks, I appreciate your help. Feature freeze is by May 15, planned release is end of May.

chuckatkins:
So, do we merge this, close it in favor of a different approach, or leave it as is and mark it WIP?

pnorbert:
Superseeded by #2214

PR #2201, opened 2020-05-04
ci: Fail if ctest_build detects any compiler warnings

bradking:
In order to handle compilers like PGI that have no `-Werror` option or equivalent, teach our CI ctest script to fail if `ctest_build` reports that any warnings were detected.


bradking:
The CI now fails the "build" step with:

```
CMake Error at /__w/ADIOS2/ADIOS2/scripts/ci/cmake/ci-common.cmake:61 (message):
  Found 1 warnings.
```

on the PGI builds.  That's what we want here.  Next I'll update this PR to actually fix the warnings.

bradking:
> Next I'll update this PR to actually fix the warnings.

I added commits to fix the warnings, and now CI is clean and verifies that there are no warnings.

PR #2206, opened 2020-05-05
ci: Auto-cancel builds when pr updates are pushed

chuckatkins:


chuckatkins:
This unfortunately doesn't work for PRs from forks due to the read-only permissions of the token when used from forks :-(

PR #2208, opened 2020-05-05
Add library suffix option

KyleFromKitware:


chuckatkins:
How would this actually get used?  Afaik the way distros typically deal with it is to just use unique install directories for different MPI installations so there's no conflict.

KyleFromKitware:
> Afaik the way distros typically deal with it is to just use unique install directories for different MPI installations so there's no conflict.

Debian appears to do a hybrid of two approaches. It puts a suffix after the library base name, but also puts a symlink to the real library inside a separate subdirectory as you described.

chuckatkins:
If there's going to be a suffix then I'd rather see it applies as an actual suffix to the library names, not just to the `adios2` portion.  i.e. `adios2_sst${ADIOS2_LIBRARY_SUFFIX}`  instead of `adios2${ADIOS2_LIBRARY_SUFFIX}_sst`.  Also does the suffix need to be applied to the installed binaries as well?

KyleFromKitware:
> If there's going to be a suffix then I'd rather see it applied as an actual suffix to the library names, not just to the adios2 portion.

The convention for hdf5 is `libhdf5_<serial|openmpi>_<component>.so`, though I suppose we can put it at the end.

KyleFromKitware:
> Also does the suffix need to be applied to the installed binaries as well?

Yes, I believe so.

chuckatkins:
> > If there's going to be a suffix then I'd rather see it applied as an actual suffix to the library names, not just to the adios2 portion.
> 
> The convention for hdf5 is `libhdf5_<serial|openmpi>_<component>.so`, though I suppose we can put it at the end.

That's fine then.  It's really a specific use case for packagers, mostly debian, so if that's how they like to have it then it's fine.

chuckatkins:
> > Also does the suffix need to be applied to the installed binaries as well?
> 
> Yes, I believe so.

The library names are fine as is then, just mangle the binary and config script names and it should be good to merge.

source/utils/adios_iotest/CMakeLists.txt
```diff
@@ -12,6 +12,7 @@ target_link_libraries(adios_iotest adios2::cxx11_mpi MPI::MPI_C)
 if(WIN32)
   target_link_libraries(adios_iotest getopt)
 endif()
+set_property(TARGET adios_iotest PROPERTY OUTPUT_NAME adios2${ADIOS2_LIBRARY_SUFFIX}_iotest)
```
chuckatkins, line 15: @pnorbert Same as above on renaming `adios_iotest` to `adios2_iotest`

CMakeLists.txt
```diff
@@ -78,6 +78,11 @@ if(NOT MSVC)
   set(CMAKE_C_STANDARD_REQUIRED True)
 endif()
 
+# Option for packagers to add a suffix after "adios" in library names (to
+# install multiple builds in parallel)
+set(ADIOS2_LIBRARY_SUFFIX "" CACHE STRING
+  "Suffix to put after \"adios\" in library names")
+
```
chuckatkins, line 85: Should be marked as advanced.

source/utils/adios_iotest/CMakeLists.txt
pnorbert: I was one time trying to just name everything adios, which is clearly not the case. I guess adios2 used everywhere and everything would be more consistent. Same with _ vs - in the names.

chuckatkins: We should just be sure to explicitly mention in the release notes the tool renaming as it could be considered an API breakage.

source/utils/CMakeLists.txt
```diff
@@ -47,10 +48,12 @@ else()
 endif()
 add_executable(adios_reorganize ${adios_reorganize_srcs})
 target_link_libraries(adios_reorganize PRIVATE adios2_core ${maybe_mgard})
+set_property(TARGET adios_reorganize PROPERTY OUTPUT_NAME adios2${ADIOS2_LIBRARY_SUFFIX}_reorganize)
```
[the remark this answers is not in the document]
chuckatkins: This is acceptable per the response in the below comment

CMakeLists.txt
```diff
@@ -78,6 +78,14 @@ if(NOT MSVC)
   set(CMAKE_C_STANDARD_REQUIRED True)
 endif()
 
+# Option for packagers to add a suffix after "adios" in library names (to
+# install multiple builds in parallel)
+set(ADIOS2_LIBRARY_SUFFIX "" CACHE STRING
+  "Suffix to put after \"adios\" in library names")
+set(ADIOS2_EXECUTABLE_SUFFIX "" CACHE STRING
```
chuckatkins, line 85: How would these two be different?  Just curious as to what the actual use case is.

KyleFromKitware: `ADIOS2_LIBRARY_SUFFIX` would be `_serial` or `_openmpi` (resulting in `libadios2_serial_core.so`) while `ADIOS2_EXECUTABLE_SUFFIX` would be `.serial` or `.openmpi` (resulting in `bpls.serial`.)

chuckatkins: Is there really a need for two separate suffixes that will generally be almost the same?  It seems like just one applying the same to both libraries and executables should be sufficient.

KyleFromKitware: The difference is that one uses a `_` while the other uses a `.`. We could have some sort of logic that consolidates the two and only includes the `_` or `.` if the suffix is non-empty.

chuckatkins: Sure, but why make them different at all?  The whole point is to just have them distinctly tagged for simultaneous installs

KyleFromKitware: This is based on the convention that HDF5 uses in Debian. It uses `libhdf5_openmpi_foo.so` and `h5cc.openmpi`.

chuckatkins: Right, but we should be able to do all the necessary separation and symlinking with `libadios2_core.so -> libadios2_openmpi_core.so ` and `bpls -> bpls_openmpi`.  Even if it's different from how the `hdf5` deb package is setup it still achieves the necessary capability to do what we need to do in the package, right?

KyleFromKitware: It does, but I worry that I may be asked to change it to match the HDF5 convention.

As a middle ground, what if we made `ADIOS2_EXECUTABLE_SUFFIX` default to the same value as `ADIOS2_LIBRARY_SUFFIX`, so that if we don't need to make them different, we can just set the one?

chuckatkins:
@eisenhauer I've been seeing some instability with the sst tests timing out waiting for the writer to start.  Do we need to increase the 1s timeout to something more forgiving, like 20s?  That would help with the vms being overloaded.

CMakeLists.txt
chuckatkins: That's acceptable then I think.

eisenhauer:
> @eisenhauer I've been seeing some instability with the sst tests timing out waiting for the writer to start. Do we need to increase the 1s timeout to something more forgiving, like 20s? That would help with the vms being overloaded.

The current default open timeout is 60s.  We drop it to 5 when testing if it works, but generally it's pretty lenient because it's more user-facing...  Can you point me to failures?

chuckatkins:
> > @eisenhauer I've been seeing some instability with the sst tests timing out waiting for the writer to start. Do we need to increase the 1s timeout to something more forgiving, like 20s? That would help with the vms being overloaded.
> 
> The current default open timeout is 60s. We drop it to 5 when testing if it works, but generally it's pretty lenient because it's more user-facing... Can you point me to failures?

https://open.cdash.org/index.php?compare1=61&filtercount=1&field1=revision&project=ADIOS&showfilters=0&limit=100&value1=b092746298c89bda6d8934aefb6a8f5040292242&showfeed=0

These look like timeout failures.  I'm rerunning them now to see if it gets past them

eisenhauer:
They didn't run long enough to be timeouts.  I'd suspect the other option "non-current SST contact file".   Filename collisions between tests?  Or contact files leftover from prior testing runs?  (The staging-common/run_tests script runs each test in a subdirectory with the same name as the testname, clearing it before test start in case it didn't get cleared at the end of the previous run because of timeout or other misadventure.)

eisenhauer:
At least, the "open failed" tests might be the above problem.  But the 300s timeouts are different.  It looks like the reader is hung.  It connected, because we got a couple of "Processing step" messages, but it got no further.  It's not impossible that this is related to the problem above, that is we're running these tests in parallel (or prior run isn't being cleaned), their filenames are conflicting, so readers are trying to connect to wrong writers, maybe getting connected to ones that are still running in some other test, etc.

chuckatkins:
> At least, the "open failed" tests might be the above problem. But the 300s timeouts are different. It looks like the reader is hung. It connected, because we got a couple of "Processing step" messages, but it got no further. It's not impossible that this is related to the problem above, that is we're running these tests in parallel (or prior run isn't being cleaned), their filenames are conflicting, so readers are trying to connect to wrong writers, maybe getting connected to ones that are still running in some other test, etc.

Digging deeper into the logs I think I see what's happening.  The filename conflict is a red herring, the root issue is the 5m timeout.  The ctest retry is causing the filename conflict since after the initial timeout failure ctest is re-running the test (we have it set to retry up to 5 times until pass) and the retries are hitting the filename conflict.  The actual failure is the 5m timeout the first time it's run though.  For the first pass through, the test fixtures force a a setup step that cleans anything out but that doesn't happen on retries since the setup is actually a separate "test".

chuckatkins:
See #2234 and rebase once addresed

chuckatkins:
@KyleFromKitware Now that both this and the external deps are merged, it would be good to have the `debian-sid` and `debian-sid-openmpi` builds add the necessary cache entries in their builds for `_serial` and `_openmpi` respectively.

KyleFromKitware:
> @KyleFromKitware Now that both this and the external deps are merged, it would be good to have the `debian-sid` and `debian-sid-openmpi` builds add the necessary cache entries in their builds for `_serial` and `_openmpi` respectively.

Please take a closer look at the diff... ;)

PR #2209, opened 2020-05-05
Use WITH_SOABI argument in Python_add_library()

KyleFromKitware:


bindings/Python/CMakeLists.txt
```diff
@@ -7,6 +7,7 @@ Python_add_library(adios2_py MODULE
   py11Operator.cpp
   py11File.cpp py11File.tcc
   py11glue.cpp
+  WITH_SOABI
```
chuckatkins, line 10: According to the documentation, the argument needs to directly follow the library type:
```
Python_add_library (<name> [STATIC | SHARED | MODULE [WITH_SOABI]]
                    <source1> [<source2> ...])
```

PR #2210, opened 2020-05-06
Print more info in common test reader in the quest for hunting down t…

pnorbert:
…he ubsan error on bp4 streaming

### DIFF
--- a/testing/adios2/engine/staging-common/TestCommonRead.cpp
+++ b/testing/adios2/engine/staging-common/TestCommonRead.cpp
@@ -70,27 +70,21 @@ TEST_F(CommonReadTest, ADIOS2CommonRead1D8)
     }
     Seconds timeOpen = std::chrono::steady_clock::now() - ts;
     EXPECT_TRUE(engine);
-    if (!mpiRank)
-    {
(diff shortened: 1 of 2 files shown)


eisenhauer:
@pnorbert Looks like this yielded some fruit:

https://open.cdash.org/testDetails.php?test=882236443&build=6524996

pnorbert:
Thanks. Not much learned, except for confirming that there are 0 steps. Is
it possible that there is an existing .bp directory at the time of the test?

eisenhauer:
Hmm.  That question made me wonder briefly if the ctest retry feature might be involved here.  As @chuckatkins pointed out WRT the HeatTransfer tests, the "pipeline" things he has been using for some of the staging tests can result in tests possibly get restarted without anything having cleaned up the files from the previous run.   But this is happening in staging-common, and the run_test python script there makes sure to run every test, restarted or not, in a clean, empty subdirectory named for the test.  So, I'd be pretty sure that there isn't an existing BP directory, unless there's something wrong with this bit of Python:
def test_setup(testname):
    testdir = testname + "_testdir"
    if os.path.exists(testdir):
        shutil.rmtree(testdir)
    os.mkdir(testdir)
    os.chdir(testdir)

PR #2213, opened 2020-05-06
Fix mpi4py required

### LINKED ISSUE #2211: Setting ADIOS2_USE_Python=ON doesn't error when mpi4py is not found
**Describe the bug**
If `ADIOS2_USE_Python` is `ON`, and `ADIOS2_USE_MPI` is `ON`, but `mpi4py` is not present, configuration does not error like it should.

**To Reproduce**
1. Make sure Python development files are installed but `mpi4py` is not
2. Configure with `ADIOS2_USE_Python=ON` and `ADIOS2_USE_MPI=ON`
3. Observe that no error occurs

**Expected behavior**
An error should occur, since the required `mpi4py` was not found

**Additional context**
This appears to be due to how `find_package(PythonModule)` works. It handles variables for `PythonModule_<module>` rather than `PythonModule`. So when `REQUIRED` is passed, `PythonModule_FIND_REQUIRED` is set to true, but `find_package_handle_standard_args()` reads `PythonModule_mpi4py_FIND_REQUIRED`, which is not set, and so the error is not triggered.

I believe that the solution is to either change how `find_package(PythonModule)` works, or write a dedicated find module for `mpi4py`.

chuckatkins:
Fixes #2211

### DIFF
--- a/cmake/FindPythonModule.cmake
+++ b/cmake/FindPythonModule.cmake
@@ -121,12 +121,19 @@ set( PythonModule_${module_NAME}_PATH
   endif()
 
   include(FindPackageHandleStandardArgs)
-  set(FPHSA_NAME_MISMATCHED TRUE)
+  set(CMAKE_FIND_PACKAGE_NAME PythonModule_${module_NAME})
+  foreach(VAR IN ITEMS REQUIRED QUIETLY VERSION COMPONENTS)
+    if(DEFINED PythonModule_FIND_${VAR})
+      set(PythonModule_${module_NAME}_FIND_${VAR} "${PythonModule_FIND_${VAR}}")
+    else()
+      unset(PythonModule_${module_NAME}_FIND_${VAR})
+    endif()
+  endforeach()
+
   find_package_handle_standard_args(PythonModule_${module_NAME}
     FOUND_VAR PythonModule_${module_NAME}_FOUND
     REQUIRED_VARS ${required_vars}
   )
-  unset(FPHSA_NAME_MISMATCHED)
 endif()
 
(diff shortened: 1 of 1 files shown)


cmake/FindPythonModule.cmake
```diff
@@ -121,12 +121,19 @@ set( PythonModule_${module_NAME}_PATH
   endif()
 
   include(FindPackageHandleStandardArgs)
-  set(FPHSA_NAME_MISMATCHED TRUE)
+  set(CMAKE_FIND_PACKAGE_NAME PythonModule_${module_NAME})
+  foreach(VAR IN ITEMS REQUIRED QUIETLY VERSION COMPONENTS)
```
KyleFromKitware, line 125: There are other variables as well. Whenever I need to do something like this, I refer to [this list](https://cmake.org/cmake/help/latest/command/find_package.html#id6).

chuckatkins: I went through the list and the individual components are handled differently so those are intentionally skipped.  The remaining are specific version components which should be okay just handling the whole version variable, right?

KyleFromKitware: If it works as is then I guess that's OK.

liangwang0734:
Sorry for referring back to this: how can we build the MPI adios2 with Python but without mpi4py?

dmitry-ganyushin:
Yes, but you need to build ADIOS2 without MPI:
cmake options:  -DADIOS2_USE_MPI=OFF    -DADIOS2_USE_Python=ON
-DPYTHON_EXECUTABLE=/usr/bin/python3

PR #2214, opened 2020-05-06
Python available_variables to allow list of specific keys

williamfgc:
Related to #2196 and #2197 to allow passing a list of case-insensitive input keys to customized variables information.
This PR does not break the current API for `available_variables`.

CC: @khuck @dyokelson

### DIFF
--- a/source/adios2/core/IO.tcc
+++ b/source/adios2/core/IO.tcc
@@ -266,7 +266,7 @@ Params IO::GetVariableInfo(const std::string &variableName,
     }
     else
     {
-        lf_Insert("SingleValue", "true");
+        lf_Insert("SingleValue", "false");
         lf_Insert("Min", helper::ValueToString(variable.Min()));
         lf_Insert("Max", helper::ValueToString(variable.Max()));
     }
--- a/source/adios2/core/IO.tcc
+++ b/source/adios2/core/IO.tcc
@@ -243,32 +243,57 @@ Params IO::GetVariableInfo(const std::string &variableName,
                            const std::set<std::string> &keys)
 {
     Params info;
+    // keys input are case insensitive
     const std::set<std::string> keysLC = helper::LowerCase(keys);
-    // small strings so passing by value
-    auto lf_Insert = [&](const std::string key, const std::string value) {
-        const std::string keyLC = helper::LowerCase(key);
-        if (keys.empty() || keysLC.count(keyLC) == 1)
-        {
-            info[key] = value;
-        }
-    };
+
+    // return empty map if only "name" key is requested
+    if (keys.size() == 1 && keysLC.count("name") == 1)
+    {
+        return info;
+    }
 
     Variable<T> &variable = *InquireVariable<T>(variableName);
 
-    lf_Insert("Type", variable.m_Type);
-    lf_Insert("AvailableStepsCount",
-              helper::ValueToString(variable.m_AvailableStepsCount));
-    lf_Insert("Shape", helper::VectorToCSV(variable.Shape()));
+    if (keys.empty() || keysLC.count("type") == 1)
+    {
+        info["Type"] = variable.m_Type;
+    }
+
+    if (keys.empty() || keysLC.count("availablestepscount") == 1)
+    {
+        info["AvailableStepsCount"] =
+            helper::ValueToString(variable.m_AvailableStepsCount);
+    }
+
+    if (keys.empty() || keysLC.count("shape") == 1)
+    {
+        // expensive function
+        info["Shape"] = helper::VectorToCSV(variable.Shape());
+    }
 
     if (variable.m_SingleValue)
     {
-        lf_Insert("SingleValue", "true");
+        if (keys.empty() || keysLC.count("singlevalue") == 1)
+        {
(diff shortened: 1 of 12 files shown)


source/adios2/core/IO.tcc
```diff
@@ -260,6 +238,41 @@ ADIOS2_FOREACH_STDTYPE_2ARGS(make_GetVariableMap)
...
+            info[key] = value;
+        }
+    };
+
+    Variable<T> &variable = *InquireVariable<T>(variableName);
+
+    lf_Insert("Type", variable.m_Type);
+    lf_Insert("AvailableStepsCount",
+              helper::ValueToString(variable.m_AvailableStepsCount));
+    lf_Insert("Shape", helper::VectorToCSV(variable.Shape()));
```
pnorbert, line 261: variable.Shape() is the call in the original GetAvailableVariables() function that called the engine.BlockInfo<T>() for every variable which went through all the blocks. I don't see how this new function avoids calling that all the time.

williamfgc: @pnorbert there is a conditional inside `lf_Insert`. Each field is populated only when the key exists in keys or if the keys set is empty (current default). The added test in this PR verifies this.

williamfgc: I could add a key for `name` to only return the map with name keys and empty sub-maps values without breaking the API. What do you think?

pnorbert: lf_Insert is too late. You are passing the result of the variable.Shape() as string, so you already parsed the blocks. I think instead of the lambda, an explicit switch case should be written in the main body of the function and only call variable.Shape() if there is a key for it.

williamfgc: I see, good catch. My bad, I missed the input l'll move the conditional outside.

source/adios2/core/IO.tcc
```diff
@@ -260,6 +238,66 @@ ADIOS2_FOREACH_STDTYPE_2ARGS(make_GetVariableMap)
...
+
+    if (keys.empty() || keysLC.count("shape") == 1)
+    {
+        // expensive function
+        info["Shape"] = helper::VectorToCSV(variable.Shape());
+    }
+
+    if (variable.m_SingleValue)
+    {
+        if (keys.empty() || keysLC.count("singlevalue") == 1)
```
pnorbert, line 276: Quick question: wasn't min/max set to the value of the single value previously? With this implementation now you cannot get the actual value with available_variables(), right?

williamfgc: Correct, that's what the `Variable::Min()` and `Variable::Max()` would return, but it was never exposed in the `AvailableVariables` front ends. This implementation shouldn't be changing that behavior to keep APIs backwards compatible, unless I made a mistake.

pnorbert: I think, now even if you ask for min/max or for everything with an empty set, single values will have no entry in the map for min and max. So maybe you need to add min/max in the true branch of that condition.

williamfgc: OK, I see. I guess it won't change current behavior. Let me add that.

williamfgc: @pnorbert should we return Max/Min for single values if the keys are empty? (warning: that would modify the returned dict in `v2.5.0`).

pnorbert: Oh, I now realize we did not give back the actual value in any form in this dictionary before. I think it would be useful to extend then the dictionary with a min/max value for them as well. 

Here is the pixie3d.bp data, an array,  a value over time, and a single string value for the entire 
runtime, with bpls listing:

```
  double   /Car_variables/Bz               11*{2, 33, 33} = 0.999978 / 1.00002
  int32_t  itime                           11*scalar = 0 / 10
  string   visit_expressions               scalar = "V:{<Car_variables/Vx>,<Car_variables/Vy>,<Car_variables/Vz>} ; B:{<Car_variables/Bx>,<Car_variables/By>,<Car_variables/Bz>} ; J:{<Car_variables/Jx>,<Car_variables/Jy>,<Car_variables/Jz>} ; Ve:{<Car_variables/Ve_1>,<Car_variables/Ve_2>,<Car_variables/Ve_3>}"
``` 

In python, right now the dictionary has no info on their value(s) except for the array:

```
>>> v['/Car_variables/Bz']
{'AvailableStepsCount': '11', 'Max': '1.00002', 'Min': '0.999978', 'Shape': '2, 33, 33', 'SingleValue': 'false', 'Type': 'double'}
>>> v['itime']
{'AvailableStepsCount': '11', 'Shape': '', 'SingleValue': 'true', 'Type': 'int32_t'}
>>> v['visit_expressions']
{'AvailableStepsCount': '1', 'Shape': '', 'SingleValue': 'true', 'Type': 'string'}
>>> 
```
 
I think it would be nice to see them for the values as well.

williamfgc: OK, will do. I remember now, this dict predates the Min Max functions that get updated on each step. There was a bug for Value which I then removed as it's not the canonical way (Get is) to get actual global values.

PR #2215, opened 2020-05-07
changed port numbers in dataman tests

JasonRuonanWang:
Required by #2184


KyleFromKitware:
I'm not sure this is enough. Don't the serial and MPI tests still use the same port number?

PR #2216, opened 2020-05-07
Add install components

KyleFromKitware:


chuckatkins:
Why switch to using a separate export set per target in lieu of the current single export set for the project?

KyleFromKitware:
> Why switch to using a separate export set per target in lieu of the current single export set for the project?

This adds the ability to install one set of bindings without installing all of them (for example, the C bindings without the C++ or Fortran bindings). The problem is that the code generated by `install(EXPORT)` checks to see if all the files that it references exist. So if the CMake code is installed without all of the targets, it will error, even if the consuming project didn't actually need the missing targets.

chuckatkins:
@bradking can you please take a look at this?

It seems okay to me but I'm not as well versed in the finer points of target exports.

bradking:
@chuckatkins this LGTM.  The component assignments make sense.  The new way of constructing the legacy `adios2::adios2` imported target based on available components looks good.  I tried installing a subset of components such that the `testing/install/C` project builds manually but `testing/install/CXX11` complains that the `CXX` component cannot be found.  It all works as expected.  Nice job @KyleFromKitware.

One decision we need to make here is whether the MPI parts should have their own installation components.  I don't think this is necessary but just want to point out the option.

KyleFromKitware:
> One decision we need to make here is whether the MPI parts should have their own installation components. I don't think this is necessary but just want to point out the option.

Since the libraries inherently require MPI to be installed, I don't know if there's much value to be added by making the MPI components separate, but I'd be glad to hear a counter-argument.

chuckatkins:
> > One decision we need to make here is whether the MPI parts should have their own installation components. I don't think this is necessary but just want to point out the option.
> 
> Since the libraries inherently require MPI to be installed, I don't know if there's much value to be added by making the MPI components separate, but I'd be glad to hear a counter-argument.

Agreed.  This goes back to the serial components of an MPI build not being appropriate to stand on their own.

PR #2223, opened 2020-05-10
Miscelaneous CI stuff

chuckatkins:


JasonRuonanWang:
@chuckatkins @eisenhauer  It seems that the staging-common ssc tests have never been run since ADIOS2_RUN_MPMD_TESTS was introduced in this PR.

eisenhauer:
Well, that's not good...

JasonRuonanWang:
I am not 100% sure whether they have been run in CI, but they definitely do not run any more on all of my local machines. They used to run on my local machines, which I remember very clearly. I am a little surprised too that I only noticed it today, after several months...

PR #2225, opened 2020-05-11
Handle pending releases when closing writer

philip-davis:
Perform normal release operations on queued Release notices from the readers at writer close, rather than just dumping them. The push-based RDMA data plane needs this because it needs to perform data transfer operations in response to release messages.


eisenhauer:
@philip-davis Would you please evaluate the attached patch as a simpler alternative to this?  Basically it keeps the existing close() mechanisms in place, but instead modifies ReleaseAndDiscardRemainingTimesteps() (which didn't do reader releases).

This does have some consequences in that the reader releases will be done more all-at-once in the shutdown process, rather than as the releases come in to rank 0 and are distributed to the other ranks.  Not sure if that makes a difference for RDMA DP.

[patch.txt](https://github.com/ornladios/ADIOS2/files/4616120/patch.txt)

PR #2235, opened 2020-05-13
Forbid transition from speculative to learned preload

eisenhauer:
Fixes #2234

PR #2236, opened 2020-05-13
Do DP Timestep Release on all reader ranks

philip-davis:
Move DP-level timestep release on the reader to SstReleaseStep, instead of FreeTimestep. This will be called unconditionally when CommMin is being used.


philip-davis:
Okay, I made a commit where I comment out centos8 in .travis.yaml, but I'll wait to push it until I can pull down the merged #2235 changes

PR #2237, opened 2020-05-13
ci: Workaround stale rpmdb lock files in CentOS 8 base image

chuckatkins:


chuckatkins:
Wait for upstream to resolve the issue in the base image

Issue #2233, opened 2020-05-12
bindings/Python: Cannot initialize serial ADIOS if MPI is enabled


dmitry-ganyushin:
Hello,

It looks like adios initialization in python  does not work with a parameter.
From examples/hello/bpWriter/helloBPWriter_nompi.py

adios = adios2.ADIOS(adios2.DebugON)
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
mpi4py/MPI/CAPI.pxi in mpi4py.MPI.PyMPIComm_Get()

TypeError: Cannot convert bool to mpi4py.MPI.Comm

The above exception was the direct cause of the following exception:

SystemError                               Traceback (most recent call last)
<ipython-input-6-f6f117954d62> in <module>()
----> 1 adios = adios2.ADIOS(adios2.DebugON)

SystemError: <built-in method __init__ of PyCapsule object at 0x7f0cfa9bfcc0> returned a result with an error set

godoywf:
@dmitry-ganyushin  that's the signature for the non-mpi build.  I'd check with @bradking as he reworked the entire MPI/non-MPI toolchain. Hope it helps.

chuckatkins:
@dmitry-ganyushin the debug parameters have been deprecated as they are no longer used internally.  The intent was to keep backwards compatible interfaces that silently dropped the argument but it looks like we may have missed that for the python bindings.  So it looks like we need to 1) add the backwards compatible interface and 2) fix the python examples to not use the deprecated function signature.

bradking:
The `examples/hello/bpWriter/helloBPWriter_nompi.py` example is named `_nompi` because it is meant for use in the case that ADIOS2 was built without MPI enabled.  If ADIOS2 is built with `ADIOS2_USE_MPI` set to `OFF` then the example works:

```python
adios = adios2.ADIOS(adios2.DebugON)
```

If ADIOS2 is built with `ADIOS2_USE_MPI` set to `ON` then the `examples/hello/bpWriter/helloBPWriter.py` example works instead:

```python
comm = MPI.COMM_WORLD
adios = adios2.ADIOS(comm, adios2.DebugON)
```

Each of those  bpWriter examples says in comments at the top "only works with MPI version" or "only works with non MPI version", though it looks like those are backwards.

That said, I've made a bunch of changes recently to make serial programs work with a MPI-enabled ADIOS2.  This issue demonstrates that the Python bindings still haven't been fixed for that.

PR #2238, opened 2020-05-13
bindings: Fix Python serial program init when ADIOS2 has MPI enabled

bradking:
The one-argument `adios2.ADIOS(...)` constructor has two overloads when MPI is enabled: `ADIOS(bool)` and `ADIOS(MPI4PY_Comm)`.  The representation of `MPI4PY_Comm` is just an integer, so a call with a `bool` may try to convert it (and fail).  Re-order the pybind11 definitions of these signatures so that the `ADIOS(bool)` is tried first.  That way a call with a `bool` will use it and a call with a `MPI4PY_Comm` will not convert and fall back to the proper variant.

Also update the test suite to always add the Serial variant of the tests and additionally enable the MPI variants when possible.  The `Bindings.Python.BPWriteReadTypes.Serial` test in particular covers the overload selection problem fixed here.

Fixes: #2233

### DIFF
--- a/bindings/Python/py11glue.cpp
+++ b/bindings/Python/py11glue.cpp
@@ -229,6 +229,15 @@ PYBIND11_MODULE(ADIOS2_PYTHON_MODULE_NAME, m)
                  const bool opBool = adios ? true : false;
                  return opBool;
              })
+        .def(pybind11::init<const bool>(),
+             "adios2 module starting point "
+             "non-MPI, constructs an ADIOS class "
+             "object",
+             pybind11::arg("debugMode") = true)
+        .def(pybind11::init<const std::string &, const bool>(),
+             "adios2 module starting point non-MPI, constructs an ADIOS class "
+             "object",
+             pybind11::arg("configFile"), pybind11::arg("debugMode") = true)
 #if ADIOS2_USE_MPI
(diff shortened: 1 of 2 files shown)


bindings/Python/py11glue.cpp
```diff
@@ -229,6 +229,15 @@ PYBIND11_MODULE(adios2, m)
                  const bool opBool = adios ? true : false;
                  return opBool;
              })
+        .def(pybind11::init<const bool>(),
+             "adios2 module starting point "
+             "non-MPI, constructs an ADIOS class "
+             "object",
+             pybind11::arg("debugMode") = true)
```
chuckatkins, line 236: Even though the debug parameter is effectively ignored, should this default to `false` instead of `true`?

chuckatkins: I guest it doesn't really matter since it's not actually used.

bradking: I thought about that but decided to keep the change to a minimum.  The old code defaulted to `true` already.

bradking:
This is blocked in CI by the problem discussed [here](https://github.com/ornladios/ADIOS2/pull/2253#issuecomment-628748135).

chuckatkins:
Debian builds disabled. Please rebase.

PR #2246, opened 2020-05-14
Added reliable mode in dataman which ensures readers receive all steps

JasonRuonanWang:


chuckatkins:
Might I make a suggestion that a different name from "secure" be used? With respect to communication and networking, the term "secure" has some implications carrying with it an expectation of some sort of encryption, authentication, signature validation, etc., which isn't the intent here.  It's much more akin to the difference in UDP vs TCP, where the term "reliability" or "reliable" vs "unreliable" is generally used to convey every message being received or only the most recent.  Maybe `fast` and `reliable` as the options?

JasonRuonanWang:
> Might I make a suggestion that a different name from "secure" be used? With respect to communication and networking, the term "secure" has some implications carrying with it an expectation of some sort of encryption, authentication, signature validation, etc., which isn't the intent here. It's much more akin to the difference in UDP vs TCP, where the term "reliability" or "reliable" vs "unreliable" is generally used to convey every message being received or only the most recent. Maybe `fast` and `reliable` as the options?

I have changed it to `reliable`

PR #2253, opened 2020-05-14
Enable test repeat in Debian builds

KyleFromKitware:


chuckatkins:
It seems like using the rolling release is inherently unstable and possibly not suitable for ci.  Suggestions?

chuckatkins:
You could put docker images in
`scripts/ci/images/debian-sid{,-openmpi}` with the dependencies already installed and then have docker hub rebuild them nightly.  That way if the image rebuild fails they wouldn't get updated but we'd all was be building against a most recent known-good configuration.

KyleFromKitware:
Yeah, I saw this issue working with Sid earlier.

We could switch to Debian Testing, which is also rolling release but more likely to be stable, because only packages that last in Sid for 10 days without bugs make it to Testing. How does that sound?

chuckatkins:
It's still an unstable base image when the code being tested is already a moving target.  Even with the testing image you can still have a broken configuration.  I'd rather have a known working base image we always use that's automatically updated.

KyleFromKitware:
OK, how do I get Dockerhub to rebuild the image nightly?

chuckatkins:
You might need admin access on the dockerhub repo.  If you get the image built and pushed with a dockerfile in this repo then I can setup the automatic rebuild separately.

KyleFromKitware:
Unfortunately, I can't build an image right now due to the Python issues in Sid. :( I've reported it to Debian but am still waiting for it to be fixed. I think for now the best thing would be to temporarily disable the Debian builds until I can get a better image uploaded.

KyleFromKitware:
Closing in favor of #2260.

PR #2255, opened 2020-05-14
Backward compatibility

pnorbert:


pnorbert:
I have added a few BP3 and BP4 datasets for this backward compatibility tests. They are small so far so I guess they are okay to store here for now.

chuckatkins:
Even if they're small it's still a bunch of binary data I'd like to avoid and not have in the git repo.  It's probably time to add git-lfs into the mix for this sort of thing.  If it can wait until after the release we can add the infrastructure to the build and test workflow.  I don't want to start it now since it could take a fair amount of iteration to get right.

However if you think we need it now before the release then I suppose this is okay in a limited capacity so long as we don't add any more.  We can then move these to lfs after when we make the shift along with additional test data sets, although these will forever be in the git history.

I'll defer to you on that decision.

chuckatkins:
Related #2257

williamfgc:
@chuckatkins do you think reusing [ADIOS2-Testing](https://github.com/ornladios/ADIOS2-Testing) to run nightly with pre-stored files makes sense? I'd be interested to contribute to more tests (which will increase the size) as they are really important for sustainability. Thanks @pnorbert for adding the tests, they are spot on.

pnorbert:
These tests can move to another repo entirely if we can make sure they are executed at least before every release. 

It is not urgent for this release, as it only confirms we are backward compatible (for these few files anyway)

chuckatkins:
> These tests can move to another repo entirely if we can make sure they are executed at least before every release.

I do think they should stay in this repo and be run wilth all of our PR testing.

> It is not urgent for this release, as it only confirms we are backward compatible (for these few files anyway)
If that's the case then I'd like to defer merging this and have the binary data use git-lfs via whatever we come up with for #2257

NAThompson:
Do we have a paid github account? I got hit with a charge at a 1GB using git-lfs with github.

chuckatkins:
> Do we have a paid github account? I got hit with a charge at a 1GB using git-lfs with github.

We do not and the transfer limits were a problematic issue with using git-lfs in the past.  What we'll do for adios2 is use a different host for the lfs repo (gitlab.kitware.com or code.ornl.gov if we can get it appropriately public) instead of the github lfs repo.

pnorbert:
@chuckatkins As we are inching towards the next release, it is time to revisit this PR.

eisenhauer:
Given #3273, I agree that it's (probably past) time to revisit this.  Raising triage level.

eisenhauer:
Actually looking at this again...

PR #2256, opened 2020-05-14
RDMA DP Preload

philip-davis:
Push mode for SST RDMA dataplane.

Enabled by 
1) Setting SST_DP_PRELOAD to 1
2) Explicitly locking definitions on both writer and reader

Creates two mailbox buffers (active and queue) on the readers. When a timestep is released, the writer pushes a new one, if available. Mailboxes switch between active and queue.

Known issues:
  - Not compatible with transformations (reduction) since that changes the write geometry of the data buffer.


eisenhauer:
Rebase on master after #2254 is merged...

chuckatkins:
Should this be targeting master?

PR #2258, opened 2020-05-14
Specify adios2_DIR when generating adios2-config

KyleFromKitware:


chuckatkins:
I assume this is to accommodate nonstandard prefixes for the CMake files?  If that's the case, the script is currently setting the `adios2_DIR` environment variable which guides `find_package` but does so with an install prefix instead of the CMake dir.  Passing this as a CMake variable overrides that so it should really just be one or the other.  I.e. change the env var to use the CMake dir instead of prefix and drop the CMake var or keep the CMake var and drop the env var.

KyleFromKitware:
> I assume this is to accommodate nonstandard prefixes for the CMake files?

Correct.

> the script is currently setting the `adios2_DIR` environment variable which guides `find_package` but does so with an install prefix instead of the CMake dir.

Sounds like that is more akin to `adios2_ROOT` than `adios2_DIR`.

> change the env var to use the CMake dir instead of prefix and drop the CMake var or keep the CMake var and drop the env var.

I think we could simply rename the environment variable `adios2_DIR` to be `prefix` or something like that, and not `export` it.

chuckatkins:
Both `adios2_DIR` and `adios2_ROOT` vars are used by `find_package`, albeit differently.  `_ROOT` works as either an env car or CMake var for both module and config mode as a search prefix; `_DIR` only works for config mode just for the package config files and as an env var both a search prefix and a full config path or as a CMake var as just a full config path.  The current script is already using `adios2_DIR` as an env var with the prefix so for this you should be able to specify the full config path instead.

Really I don't care which one of the myriad of search hints gets used here, I just don't want different ones overriding each other.

KyleFromKitware:
> as an env var both a search prefix and a full config path

TIL... https://cmake.org/cmake/help/latest/command/find_package.html#search-procedure

That does not seem intuitive to me. I thought `<package>_DIR` was exclusively for the package config file, and not a prefix, and that `<package>_ROOT` was supposed to fill the role of prefix. Thanks.

chuckatkins:
> That does not seem intuitive to me. I thought `<package>_DIR` was exclusively for the package config file, and not a prefix, and that `<package>_ROOT` was supposed to fill the role of prefix. 

Welcome to the club.  It's not intuitive for anyone really.  We discussed expanding the role of `<package>_DIR` but given that there was so much history and existing complexity behind its' use we opted to not mess with it and instead introduce `<package>_ROOT`.  Really for most use cases `<pacakge>_ROOT` takes the place of `<package>_DIR` and more with the exception of specifying the full config path (which is what you need here).

PR #2267, opened 2020-05-18
Go back using active flag in metadata index header instead of using a…

pnorbert:
… separate 'active' file, which introduced race conditions. A separate file cannot guarantee the order of the following events: removal of active file (file system's metadata server operation), and arrival of content of metadata index file on disk (a storage operation). Index content may be en route to disk but held up to indefinite time while the active file disappears, so a reader may not get the last steps believing the file is not active anymore. In contrast, active flag is written to the same file after the last steps are written, so order is guaranteed as long as the file system guarantees the order of writes to the same file.

### DIFF
--- a/source/adios2/engine/bp4/BP4Reader.cpp
+++ b/source/adios2/engine/bp4/BP4Reader.cpp
@@ -245,13 +245,15 @@ void BP4Reader::OpenFiles(const TimePoint &timeoutInstant,
         if (m_BP4Deserializer.m_RankMPI == 0)
         {
             throw std::ios_base::failure(
-                "ERROR: File " + m_Name +
-                " could not be found within timeout: " + lasterrmsg);
+                "ERROR: File " + m_Name + " could not be found within the " +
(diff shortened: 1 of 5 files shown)


chuckatkins:
Should this be targeting the release branch?

pnorbert:
@chuckatkins This docs test is new, isn't it? What can I do here to pass?

pnorbert:
Okay, I rebased current master to get the latest docs environment. 

@chuckatkins This fix should go to release. Should I make another PR to the release branch?

chuckatkins:
> > @chuckatkins This docs test is new, isn't it? What can I do here to pass?
> Okay, I rebased current master to get the latest docs environment.

Yes the docs test is new, It's a hidden feature in ReadTheDocs that I had their admins enable for our project.  They are now passing on master and release.

> @chuckatkins This fix should go to release. Should I make another PR to the release branch?

No need.  You can "edit" the pr to change the target branch, which I just did.

PR #2271, opened 2020-05-19
reduced data steps in dataman tests in order to prevent timeout

JasonRuonanWang:
A random timeout error was seen in CI when running dataman reliable test. This PR tries to prevent similar errors from happening again.


chuckatkins:
The tests that are passing are doing so in ~20s so I don't think it's the number of steps

JasonRuonanWang:
> The tests that are passing are doing so in ~20s so I don't think it's the number of steps

OK. I am still looking into other things. But these tests do not need so many steps anyway. I will just merge this and open another PR if I find the problem.

PR #2274, opened 2020-05-20
Add -F option to iotest for Fixed IO patterns. Before the first EndSt…

pnorbert:
Add -F option to iotest for Fixed IO patterns. Before the first EndStep, writer definitions or reader selections will be locked in each engine. Iotest has only fixed IO patterns anyway, this option only makes the calls to engine.Lock*() functions. 

Engines that optimize its work for fixed IO patterns can use this explicit calls from the application to turn on those optimizations.


pnorbert:
@chuckatkins Can you take a look at the docker-centos build in travis? I has failed three times this morning. Thanks.

chuckatkins:
I see the problem.  I'll get a fix up asap

chuckatkins:
See #2275 for travis fix

PR #2276, opened 2020-05-20
Add Debian Sid package build

KyleFromKitware:


KyleFromKitware:
@eisenhauer Any insights on [this](https://open.cdash.org/testDetails.php?test=885436061&build=6543473) failure?

eisenhauer:
> @eisenhauer Any insights on [this](https://open.cdash.org/testDetails.php?test=885436061&build=6543473) failure?

Quite odd that it's happening.  It's a failure in FFS dynamic code generation, which is turned off by default for ADIOS in thirdparty/ffs/CMakeLists.txt, mostly because we saw occasional failures like this and hadn't successfully tracked them down.  So I think the question is, "why is this functionality turned on?".

KyleFromKitware:
>  So I think the question is, "why is this functionality turned on?".

Because I built the entire GTkorvo stack separately and am testing against it. This gives me some insight, thanks.

KyleFromKitware:
>  It's a failure in FFS dynamic code generation, which is turned off by default for ADIOS in thirdparty/ffs/CMakeLists.txt

You mean this:

```cmake
set(FFS_CONVERSION_GENERATION_DEFAULT OFF CACHE INTERNAL "")
```

?

eisenhauer:
Yes.

KyleFromKitware:
OK, I've disabled dynamic code generation in the ffs Debian package. The tests appear to be passing now.

KyleFromKitware:
@chuckatkins Please take a look at [this build](https://open.cdash.org/buildSummary.php?buildid=6552994) and [this build](https://open.cdash.org/buildSummary.php?buildid=6552993). This is the result of running CTest inside `dpkg-buildpackage` and submitting the results to CDash. Right now, the Debian package is only running the `Install` tests (I'm still trying to figure out why `OMPI_ALLOW_RUN_AS_ROOT` isn't working, but that is an orthogonal issue.) However, I believe it may be possible to run the entire test suite from inside the package build, thus making the more traditional Debian builds redundant, and allowing them to be removed in favor of the package build. Would you have any objection to this?

PR #2288, opened 2020-05-26
Handle race condition in BP4Reader

pnorbert:
When checking data in UpdateBuffer, new step(s) may have not arrived yet. When checking active flag next, the writer may have completed write and terminated. So we may have missed a step or two. A final check on the index avoids this race condition. 

Left some debug print and value change to still make it fail in Debug builds to see in if this was actually the race condition seen in CI. This needs to be removed for the release.

### DIFF
--- a/source/adios2/engine/bp4/BP4Reader.cpp
+++ b/source/adios2/engine/bp4/BP4Reader.cpp
@@ -557,7 +557,25 @@ StepStatus BP4Reader::CheckForNewSteps(Seconds timeoutSeconds)
         }
         if (!CheckWriterActive())
         {
+            /* Race condition: When checking data in UpdateBuffer, new step(s)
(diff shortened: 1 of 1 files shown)


pnorbert:
Indeed, this was the race condition showing up in sanitizer builds
```
438: **** ADIOS2:BP4Reader::CheckForNewSteps race condition handled, found 3264 bytes of index after writer has finished ****
438: Reader Open took 4.078497502 seconds
438: Reader BeginStep t = 0 had status = StepStatus::EndOfStream after 0.331857877 seconds
438: /root/project/source/testing/adios2/engine/staging-common/TestCommonRead.cpp:390: Failure
438: Expected equality of these values:
438:   t
438:     Which is: 0
438:   NSteps
438:     Which is: 5
438: MemorySanitizer:DEADLYSIGNAL
```

PR #2293, opened 2020-05-27
add nwchem-sorting test with cheetah

lxAltria:


pnorbert:
This is going to the wrong repository. Need to go to ADIOS2-Testing
https://github.com/ornladios/ADIOS2-Testing
under performance/nwchem-sorting

Issue #2287, opened 2020-05-26
Build error on Theta


jychoi-hpc:
**Describe the bug**
I am getting the following error on Theta with intel and cmake/3.16.2.

```
[20200526:145202] -- Configuring done
[20200526:145202] CMake Error at CMakeLists.txt:27 (add_executable):
[20200526:145202]   Target "serial_with_Fortran" links to target
[20200526:145202]   "adios2::thirdparty::pugixml_impl" but the target was not found.  Perhaps a
[20200526:145202]   find_package() call is missing for an IMPORTED target, or an ALIAS target
[20200526:145202]   is missing?
[20200526:145202] 
[20200526:145202] 
[20200526:145202] CMake Error at CMakeLists.txt:27 (add_executable):
[20200526:145202]   Target "serial_with_Fortran" links to target "adios2::thirdparty::yaml-cpp"
[20200526:145202]   but the target was not found.  Perhaps a find_package() call is missing for
[20200526:145202]   an IMPORTED target, or an ALIAS target is missing?
[20200526:145202] 
[20200526:145202] 
[20200526:145202] CMake Error at CMakeLists.txt:21 (add_executable):
[20200526:145202]   Target "serial_with_CXX" links to target "adios2::thirdparty::pugixml_impl"
[20200526:145202]   but the target was not found.  Perhaps a find_package() call is missing for
[20200526:145202]   an IMPORTED target, or an ALIAS target is missing?
[20200526:145202] 
[20200526:145202] 
[20200526:145202] CMake Error at CMakeLists.txt:21 (add_executable):
[20200526:145202]   Target "serial_with_CXX" links to target "adios2::thirdparty::yaml-cpp" but
[20200526:145202]   the target was not found.  Perhaps a find_package() call is missing for an
[20200526:145202]   IMPORTED target, or an ALIAS target is missing?
[20200526:145202] 
[20200526:145202] 
[20200526:145202] CMake Error at CMakeLists.txt:21 (add_executable):
[20200526:145202]   Target "serial_with_CXX" links to target "adios2::thirdparty::pugixml" but
[20200526:145202]   the target was not found.  Perhaps a find_package() call is missing for an
[20200526:145202]   IMPORTED target, or an ALIAS target is missing?
[20200526:145202] 
[20200526:145202] 
[20200526:145202] CMake Error at CMakeLists.txt:16 (add_executable):
[...]

PR #2294, opened 2020-05-27
Fix install of static build

KyleFromKitware:
Fixes: #2287


KyleFromKitware:
Local testing confirms that this works with both internal and external dependencies.

chuckatkins:
I believe #2298 fixes the asan issue

chuckatkins:
> I believe #2298 fixes the asan issue

So, while the flags I added ion that pr should fix the issue, some tinkering needs to be done to deal with clangs argument parsing and duplicate flags.  So I'm turning off the asan build for now so we can continue with forward progress until it's resolved.

PR #2298, opened 2020-05-29
ASAN: ODR workaround

chuckatkins:
See https://github.com/google/sanitizers/issues/1017


NAThompson:
@chuckatkins : It looks like this PR is an improvement over what currently exists, even if the build is not totally green.

Is it sensible to merge it anyway?

chuckatkins:
> @chuckatkins : It looks like this PR is an improvement over what currently exists, even if the build is not totally green.
> 
> Is it sensible to merge it anyway?

Certainly no.  The additional options cause build failures which will block future ci.  The duplicate options error is an oddity of llvm but has to be fixed before merging.

Tldr; no green = no merge

chuckatkins:
@NAThompson Looks like an update to google tests cleared out the llvm argument parse error.  The most recent builds were clean, including the re-enabling of asan, only failing some formatting checks: https://open.cdash.org/index.php?compare1=61&filtercount=1&field1=revision&project=ADIOS&showfilters=0&limit=100&value1=2bc45fe8bd4fd514d8671d12a5a6a45745991c9a&showfeed=0 

After this most recent push cycles through it should be good to merge and we'll have asan re-enabled again.

NAThompson:
@chuckatkins : Just read through these diffs; looks good. I'm guessing you had to update googletest in this MR to get everything clean?

chuckatkins:
>> @NAThompson Looks like an update to google tests cleared out the llvm argument parse error.

> @chuckatkins ... I'm guessing you had to update googletest in this MR to get everything clean?

Yeah, their were some CMake issues in the version of googletest we had resulting in duplicate copies of arguments getting passed to the compiler which were fixed by updating googletest.  That in turn brought along with it some necessary updates for deprecated test macros.

NAThompson:
Ok, I just built this PR and all the tests are green:

```
build$ ctest
....
483/483 Test #483: Install.Make.CXX11 ............................................................................   Passed    1.63 sec

100% tests passed, 0 tests failed out of 483
```

Also it looks like your got all the deprecations out:

```
ADIOS2$ grep -nr 'INSTANTIATE_TEST_CASE_P' .
..//thirdparty/GTest/googletest/googletest/include/gtest/internal/gtest-internal.h:1237:    "INSTANTIATE_TEST_CASE_P is deprecated, please use "
..//thirdparty/GTest/googletest/googletest/include/gtest/gtest-param-test.h:504:#define INSTANTIATE_TEST_CASE_P
```

PR #2302, opened 2020-06-01
Reader-side DP notify fix

eisenhauer:


eisenhauer:
The only failures here are in BP4 streaming, unrelated to the actual patch here and the problem is known to @pnorbert.  The failing build isn't required for merger, so I'm going to go ahead and merge.  However, @chuckatkins , we're getting more failures on debian-sid-openmpi than I would expect if the REPEAT UNTIL:PASS functionality were enabled there.  I don't see an obvious cause for it being disabled, unless maybe the cmake is too old?

PR #2305, opened 2020-06-02
Disable float checking for now, as it seems to break in Homebrew...

KyleFromKitware:
...and isn't being used anyway.

### DIFF
--- a/cmake/DetectOptions.cmake
+++ b/cmake/DetectOptions.cmake
@@ -354,15 +354,15 @@ find_package(Threads REQUIRED)
 # Floating point detection
 include(CheckTypeRepresentation)
 
-check_float_type_representation(float FLOAT_TYPE_C)
-check_float_type_representation(double DOUBLE_TYPE_C)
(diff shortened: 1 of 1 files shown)


KyleFromKitware:
Cc: @ax3l

ax3l:
re: https://github.com/Homebrew/homebrew-core/pull/55556

KyleFromKitware:
As for the failure itself, the first thing that occurs to me is that this code is known not to work right in multi-arch OSX builds, but that would not yield the same error as what you're seeing. I'll have to take a closer look when we start actually using this functionality.

ax3l:
I am surprised as well, because the GitHub runner is just a plain `macOS: 10.13.6-x86_64` and it works on 10.14 and 10.15.

KyleFromKitware:
I don't think Xcode is the problem. Basically the way this code works is it compiles a float into a binary surrounded by some marker bytes that show CMake where to find the float. The error you found indicates that those bytes are not present, or are malformed... which is very surprising to me if it's a plain old macOS build.

PR #2310, opened 2020-06-04
Convert internal data type ids from std::string to an enumeration

bradking:
This improves type safety and reduces the runtime cost of comparisons.

The public bindings expose variable and attribute data types using string identifiers.  Preserve that by converting between the string and enumeration representations in the binding layer.

### DIFF
--- a/bindings/Python/py11IO.cpp
+++ b/bindings/Python/py11IO.cpp
@@ -111,10 +111,10 @@ Variable IO::InquireVariable(const std::string &name)
     helper::CheckForNullptr(m_IO, "for variable " + name +
                                       ", in call to IO::InquireVariable");
 
-    const std::string type(m_IO->InquireVariableType(name));
+    const Type type(m_IO->InquireVariableType(name));
     core::VariableBase *variable = nullptr;
 
-    if (type == "unknown")
+    if (type == Type::None)
     {
     }
 #define declare_template_instantiation(T)                                      \
@@ -193,9 +193,9 @@ Attribute IO::InquireAttribute(const std::string &name)
                                       ", in call to IO::InquireAttribute");
 
     core::AttributeBase *attribute = nullptr;
-    const std::string type(m_IO->InquireAttributeType(name));
+    const Type type(m_IO->InquireAttributeType(name));
 
-    if (type == "unknown")
+    if (type == Type::None)
     {
     }
 #define declare_template_instantiation(T)                                      \
@@ -264,14 +264,14 @@ std::string IO::VariableType(const std::string &name) const
 {
     helper::CheckForNullptr(m_IO, "for variable " + name +
                                       " in call to IO::VariableType");
-    return m_IO->InquireVariableType(name);
+    return ToString(m_IO->InquireVariableType(name));
 }
 
 std::string IO::AttributeType(const std::string &name) const
 {
     helper::CheckForNullptr(m_IO, "for attribute " + name +
                                       " in call to IO::AttributeType");
-    return m_IO->InquireAttributeType(name);
+    return ToString(m_IO->InquireAttributeType(name));
 }
 
 std::string IO::EngineType() const
--- a/source/utils/bpls/bpls.h
+++ b/source/utils/bpls/bpls.h
@@ -7,6 +7,7 @@
 
 #include "adios2/common/ADIOSConfig.h"
 #include "adios2/common/ADIOSMacros.h"
+#include "adios2/common/ADIOSTypes.h"
 #include "adios2/core/ADIOS.h"
 #include "adios2/core/Engine.h"
 #include "adios2/core/IO.h"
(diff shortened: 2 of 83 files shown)


source/utils/bpls/bpls.h
```diff
@@ -34,34 +35,12 @@ namespace utils
...
-    adios_unsigned_long = 54,
-    adios_real = 5,
-    adios_double = 6,
-    adios_long_double = 7,
-    adios_string = 9,
-    adios_complex = 10,
-    adios_double_complex = 11,
-    adios_string_array = 12,
-    adios_long_double_complex = 13
-};
```
bradking, line 57: @chuckatkins do you know why these had assigned values?

bindings/Python/py11IO.cpp
```diff
@@ -111,10 +111,10 @@ Variable IO::InquireVariable(const std::string &name)
     helper::CheckForNullptr(m_IO, "for variable " + name +
                                       ", in call to IO::InquireVariable");
 
-    const std::string type(m_IO->InquireVariableType(name));
+    const Type type(m_IO->InquireVariableType(name));
     core::VariableBase *variable = nullptr;
 
-    if (type == "unknown")
+    if (type == Type::None)
```
bradking, line 117: @bradking most places used an empty string for "no type", but a few used `"unknown"`.  I've converted both cases to `Type::None`.  Do you know why there was a distinction before?

source/utils/bpls/bpls.h
chuckatkins: They were initially put in thinking we'd implement them but it didn't really happen.  @KyleFromKitware has the long double support on his list but there's some other things to get to first so it will be a while.  Safe to remove.

bindings/Python/py11IO.cpp
chuckatkins: No idea.  @eisenhauer @pnorbert ?

source/utils/bpls/bpls.h
pnorbert: These are the actual byte values of encoding these types in the BP file format. 
I think I just used for something in bpls or wanted to use it. I wouldn't say it is safe to remove, there are some functions in bpls to translate between adios types and adios types (sic!)

bindings/Python/py11IO.cpp
pnorbert: Only @williamfgc  might know, but I think unknown came from adios1 where this (invalid) type value existed. I don't remember if it was ever used there other than initializing structs. I assume this is safe to get rid of as there should not be any variable object existing with unknown type in adios2.

source/utils/bpls/bpls.h
chuckatkins: My mistake then.  Leave them.

bradking: Currently this PR switches bpls from `enum ADIOS_DATATYPES` to using the new `adios2::Type` enumeration directly, so `type_to_enum` is no longer needed.  Unless `bpls` uses the actual integer values that `ADIOS_DATATYPES` had for something, why shouldn't they be removed?

Or, should I restore `enum ADIOS_DATATYPES` as it was and make `type_to_enum` convert between the two different enumerations?

pnorbert: Okay, sorry for the flame. You cleaned up bpls nicely with the new Type class. So it's okay to remove. The only places where these values count in adios2 are the engine serializers and deserializers. 

I just added this enum to bpls when adding the old bpls C program to ADIOS2 with the least amount of work (a signature of my attitude). Thanks for cleaning this up.

bindings/Python/py11IO.cpp
bradking: The new `GetTypeFromString` conversion function that maps from the old string values (still used in the public API) to the new `adios2::Type` enumeration will convert any unknown string value to `Type::None`, so if we somehow get an `"unknown"` value it will become `Type::None` and match these conditions as before.

source/utils/bpls/bpls.h
bradking: Great, thanks for explaining!

bindings/Python/py11IO.cpp
williamfgc: @pnorbert is right, at some point we tried to keep it open for supporting POD structs, but it was always pushed in the list of priorities.

williamfgc: Also, if I recall correctly the design decision at that time for using string was related to POD types as we're exploring ways of passing something generic/fixed like "struct" (POD) and something the user could set at runtime "Points3D".  Also, most string types are <= 15 characters and they are put on the stack for small string optimization (SSO). Using SSO was the recommendation from my previous employer as cost comparisons in the L-* cache stack in highly vectorized architectures are essentially undetectable. All good intentions, but POD types we not a priority in a HPC production library as there are added cost associated to offsets and memory alignment so we never went further. Hope it helps. Thanks for cleaning this.

bradking: Okay, thanks for explaining!

chuckatkins:
@pnorbert do we have any "lot's and lots of variables" benchmarks, maybe `iotest` profiles, that we could maybe see some of the potential performance impact of the reduction in string operations?

germasch:
FWIW, a thumbs up for me for this change.

Also, while kinda off-topic, I've heard something about using abseil and
`std::variant`, so I wanted to mention that about a year ago, I had
converted a lot of adios2 to using mpark's std::variant, which does
significantly simplify lots of code paths and in particular can get rid of
the FOR_EACH macro business. That mostly eliminates the need for even an
integer type, as instead the variant discriminates its content. Anyway, I'd
be more than happy to dust off my previous work, though probably not until
July as this month looks rather busy for me.

bradking:
@germasch this work was my first step toward working on avoiding the `FOR_EACH` macros.  Please push whatever you have to your fork and I'll fetch from there and take a look.

PR #2318, opened 2020-06-10
Rename internal data type enumeration from Type to DataType

bradking:
The enumeration added by #2310 is named `Type`, but `DataType` more precisely describes its purpose.

### DIFF
--- a/bindings/C/adios2/c/adios2_c_attribute.cpp
+++ b/bindings/C/adios2/c/adios2_c_attribute.cpp
@@ -49,12 +49,12 @@ adios2_error adios2_attribute_type(adios2_type *type,
             reinterpret_cast<const adios2::core::AttributeBase *>(attribute);
 
         auto type_s = attributeBase->m_Type;
-        if (type_s == adios2::helper::GetType<std::string>())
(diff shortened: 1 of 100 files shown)


bradking:
CI is happy here except for timeouts on `macos macos1015_xcode111_make` which occurred repeatedly.

bradking:
I've opened #2322 specifically to fix the new MSVC warning with no other changes.  Once that is merged this can be rebased.

ax3l:
I saw that merging this PR two days ago broke our nightly integration compiling against ADIOS2 master. We use `adios2::GetType<...>` which we thought is public API and is now renamed, what do you recommend? :)

Ref.: https://dev.azure.com/axelhuebl/openPMD-api/_build/results?buildId=392&view=results

cc @franzpoeschel

bradking:
@ax3l thanks.  I had not intended to change the public-facing name.  I've reverted that part in #2342.

pnorbert:
Thanks. It broke E3SM's adios driver too. There are functions there to map between pnetcdf and adios2 types. 

However, in the future it may be useful for exposing the DataType publicly as well. It would make coding cleaner in my opinion.

PR #2322, opened 2020-06-11
utils: Avoid unnamed struct in adios_reorganize

bradking:
MSVC now warns:

    warning C5208: unnamed class used in typedef name cannot declare
    members other than non-static data members, member enumerations,
    or member classes

### DIFF
--- a/source/utils/adios_reorganize/Reorganize.h
+++ b/source/utils/adios_reorganize/Reorganize.h
@@ -20,7 +20,7 @@ namespace adios2
 namespace utils
 {
 
-typedef struct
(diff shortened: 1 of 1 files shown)


chuckatkins:
There seems to be a consistent problem with the Mac builds getting deprovisioned no matter how many times I rerun them.  I've reported the issue upstream: https://developercommunity.visualstudio.com/content/problem/1075684/mac-builds-keep-getting-cancelled-after-20-30m.html

chuckatkins:
Replaced by #2331

PR #2324, opened 2020-06-11
[SSC] Hide data transfer from application activities

JasonRuonanWang:
Previously, SSC writers block EndStep for sending data. This PR moves MPI_Wait and MPI_Win_fence from EndStep to BeginStep. Data transfers can then be overlapped with application algorithms that happen between EndStep(N) and BeginStep(N+1).


chuckatkins:
@JasonRuonanWang  you should be able to rebase the ci failures away now.

PR #2325, opened 2020-06-11
Tweaks and test for SST engines in threads

eisenhauer:


chuckatkins:
@eisenhauer see #2332 for sanitizer help

eisenhauer:
Thanks!  I'm hopeful that I've identified the issue and am prepping a fix, but if I haven't...

chuckatkins:
> Thanks! I'm hopeful that I've identified the issue and am prepping a fix, but if I haven't...

Great!  Either way you'll need to rebase to get past the azure failures so if you wait until #2332 is merged in the next hour or so then you can go ahead and pick that up along with it in case it's not fixed.

eisenhauer:
Unfortunately this PR contains upstream merges and I don't think I've ever successfully rebased when those are in play.  (Rebasing seems to apply the upstream patches to top-level, rather than the thirdparty tree where they belong.)  So I'll likely kill this PR and start fresh...

eisenhauer:
Replaced by #2334

Issue #2320, opened 2020-06-10
Over allocation at mode::defer  


guj:
**Describe the bug**
Buffer is expanded twice when running Put(mode::defer) + PerformPuts()

**To Reproduce**

In BP4Writer.cpp::PerformPuts(): 

1.  resize buffer at  line 66 
 
2. line 84, then calls PerformPutCommon, which calls PutSyncCommon, and buffer is expanded again there.  <<I believe the counting is not quite right here.>>

**Expected behavior**

**Desktop (please complete the following information):**
 - OS/Platform:  Any
 - Build: 2.6.0 (possibly earlier versions too)

**Additional context**
Here is what I see with 1 core, writing 3 variables, var1 & var2, var3, 80MB each. 
put(var1, var1_data, deferred) 
performPuts(var1) 
=> now, peak at BufferSTL.cpp::resize(), it shows m_Buffer is a bit over 80MB

put(var2, var2_data, deferred) 
performPuts(var2) 
=> now, peak at BufferSTL.cpp::resize(), it shows m_Buffer is a bit over 240MB, expecting it to be around 160MB!!

put(var3, var3_data, deferred) 
performPuts(var3) 
=> now, peak at BufferSTL.cpp::resize(), it shows m_Buffer is a bit over 430MB, expecting it to be around 240MB!!
looks like somewhere,  byte accounting is not done correctly. The pattern is the previous variables are double counted. 

**Following up**
Was the issue fixed? Please report back.

chuckatkins:
Fixed by #2326

PR #2326, opened 2020-06-12
Fix performputs buffersize

pnorbert:
…datasize correctly in case of multiple calls of PerformPuts(). Addresses #2320

### DIFF
--- a/bindings/CXX11/adios2/cxx11/Engine.cpp
+++ b/bindings/CXX11/adios2/cxx11/Engine.cpp
@@ -199,6 +199,17 @@ ADIOS2_FOREACH_PRIMITIVE_TYPE_1ARG(declare_template_instantiation)
 ADIOS2_FOREACH_TYPE_1ARG(declare_template_instantiation)
 #undef declare_template_instantiation
 
+size_t Engine::DebugGetDataBufferSize() const
+{
+    helper::CheckForNullptr(m_Engine,
+                            "in call to Engine::DebugGetDataBufferSize");
+    if (m_Engine->m_EngineType == "NULL")
+    {
+        return 0;
+    }
+    return m_Engine->DebugGetDataBufferSize();
+}
+
 std::string ToString(const Engine &engine)
 {
     return std::string("Engine(Name: \"" + engine.Name() + "\", Type: \"" +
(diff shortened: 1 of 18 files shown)


pnorbert:
@chuckatkins This test writes 3 datasets, each 721MB (and does this twice testing BP3 and BP4). The dataset is a side effect only, so we could remove each immediately after writing, but testing has no access to thirdparty (`<adios2sys/SystemTools.hxx>` specifically). Do you have any suggestion how to add data removal either in CMakeLists.txt or in the code itself? Removing after test would still require 2.1GB of space, removing each right after writing would still require 721MB of space to write during CI.

pnorbert:
@chuckatkins NVM for this test, now it uses NULL transport to avoid writing to disk at all. 
But in general it would be good to know how to remove output files after a test.

pnorbert:
@chuckatkins @bradking Do you guys have an idea why the sanitizers fail to run this new test?
test 234/235, they print /proc/self/status and the VmSize is ridiculously big here. And they timeout, even though they don't do any IO anymore, just buffer 80MB of data  3*3*3 times.
```
234: After Open():                                 buffer size = 16384        VmSize:	107374517864 kB VmRSS:	  317036 kB 
```

bradking:
Cc: @KyleFromKitware

chuckatkins:
> @chuckatkins This test writes 3 datasets, each 721MB (and does this twice testing BP3 and BP4). ... Do you have any suggestion how to add data removal either in CMakeLists.txt or in the code itself? Removing after test would still require 2.1GB of space, removing each right after writing would still require 721MB of space to write during CI.

@pnorbert We should have the capacity on the test VMs for a 2.1gb working set, so long as it's removed after the test.  The way to accomplish that is through test fixtures.  See the docs here for how to use it: https://cmake.org/cmake/help/v3.7/prop_test/FIXTURES_REQUIRED.html

But essentially it's just:
```
# TestFoo generates some output that needs to be cleaned up
add_test(NAME TestFoo ...)

# Make a "test" that cleans up after foo
add_test(NAME TestFooCleanup
  COMMAND ${CMAKE_COMMAND} -E remove_directory /pth/to/TestFooOutput
)

# Mark it as a cleanup test for the "Foo" fixture
set_tests_properties(TestFooCleanup PROPERTIES FIXTURES_CLEANUP Foo)

# Mark the TestFoo as needing any setup or cleanup defined in the "Foo" fixture
set_tests_properties(TestFoo PROPERTIES FIXTURES_REQUIRED Foo)
```

PR #2328, opened 2020-06-15
Rework bp4 open

pnorbert:
Rework the timeout loops and separate timeout loops for opening md.idx and md.0.
Adds extra 5s wait for md.0 if md.idx is present but md.0 is not yet present.


chuckatkins:
@pnorbert you should be able to rebase the ci failures away now.

PR #2329, opened 2020-06-16

eisenhauer:


eisenhauer:
Aside from failures on unstable CI platforms, this test only showed the known BP_stream issue.

PR #2331, opened 2020-06-16
ci: Disable mac builds

chuckatkins:
The mac builds have been very flaky the past several days so I'm disabling them for now to allow other work to continue and we can re-enable them in a PR once the issue is resolve.


chuckatkins:
Replaces #2322

PR #2332, opened 2020-06-16
ci: Leverage ctest/cdash integration for sanitizers

chuckatkins:


chuckatkins:
@eisenhauer this should give you better debug information in the sanitizers

PR #2333, opened 2020-06-16
ci: Add a newer CTest to the debian-sid ci image to allow test re-runs

chuckatkins:


chuckatkins:
@KyleFromKitware This will let us have the test-reruns for the "regular" CI image and then for the packaging build we can explicitly use /usr/bin/cmake

PR #2336, opened 2020-06-17
Remove deprecated MPI_NOEXEC and fix sanitizers

chuckatkins:


chuckatkins:
@JasonRuonanWang the dataman tests weren't actually using MPI for anything; i.e. they were explicitly always using `MPI_COMM_SELF` and running with `MPI_NOEXEC` so I made them drop MPI entirely.  Is that reasonable or am I missing something?

JasonRuonanWang:
@chuckatkins For now it's completely fine. We don't have any MPI based use cases of DataMan so far. But the long-term plan is to support MPI based use cases.

chuckatkins:
@JasonRuonanWang also now that the DataMan tests are running for the serial builds they get covered by the sanitizers.  Can you please take a look at the associated DataMan failures in the `Dynamic Analysis` section in the CDash output and push the appropriate fixes back to this PR?

JasonRuonanWang:
@chuckatkins I believe #2337 will fix the CDash error. Let me know if you still see it in a while.

chuckatkins:
@JasonRuonanWang There were more sanitizer issues scattered throughout DataMan than just the uninitialized value; you can see the full set of failures here: https://open.cdash.org/index.php?compare1=61&filtercount=1&field1=revision&project=ADIOS&showfilters=0&limit=100&value1=15b45cf90b918feffbe55b0a86a18c595fe0c781&showfeed=0.  I've addressed most of them in ea03fdf.  Please take a look to make sure they're okay with you.

@JasonRuonanWang @eisenhauer There's still a few remaining tsan issues though: https://open.cdash.org/index.php?compare1=61&filtercount=1&field1=revision&project=ADIOS&showfilters=0&limit=100&value1=ea03fdf1b6e7a1947b4afa952ce82e5847b298b6&showfeed=0

@JasonRuonanWang it seems the dataman race conditions are from calling `cout` in multiple test threads without any locking, it should be easy enough to lock around it.

@eisenhauer It looks like some race conditions are showing up for SST as well.

Can you both please take a look at the sanitizer issues and push the fixes back to this branch instead of separate PRs?

eisenhauer:
> Well that's one way to get rid of it...
> 
> Is it a false alarm?

It's how it's handled in the evpath DP.  Basically there's a structure of function pointers that are intialized before the structure is returned for later use by the control plane.  They are going to be identical structures for every instance of the RDMA dp because of course the function pointers values don't change, so we just have one structure in static memory.  The first call to the LoadRDMADP function fills it, and later calls write identical values into identical locations.  It's perhaps theoretically unsafe IFF 64-bit writes/reads of addresses are not atomic *and* somehow writing the same value twice a second time somehow leaves a window where a read might sneak in and get a value that was not that same value, but there are no realistic architectures where that happens.  So, yeah, IMHO kind of like a false alarm.  We could give each RDMA DP their own copy of this identical, unchanging after initialization, structure and avoid it the TSAN warning that way, but I don't think that actually improves safety as much is just a different way to shut up TSAN...

chuckatkins:
@eisenhauer thanks for cleaning up the sst tsan issues.

@JasonRuonanWang can you work through the remaining DataMan tsan issues?  I think most of them are just in the test code itself.

chuckatkins:
@JasonRuonanWang I believe all the DataMan issues are clean now; thanks!

@eisenhauer still a few sst failures left.

eisenhauer:
Of the SST issues, some are inside libfabric, in fi_getinfo() which the RDMA data plane is calling trying to see if it can run there.  In this test, two threads each with a different ADIOS object are both calling into it and likely libfabric has some state.  I find release notes for libfabric that seem to indicate that fi_getinfo being "thread safe" may vary by provider.  (At least a 2019 v1.7.1 note says that it has been made thread-safe for verbs.)  Thoughts on handling?

eisenhauer:
Looks like the other problems will require FFS upstream fixes.  They are not onerous, but since upstream and rebasing don't play well together I wanted to confirm how you'd like that done.  Push an upstream update here?  Or push it to master and rebase this?

chuckatkins:
It sounds like we can't guarantee that fi_getinfo is thread safe given we can operate over different providers.  The same issue would arise if it wasn't a reader and writer but multiple distinct writers in the same application, which is arguably a more likely use case.  Can you use a static mutex accessable to all threads it's being called from?

chuckatkins:
You can just push it here.  The scope of the pr expanded a bit since it  actually gates passing on the sanitizers being clean, which in turn means the sanitizer issues need resolving in this PR as well.

eisenhauer:
On the libfabric issues.  Yes, certainly, but generally I'd consider whatever we do a patch to get through TSAN now.  Really running multiple RDMA streams likely requires more extensive cooperation between ADIOS instances to manage shared resources...  I'll drop in a patch, but that and the FFS things are going to have to wait at least until the afternoon.

chuckatkins:
@eisenhauer rather churning through changes in ffs, does it instead just make more sense to assume that the `FMregister_data_format` calls aren't thread safe and just mutex the call inside `doFFSFormatRegistration`?

eisenhauer:
Ugh, and one more I missed.  There's a static value that tells us if our current host is bigendian, initialized once.  That could be done in CMake now, but the code predates cmake by many years...

chuckatkins:
@eisenhauer the most recent failure is mine.  I've updated with the fix and we should be good to merge after this.

@JasonRuonanWang @eisenhauer Thank you both for the assist in this pr

PR #2345, opened 2020-06-19
Rearrange TestThreads; avoid blowing stack size on CLang

eisenhauer:


chuckatkins:
I believe this will conflict with #2344 so you may want to either pull that into this pr or rebase after it's merged.  Which would you rather?

eisenhauer:
I was just writing a comment on your PR.   I've killed the action on #2345, will wait for #2344 to merge and then rebase this.

Issue #2339, opened 2020-06-17
BP4 invalid blockID error


caitlinross:
@pnorbert  I've run into a bug in ADIOS with BP4 engine and I think it may be related to #1639 (I'm at least getting the same error as that issue).

I'm using ADIS for reading alongside a running XGC sim. In ADIS, I'm using Begin/EndStep calls and ensure that I'm only doing Get calls once BeginStep has returned OK. If I use this approach to read a BP file that's no longer being written to, there's no problem. If I start the reader at the same time as the simulation, first step runs fine, but after that I hit this error:
```
terminate called after throwing an instance of 'std::invalid_argument'
  what():  ERROR: invalid blockID 0 from steps start 0 in variable dpot, check argument to Variable<T>::SetBlockID, in call to Get
```

I also tried letting the sim get more of a head start and starting the reader after about 5 or 6 steps had already been written. I thought maybe there was somehow a bug that was causing BeginStep to return OK when it shouldn't, and that I was trying to do Get calls before those variables were actually in the file, but that doesn't appear to be the case. 
Here's what I did:
- let simulation complete steps 0-4
- confirmed with bpls that all 5 steps were in the file
- started reader (while simulation was working on step 5)
- reader ran fine until step 5, where I got the above error (even though the simulation was already writing step 8 or 9 at that point and all the previous steps were in the file according to bpls)

I added in some debugging output into ADIOS and think I managed to track down the problem (or at least part of it). It looks like `m_StepsStart` in the variable gets reset incorrectly at some point due to `m_FirstStreamingStep` not being set correctly. Here's some output:
```
[ADIS] reader.BeginStep
BP4Reader::InitBuffer
BP4Deserializer::DefineVariableInEngineIOPerStep m_StepsStart set to 0
VariableBase::ResetStepsSelection zeroStart = 0, m_StepsStart 0, m_FirstStreamingStep 1
    setting m_StepsStart to 0
[ADIS] ADIOS stepstatus OK
[ADIS] source 3d is on step 0
[ADIS] about to make adios Get call
BP4Deserializer::InitVariableBlockInfo var dpot m_StepsStart 0, stepsCount 1
[ADIS] Get call done
[ADIS] reader.EndStep

[ADIS] reader.BeginStep
VariableBase::ResetStepsSelection zeroStart = 0, m_StepsStart 0, m_FirstStreamingStep 0
    setting m_StepsStart to 1
[ADIS] ADIOS stepstatus OK
[...]

Issue #1639, opened 2019-07-28
BP4 streaming and Local arrays lead to exception


pnorbert:
When new steps are found in BP4 streams, something goes wrong with the processing. 

```
C++ exception with description "ERROR: invalid blockID 1 from steps start 0 in variable i8, check argument to Variable<T>::SetBlockID, in call to Get
```

To reproduce:

- Enable Staging.*.Local.BPS tests
in testing/adios2/engine/staging-common/CMakeLists.txt, line 233, comment out:
`list (FILTER BP4_STREAM_TESTS EXCLUDE REGEX ".*Local.BPS$")`

- Add some sleeptime (100ms) in testing/adios2/engine/staging-common/TestCommonWriteLocal.cpp
```
#include <thread>
...
        std::this_thread::sleep_for(std::chrono::milliseconds(100));
        engine.BeginStep();
```

- run ctest

```
$ ctest -R Staging.*.Local.BPS.BP4_stream
Test project /home/adios/work/ADIOS2/build.master
    Start 610: Staging.1x1.Local.BPS.BP4_stream
1/5 Test #610: Staging.1x1.Local.BPS.BP4_stream ...***Failed    3.27 sec
    Start 626: Staging.2x1.Local.BPS.BP4_stream
2/5 Test #626: Staging.2x1.Local.BPS.BP4_stream ...***Failed    3.36 sec
    Start 627: Staging.1x2.Local.BPS.BP4_stream
3/5 Test #627: Staging.1x2.Local.BPS.BP4_stream ...***Failed    2.30 sec
    Start 628: Staging.3x5.Local.BPS.BP4_stream
4/5 Test #628: Staging.3x5.Local.BPS.BP4_stream ...***Failed    1.69 sec
    Start 629: Staging.5x3.Local.BPS.BP4_stream
5/5 Test #629: Staging.5x3.Local.BPS.BP4_stream ...***Failed    2.16 sec

0% tests passed, 5 tests failed out of 5
```

- run single test
```
rm -rf local.bp; mpirun -n 5 ./bin/TestCommonWriteLocal bp4 local : -n 3 ./bin/TestCommonReadLocal bp4 local OpenTimeoutSecs:2.0008
```

pnorbert:
Bringing back this issue into attention. It was never fixed. It sometimes shows up in CI with test
[...]

PR #2348, opened 2020-06-24
Tests for reading variables with Steps

pnorbert:
Added tests for reading back global/local arrays with different modes: File vs Stream, All steps vs step by step, and block-by-block.

Exposes bugs in #1639 and #2339 and also exposes read problems with variables that are not written in the first step and/or not written every step.

A lot of tests fail in this PR. Any attempt to solve the handling of steps in BP3/BP4 must pass these tests.
More tests needs to be created, especially for in-situ reading where new steps are becoming available during reads.


pnorbert:
@guj @caitlinross

PR #2349, opened 2020-06-25
Add rank decomposition by dimensional ratio to iotest

philip-davis:
Currently, the iotest x y z rank decomposition is done with the -d flag, using literal dimensions; so `-d 2 4 2 `defines a 2x4x2 decomposition, and the number of ranks for the application must be 16. This PR adds the option to specify the ratio of the dimensions of ranks by providing a comma-separated list to the -d flag, so that `-d 1,2,1` would have the same results ad `-d 2 4 2` if 16 ranks were provided, but would also work for a scaled up or scaled down number of ranks (for example, `-d 1,2,1` with 2 ranks would be equivalent to `-d 1 2 1`; `-d 1,2,1` with 54 ranks would be equivalent to `-d  3 6 3`; with 128 ranks, `-d 4 8 4`; etc.)

The reason for this change is to work better with Cheetah. It is not possible to correlate options that are separated by spaces inside a sweep in Cheetah, and it does not seem to be possible to correlate process count with different option values without defining a sweep for each valid combination (at least this is my understanding after some correspondence with @kshitij-v-mehta.) I wrote this issue up more [here](https://github.com/philip-davis/adios-patterns-planar/issues/1).

This change is backwards compatible.

### DIFF
--- a/source/utils/adios_iotest/settings.cpp
+++ b/source/utils/adios_iotest/settings.cpp
@@ -130,8 +130,6 @@ int Settings::rescaleDecomp()
 
     throw std::invalid_argument(
         "decomposition ratios must scale up to process count");
-
-    return (0);
 }
 
 int Settings::processArgs(int argc, char *argv[])
--- a/source/utils/adios_iotest/settings.cpp
+++ b/source/utils/adios_iotest/settings.cpp
@@ -88,9 +88,10 @@ size_t Settings::stringToNumber(const std::string &varName,
 
 int Settings::parseRatios(const char *arg)
 {
-    char *argCopy = strdup(arg);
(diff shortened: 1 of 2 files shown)


pnorbert:
Can you use a different character than -d? One option with two meanings is confusing.
How about -D?

pnorbert:
And then maybe it would be nice to have both the same input style
-d 2,4,2
-D 1,2,1

philip-davis:
I will make those changes. I assume we still want `-d 2 4 2` to work as before (interpret the same as `-d 2,4,2` to preserve backwards compatibility.

philip-davis:
I have made the requested changes. Both -D and -d accept either a space- or comma-separated list of arguments. The usage notice instructs comma-separated.

PR #2362, opened 2020-07-06
bpls: Minor cleanup of variable and attribute lookup

bradking:
Follow up #2350.


bradking:
CI is finally happy here after a few retries.

PR #2369, opened 2020-07-09
WIP: Introduce and use an IO::ForEachAttribute method

bradking:
Introduce an alternative to the `ADIOS2_FOREACH_ATTRIBUTE_STDTYPE_1ARG` macro and demonstrate its use in a few places.

Please do not merge yet.  This is WIP for discussion.

### DIFF
--- a/source/adios2/core/Attribute.h
+++ b/source/adios2/core/Attribute.h
@@ -55,6 +55,61 @@ private:
     std::string DoGetInfoValue() const noexcept override;
 };
 
+template <typename V>
+void AttributeVisit(AttributeBase const &a, V const &v)
+{
+    switch (a.m_Type)
+    {
+    case DataType::None:
+        break;
+    case DataType::Int8:
+        v(static_cast<Attribute<int8_t> const &>(a));
+        break;
+    case DataType::Int16:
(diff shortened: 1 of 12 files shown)


chuckatkins:
I'm concerned for large metadata scenarios with hundreds of thousands of variables and attributes that this inverts the switch to the inner loop than the outer loop meaning the switch branching happens for every attribute rather then once on the outer loop.

source/adios2/core/Attribute.h
```diff
@@ -55,6 +55,61 @@ class Attribute : public AttributeBase
...
+        break;
+    case DataType::DoubleComplex:
+        v(static_cast<Attribute<std::complex<double>> const &>(a));
+        break;
+    case DataType::String:
+        v(static_cast<Attribute<std::string> const &>(a));
+        break;
+    case DataType::Compound:
+        // No compound attributes.
+        break;
```
NAThompson, line 109: Shouldn't we future-proof this a bit by adding a `case default` handler?

NAThompson:
@bradking : Could you put a paragraph into an `.rst` that explains why this is important? I couldn't back it out by reading through the code.

PR #2370, opened 2020-07-09
Fix documentation that append mode works with BP4 engine. 

### LINKED ISSUE #2367: Is Append mode supported?
In the [documentation](https://adios2.readthedocs.io/en/latest/components/components.html#opening-an-engine), it states that `adios2::Mode::Append` is under development. However, there are [unit](https://github.com/ornladios/ADIOS2/blob/003dc7eb4d54cb791f4a8a25af1ed1970a2a6bb3/testing/adios2/engine/bp/TestBPWriteAppendReadADIOS2.cpp#L326) tests for this mode, and it is exposed in the [Python interface](https://github.com/ornladios/ADIOS2/blob/master/bindings/Python/py11glue.cpp#L129). Yet the [comments](https://github.com/ornladios/ADIOS2/blob/master/bindings/CXX11/adios2/cxx11/IO.h#L255) say it is not supported.

Is append mode supported or not?

NAThompson:
Fixes #2367.

### DIFF
--- a/bindings/CXX11/adios2/cxx11/IO.h
+++ b/bindings/CXX11/adios2/cxx11/IO.h
@@ -252,7 +252,7 @@ public:
      * Open an Engine to start heavy-weight input/output operations.
      * @param name unique engine identifier
      * @param mode adios2::Mode::Write, adios2::Mode::Read, or
-     *             adios2::Mode::Append (not yet support)
+     *             adios2::Mode::Append
      * @return engine object
      */
     Engine Open(const std::string &name, const Mode mode);
@@ -265,7 +265,7 @@ public:
      * MPI Collective function as it calls MPI_Comm_dup
      * @param name unique engine identifier within IO
      * @param mode adios2::Mode::Write, adios2::Mode::Read, or
-     *             adios2::Mode::Append (not yet support)
+     *             adios2::Mode::Append
      * @param comm new communicator other than ADIOS object's communicator
      * @return engine object
      */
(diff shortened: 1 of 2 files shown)


NAThompson:
@JasonRuonanWang : The append mode was completed a long time ago, but the docs weren't updated.

I think a more understandable title would be "Document that Append mode has been completed"; my bad there.

lwan86:
@NAThompson   I only implemented the append mode for BP4. BP3 still doesn't support this mode. Like you said, the title of this commit is confusing. Could you change it? BTW, since you pointed out the docs haven't been updated, could you also clarified this in the docs? Thanks!

NAThompson:
@lwan86 : Done.

JasonRuonanWang:
I don't even think this actually "Document that append mode has been completed". As @lwan86 said, if you want to state it in that way, you probably also need to really go through the *documentation* and see if there is anything that needs to be changed.

NAThompson:
@JasonRuonanWang : The comments are extracted and put in the documentation.

pnorbert:
@guj Is Append mode supported by the HDF5 engine? Just curious, as this never came up, but it could easily support it by starting from N+1 steps if a file already has N steps.

williamfgc:
As far as I remember only BP4 (default engine) supports this, it was dropped for BP3 as it made more sense to put efforts as BP4 was moving towards default status. Append is not a simple flag as the `Substreams` parameters must match the original dataset. Merging different MPI aggregation patterns in a single dataset is not a trivial thing to implement. I'll try to find some time to help @NAThompson with the current docs as it's a task on its own as this is my fault. I'd encourage engine owners to review current docs and update.

guj:
No, currently not. Might make sense to make it consistent with ADIOS flags.

- Junmin

NAThompson:
@pnorbert: With @guj's comments in hand, it appears that this documentation is correct.

pnorbert:
Thanks all for the clarification. I will create an enhancement issue for HDF5 engine, so we don't forget about it.

PR #2371, opened 2020-07-09
Update documentation.

NAThompson:
- Joined arrays have been implemented.
- Follow documentation advice by using fixed width types in the docs.
- Properly typeset more code calls.


pnorbert:
Actually, I don't find a single test which tests Joined Arrays. There are some interface tests for defining them (not writing them), but the secret sauce is in reading back arrays written like this.

NAThompson:
@pnorbert : Nice catch, I only looked through the source and bindings for joined array code paths.

PR #2373, opened 2020-07-10
Print errno and strerror in POSIX transport to see better error messages

pnorbert:
-- don't know what is happening on the NVMe

### DIFF
--- a/source/adios2/toolkit/transport/file/FilePOSIX.cpp
+++ b/source/adios2/toolkit/transport/file/FilePOSIX.cpp
@@ -10,6 +10,8 @@
 #include "FilePOSIX.h"
 
 #include <cstdio>      // remove
+#include <cstring>     // strerror
+#include <errno.h>     // errno
 #include <fcntl.h>     // open
 #include <stddef.h>    // write output
 #include <sys/stat.h>  // open, fstat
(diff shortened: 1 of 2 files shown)


source/adios2/toolkit/transport/file/FilePOSIX.cpp
```diff
@@ -59,7 +59,9 @@ void FilePOSIX::Open(const std::string &name, const Mode openMode,
 {
     auto lf_AsyncOpenWrite = [&](const std::string &name) -> int {
         ProfilerStart("open");
+        errno = 0;
```
NAThompson, line 62: I'm confused about this. You set `errno` to zero, then assign `m_Errno` to `errno`, without any intervening assignments. So `m_Errno` is always zero.

NAThompson: Oh wow I just read about `errno`; I guess it behaves like an `eflags` register.

pnorbert: There is an intervening system function call that sets 'errno' to non-zero if there is an error. It's standard practice (for C programmers at least) to zero out errno before system calls to not catch errors from the past, hence the errno=0 statement here. m_Errno is used to remember the (possible) error for the call right after this and checked later (possibly the call in happens in a thread, line 63, but the check is in main thread much later)

PR #2372, opened 2020-07-10
Link to Software X paper in README.md

NAThompson:
Also, use `ADIOS2` in preference to `ADIOS 2`, matching the `grep` findings of higher frequency occurrence of the former in the documentation.

Remove reference to required `cmake` version as this tends to go stale.

PR #2377, opened 2020-07-14
Link to Software X paper for ADIOS in docs.

NAThompson:
This is the same PR as #2372, but I didn't use the github editor to do it.


NAThompson:
The build failure is due to #2378

PR #2382, opened 2020-07-16
Use pip3 for Python files.

NAThompson:
Although we may want to retain Python2 compatibility in the ADIOS package, there is no harm in running `pip3` on the files, and also, we have Python3 files which require `pip3`. So just use `pip3` everywhere.


NAThompson:
In a conversation with Norbert, we have decided to only support Python3 going forward.

chuckatkins:
We'll need to go ahead and migrate the ci images to drop python 2, but that sounds good

chuckatkins:
Closing in favor of #2407

PR #2390, opened 2020-07-22
Enforce invariants for inline engine

NAThompson:
The inline engine must have exactly one writer, and exactly one reader.
If a user attempts to add multiple readers or multiple writers, throw an exception.
This is a first step towards unifying the engine APIs, since currently we require
the user to specify which engine is the reader with a key.

Enforcing this invariant will allow us to remove the parameter keys, since we will
be able to uniquely identify the reader and writer via their respective modes.

As a bit of unrelated cleanup: throw exceptions when users request unsupported features,
like append mode or sync mode for the inline engine.

### DIFF
--- a/source/adios2/core/IO.cpp
+++ b/source/adios2/core/IO.cpp
@@ -572,7 +572,7 @@ Engine &IO::Open(const std::string &name, const Mode mode, helper::Comm comm)
         }
     }
 
-    // For the inline engine, there must be exactly 1 reader, and exactly 2
+    // For the inline engine, there must be exactly 1 reader, and exactly 1
     // writer.
     if (engineTypeLC == "inline")
     {
--- a/source/adios2/core/IO.cpp
+++ b/source/adios2/core/IO.cpp
@@ -572,6 +572,51 @@ Engine &IO::Open(const std::string &name, const Mode mode, helper::Comm comm)
         }
     }
 
+    // For the inline engine, there must be exactly 1 reader, and exactly 2
+    // writer.
+    if (engineTypeLC == "inline")
+    {
+        if (mode == Mode::Append)
+        {
+            throw std::runtime_error(
+                "Append mode is not supported for the inline engine.");
+        }
+
+        // See inline.rst:44
+        if (mode == Mode::Sync)
+        {
+            throw std::runtime_error(
+                "Sync mode is not supported for the inline engine.");
+        }
+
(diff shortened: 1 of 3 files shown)


NAThompson:
@caitlinross: Could you review?

pnorbert:
Isn't this code included in PR #2392?

NAThompson:
@pnorbert : Yes; but this is logically independent. We may want these checks even if we don't want to remove key-based lookup of the inline engines. In addition, merging this should leave the diff of #2392 with only a single commit.

source/adios2/core/IO.cpp
```diff
@@ -572,6 +572,51 @@ Engine &IO::Open(const std::string &name, const Mode mode, helper::Comm comm)
...
+    if (engineTypeLC == "inline")
+    {
+        if (mode == Mode::Append)
+        {
+            throw std::runtime_error(
+                "Append mode is not supported for the inline engine.");
+        }
+
+        // See inline.rst:44
+        if (mode == Mode::Sync)
```
caitlinross, line 586: I don't think it does, but to be sure, this doesn't affect single value variables that get read in sync mode by default does it (See `PutSyncCommon` in `InlineWriter.tcc`)? Actually I didn't even realize until now that Sync/Deferred are technically valid options to pass to `IO::Open`. I thought they were just for `Put`/`Get`.

caitlinross:
@NAThompson Looks good to me. The one comment I had is really more confusion on my part and not necessarily something that needs to change.

source/adios2/core/IO.cpp
pnorbert: Well, they should not be semantically valid options. It looks like Mode is an enum for two sets of options.

NAThompson: I was confused about this as well; with Norbert's comments it appears to be a slight inelegance in the code. I'll therefore also prohibit opening an engine in deferred mode in a separate PR since I don't want to get the commit graph all messed up for #2392.

caitlinross: So should the check for mode being Sync or Deferred be done for all engines instead of only inline?

NAThompson: @caitlinross: Good question; it seems to me that it should be for all engines, but I'll leave @pnorbert to make that decision.

PR #2392, opened 2020-07-23
Identify via mode

NAThompson:
The inline engine now uses the requested mode to identify the correct reader and writer.
This streamlines the API so that users can more easily swap between engines.

Note: It is not obvious from the diff, but no backwards incompatible changes were made.


NAThompson:
@caitlinross: Could you review?

PR #2397, opened 2020-07-28
Add stream reading mode for BP4 metadata processing. Read and parse o…

pnorbert:
…nly a limited metadata at a time (max 16MB but min 1 step). FileStream turns this on automatically


eisenhauer:
Nice!

PR #2400, opened 2020-07-29
u8 string literals are ill-formed in C++20

NAThompson:
See http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2019/p1423r0.html

### DIFF
--- a/testing/adios2/engine/bp/TestBPWriteReadAttributes.cpp
+++ b/testing/adios2/engine/bp/TestBPWriteReadAttributes.cpp
@@ -948,11 +948,6 @@ TEST_F(BPWriteReadAttributes, WriteReadStreamVar)
                                         separator);
         io.DefineAttribute<std::string>("smile", "\u263A", var2.Name(),
                                         separator);
-
-        io.DefineAttribute<std::string>("utf8", std::string(u8"महसुस"),
-                                        var1.Name(), separator);
-        io.DefineAttribute<std::string>("utf8", std::string(u8"महसुस"),
-                                        var2.Name(), separator);
 #endif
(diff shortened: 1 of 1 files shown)


williamfgc:
Looks good to me, it was just a proof that adios2 supports utf-8 as much as C++ does.

pnorbert:
So how does now ADIOS2 support non-English users to put utf-8 text in the data? I know these are just tests and we don't know if anyone is using UTF-8 to store data, but the goal should be to make it work with both C++11 and C++20 compilers.

NAThompson:
@pnorbert : There is no problem with putting utf8 in strings; it's just the u8 prefix that causes the problem. See [here](https://github.com/boostorg/multiprecision/blob/6447d5326dcaffd3f8114cc7f8722fae24694a4a/include/boost/multiprecision/pslq.hpp#L87) for an example of using utf8 in C++ strings.

williamfgc:
@NAThompson is there is a portable way between C++11 and C++20 to do this? I always worked on the assumption that C++ will remain backwards compatible, thanks for bringing this up. I think @pnorbert brings a very valid point. FYI, I'm not aware if we test with more recent C++ standards.

pnorbert:
Then instead of removing these tests, can you just fix them?

NAThompson:
> FYI, I'm not aware if we test with more recent C++ standards.

I checked for this; we do not test other C++ standards; moreover we force C++-11 and C99; see [here](https://github.com/ornladios/ADIOS2/blob/f5da12f0a3fde7dcd6cfa3ab1af8289f74b0b8e9/CMakeLists.txt#L73).

NAThompson:
Closing in favor of #2404

PR #2404, opened 2020-07-31
u8 string literals are ill-formed in C++20

NAThompson:
See http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2019/p1423r0.html


NAThompson:
@pnorbert : This is ready to go.

PR #2405, opened 2020-08-03
Added 3 64-bit MPI_Gatherv implementations

JasonRuonanWang:
We have had problems for aggregating metadata larger than 2GB. This PR tries to provide a few 64-bit MPI_Gatherv implementations which can handle gather operations larger than 2GB. 

There are three subroutines:

**Gatherv64**

This uses two-sided MPI, nonblocking Isend and Irecv underneath. The send count on each rank must match the receive counts on the root rank, otherwise it will cause unmatched Isend and Irecv paris and therefore it will run into deadlock. 

**Gatherv64OneSidedPush**

This uses one-sided MPI, MPI_Put. It allows the send count on each rank and the receive counts on the root rank to be unmatched. When they are unmatched, the send count on each rank will be effective.

**Gatherv64OneSidedPull**

This uses one-sided MPI, MPI_Get. It allows the send count on each rank and the receive counts on the root rank to be unmatched. When they are unmatched, the receive counts on the root rank will be effective.

### DIFF
--- a/source/adios2/helper/adiosComm.h
+++ b/source/adios2/helper/adiosComm.h
@@ -400,6 +400,26 @@ public:
                          Datatype recvtype, int root,
                          const std::string &hint) const = 0;
 
+    virtual void Gatherv64(const void *sendbuf, size_t sendcount,
+                           Datatype sendtype, void *recvbuf,
+                           const size_t *recvcounts, const size_t *displs,
+                           Datatype recvtype, int root,
+                           const std::string &hint) const = 0;
+
+    virtual void Gatherv64OneSidedPush(const void *sendbuf, size_t sendcount,
+                                       Datatype sendtype, void *recvbuf,
+                                       const size_t *recvcounts,
+                                       const size_t *displs, Datatype recvtype,
+                                       int root,
+                                       const std::string &hint) const = 0;
+
+    virtual void Gatherv64OneSidedPull(const void *sendbuf, size_t sendcount,
+                                       Datatype sendtype, void *recvbuf,
+                                       const size_t *recvcounts,
+                                       const size_t *displs, Datatype recvtype,
+                                       int root,
+                                       const std::string &hint) const = 0;
+
     virtual void Reduce(const void *sendbuf, void *recvbuf, size_t count,
                         Datatype datatype, Comm::Op op, int root,
                         const std::string &hint) const = 0;
--- a/source/adios2/helper/adiosComm.h
+++ b/source/adios2/helper/adiosComm.h
@@ -218,6 +218,23 @@ public:
                  const size_t *recvcounts, const size_t *displs, int root,
                  const std::string &hint = std::string()) const;
 
+    template <typename TSend, typename TRecv>
+    void Gatherv64(const TSend *sendbuf, size_t sendcount, TRecv *recvbuf,
+                   const size_t *recvcounts, const size_t *displs, int root,
+                   const std::string &hint = std::string()) const;
+
+    template <typename TSend, typename TRecv>
+    void Gatherv64OneSidedPush(const TSend *sendbuf, size_t sendcount,
+                               TRecv *recvbuf, const size_t *recvcounts,
+                               const size_t *displs, int root,
+                               const std::string &hint = std::string()) const;
+
+    template <typename TSend, typename TRecv>
+    void Gatherv64OneSidedPull(const TSend *sendbuf, size_t sendcount,
+                               TRecv *recvbuf, const size_t *recvcounts,
+                               const size_t *displs, int root,
+                               const std::string &hint = std::string()) const;
+
     template <typename T>
     void Reduce(const T *sendbuf, T *recvbuf, size_t count, Op op, int root,
                 const std::string &hint = std::string()) const;
--- a/source/adios2/helper/adiosComm.inl
+++ b/source/adios2/helper/adiosComm.inl
@@ -217,6 +217,38 @@ void Comm::Gatherv(const TSend *sendbuf, size_t sendcount, TRecv *recvbuf,
                            CommImpl::GetDatatype<TRecv>(), root, hint);
 }
 
+template <typename TSend, typename TRecv>
+void Comm::Gatherv64(const TSend *sendbuf, size_t sendcount, TRecv *recvbuf,
+                     const size_t *recvcounts, const size_t *displs, int root,
+                     const std::string &hint) const
+{
+    return m_Impl->Gatherv64(sendbuf, sendcount, CommImpl::GetDatatype<TSend>(),
+                             recvbuf, recvcounts, displs,
+                             CommImpl::GetDatatype<TRecv>(), root, hint);
+}
+
+template <typename TSend, typename TRecv>
+void Comm::Gatherv64OneSidedPush(const TSend *sendbuf, size_t sendcount,
+                                 TRecv *recvbuf, const size_t *recvcounts,
+                                 const size_t *displs, int root,
+                                 const std::string &hint) const
+{
+    return m_Impl->Gatherv64OneSidedPush(
+        sendbuf, sendcount, CommImpl::GetDatatype<TSend>(), recvbuf, recvcounts,
+        displs, CommImpl::GetDatatype<TRecv>(), root, hint);
+}
+
+template <typename TSend, typename TRecv>
+void Comm::Gatherv64OneSidedPull(const TSend *sendbuf, size_t sendcount,
+                                 TRecv *recvbuf, const size_t *recvcounts,
+                                 const size_t *displs, int root,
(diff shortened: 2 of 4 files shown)


source/adios2/helper/adiosComm.h
```diff
@@ -400,6 +400,26 @@ class CommImpl
                          Datatype recvtype, int root,
                          const std::string &hint) const = 0;
 
+    virtual void Gatherv64(const void *sendbuf, size_t sendcount,
+                           Datatype sendtype, void *recvbuf,
+                           const size_t *recvcounts, const size_t *displs,
+                           Datatype recvtype, int root,
+                           const std::string &hint) const = 0;
+
```
bradking, line 408: The `helper::Comm` abstraction's methods use templating so that callers don't need to pass `sendtype` or `recvtype` explicitly.  If we add `Gatherv64` that should be done here too instead of using `void*`.

However, we shouldn't need to add `Gatherv64`.  The `sendcount` argument in plain `Gatherv` is already a `size_t`.  Its existing implementation casts that to `int`, which is the cause of the existing limitation.  Instead of adding `Gatherv64`, you just need to update the implementation of `CommImplMPI::Gatherv` to use the block approach.  IIRC some of the other methods in `CommImplMPI` already do that.

JasonRuonanWang: > The `helper::Comm` abstraction's methods use templating so that callers don't need to pass `sendtype` or `recvtype` explicitly. If we add `Gatherv64` that should be done here too instead of using `void*`.
> 
> However, we shouldn't need to add `Gatherv64`. The `sendcount` argument in plain `Gatherv` is already a `size_t`. Its existing implementation casts that to `int`, which is the cause of the existing limitation. Instead of adding `Gatherv64`, you just need to update the implementation of `CommImplMPI::Gatherv` to use the block approach. IIRC some of the other methods in `CommImplMPI` already do that.

The reason to have some extra Gatherv64 implementations rather than modifying the existing Gatherv is because for use cases which does not exceed 2GB, the original Gatherv is still more optimal. If we completely remove the original Gatherv, it means that these use cases will get worse performance.

Due to the fundamental mechanism of MPI, it's also undesirable to determine 32-bit or 64-bit at runtime. Because in some cases, it's possible that only a few of the senders actually exceed the 32-bit limitation. If we determine this locally, it will cause unmatched MPI operations across all ranks. If we determine this globally, it will introduce extra MPI global operations than what is necessary to perform the Gatherv itself. 

So currently the most optimal way I can think of is to keep these 32-bit and 64-bit implementations, and let specific engines to determine which to use based on it's scopes, use cases, assumptions, etc.

JasonRuonanWang: > The `helper::Comm` abstraction's methods use templating so that callers don't need to pass `sendtype` or `recvtype` explicitly. If we add `Gatherv64` that should be done here too instead of using `void*`.

The missing templated functions were added.

JasonRuonanWang:
@pnorbert Is this what's needed for your  use case?

source/adios2/helper/adiosComm.h
bradking: The existing `Gatherv` takes a `size_t` count and should not fail when given a size larger than 32-bits.  Therefore `Gatherv`'s implementation should be fixed to support 64-bit sizes somehow.  If a non-MPI backend is added in the future it might not need any special help or chunking.

To retain efficiency for callers that *know* their sizes fit in 32 bits we could offer a `Gatherv32` that takes a `uint32_t` as a size (or if MPI is actually limited by a signed `int` then maybe even `Gatherv31` taking a `int32_t`).

JasonRuonanWang: > The existing `Gatherv` takes a `size_t` count and should not fail when given a size larger than 32-bits. Therefore `Gatherv`'s implementation should be fixed to support 64-bit sizes somehow. If a non-MPI backend is added in the future it might not need any special help or chunking.
> 
> To retain efficiency for callers that _know_ their sizes fit in 32 bits we could offer a `Gatherv32` that takes a `uint32_t` as a size (or if MPI is actually limited by a signed `int` then maybe even `Gatherv31` taking a `int32_t`).

The original MPI_Gatherv is for sure limited by a signed int, which is exactly the reason why I am doing the whole lot of this. Not only MPI_Gatherv, but also all MPI functions that take int as the size, are limited by a singed int. If you think in that way, then you will need to re-implement EVERY SINGLE MPI function in the Comm class to support 64-bit. Is that what you think we are supported to do? Even if the answer is yes, at least it's far beyond the scope of this PR...

germasch: My 2 cents: Since there is already an interface that supports 64-bit offsets/sizes, one shouldn't introduce another interface. In that case one would end up with two functions essentially doing the same thing, taking the same data types, only that the existing one, while taking 64-bit values, would actually not work when actual values are >= 32 bits.

Basically, all you have to is to go to the current implementation of Gatherv and make sure that all of the offsets/counts fit into 32 bits (or maybe 31). That, in fact, should be done anyway. And if they don't fit, use the new 64-bit-safe implementations, otherwise continue to use the standard MPI implementation.

JasonRuonanWang: > Basically, all you have to is to go to the current implementation of Gatherv and make sure that all of the offsets/counts fit into 32 bits (or maybe 31). That, in fact, should be done anyway. 

If you mean checking this at runtime, of course I thought about this. But as I mentioned, doing this only at local ranks will result in unmatched MPI operations across the world communicator, whereas doing this globally will introduce an extra MPI collective operation which is not needed if some engines know exactly what they are doing. 

In short, if I follow your suggestions, we will immediately see a performance drop for all applications.

germasch: Yeah, I guess that's a good argument, mismatched calls obviously is obviously a bad outcome. Well, so the one thing I'd definitely still suggest is to make the version that uses `MPI_Gatherv` throw an error if any arguments are out of range. But it does look like it's not easily possible to have a single interface and decide at runtime.

The iffy thing with this is that errors will only occur in rare circumstances, so will be difficult to test for. If one has `GatherV` and `GatherV64`, then the question is, can you ever use `GatherV` safely, ie. guarantee at compile time that the args will never be > 32bit?

Not sure what the best solution is. This would be all much easier if they could have settled on a 64-bit solution within the MPI standard...

JasonRuonanWang: @germasch If an engine blindly uses Gatherv, then it's for sure unsafe. But the engine can determine this in several ways. It can do an MPI collective operation before calling Gatherv and then call Gatherv or Gatherv64 accordingly if it does not care about the overhead. If it does care, however, it can ask users to pass a static engine parameter, to specify whether > 32 bit case is ever supposed to happen. For most of the applications, this should be a statically known knowledge. 

The reason why I am thinking it in such a uncommon way, is also because the actual applications that need 64 bit are very rare, and they probably only need Gatherv for aggregating metadata, but don't need other MPI functions being operated at 64 bit. If just for one or two this kind of apps, we re-implement the whole MPI software stack, it does not sound like a smart thing to do. If all other functions are actually 32-bit but with a 64-bit wrapper, why do we make Gatherv such an exception by have Gatherv32 and Gatherv? In the current status of the Comm class, having a Gatherv and a Gatherv64 is a very natural thing, unless all other 32-bit MPI functions with 64-bit wrappers are already addressed correctly.

bradking: > you will need to re-implement EVERY SINGLE MPI function in the Comm class to support 64-bit. Is that what you think we are supported to do?

Yes, eventually.  That was my intention when introducing the `Comm` abstraction.  IIRC some of the methods have the chunking already.

> if I follow your suggestions, we will immediately see a performance drop for all applications.

My suggestion is to re-implement the existing `Gatherv` using your 64-bit implementation, and introduce a new `Gatherv31` method that has the `int`-based limitation builtin to its interface.  This change is inside our internal API and does not need to affect applications.  Just update all our internal call sites (there are only a couple of them) to switch from `Gatherv` to `Gatherv31`.

JasonRuonanWang: @bradking Sounds reasonable. Do you want me to do it in this PR?

bradking: I suggest opening a new PR to rename the existing `Gatherv` to `Gatherv31`, change its signature, and update all its call sites in one step.  That way we can be sure no call sites accidentally use the 64-bit variant.  After that is merged, rebase this PR to restore the `Gatherv` with its original signature and use the chunking implementation.

JasonRuonanWang:
Ok, since I myself don't need this at all, and I am only here to help, if this is too complicated, I will just close it for now and wait for a smarter solution.

PR #2407, opened 2020-08-03
Migrate to python3 exclusively

chuckatkins:


chuckatkins:
@jasonwang the remaining test failure here is with ssc and it seems all the expected values are off by one.  Any idea what could be happening here?

JasonRuonanWang:
I actually cannot reproduce this. I checked out your branch on my local machine and everything worked perfectly...

PR #2410, opened 2020-08-04
Process metadata really step-by-step in BP4 when StreamReader is on. …

pnorbert:
…So far, it was reading in metadata in chunks but it processed the whole chunk (multiple steps) at once. For true semantics, we want Variables and Attributes show up in their proper step and since the IO map holds all entries of all processed steps, the current processing can show variables in later steps too early. Now the chunks are kept in memory but only processed step by step.


pnorbert:
Do not merge until Greg's attribute test is added with BP4

pnorbert:
Well, merge it anyway, because that test cannot be modified for all engines but now it works as intended with StreamReader=ON

PR #2412, opened 2020-08-05
Require C++14 support

NAThompson:
This is a "discussion commit", in that I have a compelling usecase for the C++-14 `[[deprecated]]` tag. Since VTK-m is now requiring C++14 support, it might be sensible to reap the benefit of increasing the C++ standard.

The compiler support for the `[[deprecated]]` tag is [green across the board](https://en.cppreference.com/w/cpp/compiler_support).

In addition, I updated the C standard to 11; I was unaware of reasoning behind the requirement for use C99 so this can be reverted if necessary.


eisenhauer:
I'd love to have deprecated in the context of the debugFlag API changes.  But, I do worry about the nature of the HPC machines that are ADIOS's raison d'etre...   Odd, specialized compilers aren't uncommon there...

NAThompson:
@eisenhauer : Yeah `debugMode` is another great use case. Does anyone know the `xl` compiler version on Summit? The link above claims the `xl` compiler supports `[[deprecated]]` since version 13.1.2.

eisenhauer:
Summit xl is 16.1, but that's a relatively new machine.  Constraints on older ones?  Rhea's default gcc is still 4.8.  (Newer available, but...)

NAThompson:
Ok, then on Rhea they would have to `mload` a new gcc; looks like the `[[deprecated]]` became available in `gcc 4.9`. However, 4.8 is super antiquated; I would be very surprised if most users weren't already `mload`ing a new version since 4.8 had numerous C++-11 bugs, IIRC.

eisenhauer:
That it's still the default probably tells us something though.  IMHO, deprecated flag is nice, love to have it, but in the end it's nicer error messages.  Makes it hard to justify a risk of not being able to build for some user on some platform.

germasch:
My 2c: `[[deprecated]]` is nice, but hardly essential. A lot of it can be emulated using compiler-specific ways of doing the same thing (`__attribute__((deprecated))`), which obviously is more painful to use, though, since it needs macro-wrapping.

There is, however a C++14 feature that adios2 could greatly benefit from, ie., lambda function taking `auto` arguments. Using this together with `std::variant` (e.g. from abseil), one could greatly simplify all those macro-generated FOR_EACH loops. But I do see the point that compiler support may be iffy (I don't really know, though, all the systems I'm on except for one ancient Cray do have recent enough compilers.)

NAThompson:
> My 2c: [[deprecated]] is nice, but hardly essential. A lot of it can be emulated using compiler-specific ways of doing the same thing (__attribute__((deprecated))), which obviously is more painful to use, though, since it needs macro-wrapping.

[Here's](https://stackoverflow.com/a/49037913/904050) what we'd have to do it we wanted to portably deprecate any part of the API. Personally I think this is more painful and more likely to break user code than requiring C++14 support.

As to essential; well, the long term health of any project depends on being able to undo mistakes. Currently we have no mechanism to do this.

> But I do see the point that compiler support may be iffy (I don't really know, though, all the systems I'm on except for one ancient Cray do have recent enough compilers.)

In my own projects I've moved to C++17; and no one has complained. In any case, we do have a spack build, so if users do insist on using antiquated compilers there are released versions they can use.

NAThompson:
So the build is failing due to the `xl` compiler linking to `libstdc++` 4.8. If we looks at the [spack](https://github.com/spack/spack/blob/acfa2ea0188129fd1e683fe8e146256b1a695d5c/var/spack/repos/builtin/packages/adios2/package.py#L69) script, it shows that ADIOS2 cannot be compiled with gcc 4.7. So the only gcc version we'd lose is 4.8. Similarly, only PGI 14 is lost, but no intel compiler versions would be lost via this change.

I'll need to log into Summit to understand if we can link to a different `libstdc++`.

chuckatkins:
> So the build is failing due to the `xl` compiler linking to `libstdc++` 4.8. If we looks at the [spack](https://github.com/spack/spack/blob/acfa2ea0188129fd1e683fe8e146256b1a695d5c/var/spack/repos/builtin/packages/adios2/package.py#L69) script, it shows that ADIOS2 cannot be compiled with gcc 4.7. So the only gcc version we'd lose is 4.8. Similarly, only PGI 14 is lost, but no intel compiler versions would be lost via this change.
> 
> I'll need to log into Summit to understand if we can link to a different `libstdc++`.

You *can* configure PGI to use a different gcc backend to match a newer libstdc++, this is something we had to do in Titan and should be in the docs on how to do it.  Dropping 4.8 is a particular problem though.  It's the system compiler on EL7 which is widely deployed and even if we use a newer C++ language version most commercial compilers like intel, pgi, and XL, rely on a system provided gcc and libstdc++ for their standard library implementation.

NAThompson:
@chuckatkins : Do we have any data on how many users do source builds vs how many consume ADIOS2 via spack? Does spack have download statistics?

ax3l:
For what it's worth, we will downstream switch to C++14 in openPMD-api very soon, too.

PIConGPU and WarpX are already C++14 and newer.
Deprecated attributes pre C++14 are very inconsistent for various use cases and I just filed C++11 bugs again for the major compilers. If you switch to C++14, this will definitely make it easier and we don't mind.

Python stacks are also with manylinux_2010 and newer free of GCC 4.8 these days.

PR #2413, opened 2020-08-05
Add 64-bit MPI_Gatherv into SSC helper

JasonRuonanWang:


chuckatkins:
The build failures you're getting are related to dropping python 2 support in #2407 since it's also changing the docker images we use in CI.  Once it's merged you can rebase and the CI falures should go away.

PR #2414, opened 2020-08-06
Compile C code with CC

NAThompson:
Compiling C code with the CXX compiler is deprecated and will be an error.
Also, use a uniform minimum CMake version throughout the project.

### DIFF
--- a/cmake/install/post/generate-adios2-config.sh.in
+++ b/cmake/install/post/generate-adios2-config.sh.in
@@ -103,7 +103,7 @@ fi
 
 for variant in $variants
 do
-  echo "Extacting ADIOS flags for C bindings (${variant})"
+  echo "Extracting ADIOS flags for C bindings (${variant})"
   echo "  Without ADIOS"
   make_target_flags ${variant}_without_C
   without_C_CFLAGS=$(head -1 ${variant}_without_C.flags)
@@ -119,7 +119,7 @@ do
   echo ADIOS2_C_CFLAGS_${variant}=\"${ADIOS2_C_CFLAGS}\" >> adios2.flags
   echo ADIOS2_C_LDFLAGS_${variant}=\"${ADIOS2_C_LDFLAGS}\" >> adios2.flags
 
-  echo "Extacting ADIOS flags for C++ bindings (${variant})"
+  echo "Extracting ADIOS flags for C++ bindings (${variant})"
   echo "  Without ADIOS"
   make_target_flags ${variant}_without_CXX
   without_CXX_CXXFLAGS=$(head -1 ${variant}_without_CXX.flags)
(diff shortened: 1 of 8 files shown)


pnorbert:
@chuckatkins Was this just an oversight, or did you have reasons to use CXX for C pieces?

NAThompson:
Looks like the CMake version diffs, which I thought were trivial and harmless, actually induced a bloodbath:

```
CMake Error at /opt/cmake/share/cmake-3.18/Modules/FindPackageHandleStandardArgs.cmake:165 (message):
  Could NOT find Python (missing: Python_INCLUDE_DIRS Python_LIBRARIES
  Python_NumPy_INCLUDE_DIRS Development NumPy Development.Module
  Development.Embed) (found version "2.7.5")
...
-- Found EnvModules: /opt/ohpc/admin/lmod/lmod/libexec/lmod  
Lmod has detected the following error: The following module(s) are unknown:
"py2-numpy"
```

So I'll either wait for #2407 to be merged or revert those diffs.

chuckatkins:
The build failures you're getting are, as you guessed, related to dropping python 2 support in #2407 since it's also changing the docker images we use in CI.  Once it's merged you can rebase and the CI failures should go away.

chuckatkins:
I think you rebased on the python branch instead of master

NAThompson:
No, I rebased on `master`; just checked my `zsh` history.

chuckatkins:
Something was squirrely with the rebase since it included the commits from the python pr.  I update my master locally and re-did the rebase and not ti looks like it should with just your commits.  Not quite sure what happened but it's good to go now.

chuckatkins:
These changes might break with a static build since the c++ libraries will have to be pulled in transitively.  Can you make sure this works with `-DBUILD_SHARED_LIBS=OFF`?

NAThompson:
@chuckatkins : I'm stunned at how great of a catch that was; it did indeed break with `-DBUILD_SHARED_LIBS=OFF`.

chuckatkins:
I'm the meantime I'll get our static ci build re-enabled

chuckatkins:
@NAThompson 

So, looking at the details in all of this, I believe the existing behavior is correct.  The C files are built with `CC`; `CXX` is only 
used as the linker, which is the desired behavior.  Was this failing for you before?

As for the CMake versions, while we require 3.12 to build, it should be backwards compatible with consumers down to 3.1.  So leaving the install tests setting a lower version is good because it sets the policy behaviors to act like an older version.

We should, however, fix the typos and bump the dashboard infrastructure script to `2.8.12`.

NAThompson:
@chuckatkins : To reproduce the failure, bump the C version to 11 and the C++ version to 14. That should get the failure I observed, and also give the intelligible deprecation message. (It's also why this is not a high priority.)

chuckatkins:
@NAThompson I switched from C99 & CXX11 to C11 & CXX14 and wasn't able to reproduce the problem with gcc 8.3.0; the `Install.Make.C` passed just fine. That being said, I can see this would probably fail specifically on the `_test_2` scenarios since they don't use separate compile and link steps.  Really the combination of compile and link steps for a C executable using `CC` are:

|  | static | shared |
| -- | -- | --
| **CC compile + link** | Unsupported | Supported |
| **CC compile -> CC link** | Unsupported | Supported |
| **CC compile -> CXX link** | Supported | Supported |

Given that we don't really need to wory about that right now since `CXX` works just for for compiling C99 code then I think we can punt any rework until the time is necessary.  Worst case it's a single test we need to work through.

Given that, I'm goign to go ahead and close this and I'll make a separate pr to address the spelling and cmake version issues

PR #2193, opened 2020-05-01
ci: Add spack images to CI testing

chuckatkins:

PR #2416, opened 2020-08-10
DataSpaces Engine CI

philip-davis:
This fixes a DataSpaces engine build issue caused by a change to the m_Type member in VariableBase.

A larger issue being addressed by this PR is that the DataSpaces engine is not in CI, so the build failure was not detected for some time. This WIP PR will add build and testing support for the DataSpaces engine into CI.

### DIFF
--- a/scripts/ci/images/el7/Dockerfile
+++ b/scripts/ci/images/el7/Dockerfile
@@ -29,6 +29,15 @@ ENV PATH /opt/sz/2.1.8.3/bin:${PATH}
 ENV LD_LIBRARY_PATH /opt/sz/2.1.8.3/lib64:${LD_LIBRARY_PATH}
 ENV CMAKE_PREFIX_PATH /opt/sz/2.1.8.3:${CMAKE_PREFIX_PATH}
 
+#Install DataSpaces
+WORKDIR /opt/dspaces
+RUN curl -L https://dataspaces.rdi2.rutgers.edu/downloads/dataspaces-1.8.0.tar.gz | tar -xvz && \
+    mkdir install
+    cd 1.8.0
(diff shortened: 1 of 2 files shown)


scripts/ci/images/el7/Dockerfile
```diff
@@ -29,6 +29,17 @@ ENV PATH /opt/sz/2.1.8.3/bin:${PATH}
 ENV LD_LIBRARY_PATH /opt/sz/2.1.8.3/lib64:${LD_LIBRARY_PATH}
 ENV CMAKE_PREFIX_PATH /opt/sz/2.1.8.3:${CMAKE_PREFIX_PATH}
 
+#Install DataSpaces
+WORKDIR /opt/dspaces
+RUN curl -L https://dataspaces.rdi2.rutgers.edu/downloads/dataspaces-1.8.0.tar.gz | tar -xvz && \
+    mkdir install
+    cd 1.8.0
+    ./configure CC=mpicc CXX=mpic++ FC=mpifort --prefix=/opt/dspaces/1.8.0
+    make -j
```
chuckatkins, line 38: The `el7` build is without MPI so it needs to go in a different image.  Also using `make -j` instead of `make -jN` means `j = infinite`, not great.  See the above SZ build for using the correct `j` value.

chuckatkins:
I've added dataspaces to the space ci build in #2193 which I should get fixed and merged today so I'd rather just wait for that to be merged and then rebase, adding the `ADIOS2_USE_DATASPACES=ON` in the CTest script for the build.

philip-davis:
Ah, okay, I didn't see your review until this last commit. I think I'll kill this PR and split out the engine fix and a CI update that is based on your spack ci build.

PR #2419, opened 2020-08-11
let ssc use pgi compiler

JasonRuonanWang:
An urgent request from HBPS project. XGC has been compiled only using PGI compiler, but SSC never tried that because nlohmann::json used to break PGI. 

I just tried on Summit using PGI, and it actually worked. Now let's see if it works with CI.

### DIFF
--- a/cmake/DetectOptions.cmake
+++ b/cmake/DetectOptions.cmake
@@ -179,7 +179,7 @@ endif()
 
 # SSC
 # SSC currently breaks the PGI compiler
-if(NOT (CMAKE_CXX_COMPILER_ID STREQUAL "PGI") AND NOT MSVC)
+if(NOT MSVC)
     if(ADIOS2_HAVE_MPI)
         if(ADIOS2_USE_SSC STREQUAL AUTO)
             set(ADIOS2_HAVE_SSC TRUE)
--- a/source/adios2/engine/ssc/SscHelper.cpp
+++ b/source/adios2/engine/ssc/SscHelper.cpp
@@ -48,11 +48,7 @@ size_t TotalDataSize(const Dims &dims, DataType type, const ShapeID &shapeId)
     {
         return GetTypeSize(type);
     }
-    else
-    {
-        throw(std::runtime_error("ShapeID not supported"));
-    }
-    return 0;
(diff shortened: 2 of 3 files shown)


JasonRuonanWang:
@chuckatkins Is it possible to turn off the warnings reported by PGI complier? It reports lots of false warnings which prevents lots of very common usage from being merged...

chuckatkins:
> @chuckatkins Is it possible to turn off the warnings reported by PGI complier? It reports lots of false warnings which prevents lots of very common usage from being merged...

The warning about unreachable code isn't a false alarm, it's legitimate; there's no code path that the `return 0;` will ever execute.  The function in question can be fixed by changing
```c++
    else
    {
        throw(std::runtime_error("ShapeID not supported"));
    }
    return 0;
```
to
```c++
    throw(std::runtime_error("ShapeID not supported"));
```

JasonRuonanWang:
> The warning about unreachable code isn't a false alarm, it's legitimate; there's no code path that the `return 0;` will ever execute. The function in question can be fixed by changing
> 
> ```c++
>     else
>     {
>         throw(std::runtime_error("ShapeID not supported"));
>     }
>     return 0;
> ```
> 
> to
> 
> ```c++
>     throw(std::runtime_error("ShapeID not supported"));
> ```

I didn't see this. What I have seen is that it was complaining about the pugixml having some functions declared but never referenced...

chuckatkins:
The warnings coming from third party libraries should be suppressed/ ignored in the ci builds.  Can you point me to a build that's failing from it?  Its possible the regex isn't matching everything it needs to ignore and might need to be tweaked.

JasonRuonanWang:
> The warnings coming from third party libraries should be suppressed/ ignored in the ci builds. Can you point me to a build that's failing from it? Its possible the regex isn't matching everything it needs to ignore and might need to be tweaked.

Oh okay. I didn't know about that. I will re-work on it and let you know if there is anything I cannot solve. Thanks.

chuckatkins:
If you look at the output on the dashboard you can see lines like `CTest warning ignored` and `CTest warning matched` to see which of the warnings in the output are actually causing the build to fail.

PR #2428, opened 2020-08-13
Double-pointer Get

NAThompson:
@pnorbert : Do you think `Get(Variable<T>, T**)` should return an `adios2::Dims` (local dimensions).


pnorbert:
> Do you think `Get(Variable<T>, T**)` should return an `adios2::Dims` (local dimensions).
That's a neat idea. I am not sure if we need it, but I think it makes programming easier in case the user does not want to set the boundaries. Trouble is: what does that mean in other engines, which can read anything from a global space. What should they read when user says nothing? By default, they read the whole variable, which is not a smart thing in parallel codes. So for now I lean to say no to this suggestion.

NAThompson:
@pnorbert : Thanks; that's a good point.

And actually, it is not a breaking change to return an `adios2::Dims` should we wish to undo this decision in the future.

Issue #2420, opened 2020-08-11
Defective bp4 dataset with v2.6.0, works with v2.5.0


williamfgc:
**Describe the bug**
BP4 dataset generated using the same code  

**To Reproduce**
Attached a good file and a bad file in `adios2_bug_data.tar.gz`
good file (with 2.5.0): ex9-p-periodic-hexagon.mesh_o3_250.bp
bad file (with 2.6.0): ex9-p-periodic-hexagon.mesh_o3_260_bad.bp
bp4dgb output for bad file: bp4dgb.log with message `ERROR: Var Name string length (8) is longer than the limit to stay inside the block (-12.0)`
[adios2_bug_data.tar.gz](https://github.com/ornladios/ADIOS2/files/5059696/adios2_bug_data.tar.gz)

**Expected behavior**
File contents should be the same and correct

**Desktop (please complete the following information):**
 - OS/Platform: Ubuntu 18.04
 - gcc 8

**Additional context**
Files generated from MFEM-adios2 integration in ExAM using exactly the same source code for adios2 v2.5.0 and v2.6.0 this could affect other apps since these were working fine with v2.5.0 standard production features.

williamfgc:
@pnorbert FYI, this has been stable with v2.5.0 all across.

pnorbert:
The md.idx of the 260 version is already corrupt. The offsets into the md.0 file should be monotonous and completely filling the space (EndPtr of step N = PGPtr of step N+1) but here record 6 is not continuing record 5 and later you can see many jumps backwards. Was this a single run writing this file? I don't have a clue how it is possible to create such an index. 
```
$ bp4dbg ex9-p-periodic-hexagon.mesh_o3_260_bad.bp/md.idx 
========================================================
    Index Table File: ex9-p-periodic-hexagon.mesh_o3_260_bad.bp/md.idx
========================================================
----------------------------------------------------------------------------------------------------------------------
|        Version string            | Major | Minor | Patch | unused | Endian | BP version | Active |     unused      |
|          32 bytes                |   1B  |   1B  |   1B  |   1B   |   1B   |     1B     |   1B   |       25B       |
+--------------------------------------------------------------------------------------------------------------------+
| ADIOS-BP v2.6.0 Index Table      |   2   |   6   |   0   |        |  yes   |      4     |   no   |                 |
----------------------------------------------------------------------------------------------------------------------
 
-----------------------------------------------------------------------------------------------------------
[...]

PR #2434, opened 2020-08-17
bindings: Help C++11 clients avoid calling serial constructor with MPI_Comm

bradking:
Client code that does not enable `ADIOS2_USE_MPI` may accidentally try to pass a `MPI_Comm` instance to our constructor.  If the type the MPI library uses to implement MPI_Comm happens to be convertible to one of the types offered by our serial constructors (e.g. `bool`), the invocation will appear to succeed but will ignore the communicator passed.  Add explicitly deleted constructors that have better conversions for common `MPI_Comm` types.

With this, an attempt in client code to call the `adios2::ADIOS` constructor with a `MPI_Comm` object will fail to compile with an error such as:

    error: use of deleted function 'adios2::ADIOS::ADIOS(void*)'
        adios2::ADIOS adios(comm);
    ...note: declared here
        ADIOS(void *define_ADIOS2_USE_MPI_to_use_MPI_constructor) = delete;
        ^~~~~

### DIFF
--- a/bindings/CXX11/adios2/cxx11/ADIOS.h
+++ b/bindings/CXX11/adios2/cxx11/ADIOS.h
@@ -62,6 +62,20 @@ public:
      */
     ADIOS(const std::string &configFile, MPI_Comm comm,
           const bool debugMode = true);
+#else
+    // Client code that does not enable ADIOS2_USE_MPI may accidentally
+    // try to pass a MPI_Comm instance to our constructor.  If the type
+    // the MPI library uses to implement MPI_Comm happens to be convertible
+    // to one of the types offered by our serial constructors (e.g. 'bool'),
+    // the invocation will appear to succeed but will ignore the
+    // communicator passed.  Add explicitly deleted constructors that have
+    // better conversions for common MPI_Comm types.
+
+    // openmpi uses 'ompi_communicator_t*' for MPI_Comm
+    ADIOS(void *define_ADIOS2_USE_MPI_to_use_MPI_constructor) = delete;
+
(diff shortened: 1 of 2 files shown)


bradking:
For reference, this came up in https://github.com/ornladios/ADIOS2/issues/2420#issuecomment-675041926.

bradking:
It's not needed for `IO::Open` because there are no serial overloads that will be selected by passing a `MPI_Comm` incorrectly.  When passed, the communicator is always the third argument, so it should already be a compile error if `-DADIOS2_USE_MPI` is not enabled.

bradking:
> would a explicit constructor be more effective than deleting implicit constructors?

No.  An explicit constructor only prevents implicit calls to that constructor.  In this case the call sites are already calling the constructor explicitly.  The undesired conversion is taking place in the overload selection of that explicit call.

williamfgc:
@bradking I see, makes sense. Thanks for the explanation. A minor thing, should this be put in the `core::` namespace? That way it propagates to the Python bindings, but I'm not sure if we'd see something similar with Python. Otherwise, it looks good. Thanks! The main issue is that this was failing silently.

williamfgc:
Just pickin' your brains @bradking and @chuckatkins , is non-target legacy CMake going to have its Python 2 moment?

bradking:
> should this be put in the `core::` namespace?

The `core::ADIOS` constructors are agnostic and take only the `helper::Comm` abstraction, so this problem cannot occur internally.  #2238 already took care of a similar public-facing case for the Python bindings.

PR #2436, opened 2020-08-18
Inline simplifications.

NAThompson:


NAThompson:
> Add a unit test to ensure that when selecting multiple blocks, that failure occurs gracefully. Or if selecting blocks that aren't available on the current rank.

(The test we have validates the successful case but does not validate the failure mode.)

chuckatkins:
@NAThompson any idea when we can get the remaining test cases so this can get merged?

chuckatkins:
@NAThompson ping?  Just waiting on some additional test cases so we can merge this and make use if it with the inline engine.

chuckatkins:
Rebased on current `release_27`

PR #2448, opened 2020-09-01
Added SSC tests to ensure different ways of calling LockWriterDefinitions and LockReaderSelections

JasonRuonanWang:


eisenhauer:
Hey Jason!  Is there anything SSC-specific about these tests?  Always happy to have more tests to run against all staging engines.  I'm happy to help shoehorn this into the framework in staging-common if you'd like...

JasonRuonanWang:
> Hey Jason! Is there anything SSC-specific about these tests? Always happy to have more tests to run against all staging engines. I'm happy to help shoehorn this into the framework in staging-common if you'd like...

No there is nothing SSC-specific. These tests just simply move the LockWriterDefinitions and LockReaderSelections around and ensure that wherever users call them it always work. If you like to take them and put into the staging-common tests then just feel free to take them. The reason why I prefer keeping a copy of these tests in the SSC tests is because I want to keep every test as an independent code, with very straight-forward logic, and in this case if there is something wrong, it's very easy to identify the problem and solve it. Staging-common tests are great in the sense that it's flexible to change parameters, so every engine can re-use the same set of tests, but the way they handle different configurations, parameters, etc and the way they are launched are a little too complicated to me when I want to isolate problems quickly.

PR #2453, opened 2020-09-03
H5 vol 

guj:


chuckatkins:
Was there a need to write the whole thing in C rather than C++ and just `extern "C"` the api that hdf needs?

chuckatkins:
@guj it looks like when rebased on the current master / release there's a build failure.  If you can resolve that issue then I can work through the rest of the code and cmake re-organization.

guj:
> @guj it looks like when rebased on the current master / release there's a build failure. If you can resolve that issue then I can work through the rest of the code and cmake re-organization.

done.

chuckatkins:
@guj Reorg of build complete.  Please take a look.  The example you had was too brittle to use as an actual test so please add some proper tests in `testing/h5vol`; you can test for it's enabling via the CMake variable `ADIOS2_HAVE_HDF5_VOL`.

chuckatkins (approved):

guj:
> @guj Reorg of build complete. Please take a look. The example you had was too brittle to use as an actual test so please add some proper tests in `testing/h5vol`; you can test for it's enabling via the CMake variable `ADIOS2_HAVE_HDF5_VOL`.

done!

PR #2456, opened 2020-09-10
auto call BeginStep in SscReader::Close to match collective calls in SscWriter::Close

JasonRuonanWang:


pnorbert:
I did not look at this yet but it does not sound right. It sounds like a
hack.
An extra BeginStep in reader should be called and EndOfStream status should
be checked in a proper code.

JasonRuonanWang:
@pnorbert Actually, this fixed the iotest effis cheetah workflow bug... 

So I would suppose that in that workflow, iotest actually does not call the extra Reader::BeginStep. That's the only reasonable explanation why this little hack solved everything...

JasonRuonanWang:
@pnorbert But in any way, this little hack does not affect anything that's already working, while being able to fix the iotest thing.

Issue #2472, opened 2020-09-29
`adios2_steps` returns 0


hr87:
**Describe the bug**
Fortran binding for `adios2_steps` returns the number 0, even though there are 14 steps in the file (checked with `bpls`)

**To Reproduce**
```Fortran
program test
    use adios2, only: adios2_adios, &
                      adios2_io, &
                      adios2_engine &
                      adios2_init, &
                      adios2_declare_io, &
                      adios2_open, &
                      adios2_steps, &
                      adios2_mode_read

    type(adios2_adios) :: adios
    type(adios2_io) :: adios_io
    type(adios2_engine) :: adios_engine
    integer(8) :: steps
    integer :: adios_return
    logical, parameter :: adios_debug_mode = .True.
    character(len=*), parameter :: 'test_input'

    call adios2_init(adios, adios_debug_mode, adios_return)
    call adios2_declare_io(adios_io, adios, 'lo_system_reader', adios_return)
    call adios2_open(adios_engine, adios_io, restart_file, adios2_mode_read, adios_return)
    call adios2_steps(steps, adios_engine, adios_return)

    if (adios_return /= 0) then
        print *, 'ERROR'
    else
        print *, steps
    end if

end program test
```

**Expected behavior**
The routine should return the actual number of steps, or an exception if it can't (e.g. streaming mode)

**Desktop:**
LANL HPC Snow:   gcc/8.3.0   cmake/3.17.2  
No-MPI version

pnorbert:
This should be now fixed in the master branch.

PR #2473, opened 2020-09-29
Redirect f2c to correct call

williamfgc:
Addresses #2472    @pnorbert  looks like a terrible oversight from my side, the Fortran bindings need more testing on the read side.

### DIFF
--- a/bindings/Fortran/f2c/adios2_f2c_engine.cpp
+++ b/bindings/Fortran/f2c/adios2_f2c_engine.cpp
@@ -53,7 +53,7 @@ void FC_GLOBAL(adios2_steps_f2c, ADIOS2_STEPS_F2C)(int64_t *steps,
 {
     *steps = -1;
     size_t stepsC;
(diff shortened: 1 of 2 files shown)


pnorbert:
I am playing with the example provided in #2472 and there is a test for it, Bindings.Fortran.BPWriteTypes.Serial`
However, the example fails on that test's output (0 steps), while the test itself passes (3 steps)
So this PR will fix that?

pnorbert:
Yes, it fixes it. Thanks.

pnorbert (approved):

pnorbert:
Ahh, the test just prints the nsteps if it is not correct, but does not fail.

pnorbert:
adios2/bindings/fortran/TestBPWriteTypes.F90: 234
Can you fix the test as well for this PR?

williamfgc:
@pnorbert great catch. Now, I'm wondering what exactly I was trying to do :)

PR #2486, opened 2020-10-09
Get function in python binding to read a string

dmitry-ganyushin:
Fix for issue 2338: added a function to read a string in python full API


williamfgc:
@dmitry-ganyushin thanks for looking into it as it's a long standing issue. I'd add tests to make sure the C++/Python translation works in the low-level API.

dmitry-ganyushin:
Sure, I will add reading/writing tests later

PR #2490, opened 2020-10-19
Fortran: Fix submodule and preprocessor issues with Cray Fortran

chuckatkins:


bindings/Fortran/modules/adios2_adios_init_mod_mpi.F90
```diff
@@ -7,9 +7,9 @@
 !
 
 #ifdef ADIOS2_HAVE_FORTRAN_SUBMODULES
-# define ADIOS2_MODULE_PROCEDURE module &
+# define ADIOS2_MODULE_SUBROUTINE module subroutine
 #else
-# define ADIOS2_MODULE_PROCEDURE
+# define ADIOS2_MODULE_SUBROUTINE subroutine
```
germasch, line 12: Seems like an opportunity to consolidate this repeated snippet?

bindings/Fortran/modules/adios2_adios_init_mod_mpi.F90
```diff
@@ -33,8 +33,7 @@ module adios2_adios_init_mod_mpi
 
 contains
 
-    ADIOS2_MODULE_PROCEDURE
-    subroutine adios2_init_mpi(adios, comm, adios2_debug_mode, ierr)
+    ADIOS2_MODULE_SUBROUTINE adios2_init_mpi(adios, comm, adios2_debug_mode, ierr)
```
germasch, line 36: I guess I don't know whether this was needed (ie., whether there were problems with the continuation line?). But I'm pretty sure this is going to confuse editors that indent Fortran `subroutine` .. `end subroutine`.

I just tried, and yeah, at least emacs ends up being confused. It's a minor issue, but maybe worth considering.

chuckatkins: No issue with the continuation but the cray fortran compiler specifically has problems with the empty macro when submodules are disabled.

chuckatkins: The issue seems to be having an empty macro on a line by itself.  Does this work better in the editor?  It just works around the issue in a different way:
```fortran
#ifdef ADIOS2_HAVE_FORTRAN_SUBMODULES
# define ADIOS2_MODULE_PREFIX module
#else
# define ADIOS2_MODULE_PREFIX
#endif
...
contains
    ADIOS2_MODULE_PREFIX subroutine adios2_open_new_comm(engine, io, name, adios2_mode, comm, ierr)
...
    end subroutine
```

efdazedo:
Is it possible to not use "mpi" as the name of the sub-module but use say "adios2_mpi" instead? 
Will it break user application code with this change?

The submodule name "mpi" may conflict with MPI module "mpi".


./bindings/Fortran/modules/adios2_adios_init_mod_mpi.F90:submodule ( adios2_adios_init_mod ) **mpi**   
./bindings/Fortran/modules/adios2_io_open_mod_mpi.F90:submodule ( adios2_io_open_mod ) **mpi**

bindings/Fortran/modules/adios2_adios_init_mod_mpi.F90
germasch: `ADIOS2_MODULE_PREFIX subroutine ...` works for me -- I like it ;)

chuckatkins:
> Is it possible to not use "mpi" as the name of the sub-module but use say "adios2_mpi" instead?
> Will it break user application code with this change?
> 
> The submodule name "mpi" may conflict with MPI module "mpi".
> 
> ./bindings/Fortran/modules/adios2_adios_init_mod_mpi.F90:submodule ( adios2_adios_init_mod ) **mpi**
> ./bindings/Fortran/modules/adios2_io_open_mod_mpi.F90:submodule ( adios2_io_open_mod ) **mpi**

The submodules aren't user-facing, it's an internal implementation detail, essentially private modules for adios2, so a user shouldn't ever encounter a conflict.  The reason it shows up for cray fortran is because the module files get generated only containing the name of the submodule and not the parent module, i.e. instead of `adios2_adios_init_mod@mpi.smod` you just get `MPI.mod`.  If the module file naming for cray fortran encoded the parent module name in the file then this would be a non-issue.  Even if the module name were to change to something else (still not user-facing), something like `mpi_mod` and `serial_mod`, we'd still run into the other problem of both the `adios2_adios_init_mod@mpi_mod` and `adios2_io_open_mod@mpi_mod` having the same submodule name and thus filename for cray-fortran, `MPI_MOD.mod`, and one of the two module files will overwrite the other or just get corrupted in the build.  The least invasive way to work around the cray fortran issue is to simply disable the use of submodules for cray fortran.

efdazedo:
If the sub module name is prefixed by the parent module name then we can avoid this problem? The sub module name is not visible to the user program so perhaps the developer can use a longer name?

bradking (dismissed):
I don't recall for sure the reason I originally put the line continuation in the macro rather than putting it at the beginning of the line at the usage sites as proposed here.  It might have been cosmetic, but it might have had something to do with CMake's module dependency extraction.  Please check that rebuilds still work with both Ninja and Makefiles.

bindings/Fortran/modules/adios2_adios_init_mod_mpi.F90
chuckatkins: Perhaps, though I'd rather save re-structuring for the sake of cleanup to a separate issue

chuckatkins:
> Please check that rebuilds still work with both Ninja and Makefiles.

AFAICT it's working like it's supposed to.

chuckatkins:
@efdazedo I was initially opposed to renaming the submodules because it seemed like a kludgey workaround for an oddball implementation of submodule file naming in a specific compiler.  The more I thought about it though it the more it makes sense to do so given how submodules work in fortran, i.e. it's really separating out the implementation and not having anything to do with a hierarchical data model.  I also dug around a bit and it seems like encoding the full module set in a submodule name is pretty common practice, so I went ahead and made the associated change.

bradking (approved):

chuckatkins:
@efdazedo This should now have all the appropriate changes to build with Cray FTN.  Please try it out to confirm that it works for you.

PR #2502, opened 2020-11-02
Added a unittest for reading a string with high-level API in python

dmitry-ganyushin:


chuckatkins:
The pgi builds have been removed in master.  Please rebase.

PR #2503, opened 2020-11-02
Fortran: Fix module install rules for Cray Fortran

chuckatkins:


chuckatkins:
@efdazedo This should fix the bad Fortran flags you're getting from `adios2-config`

chuckatkins:
Blocked on #2501

chuckatkins:
@pnorbert @eisenhauer any idea why we'd be getting such different floating point numbers?

https://open.cdash.org/test/279486827

eisenhauer:
In a word, no.  But we also got something like this on my FFS-related PR (https://open.cdash.org/test/279513230).  I went ahead and merged that because whatever it was, it was happening in BP marshalling and the FFS PR didn't touch that stuff.  But now you're seeing something similar in an FFS marshalling case.  So either it's something in SST (the chunks of zeroes are sspicious), or something numerical in heatTransfer that we just haven't seen before (how SST seems unlikely to produce close-but-not-quite right floats in two different marshaling methods), or something external to all of it.

It look like this PR does *not* contain any of the FFS changes that I merged this morning, so whatever is going on just happened to hit that PR as well as this one.  I have no solid guesses as to what that might be.  @pnorbert ?

eisenhauer:
So, I tried a PR to diagnose the HeatTransfer issues and got 7 completely different failures, mostly MPI failures where it dies with a out-of-memory error (e.g.  https://open.cdash.org/test/280352900). I think we have to consider the likelihood that this is a problem with this VM, not with the code.

chuckatkins:
> So, I tried a PR to diagnose the HeatTransfer issues and got 7 completely different failures, mostly MPI failures where it dies with a out-of-memory error (e.g. https://open.cdash.org/test/280352900).

That was particularly insightful.  The default size of `/dev/shm` for docker containers is 64mb.  Increasing it to 1g (I believe the containers get 4g or 8g of ram) seems to have resolved the issue.

PR #2506, opened 2020-11-03
CI: Remove PGI builds from CI

chuckatkins:
The licensing model for PGI has changed and community licenses are no
longer available.  This removes PGI from the CI pipeline until we can
come up with a way to enable it.


chuckatkins:
Being addressed in #2501

PR #2508, opened 2020-11-05
Added "std::vector<size_t> GetAbsoluteSteps(const Variable<T> variabl…

pnorbert:
…e) const" to the Engine class as requested by OpenPMD devs.


pnorbert:
@ax3l @franzpoeschel This was easy to provide per variable assuming you have the Variable<T> object from InquireVariable()

PR #2509, opened 2020-11-05
Update ci images and re-enable PGI

chuckatkins:


chuckatkins:
@eisenhauer I've got this cleaned up with only 2 files now in sst generating PGI warnings.  Can you please take a look and push a fix to this branch?

https://open.cdash.org/viewBuildError.php?type=1&buildid=6872771

chuckatkins:
Thanks for the help @eisenhauer this looks good.  All the zfp stuff is broken of course, but that's unrelated to sst.  I can debug from here.

chuckatkins:
Looks like zfp is actually broken with current PGI.  I've fixed our images to use gcc specifically for zfp until the issue is fixed.  See LLNL/zfp#112.

PR #2516, opened 2020-11-17
Do not shrink serialization buffer upon EndStep – Fix Performance Bug

franzpoeschel:
The BP4 serialization engine maintains the same serialization buffer `format::BP4Serializer` across steps. This allows to avoid heavyweight allocation operations after a certain "warmup" phase. Methods such as [`format::BPBase::ResizeBuffer`](https://github.com/ornladios/ADIOS2/blob/addcb471f0341c93ba7a5f99a1a58a571a374f8a/source/adios2/toolkit/format/bp/BPBase.cpp#L305) support this idea by taking care not to shrink the buffer.

The method [`format::BP4Serializer::SerializeDataBuffer`](https://github.com/ornladios/ADIOS2/blob/addcb471f0341c93ba7a5f99a1a58a571a374f8a/source/adios2/toolkit/format/bp/bp4/BP4Serializer.cpp#L375) (called upon `Engine::EndStep`) does, however, pay no such attention.

This has noticeable consequences in writing to ADIOS2 via openPMD in PIConGPU. (Possibly related to the fact that the openPMD plugin in PIConGPU will repeatedly call `Engine::PerformPuts`, leading to a low buffer "fill status" upon `Engine::EndStep`, resulting in an actual shrinking of the buffer). While the buffer is never actually reallocated (which would happen only upon `std::vector::shrink_to_fit`), resizing it back up to the needed size in the following step involves zero-initialization. This means that ADIOS spends a lot of time per step just filling gigabytes of main memory with zeros. Quick measurements suggest a speedup down from ~1 minute for a data dump to ~30 seconds after fixing this.

This PR quickly fixes the issue by adding conditionals to avoid shrinking the buffer.

### DIFF
--- a/source/adios2/toolkit/format/bp/bp4/BP4Serializer.cpp
+++ b/source/adios2/toolkit/format/bp/bp4/BP4Serializer.cpp
@@ -372,13 +372,24 @@ void BP4Serializer::SerializeDataBuffer(core::IO &io) noexcept
     if (attributesSizeInData)
     {
         attributesSizeInData += 12; // count + length + end ID
-        m_Data.Resize(position + attributesSizeInData + 4,
-                      "when writing Attributes in rank=0\n");
+        // Take care not to shrink the buffer size in this.
+        // Otherwise, growing the buffer again later on is expensive
+        // - not because of reallocation, but because of (re)initialization
+        // with zeros by BufferSTL::Resize.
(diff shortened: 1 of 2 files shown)


pnorbert (approved):

ax3l:
Great find and fix, @franzpoeschel ! Thank you! :rocket: :sparkles:

PR #2496, opened 2020-10-24
update mgard due to its API change

qliu21:

PR #2517, opened 2020-11-17
update mgard due to its API change

eisenhauer:
Essentially a rebase of #2496 (rebasing that directly doesn't work well because of the thirdparty updates via update.sh)


qliu21:
please merge and close.

PR #2574, opened 2021-01-07
Added a description of the Group API and step selection.

dmitry-ganyushin:
Added documentation for the Group API and step selection


dmitry-ganyushin:
Change the parameter name to "FilterSteps".

PR #2579, opened 2021-01-11
utils: Simplified wrappers for adios2_reorganize

chuckatkins:


source/utils/adios_reorganize/adios2_reorganize_wrapper
```diff
@@ -0,0 +1,30 @@
...
+fi
+
+input="$1"
+output="$2"
+base_dir="$(dirname $(readlink -f "${BASH_SOURCE}"))"
+wrapper=$(basename "$0")
+case ${wrapper} in
+  bp2h5@ADIOS2_EXECUTABLE_SUFFIX@)
+    reorg_tool="${base_dir}/adios2_reorganize@ADIOS2_EXECUTABLE_SUFFIX@"
+    in_engine=BPFile
```
pnorbert, line 16: Can you please use FileStream? That will  properly read variables and attributes popping up later so it should be a bit more portable for all cases than just file-based reader. Optionally, if you want to enforce BP file reading, you can use BPFile and add an option "StreamReader=ON"

chuckatkins: Fixed.  Given that the wrapper is "bp2h5" then I do want to enforce the BP engine.  We could easily use this same mechanism to add other wrappers in the future for pre-baked configurations.

pnorbert: Okay but we need the same option for the bp2h5_mpi tool as well

chuckatkins: Doh!  Missed that.

chuckatkins: fixed

PR #2584, opened 2021-01-13
Add threading into ssc writer to optimize first step metadata manipulation

JasonRuonanWang:


JasonRuonanWang:
@chuckatkins @pnorbert 

Has anyone seen this error before? It does not seem to be something caused by this PR itself.

  Could not find compiler set in environment variable FC:

  gfortran.
  cmake/DetectOptions.cmake:140 (enable_language)
  CMakeLists.txt:137 (include)

chuckatkins:
There's an issue with the Mac build right now.  I'll have it fixed shortly

chuckatkins:
Fixed.  Please rebase on master.

PR #2588, opened 2021-01-16
Add threading to SSC reader

JasonRuonanWang:


JasonRuonanWang:
This, together with #2584, will help apps that have considerable computing time between every step, such as WarpX, hide data transfer time behind computing time, and thus improve the performance. 

It requires that every app in the workflow only opens a single SSC stream.

Issue #2018, opened 2020-03-03
C-Blosc: Support for Large Vars


ax3l:
**Describe the bug**

Writing for individual variables with c-blosc (1) that write very large data fails, since c-blosc has a 2Gbyte limit.

**To Reproduce**
Based on #2010, use this diff
```diff
diff --git a/testing/adios2/performance/largevar/TestLargeVarWrite.cpp b/testing/adios2/performance/largevar/TestLargeVarWrite.cpp
index 59d4ccad..4c1acef6 100644
--- a/testing/adios2/performance/largevar/TestLargeVarWrite.cpp
+++ b/testing/adios2/performance/largevar/TestLargeVarWrite.cpp
@@ -4,6 +4,7 @@
  */
 
 #include <memory>
+#include <random>
 #include <stdexcept>
 #include <string>
 
@@ -29,6 +30,11 @@ TEST_F(Write, Plain1D_1B)
 
     std::unique_ptr<float[]> dataBig(new float[Nx]);
 
+    auto const seed = std::random_device{}();
+    std::default_random_engine gen{ seed };
+    std::uniform_real_distribution< float > dist( 0.0f, 1.0f );
+    std::generate_n( dataBig.get(), Nx, [&]{ return dist( gen ); } );
+
 #ifdef ADIOS2_HAVE_MPI
     MPI_Comm_rank(MPI_COMM_WORLD, &mpiRank);
     MPI_Comm_size(MPI_COMM_WORLD, &mpiSize);
@@ -53,6 +59,17 @@ TEST_F(Write, Plain1D_1B)
 
     auto varBig = io.DefineVariable<float>("varBig", shape, start, count,
                                            adios2::ConstantDims);
+    // add operations
+    adios2::Operator BloscOp =
+        adios.DefineOperator("BloscCompressor", adios2::ops::LosslessBlosc);
+
+    varBig.AddOperation(BloscOp,
+                         {{adios2::ops::blosc::key::compressor, "zstd"}});
+    varBig.AddOperation(BloscOp,
+                         {{adios2::ops::blosc::key::clevel, "1"}});
+    varBig.AddOperation(BloscOp,
+                         {{adios2::ops::blosc::key::doshuffle, "BLOSC_BITSHUFFLE"}});
+    // FIXME how to set the compressor threads?
 
     adios2::Engine writer = io.Open(fname, adios2::Mode::Write);
```

And execute. Repeat with a smaller array size in `Nx`, e.g. divide by `4`.

[...]

PR #2592, opened 2021-01-20
improve Blosc compression operator

psychocoderHPC:
fix: #2018

The current implementation of the Blosc ADIOS2 operator is due to limitations of Blosc not supporting adios variables larger than 2GiB.
ADIOS1 was originally using a chunked format to work around this limitation.

- Add option `threshold` to set a threshold under which data will only be copied instead of compressed (equivalent to Blosc in ADIOS1).
- Use chunk format to avoid 2GiB limit for variables (datasets).
- Add support to decompress Blosc data written with ADIOS before this PR.
- - update blosc tests

The header written to decompressed the date is different from the ADIOS1 implementation. 
Compared to ADIOS1 this PR implements compression and decompression much cleaner.

@maintainer Where do I should document the new option `threshold` for the Blosc operator. I have not found anything in the documentation.

Thanks to @franzpoeschel for providing the test code [testBlosc.txt](https://github.com/ornladios/ADIOS2/files/5842414/testBlosc.txt)
.
CC-ing: @ax3l

### DIFF
--- a/source/adios2/operator/compress/CompressBlosc.h
+++ b/source/adios2/operator/compress/CompressBlosc.h
@@ -6,6 +6,7 @@
  *
  *  Created on: Jun 18, 2019
  *      Author: William F Godoy godoywf@ornl.gov
+ *              Rene Widera r.widera@hzdr.de
  */
 
 #ifndef ADIOS2_OPERATOR_COMPRESS_COMPRESSBLOSC_H_
@@ -34,12 +35,15 @@ public:
 
     ~CompressBlosc() = default;
 
+    size_t BufferMaxSize(const size_t sizeIn) const final;
+
     /**
      * Compression signature for legacy libraries that use void*
      * @param dataIn
      * @param dimensions
      * @param type
-     * @param bufferOut
+     * @param bufferOut format will be: 'DataHeader ; (BloscCompressedChunk |
+     * UncompressedData), [ BloscCompressedChunk, ...]'
      * @param parameters
      * @return size of compressed buffer in bytes
      */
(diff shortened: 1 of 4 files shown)


pnorbert:
Thank you for fixing this compressor. Unfortunately, you caught us with the docs, there is no section on operators/compressors. 
@chuckatkins Can you please add redirect this to the patch release?

williamfgc:
@psychocoderHPC thanks, that was long overdue, only bzip2 was doing that and the hope was to migrate to blosc2 eventually as in [this issue](https://github.com/Blosc/c-blosc/issues/67#issuecomment-263234555). @pnorbert there is almost nothing in the [docs](https://adios2.readthedocs.io/en/latest/components/components.html#operator) on compression operations since each engine supports its own set.

psychocoderHPC:
> @psychocoderHPC Can you please extend the tests in `testing/adios2/engine/bp/operations/TestBPWriteReadBlosc.cpp` so that the threshold settings are tested? You need to look at this test anyway as the memory sanitizer fails on it...

I updated the blosc tests and added test combinations for `shuffling` and `threshold`

source/adios2/operator/compress/CompressBlosc.h
```diff
@@ -60,6 +64,50 @@ class CompressBlosc : public Operator
...
+                                   void *dataOut, const size_t sizeOut,
+                                   Params &info) const;
+
+    /** Decompress data written before ADIOS2 supported large variables larger
+     * 2GiB. */
+    size_t DecompressOldFormat(const void *bufferIn, const size_t sizeIn,
+                               void *dataOut, const size_t sizeOut,
+                               Params &info) const;
+
+    class __attribute__((packed)) DataHeader
```
psychocoderHPC, line 80: @pnorbert I fixed hopefully the `ubsan` issue in the CI by disallowing the compiler to add padding. 
The disadvantage is that there is no standard attribute in C++ :-(
Is this way ok? If not, do you have a suggestion on how I can fix this issue instead?

psychocoderHPC:
>  that and the hope was to migrate to blosc2 eventually as in [this issue](https://github.com/Blosc/c-blosc/issues/67#issuecomment-263234555).

I also hoped to switch to blosc2 and it is solving the 2GiB issue but unluckily as I know they now support larger size-types but you need to explicitly pack the data with the framing API and there is no simple function available which is doing it transparently.
Therefore I decided to stay with the well-tested approach from ADIOS1 with blosc1.

source/adios2/operator/compress/CompressBlosc.h
williamfgc: if it helps, there is an [internal helper function](https://github.com/ornladios/ADIOS2/blob/master/source/adios2/helper/adiosMemory.h#L244) to add padding in a portable way.

psychocoderHPC: I am not sure if this helps. The problem is that maybe the ptr for the output is not aligned during the compression. If I use the helper and shift my data few bytes and later during decompress I get the data within a array with a different base address. If so it would be not possible to find the data because the required padding is different.

If I understand the ubsan error for the old code correctly and what I found in the net the packed attribute is required to guarantee that the struct has the same layout even if you compile blosc on a different system. This attributes forbits the compiler to transform the struct which is IMO correct.

I can also remove the header class und work with two pure integers but I found it better readable to use the header class because it allowes in the code to shift with `sizeOf(Header)` instead using a magic number 8byte.

I am not sure which solution guarantee portability.

psychocoderHPC: Maybe it is enough if I align the struct to 4 byte.
Is ADIOS guaranteeing that my pointers given to compress operators are 4 byte aligned?

chuckatkins:
I've rebased the topic on release_27 as a candidate for the 2.7.1 patch release.  Please update your local branch with the remote.

source/adios2/operator/compress/CompressBlosc.h
williamfgc: > The problem is that maybe the ptr for the output is not aligned during the compression. 

The function I shared can verify that returning the required padding in bytes (if 0 no padding is required).

> Is ADIOS guaranteeing that my pointers given to compress operators are 4 byte aligned?

That's a good question. It depends on the serialization buffer consuming this interface. The BP buffer is `char` so it shouldn't be a problem. I'd check with @pnorbert for other existing strategies.

> I am not sure which solution guarantees portability.

Yup, that's the challenge.

chuckatkins (approved):

source/adios2/operator/compress/CompressBlosc.h
ax3l: It looks like Blosc is not covered by CI and thus breaking on MSVC: #2745

PR #2618, opened 2021-02-12
BP4 engine: Resize data buffer to its final content size after aggreg…

pnorbert:
…ation, which modifies the size for receiving other process' buffer. That size may be very different in case processes have vastly different data sizes.


pnorbert:
New E3SM manual block-merging across processes leads to vastly different buffer sizes, 10GB vs 26KB on some processes. After aggregation, the process with 10GB buffer is resized to 26KB which then causes a lot of small resizes as, in the next step, the 500 variables are put in. This PR resets the buffer to whatever that process was actually using.

pnorbert:
@tkurc This PR should fix your problem with aggregation

Issue #2620, opened 2021-02-15
Char type is not portable


franzpoeschel:
**Describe the bug**

A variable or attribute defined by `IO::DefineVariable<char>()` / `IO::DefineAttribute<char>()` will be converted to ADIOS2 types `int8_t` or `uint8_t` at least in engines SST and BP4. This makes the `char` type non-portable between platforms where `char` has a different signedness.

[CppReference](https://en.cppreference.com/w/cpp/language/types) on `char`:
> has the same representation and alignment as either signed char or unsigned char, but is always a distinct type

Unlike other implementation-defined C++ types such as `int`, `char` is not an alias for a more specific type:
```cpp
std::is_same<int, int32_t>::value // -> true on some platforms
std::is_same<char, unsigned char>::value // -> false on all platforms
std::is_same<char, signed char>::value // -> false on all platforms
```
Because of this, and also since `char` variables may be used to store strings in a variable (see `bpls -S`), making the `char` type portable would be a sensible addition to ADIOS2.

**To Reproduce**
Writer:
```cpp
#include <adios2.h>
#include <iostream>
#include <string>
#include <vector>

int main()
{
    std::string engine_type = "bp4";

    adios2::ADIOS adios;
    adios2::IO IO = adios.DeclareIO( "IO" );
    IO.SetEngine( engine_type );
    adios2::Engine engine = IO.Open( "strings.bp", adios2::Mode::Write );

    engine.BeginStep();

    std::string scalarStringValue{ "some string" };
    auto scalarString = IO.DefineVariable< std::string >( "scalarString" );
    engine.Put( scalarString, scalarStringValue );

    char scalarCharValue = 'a';
    auto scalarChar = IO.DefineVariable< char >( "scalarChar" );
    engine.Put( scalarChar, scalarCharValue );

    std::vector< char > arrayCharValue( 10, 'a' );
    auto arrayChar =
        IO.DefineVariable< char >( "arrayChar", { 10 }, { 0 }, { 10 } );
    engine.Put( arrayChar, arrayCharValue.data() );

    IO.DefineAttribute< char >( "charAttribute", 'a' );
    IO.DefineAttribute< std::string >( "stringAttribute", "some string" );

    engine.EndStep();
    engine.Close();
}
```
Reader:
```cpp
#include <adios2.h>
#include <iostream>
#include <string>
#include <vector>

int main()
{
    std::string engine_type = "bp4";

    adios2::ADIOS adios;
[...]

PR #2627, opened 2021-02-18
Add 'char' to supported types. Only added to C++ and tested with BP e…

pnorbert:
…ngines. Python can read it back but converts it to numpy single-byte integer type.

Implements request in #2620.

```
$ bin/Test.Engine.BP.WriteReadADIOS2.Serial "--gtest_filter=BPWriteReadTestADIOS2.ADIOS2BPWriteRead1D8" "BP4"
Note: Google Test filter = BPWriteReadTestADIOS2.ADIOS2BPWriteRead1D8
[==========] Running 1 test from 1 test suite.
[----------] Global test environment set-up.
[----------] 1 test from BPWriteReadTestADIOS2
[ RUN      ] BPWriteReadTestADIOS2.ADIOS2BPWriteRead1D8
[       OK ] BPWriteReadTestADIOS2.ADIOS2BPWriteRead1D8 (6 ms)
[----------] 1 test from BPWriteReadTestADIOS2 (6 ms total)

[----------] Global test environment tear-down
[==========] 1 test from 1 test suite ran. (7 ms total)
[  PASSED  ] 1 test.
$ ./bin/bpls ADIOS2BPWriteRead1D8.bp/ -ld ch -n 8
  char      ch       3*{8} = B / z
    (0,0)    b c d z a B C D
    (1,0)    c d e a b C D E
    (2,0)    d e f b c D E F
```

### DIFF
--- a/source/adios2/common/ADIOSTypes.inl
+++ b/source/adios2/common/ADIOSTypes.inl
@@ -79,6 +79,16 @@ struct TypeInfo
     using ValueType = T;
 };
 
+// Hack "char" type into this convoluted struct definition which is the key
+// to translate between bindings API types to supported stdtypes
+template <>
+struct TypeInfo<char,
+                typename std::enable_if<std::is_integral<char>::value>::type>
+{
+    using IOType = char;
+    using ValueType = char;
+};
+
 template <typename T>
 struct TypeInfo<T, typename std::enable_if<std::is_integral<T>::value>::type>
 {
--- a/source/adios2/common/ADIOSTypes.inl
+++ b/source/adios2/common/ADIOSTypes.inl
@@ -83,7 +83,7 @@ struct TypeInfo
 // to translate between bindings API types to supported stdtypes
 template <>
 struct TypeInfo<char,
-                typename std::enable_if<std::is_integral<char>::value>::type>
+                typename std::enable_if<std::is_same<char, char>::value>::type>
 {
     using IOType = char;
     using ValueType = char;
--- a/source/adios2/common/ADIOSMacros.h
+++ b/source/adios2/common/ADIOSMacros.h
@@ -33,6 +33,7 @@
  </pre>
 */
 #define ADIOS2_FOREACH_ATTRIBUTE_PRIMITIVE_STDTYPE_1ARG(MACRO)                 \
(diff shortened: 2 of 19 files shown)


pnorbert:
The hack to add `char` was to avoid template-matching it as an integral type to `uint8_t` or `int8_t` between the binding and the core `IO.DefineVariable()` functions. I "solved" this by adding one more funny TypeInfo struct in `adiosTypes.inl`

```
template <>
struct TypeInfo<char,
                typename std::enable_if<std::is_integral<char>::value>::type>
{
    using IOType = char;
    using ValueType = char;
};
```

source/adios2/common/ADIOSTypes.inl
```diff
@@ -79,6 +79,16 @@ struct TypeInfo
     using ValueType = T;
 };
 
+// Hack "char" type into this convoluted struct definition which is the key
+// to translate between bindings API types to supported stdtypes
+template <>
+struct TypeInfo<char,
+                typename std::enable_if<std::is_integral<char>::value>::type>
```
chuckatkins, line 86: I think you want to use `is_same` here instead of `is_integral`:
```
template<typename T>
struct TypeInfo<T, typename std::enable_if<std::is_same<T, char>::type>
{
  using IOType = T;
  using ValueType = T;
};

Comment (pnorbert) [reply source/adios2/common/ADIOSTypes.inl:86]: Your example leads to:

```
/home/adios/ADIOS2/bindings/CXX11/adios2/cxx11/Variable.h: In instantiation of ‘class adios2::Variable<char>’:
/home/adios/ADIOS2/bindings/CXX11/adios2/cxx11/Engine.h:453:1:   required from here
/home/adios/ADIOS2/bindings/CXX11/adios2/cxx11/Variable.h:130:11: error: ambiguous template instantiation for ‘struct adios2::TypeInfo<char, void>’
  130 |     using IOType = typename TypeInfo<T>::IOType;
```

because there is another definition that matches the T=char type. But this silly declaration works fine:

template <>
struct TypeInfo<char,
                typename std::enable_if<std::is_same<char, char>::value>::type>
{
    using IOType = char;
    using ValueType = char;
};

PR #2630, opened 2021-02-19
ci: Add LAMMPS in contract testing

chuckatkins:


chuckatkins:
@pnorbert please take a look

PR #2635, opened 2021-02-21
fixed mgard dimensions

JasonRuonanWang:
The dimensions in mgard operator should be set to [1,1,1] by default rather than [0,0,0]. Otherwise in use cases of less than 3 dimensional data, mgard library will throw exception "all dimensions must be positive".


qliu21:
I am fine with it. Just FYI, @JasonRuonanWang. MGARD is being changed drastically and I will update the adios side once it is released. stay tuned.

JasonRuonanWang:
> I am fine with it. Just FYI, @JasonRuonanWang. MGARD is being changed drastically and I will update the adios side once it is released. stay tuned.

Thanks for the info. I will merge this first because I have some use cases which need compression 2D data.

PR #2639, opened 2021-02-23
Address compile warnings across the project (Former: Enable -Wall)

chuckatkins:
Address compile warnings across the project

- Add multiple EXPECT_THROW in tests
- Corrects a integer overflow in InSituMPISchedules.cpp
- Corrects a printf modifier in H5VolUtil.cpp
- Enforce coding style across the project

Co-authored-by: Chuck Atkins <chuck.atkins@kitware.com>
Co-authored-by: Vicente Adolfo Bolea Sanchez <vicente.bolea@kitware.com>


guj:
OK Chuck, 

I tried my best to remove the hdf5 related warnings and I also corrected those. in BP tests along the way. I will leave the rest to you. Especially the ones in the adiosMemory.inl.   Looks like no one is using that function.   
Cheers,

source/h5vol/H5VolUtil.c
```diff
@@ -0,0 +1,232 @@
...
+    {
+        return NULL;
+    }
+
+    void *p = calloc(n, s);
+    if (!p)
+    {
+        fprintf(stderr, "[%s:%ld]Out of memory at calloc (%ld, %ld)\n",
+                __FILE__, line, (unsigned long)n, (unsigned long)s);
+        exit(EXIT_FAILURE);
```
chuckatkins, line 16: Do you really want to call `exit` here and terminate the entire program?

source/h5vol/H5VolUtil.c
```diff
@@ -0,0 +1,232 @@
...
+}
+
+void safe_free(void *p)
+{
+    if (p == NULL)
+    {
+        return;
+    }
+    free(p);
+    p = NULL;
```
chuckatkins, line 45: Since `*p` is passed by value then setting it to `NULL` doesn't actually do anything here.  The same effect would be achieved from
```c++
void safe_free(void *p)
{
    if (p)
    {
        free(p);
    }
}
```

chuckatkins (commented):
This is great @guj; Thank you!  The rest of the errors seem to be outside  the scope of HDF5 so I'll work through the rest.

source/h5vol/H5VolUtil.c
germasch: Random comment: `safe_free` does do exactly the same as `free` -- according to the C standard (at least since 89), `free()` does nothing if `NULL` is passed to it.

source/h5vol/H5VolUtil.c
```diff
@@ -0,0 +1,232 @@
...
+    {
+        return NULL;
+    }
+
+    void *p = malloc(n);
+    if (!p)
+    {
+        fprintf(stderr, "[%s:%ld]Out of memory at malloc (%ld bytes)\n",
+                __FILE__, line, (unsigned long)n);
+        exit(EXIT_FAILURE);
```
germasch, line 33: FWIW, since C99 one can use `(%zu bytes)` to print a `size_t` argument, then you don't need to cast. (I don't know if you need to support older standards, though.)

source/h5vol/H5VolUtil.c
guj: Thanks for the catch! I fixed it.

source/h5vol/H5VolUtil.c
guj: Well, yes, nothing will succeed when there is not enough memory

PR #2647, opened 2021-03-12
Added ADIOS() constructor in the CXX11 public API so that hostLanguag…

pnorbert:
…e can be passed. Needed by pbdR hola package to ensure data is treated column-major.

See https://github.com/RBigData/hola/blob/master/src/init.cpp#L27-L35 how it calls the C++11 constructor

### DIFF
--- a/bindings/CXX11/adios2/cxx11/ADIOS.cpp
+++ b/bindings/CXX11/adios2/cxx11/ADIOS.cpp
@@ -21,6 +21,12 @@ ADIOS::ADIOS(const std::string &configFile, const bool debugMode)
 
 ADIOS::ADIOS(const bool debugMode) : ADIOS("", "C++") {}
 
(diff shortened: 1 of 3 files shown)


pnorbert:
@wrathematics Let me know if this is working for hola. You need to pass an empty string "" now in all cases, and pass true or false as 4th argument to ensure the new signature is matched.
```
    if (config == R_NilValue)
      ad = new adios2::ADIOS("", comm, "R", false);
    else
    {
      std::string config_string = CHARPT(config, 0);
      ad = new adios2::ADIOS(config_string, comm, "R", false);
    }
```

wrathematics:
Is this change going to be added to the existing 2.7.1 release? If not, could you go ahead and bump the version? I added a configure test for minimum version and I need to know what to use.

PR #2654, opened 2021-03-23
adios_iotest/settings: avoid an off-by-one

mathstuf:
`argCopy`'s allocation did not save space for the `NUL` terminating
byte. Instead of manually calculating memory sizes, use `strdup`.

### DIFF
--- a/source/utils/adios_iotest/settings.cpp
+++ b/source/utils/adios_iotest/settings.cpp
@@ -88,12 +88,16 @@ size_t Settings::stringToNumber(const std::string &varName,
     return retval;
 }
 
+#ifdef _WIN32
+#define strdup _strdup
+#endif
+
 int Settings::parseCSDecomp(const char *arg)
 {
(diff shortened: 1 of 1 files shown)


chuckatkins:
@mathstuf can you please address the msvc failure?

mathstuf:
The logs seem to have expired. Is there a way to rerun just the failed bits?

mathstuf:
I can try and rebase I guess; `master` looks OK.

source/utils/adios_iotest/settings.cpp
```diff
@@ -88,12 +88,16 @@ size_t Settings::stringToNumber(const std::string &varName,
     return retval;
 }
 
+#ifdef _MSC_VER
```
chuckatkins, line 91: should be `#ifdef _WIN32` as it's a windows issue, not an msvc issue.

chuckatkins (changes requested):

chuckatkins (approved):

PR #2679, opened 2021-04-20
Added a test for the issue 2500

dmitry-ganyushin:
Added a test to check that an exception is thrown if InquireVariable is done outside the BeginStep/EndStep block.


dmitry-ganyushin:
Closed in favor of the PR#2682

PR #2680, opened 2021-04-22
prevent cmake failure when no Fortran compiler is present

germasch:
`ADIOS2_USE_Fortran` is `AUTO` by default, so even if no Fortran compiler was found, it'd still try to check a Fortran compiler flag, which failed without a Fortran compiler being enabled:

```
CMake Error at /opt/local/share/cmake-3.19/Modules/Internal/CheckCompilerFlag.cmake:52 (message):
  check_compiler_flag: Fortran: needs to be enabled before use.
```

The right thing to use is `ADIOS2_HAVE_Fortran`.


germasch:
@chuckatkins This is a pretty trivial fix ;)

PR #2681, opened 2021-04-23
fix build error due to missing <string> include

germasch:


germasch:
@eisenhauer: Maybe you could take a look at this PR, too? It has essentially zero chance of breaking anything, and it does fix a build error I encountered (on Mac OS, FWIW).

eisenhauer:
Looks good to me…

PR #2682, opened 2021-04-23
Added a test to address issue 2500

dmitry-ganyushin:
Check that the reader throws an exception if InquireVariable is outside of the BeginStep/EndStep Block in C++ with FileStream engine.


eisenhauer:
You confused me here.  I think you mean InquireVariable, not DefineVariable...

dmitry-ganyushin:
I corrected, thanks!

PR #2695, opened 2021-05-12
Tune linearizeindex

pnorbert:
The discussed read test went from 22.8s to 4.5s on my PC. 
The consequences are: 
- 24 as max dimension limit set in ADIOSTypes.h, and the first const array in ADIOSTypes.h named DimZeros.
- LinearIndex and EndToCount works on size_t * pointers and number of dimensions must be passed by the caller

### DIFF
--- a/source/adios2/common/ADIOSTypes.h
+++ b/source/adios2/common/ADIOSTypes.h
@@ -188,6 +188,12 @@ constexpr float DefaultBufferGrowthFactor = 1.05f;
  *  2Gb - 100Kb (tolerance)*/
 constexpr size_t DefaultMaxFileBatchSize = 2147381248;
 
+/** maximum number of dimensions supported */
+constexpr size_t MaxDimensions = 24;
+
+static const size_t DimZeros[MaxDimensions] = {
+    0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0};
+
 constexpr char PathSeparator =
 #ifdef _WIN32
     '\\';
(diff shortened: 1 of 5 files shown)


eisenhauer:
is there any magic about 24?

pnorbert:
nothing about 24, it's just big enough to cover all known use cases and some.
LinearIndex() and its callers create an array of MaxDims on the stack instead of creating a vector and instead of malloc-ing an arbitrarily big array. Ugly but much faster. Any idea how to make this fast and generic?

dmitry-ganyushin:
I would test it on Summit first.
> The discussed read test went from 22.8s to 4.5s on my PC.
>

eisenhauer:
> Any idea how to make this fast and generic?
No, and I fully support some limit on dimensions, however large.  The stack allocation rather than malloc is a handy trick I've used before.  In some circumstances I've used stack allocation for moderate sized temporary structures, with a fallback to malloc if the structure got too large, but that does tend to complicate the code and shouldn't be necessary here.   Honestly, I can get some improvements in BP5 performance for special-casing 1 or 2 dimensional cases, which I have to think would cover a lot of our use cases.  (I haven't done this yet.)

germasch:
I kinda don't see anyone ever needing more than 24 dims (definitely not me).

One way to keep this general, though, would be to something akin to small string optimization for `std::vector`. Ie., Make something akin to `std::vector` but containing some storage inside the class for, e.g., up to 10 dims that can be used without any additional dynamic allocation overhead. In the rare case that someone really wants 30 dims, that vector could then fall back to dynamic allocation.

pnorbert:
Do not merge. Better implementation without constants coming.

PR #2706, opened 2021-05-17
add python requirements file to help automate delta installation

lastephey:
Hi ADIOS developers,

This PR adds a standard Python `requirements.txt` file to help automate installation

The specific use case here is to install Delta in a container in an automated way

Please let me know if you have any comments or questions

Thank you,
Laurie


JasonRuonanWang:
@chuckatkins @pnorbert 

We are trying to put ADIOS2 into our automated container installation of the delta-fusion application. The whole workflow is very much Python based, so having a standard python requirements file would make things much easier. I will merge this PR if there is no oppositions.

chuckatkins:
Does the file format allow for comments?  If so it would be good to have a single comment just saying what the file is.  It's obvious for python only packaged but since adios lives in many different packaging worlds it'd be good to have a comment in the file for the non-python world as to what it's for.

PR #2707, opened 2021-05-18
Fix clip contiguous memory

pnorbert:
Copy from read buffer to user target as much as possible at once instead of line-by-line. 
Only  fixing row-major reading so far, making a PR to test all platforms.


williamfgc:
Thanks @pnorbert this is one of the things that was never optimized and long overdue.

PR #2714, opened 2021-05-24
fix unresolved symbol errors when compiling with gcc10

germasch:
Building adios 2.6.0 with spack (2.6.0 being the latest/default version supported in spack), and using gcc-10.2 as compiler, I end up with

```
[ 83%] Linking CXX executable ../../bin/adios2_reorganize
cd /tmp/kaig1/spack-stage/spack-stage-adios2-2.6.0-sgqcjhbf5wd25edgurprcjrfuuephugn/spack-build-sgqcjhb/source/utils && /autof
s/nccs-svm1_sw/summit/.swci/0-core/opt/spack/20180914/linux-rhel7-ppc64le/gcc-4.8.5/cmake-3.20.2-24ualfzy6em6ws5sbiu7rlgcuiono
drm/bin/cmake -E cmake_link_script CMakeFiles/adios_reorganize.dir/link.txt --verbose=1
/autofs/nccs-svm1_home1/kaig1/build/spack/lib/spack/env/gcc/g++ -O3 -DNDEBUG CMakeFiles/adios_reorganize.dir/adios_reorganize/
main.cpp.o CMakeFiles/adios_reorganize.dir/adios_reorganize/Reorganize.cpp.o CMakeFiles/adios_reorganize.dir/Utils.cpp.o -o ..
/../bin/adios2_reorganize  -Wl,-rpath,/tmp/kaig1/spack-stage/spack-stage-adios2-2.6.0-sgqcjhbf5wd25edgurprcjrfuuephugn/spack-b
uild-sgqcjhb/lib64:/autofs/nccs-svm1_home1/kaig1/build/spack/opt/spack/linux-rhel7-power9le/gcc-10.2.0/zfp-0.5.5-xsflvsxe3y3we
ghpuigt7cuculofltkv/lib64: ../../lib64/libadios2_core.so.2.6.0 -lpthread -Wl,-rpath-link,/tmp/kaig1/spack-stage/spack-stage-ad
ios2-2.6.0-sgqcjhbf5wd25edgurprcjrfuuephugn/spack-build-sgqcjhb/lib64:/autofs/nccs-svm1_home1/kaig1/build/spack/opt/spack/linu
x-rhel7-power9le/gcc-10.2.0/zfp-0.5.5-xsflvsxe3y3weghpuigt7cuculofltkv/lib64
../../lib64/libadios2_core.so.2.6.0: undefined reference to `int adios2::format::DataManSerializer::GetData<long>(long*, std::
__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::vector<unsigned long, std::allocator<u
nsigned long> > const&, std::vector<unsigned long, std::allocator<unsigned long> > const&, unsigned long, std::vector<unsigned
 long, std::allocator<unsigned long> > const&, std::vector<unsigned long, std::allocator<unsigned long> > const&)'
../../lib64/libadios2_core.so.2.6.0: undefined reference to `int adios2::format::DataManSerializer::GetData<unsigned short>(un
signed short*, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::vector<unsigned lo
ng, std::allocator<unsigned long> > const&, std::vector<unsigned long, std::allocator<unsigned long> > const&, unsigned long,
std::vector<unsigned long, std::allocator<unsigned long> > const&, std::vector<unsigned long, std::allocator<unsigned long> >
const&)'
[...]
```
and similar errors for other template parameters to `GetData<>`. This PR fixes this error.

I'm not sure that adios-2.6 is being maintained, but in any case, with the PR here, I can reference it as a patch in the adios2 spack config to make things work.

For whatever reason, 2.7.1, if selected manually, installs fine via spack as-is.


germasch:
FWIW, the CI failures aren't this PR's fault, and they may be caused by the fact that this PR is against an old branch (I think it'd apply fine against current master, but since I don't see the problem there, I don't see a need to apply it there). Again, I don't necessarily care about it being merged into the old release branch, as I suppose there isn't going to be a 2.6.1 release anytime soon (or ever). 

But it'd be nice to okay this change here so that it can be referenced from the spack recipe since without it spack's latest release produces the error above when installing adios2's default version (which is 2.6.1 in spack 16.2) in my case (which is gcc-10 on summit).

JasonRuonanWang:
Just curious, any particular reason for sticking with 2.6 instead of moving to 2.7 or the latest master?

germasch:
> Just curious, any particular reason for sticking with 2.6 instead of moving to 2.7 or the latest master?

The reason is just that adios2-2.6.0 the latest (and default) version that the latest spack relase (0.16.2) installs -- and for me on Summit that fails. So my goal is to fix this issue in the spack recipe for adios2 (it allows to add patches, and this PR is what I'm using as the patch). It certainly wouldn't hurt to have the spack recipe updated to know about 2.7, but that's a separate issue, and I'm not really sure about spack's policy on updating default versions of their packages -- I've seen that adios2-2.7 is supported in spack's `develop` branch, but I don't really want to update my spack to an evolving development branch just to get an adios2 that builds.

chuckatkins:
rebased

germasch:
The thing about this is that it's against 2.6 intentionally, since that's where I got the unresolved symbols, while 2.7 worked fine (not that I understand why). The main intention of having this PR is just to have a patch url that I can reference in the spack package to fix the issue I had.

chuckatkins:
> The thing about this is that it's against 2.6 intentionally, since that's where I got the unresolved symbols, while 2.7 worked fine (not that I understand why). The main intention of having this PR is just to have a patch url that I can reference in the spack package to fix the issue I had.

Okay, I misunderstood I think.  I'll cycle this through today and get it merged along with the respective spack update.  Just ignore the churn you might see on this and I twiddle with git and ci things today.

chuckatkins:
Applied directly to `release_26`: https://github.com/ornladios/ADIOS2/commit/5a9b505bd82f8b918313d99b9521d0148ca01131

PR #2715, opened 2021-05-25
Blosc: Warn Unknown Parameters

ax3l:
Warn the user if unknown parameters are passed to the blosc compressor. Otherwise, typos get quickly unnoticed.

cc @psychocoderHPC @franzpoeschel

### DIFF
--- a/source/adios2/operator/compress/CompressBlosc.cpp
+++ b/source/adios2/operator/compress/CompressBlosc.cpp
@@ -20,6 +20,7 @@ extern "C" {
 #include <algorithm>
 #include <cassert>
 #include <cstring>
+#include <iostream>
 
 namespace adios2
 {
@@ -139,6 +140,12 @@ size_t CompressBlosc::Compress(const void *dataIn, const Dims &dimensions,
                 value, "when setting Blosc threshold parameter\n"));
             if (thresholdSize < 128u)
                 thresholdSize = 128u;
+        }
+        else
+        {
+            std::cerr << "ADIOS WARNING: Unknown parameter keyword '" << key
+                      << "' with value '" << value
+                      << "' passed to Blosc compression operator." << std::endl;
         }
     }
 


eisenhauer:
Unfortunately, parameters don't work like that.  Blosc gets all the parameters passed to the engine.  It can pull of the ones that it understands, but it can't assume that the ones that it doesn't understand are invalid.  They may be typos, or they may just be meant for another ADIOS component.  To properly detect invalid parameters, we have to have a global registry or something like that.  We've talked about it, but not done it.  We don't seem to have an issue on it though.  May I suggest closing this PR and opening an issue instead?

psychocoderHPC:
> Unfortunately, parameters don't work like that. Blosc gets all the parameters passed to the engine. It can pull off the ones that it understands, but it can't assume that the ones that it doesn't understand are invalid. They may be typos, or they may just be meant for another ADIOS component. To properly detect invalid parameters, we have to have a global registry or something like that. We've talked about it, but not done it. We don't seem to have an issue on it though. May I suggest closing this PR and opening an issue instead?

If I do not miss something the parameters passed are hierarchically organized and belong to a compressor. Invalid parameters should not be part of another compressor. IMO throwing a warning for an unknown compressor should be fine.

```
//...
    auto blosc = adios.DefineOperator( "compression", "blosc" );
    adios2::Params bloscParams{ { "compressor", "zstd" /*"blosclz"*/ },
                                { "compression_level", "7" }, { "threshold","200" }};

    std::vector< datatype > v( extent );
    std::iota( v.begin(), v.end(), 0 );
    {
        // write
        adios2::Engine engine = IO.Open( "blosctest.bp", adios2::Mode::Write );
        engine.BeginStep();
        auto othervariable = IO.DefineVariable< datatype >(
            "iota", { extent }, { 0 }, { extent } );
        // the parameters should only be visible for blosc
        othervariable.AddOperation( blosc, bloscParams );
        engine.Put( othervariable, v.data() );
        engine.EndStep();
        engine.Close();
    }
//...
```

eisenhauer:
Hmm.  Looks like you may be right.  Not sure what I was recalling...

JasonRuonanWang (approved):
Okay I had a check in the compression operator code. The compressor parameters are generated when user applications add operations to a variable. This is separated from the engine parameters or any other parameters. So in principle this map should not include anything other than the designed parameters for the specific compressor. So it should be fine to give some notifications when an invalid parameter is caught. 

The only thing we may need to think later is whether we should output it to std::cerr or throw an exception. Also we need to add this for other compressors as well.

PR #2676, opened 2021-04-15
add support for libpressio as an operator

robertu94:
Hi,

I developed [libpressio](https://github.com/CODARcode/libpressio), a generic abstraction for dense tensors for several of the error bounded lossy compressors including SZ, ZFP, and others not currently supported by ADIOS.  While I am aware that there is support for SZ/ZFP already in ADIOS2, this support adds a number of features that currently aren't offered by the current bindings:

+ GPU support for ZFP, and soon for SZ (GPU support for SZ when merged will not require changes to this code).
+ Several currently unsupported options for both SZ and ZFP including those for SZ's "protect_value_range" and new "modes" like ROIBIN-SZ, ZFP's expert mode, and more.
+ Supports additional several additional compressors like SZAuto, fpzip, other image formats via image magic, as well as meta-compressors such as FRaZ/LibPressio-Opt which automatically configure compressors.

I'm relatively sure that I missed some things that I need to tweak to make this play nicely in the ADIOS-2 ecosystem (unit-test jump out to me), so please let me know what I'm missing and what I need to fill in.

I did write an "executable" test based on the ZFP ones which looks like this:

```cpp
  int size, rank;
  MPI_Init(&argc, &argv);
  MPI_Comm_size(MPI_COMM_WORLD, &size);
  MPI_Comm_rank(MPI_COMM_WORLD, &rank);

  adios2::ADIOS adios(MPI_COMM_WORLD);
  adios2::IO io = adios.DeclareIO("TestIO");

  const int nx=5, ny=5;
  const adios2::Dims 
      shape{nx, ny*static_cast<size_t>(size)},
      start{0, ny*static_cast<size_t>(rank)},
      count{nx, ny};
  size_t totals = size*nx*ny;

  auto temp = io.DefineVariable<float>("temp", shape, start, count, adios2::ConstantDims);
  auto prec = io.DefineVariable<float>("prec", shape, start, count, adios2::ConstantDims);
  auto time_var = io.DefineVariable<double>("time", {1}, {0}, {1}, adios2::ConstantDims);

  adios2::Operator lp_op = adios.DefineOperator("LibPressioOperator", adios2::ops::LossyLIBPRESSIO);
  prec.AddOperation(lp_op, {{"compressor_id", "sz"},
                            {"config:sz:rel_err_bound", "1e-6"},
                            {"config:sz:error_bound_mode_str", "rel"}});
  temp.AddOperation(lp_op, {{"compressor_id", "sz"},
[...]

PR #2742, opened 2021-06-02
add support for libpressio as an operator

JasonRuonanWang:
This PR is to replace #2676 which got automatically closed due to having the same branch name as master. Please refer to #2676 for original information @robertu94 provided.


robertu94:
@JasonRuonanWang weird, I've never seen that cause a problem before.  It look's like you've done what you've asked me to do.  Do you need me to do anything else?

JasonRuonanWang:
> @JasonRuonanWang weird, I've never seen that cause a problem before. It look's like you've done what you've asked me to do. Do you need me to do anything else?

I think it's fine now. But if you want to make updates please use this branch instead of master :)

robertu94:
@JasonRuonanWang I can't see the output of the msan run, but since the LibPressio code isn't built by default, I'm guessing that the test failure wasn't caused by this PR

JasonRuonanWang:
It's caused by a cmake warning. I will look into it later

```
CMake Warning at cmake/DetectOptions.cmake:112 (find_package):
  By not providing "FindLibPressio.cmake" in CMAKE_MODULE_PATH this project
  has asked CMake to find a package configuration file provided by
  "LibPressio", but CMake did not find one.

  Could not find a package configuration file provided by "LibPressio" with
  any of the following names:

    LibPressioConfig.cmake
    libpressio-config.cmake

  Add the installation prefix of "LibPressio" to CMAKE_PREFIX_PATH or set
  "LibPressio_DIR" to a directory containing one of the above files.  If
  "LibPressio" provides a separate development package or SDK, be sure it has
  been installed.
  CMakeLists.txt:144 (include)


CMake Warning at cmake/DetectOptions.cmake:113 (find_package):
  By not providing "Findstd_compat.cmake" in CMAKE_MODULE_PATH this project
  has asked CMake to find a package configuration file provided by
  "std_compat", but CMake did not find one.

  Could not find a package configuration file provided by "std_compat" with
  any of the following names:

    std_compatConfig.cmake
    std_compat-config.cmake

  Add the installation prefix of "std_compat" to CMAKE_PREFIX_PATH or set
  "std_compat_DIR" to a directory containing one of the above files.  If
  "std_compat" provides a separate development package or SDK, be sure it has
  been installed.
  CMakeLists.txt:144 (include)
```

robertu94:
That warning is caused by not being able to find the cmake find_package dependencies for  LibPressio and it's only required dependency std_compat.

I'm surprised this didn't cause a warning/error in the other configurations since it did here.

How would you prefer to install LibPressio for unit/integration testing?

JasonRuonanWang:
We probably need to write a FindLibPressio.cmake file and a Findstd_compat.cmake, and put them in ADIOS2/cmake to tell cmake how to find the header and library files of LibPressio. If you look at cdash, many tests had these two warnings. 

I am fine with any method to install it. Once it works I will add it to DataMan engine and write some tests.

robertu94:
These files get generated by LibPressio Cmake files. We just need to install the in a place that cmake looks for them. 


https://cmake.org/cmake/help/latest/command/find_package.html

JasonRuonanWang:
I guess the warning is for cases where there is no LibPressio installed. I am sure if you actually have it installed in somewhere and export that directory to CMAKE_MODULE_PATH then this warning will go away. But it should allow tests to run warning-free if there is actually does not find LibPressio. If you look at the directory ADIOS2/cmake, there are many FindBlahblah.cmake which I guess deals with this situation. I am not a cmake expert, this is only my guess :)

germasch:
The solution may just be to do
```
if(ADIOS2_USE_LIBPRESSIO STREQUAL AUTO) 
 find_package(LibPressio QUIET CONFIG)
 find_package(std_compat QUIET CONFIG)
elseif(ADIOS2_USE_LIBPRESSIO)
  find_package(LibPressio REQUIRED CONFIG)
  find_package(std_compat REQUIRED CONFIG)
endif()
```

Since it sounds like those packages do install their own cmake configs, no find module is required, and by telling cmake so (`CONFIG`), those warnings should go away.

I do think there's another problem, though: If only LibPressio is found, but not std_compat, it'll still try to enable it.

JasonRuonanWang:
@germasch Thanks for the point. I will just try this change and see how it works.

JasonRuonanWang:
@robertu94 It seems to pass all the tests now. Another question, what is the std_compat library? I was trying to find a reference in LibPressio but it does not have a link or something that I can refer to. I saw this repository https://github.com/thierry-martinez/stdcompat but I am not sure if it is the library that LibPressio requires...

robertu94:
@JasonRuonanWang std_compat contains a bunch of backports for c++14,17, and 20 library features that LibPressio uses to support older compilers. It's a separate package because some other applications/libraries that my collaborators use it independently of LibPressio

https://github.com/robertu94/std_compat

JasonRuonanWang:
@robertu94 Sounds great. But I couldn't get it compiled on my Mac systems. I have opened an issue in your repository. Can you help have a look? Thanks.

PR #2755, opened 2021-06-07
CI:  Enforce -Werror & -Wall in cpp sources

vicentebolea:
Part 2 of: https://github.com/ornladios/ADIOS2/issues/2766

Followup of https://github.com/ornladios/ADIOS2/pull/2639/files

Notes from this PR:

Wall -Werror enabled for C++ in all builds but:

1. Accent + XL builds
2. Power8 builds
3. Debian buids
4. Using intel compiler translation units that include `nlohmann_json`


vicentebolea:
@chuckatkins  I will rebase this onto master once your branch is merged :)

thirdparty/nlohmann_json/nlohmann_json/src/single_include/nlohmann_json.hpp
```diff
@@ -0,0 +1,36 @@
+#ifndef __NLOHMANN_JSON_HPP__
```
vicentebolea, line 1: @chuckatkins I added this wrapper for the nlohmann_json library in order to be able to disable warnings in the library

thirdparty/nlohmann_json/nlohmann_json/CMakeLists.txt
```diff
@@ -1,7 +1,14 @@
...
 target_sources(nlohmann_json INTERFACE
-  $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/src/single_include/nlohmann/json.hpp>
+  $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/src/single_include/upstream/nlohmann/json.hpp>
 )
+
+target_sources(nlohmann_json INTERFACE
+  $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/src/single_include/nlohmann_json.hpp>
+)
+
+target_compile_options(nlohmann_json INTERFACE $<$<AND:$<COMPILE_LANGUAGE:CXX>,$<CXX_COMPILER_ID:Intel>>:-Wno-all -Wno-error>)
```
vicentebolea, line 10: @chuckatkins Since Intel compiler does not like warning pragmas I had to disable all the warnings for TU including nlohmann_json, which are a handful of them.

testing/install/CMakeLists.txt
```diff
@@ -17,9 +17,6 @@ function(add_install_cmake_test name)
     "-DCMAKE_C_COMPILER=${CMAKE_C_COMPILER}"
     "-DCMAKE_CXX_COMPILER=${CMAKE_CXX_COMPILER}"
     "-DCMAKE_Fortran_COMPILER=${CMAKE_Fortran_COMPILER}"
-    "-DCMAKE_C_FLAGS=${CMAKE_C_FLAGS}"
```
vicentebolea, line 20: @chuckatkins This was needed to make this test work. Note that the only flags that were being passed was -Wall -Werror. It seems that we did not need this.

bindings/CXX11/adios2/cxx11/fstream/ADIOS2fstream.tcc
```diff
@@ -15,6 +15,17 @@
 
 #include "adios2/core/Stream.h"
 
+#if defined(_MSC_VER) && !defined(__clang__)
```
vicentebolea, line 18: @chuckatkins some of the files have those pragma guards which disables a whole warning in the file. The reason of doing this instead of solving the warning is that either solving the warning added another problem or that for the sake of code clarity, getting rid of the warning obscured the code.

vicentebolea (commented):
@chuckatkins I think it is ready to go, there remains to add Wall Werror for Fortram which we can address it following up this PR.

testing/install/CMakeLists.txt
chuckatkins: We need to keep this in.  If a user specifies certain compilation flags then leaving this out can cause the adios-config script generated at install time to produce incorrect flags.

vicentebolea: I see, right! Find this resolved at the last push

bindings/Python/py11File.cpp
```diff
@@ -216,7 +216,7 @@ pybind11::array File::Read(const std::string &name, const Dims &start,
             m_Stream->Read<std::string>(name, blockID).front();
         pybind11::array_t<char> pyArray(Dims{value.size()});
         std::copy(value.begin(), value.end(), pyArray.mutable_data());
-        return pyArray;
+        return std::move(pyArray);
```
chuckatkins, line 219: shouldn't C++11 return value optimization guarantee this already?

vicentebolea: you are right, however, still, not explicitly using std::move brought warnings in some of the builds

vicentebolea:
@chuckatkins I think it is ready for re-review :)

chuckatkins:
So, for the sake of forward progress, I'm going to go ahead and merge this.  However, the json wrapper needs to shuffled around a bit to make sure the to update script works as it's expected to.  So there'll just need to be a follow up with that and another that drops a few of the outstanding Werror inclusions.

PR #2759, opened 2021-06-09
Update format comment, add note about endian support

eisenhauer:


germasch:
Pretty much totally unrelated, but as I see bp5 work going on, I figured I might as well put in a request to have the file format store a per-dataset flag as to whether its data is in row-major or column-major order. Having this presumed to be known just by the language used is not really the case anymore these days -- to name some examples, Kokkos supports both orders, as do many other C++ multi-d libraries, and so does Python/numpy. (Even in my C days, actually, I'd have a multi-d data structure that was could be, and usually was, col-major).

eisenhauer:
> Pretty much totally unrelated, but as I see bp5 work going on, I figured I might as well put in a request to have the file format store a per-dataset flag as to whether its data is in row-major or column-major order. Having this presumed to be known just by the language used is not really the case anymore these days -- to name some examples, Kokkos supports both orders, as do many other C++ multi-d libraries, and so does Python/numpy. (Even in my C days, actually, I'd have a multi-d data structure that was could be, and usually was, col-major).

Agreed.  This requires an addition to the API somewhere, so might need some thought.  I'm assuming that it would be sufficient to have an IO-wide flag that declared Row- or Column-major?   I.E. not something that was variable by variable?  Still needs some restructuring, because ADIOS is just passing the host language around and then inferring the RC order from that.  I think that's the only thing we infer from that.  Open an issue so that it gets handled and not forgotten when this PR is merged?

Issue #1661, opened 2019-08-05
data layout / BP-X


germasch:
This is somewhat BP-X related, but also a more generic API issue, so I figured I'll put the details here.

## Language-independent handling of row-major vs col-major data layout

Here's a proposal to increase flexibility in handling data layout while keeping the existing behavior / APIs unchanged.

### Rationale

Currently, the ADIOS2 API is built around the assumption that the language of the API bindings determines the data layout that the application is using, in particular Fortran/Matlab/R: column-major; otherwise: row-major (see `adiosSystem.cpp`). That covers most common applications. In particular reading and writing data from the same language works nicely. However, mixing languages or data layouts is more complicated.

When Fortran writes, say, a 400x200 (column-major) array, it will be presented in the C/C++ API as a 200x400 row-major array. In order to present the data as a row-major array, there are only two options: reverse the shape, or transpose the data. Since transposition is potentially expensive, ADIOS2 make the perfectly reasonable choice to instead reverse the dimensions.

However, the assumption that the language of the API determines the data layout used is not always valid. For example:

- The main part of a legacy code remains written in Fortran, hence   it's using column-major data layout for its arrays. However, some   parts of the code are rewritten in C++, e.g., in order to enable use  of CUDA / Kokkos (this is currently happening for XGC). At this  point, if one writes data using the adios2 C++ API, it makes the  assumption that the data layout is row-major, even though it  actually is not. This can be worked around on the application side  by pretending that the column-data is row-major if the app reverses  dims itself. However, one loses the information whether a dataset is  actually row-major, or just pretending to be.

- Essentially the same issue happens in a pure C++ code that uses data  structures that aren't (necessarily) row-major, like, e.g., xtensor  or Kokkos Views. Again, it can be worked around by reversing dims on  the application side, but the info what the real layout is gets  lost.

- A general visualization app, say written in C++, sees a data set of  size (200x400). It doesn't know whether that means there are really  200 points in x direction and 400 points in y direction, or whether  it is looking at an actual 400x200 data set that had been written  using the Fortran API. This is somewhat solved for the unique  language->layout mapping case by #1573, though (a) the app has to go  down to the block level to check the `IsReverseDims` flag and (b)  the `IsReverseDims` flag doesn't directly tell the original data  layout, but requires additional logic from the application. The viz  app then has to reverse dims itself again to read data into a data  structure using the original layout.

[...]

PR #2763, opened 2021-06-12
Restructure to allow direct specification of array order

eisenhauer:
@germasch Please look to see if this would reference the concerns you raise about array ordering in  #2759.
@pnorbert, @JasonRuonanWang, et al.  I think we have a notorious shortage of cross-language tests, so perhaps this needs extra attention before merge.  The basic approach is minimalistic.  There's a new parameter to DeclareIO that has an optional value specifying the array element ordering.  The default is Auto, meaning that it's determined by the language specified when the ADIOS object is created, but it can also be RowMajor or ColumnMajor.  This does *not* change the format of BP3/4 files on disk, where the language specified serves as a stand-in for the array ordering.  Instead we store C++ for RowMajor and Fortran for ColumnMajor.

### DIFF
--- a/source/adios2/engine/bp3/BP3Writer.cpp
+++ b/source/adios2/engine/bp3/BP3Writer.cpp
@@ -222,7 +222,8 @@ void BP3Writer::InitBPBuffer()
     else
     {
         m_BP3Serializer.PutProcessGroupIndex(
-            m_IO.m_Name, m_IO.m_HostLanguage,
+            m_IO.m_Name,
+            (m_IO.m_ArrayOrder == ArrayOrdering::RowMajor) ? "C++" : "Fortran",
             m_FileDataManager.GetTransportsTypes());
     }
 }
--- a/bindings/C/adios2/c/adios2_c_adios.cpp
+++ b/bindings/C/adios2/c/adios2_c_adios.cpp
@@ -17,6 +17,30 @@
 extern "C" {
 #endif
 
+adios2::ArrayOrdering adios2_ToArrayOrdering(const adios2_arrayordering Corder)
+{
+    adios2::ArrayOrdering order = adios2::ArrayOrdering::Auto;
+    switch (Corder)
+    {
+
+    case adios2_arrayordering_rowmajor:
+        order = adios2::ArrayOrdering::RowMajor;
+        break;
+
+    case adios2_arrayordering_columnmajor:
+        order = adios2::ArrayOrdering::ColumnMajor;
+        break;
+
+    case adios2_arrayordering_auto:
+        order = adios2::ArrayOrdering::Auto;
+        break;
+
+    default:
+        break;
+    }
+    return order;
+}
+
 adios2_adios *adios2_init_config_glue_serial(const char *config_file,
                                              const adios2_debug_mode debug_mode,
                                              const char *host_language)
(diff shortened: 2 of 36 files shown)


JasonRuonanWang (commented):
Looks good to me as long as it passes all the tests. For cross-language tests, I think we are good for now. At least I have been often blocked by the Fortran tests when I wanted to merge something, which is saying the Fortran tests are pretty effective. I don't mind if you want to add new tests though :)

eisenhauer:
This PR lacks:  Python bindings, and tests of its actual functionality.

germasch:
This looks like great step forward, thanks @eisenhauer. Let me mention @rmchurch since he also independently ran in to related issues a little while ago.

For reference, there's this issue with more of my thoughts back in the day: #1661

For the use cases, I have, this PR would do all that's needed -- the case I have are Fortran-order arrays that I'm using in C++. I  also think it's not easily possible to have more fine-grained control in the existing BP3/BP4 formats, and so I think it's fine to have a one ordering per file limitation there. (It might be in theory possible to get finger control by (ab)using progress groups, but that doesn't seem worth it.)

In general, though, and that's why I mentioned it for BP5, I think it'd be useful to have the capability to specify the ordering for each dataset (ie., definitely per-variable, maybe even per-variable per-step). I don't think it's urgent to implement a corresponding API for it right now -- but if it's something that's not supported by the file format, it won't be possible to add it later without compatibility issues.

Here are some examples why I find it's useful:
* Kokkos, e.g., supports both layouts, on a per-dataset basis. It's certainly conceivable that both get used within the same code, e.g., when part of the data is kept in Fortran layout during a port to C++ effort, while other parts may be completely migrated and use standard C++ layout. Even if that weren't the case, presumably if someone implements a shim layer to write Kokkos arrays to ADIOS2, they'd know what layout to use at the time the dataset is written, not yet when the file is opened.
* Better python support. Numpy arrays can be either layout, and it's transparent to the user unless one looks (using `<array>.flags`. It's quite easily possible to have both layouts used under the hood, in particular in data analysis, where the data probably/hopefully corresponds to the layout on disk. E.g., if one uses h5py to read a dataset written in HDF5 from Fortran, one ends up with the data with the dimensions in `zyx` order (because HDF5 sadly doesn't support multiple data layouts, either -- last I checked). So usually what one does is take the transpose after reading, which is cheap because it pretty much only flips the ordering flag in numpy. If one now writes this data to adios2, it'll be in Fortran order, and it'd be nice if it could be written with adios2.

In terms of API, I think what you did can naturally be extended, and remains backward compatible:
* For an IO, if the ordering is not specified, fall back to the host language (ie., current behavior)
* For a variable, if the order is not specified, fall back to the IO's ordering.

So the only additional piece is a new API function like `var.setLayout(..)`, and a corresponding `getLayout()`. As that's an addition, it doesn't break any existing usage, and there may not even be too much urgency implementing it. The only drawback is that if the file is actually BP3/BP4, and the IO is opened as, e.g., row-major, if one were to try to `var.setLayout(columnMajor)`, it can't be done. I think there are two options on how to handle it: (1) throw an exception or (2) transpose the dims (my preference being the former, as (2) leads to surprises when reading back data).

source/adios2/engine/bp3/BP3Writer.cpp
```diff
@@ -218,7 +218,8 @@ void BP3Writer::InitBPBuffer()
     else
     {
         m_BP3Serializer.PutProcessGroupIndex(
-            m_IO.m_Name, m_IO.m_HostLanguage,
+            m_IO.m_Name,
+            (m_IO.m_ArrayOrder == ArrayOrdering::RowMajor) ? "C++" : "Fortran",
```
guj, line 222: Since m_HostLanguage is still around under the IO class,  can we use it instead of deriving it from m_ArrayOrder?  I noticed this in a few places.

guj (commented):
Thanks Greg for the big effort!

source/adios2/engine/bp3/BP3Writer.cpp
eisenhauer: Well, the idea is that the ArrayOrdering might not match the host language, so if they've change the array ordering in a C++ to be ColumnMajor (because that's how the app views the data) and then we write out using "C++", that ArrayOrdering specification will be effectively lost (because we don't have ArrayOrder directly in the BP3/4 header).  BP5 won't use Language as a stand-in for ArrayOrder, but in the old formats that it's in we use C++ as a standin for RowMajor and Fortran as a stand-in for ColumnMajor.

guj: Hmm this host language parameter, derived from array ordering, is later used to determine column/row major again in the serializers..  Might as well pass the arraying ordering directly to the serializers :)

eisenhauer: Yep.  For BP3/4 that the hostlanguage is used for that is a historical artifact.  We could rewrite them, but unless we change their file formats it's going to be converted to language specification before going to disk.  So, this is the minimalist approach to leaving those formats the same, but evolving so that new formats can use it directly.

guj: Understood. Thanks Greg

eisenhauer:
I'm going to go ahead and merge this, as it passes all existing tests.

PR #2765, opened 2021-06-14
Table engine: added error message for compressor parameter

JasonRuonanWang:


JasonRuonanWang:
@rtobar This should address the issue you have seen.

PR #2777, opened 2021-06-25
BP4 engine capable of using device buffers with Put

anagainaru:
Write side only

Users need to specify the memory space for an ADIOS variable (by default Host) and call Put in the same way as before.
All consecutive Puts will use the same memory space unless updated.

```c++
      bpWriter.BeginStep();
      data.SetMemorySpace(adios2::MemorySpace::CUDA);
      bpWriter.EndStep();
```

For now only CUDA is supported, when/if we allow Kokkos, Sycl, etc buffers to be submitted, the m_IsGPU parameter inside variables need to be more than boolean.

### DIFF
--- a/cmake/DetectOptions.cmake
+++ b/cmake/DetectOptions.cmake
@@ -141,6 +141,16 @@ endif()
 
 set(mpi_find_components C)
 
+# Cuda
+if(ADIOS2_USE_CUDA STREQUAL AUTO)
+  find_package(CUDA)
+elseif(ADIOS2_USE_CUDA)
+  find_package(CUDA REQUIRED)
+endif()
+if(CUDA_FOUND)
+  set(ADIOS2_HAVE_CUDA TRUE)
+endif()
+
 # Fortran
 if(ADIOS2_USE_Fortran STREQUAL AUTO)
   include(CheckLanguage)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -147,6 +147,7 @@ adios_option(MGARD     "Enable support for MGARD transforms" AUTO)
 adios_option(PNG       "Enable support for PNG transforms" AUTO)
 adios_option(MPI       "Enable support for MPI" AUTO)
 adios_option(DAOS      "Enable support for DAOS" AUTO)
+adios_option(CUDA       "Enable support for Cuda" AUTO)
 adios_option(DataMan   "Enable support for DataMan" AUTO)
 adios_option(DataSpaces "Enable support for DATASPACES" AUTO)
 adios_option(SSC       "Enable support for SSC" AUTO)
@@ -174,7 +175,7 @@ if(ADIOS2_HAVE_MPI)
 endif()
 
 set(ADIOS2_CONFIG_OPTS
-    Blosc BZip2 ZFP SZ MGARD PNG MPI DataMan DAOS Table SSC SST BP5 DataSpaces ZeroMQ HDF5 HDF5_VOL IME Python Fortran SysVShMem Profiling Endian_Reverse LIBPRESSIO
+    Blosc BZip2 ZFP SZ MGARD PNG MPI DataMan CUDA DAOS Table SSC SST BP5 DataSpaces ZeroMQ HDF5 HDF5_VOL IME Python Fortran SysVShMem Profiling Endian_Reverse LIBPRESSIO
 )
 GenerateADIOSHeaderConfig(${ADIOS2_CONFIG_OPTS})
 configure_file(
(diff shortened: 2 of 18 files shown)


cmake/DetectOptions.cmake
```diff
@@ -141,6 +141,16 @@ endif()
...
 
+# Cuda
+if(ADIOS2_USE_CUDA STREQUAL AUTO)
+  find_package(CUDA)
+elseif(ADIOS2_USE_CUDA)
+  find_package(CUDA REQUIRED)
+endif()
+if(CUDA_FOUND)
+  set(ADIOS2_HAVE_CUDA TRUE)
+endif()
```
germasch, line 152: `find_package(CUDA)` has been deprecated, and it's goal was to enable CUDA as a language, but what you need here is really just the toolkit. There is `find_package(CUDAToolkit)` now, but it's a relatively recent addition to cmake, so you might have to bump up the `cmake_minimum_required` to make sure it's available. In any case, you have cmake experts on your team, I'm sure they know how to handle this best ;)

germasch (commented):
I think this is definitely useful functionality 👍 

I'm wondering whether you'd want to support "cuda-aware ADIOS2" in a way that mirrors "cuda-aware MPI" (maybe in addition to specifying memory space explicitly)? With cuda-aware MPI, there is no need to specify the memory space, it's just exactly the usual MPI API. What it does, though, is inspect the pointer to figure out what the underlying memory space is -- the same idea should be possible here.

cmake/DetectOptions.cmake
anagainaru: CUDA aware MPI relies on the Unified Virtual Addressing (UVA) feature introduced in CUDA 4.0 to detect if a buffer resides on the device or host. I chose to use a memory space because we might want to extend this to AMD for Frontier (with either HIP, Kokkos, etc) and I am not aware of any solution that can detect where a buffer has been allocated across all architectures.

Thanks for pointing me to `CUDAToolkit`. Will bring it up at our next ADIOS meeting.

germasch: HIP supports unified virtual addressing, just like cuda (I'd consider cuda 4.0 historic by now, I don't think one has to worry about compatibility to pre cuda-4).

SYCL also supports something similar ("unified shared memory"). However, the native way in sycl, I guess, is to use sycl buffers -- but adios2 doesn't have an API to deal with buffers directly at this time, so one has to pass in pointers in any case.

Anyway, I'm not saying it's a bad idea to have the feature to specify a memory space explicitly -- but it'd be more application-friendly if the default memory space would be "auto" aka "UVA" or something, so that by default things just work, as they do in MPI.

williamfgc:
@anagainaru thanks for adding this cool feature. I'll encourage adding something in the docs section once it's done. That way we can use it properly depending on hardware. Bonus point for thinking of different underlying programming models (SyCL, Kokkos, CUDA, AMD). Would this apply to NVLink?

williamfgc:
Also, I haven't looked in detail, but would the Span Put signature work with it? Memory bound device-based apps could benefit. Thanks.

anagainaru:
> Would this apply to NVLink?

You mean for streaming? The PR uses the BP engine so it will only apply to GPU-to-storage.

> Also, I haven't looked in detail, but would the Span Put signature work with it? Memory bound device-based apps could benefit. Thanks.

It is not, I wanted to start the discussion and be sure of the design first. But I am working on it so eventually the Span Put will work with it. The code in this PR is only for continuous buffer Puts.

cmake/DetectOptions.cmake
anagainaru: Yes I do agree that default usage would be nice (even though it will be a headache to implement) :) I'll discuss this with the others and figure out what we can do.

eisenhauer (changes requested):
In the interest of keeping non-type-specific things in VariableBase rather than Variable, can I suggest moving m_IsGPU and the associated SetMemorySpace function to VariableBase?

JasonRuonanWang:
I was just trying the github built-in conflict-resolve feature, didn't realize how it would merge back to the branch. If it doesn't work please remove my commit.

eisenhauer:
It might make sense to start with current master and just cherry-pick the necessary commits from this to make a new branch that's not so stale.  (I did a quick try at a rebase, but conflicts made it go badly and what I ended up with was missing functions, so something went wrong.  The cherry-pick might yield more obvious results.)

anagainaru:
Yes I agree, will cherry pick the commits and apply it on the latest version.

williamfgc:
@pnorbert CMake support is moving to [FindCUDAToolkit](https://cmake.org/cmake/help/latest/module/FindCUDAToolkit.html).

anagainaru:
History is clean, once it is ready to be merged

eisenhauer:
> History is clean, it is ready to be merged

Sounds good.  Lets let it run through CI and get it in.

PR #2786, opened 2021-07-12
available_variables and available_attributes for the C API

dmitry-ganyushin:


bindings/C/adios2/c/adios2_c_io.cpp
```diff
@@ -874,6 +874,60 @@ adios2_error adios2_remove_all_attributes(adios2_io *io)
...
+            io, "for adios2_io, in call to adios2_available_variables");
+        std::map<std::string, std::map<std::string, std::string>> varInfo =
+            reinterpret_cast<adios2::core::IO *>(io)->GetAvailableVariables();
+        *size = varInfo.size();
+        char **names = (char **)malloc(*size * sizeof(char *));
+
+        size_t cnt = 0;
+        for (auto var : varInfo)
+        {
+            int len = var.first.length();
```
williamfgc: `len` should stay as `std::size_t` as returned by `length()` and the input type to `malloc`

williamfgc (commented):
@dmitry-ganyushin Great work. It should be documented that the returned `char**` must be manually deallocated by the user. I remember @pnorbert needed something similar for the Matlab bindings that use the C APIs.

bindings/C/adios2/c/adios2_c_io.cpp
dmitry-ganyushin: Thanks a lot! The main question is whether we should have unicode strings or not.

williamfgc: If I remember correctly unicode was supported as-is in the C++ `std::string` side. You can check with @pnorbert .

PR #2790, opened 2021-07-15
Plugin engine

caitlinross:


caitlinross:
@chuckatkins

examples/plugins/engine/ExampleWritePlugin.h
```diff
@@ -0,0 +1,69 @@
...
+    void DoPutSync(Variable<T> &variable, const T *values) override;           \
+    void DoPutDeferred(Variable<T> &variable, const T *values) override;
+    ADIOS2_FOREACH_STDTYPE_1ARG(declare)
+#undef declare
+
+    void DoClose(const int transportIndex = -1) override;
+
+private:
+    std::ofstream m_Log;
+    std::unique_ptr<BP4Writer> m_Writer;
```
JasonRuonanWang, line 63: Just curious. Is this assuming that all potential plugin engines will use BP4? What if a plugin engine wants to write data in its own way?

caitlinross: No, this is just an example plugin, there's no reliance on any specific engines by the actual PluginEngine (as in the piece that loads/runs the plugin code). But this example is being updated to actually do its own work instead of using BP under the covers.

caitlinross:
@chuckatkins Sorry, I thought I tagged you last week that this was done (fixed the dashboard issues and plugin install test), but I guess not. Anyway I just rebased and am rerunning tests because there was still one test failure, but it was for `HeatTransfer.SST.FFS.MxN.Validate`, which I don't really see how this PR would affect that test.

chuckatkins (approved):

PR #2798, opened 2021-07-24
accomodate std::string values for marshalling

guj:


eisenhauer (approved):
Hi Junmin.  Thanks for submitting.  I was poking at this Friday, but wanted to have a test that covered it and didn't finish that.  However, if this fixes a WarpX problem, then I think we can merge and add the test later.

PR #2803, opened 2021-07-28
Bp5 span

pnorbert:


pnorbert:
Hi @williamfgc, do you remember why there is a BufferID in the Span functionality at the user API level? BP3 and BP4 does not use it and neither BP5. BP5->BufferV->ChunkV now needs two levels of reference into the vector of allocated chunks (integer BufferIdx and then a payload position within a chunk) but this is not a user level information, rather it is hidden inside the core::Variable::Span structure. 

If there is no particular reason to keep the user facing BufferID around, I would rework the API to eliminate it. But I don't remember why it is there in the first place. 

Thanks

williamfgc:
> do you remember why there is a BufferID in the Span functionality at the user API level?

@pnorbert good question. If I recall correctly that came from some underlying transport libraries ability to process multiple buffers in a queue style to process them asynchronously and independently (e.g. ZeroMQ, MPI). Essentially, a buffer hierarchy that way `BufferID0` is more relevant than `BufferID1` as it might carry information that was needed first (e.g. metadata, smaller variables,etc). It was never implemented at the transport layer that way for files in BP3, BP4, BP5 as it doesn't make much difference for the POSIX backend, and I'm not sure if SST can do that (maybe through different sockets). It's one of the design decisions that were influence by research ideas, more than immediate need. Hope it helps.

eisenhauer (approved):

pnorbert:
> 
> 
> > do you remember why there is a BufferID in the Span functionality at the user API level?
> 
> @pnorbert good question. If I recall correctly that came from some underlying transport libraries ability to process multiple buffers in a queue style to process them asynchronously and independently (e.g. ZeroMQ, MPI). Essentially, a buffer hierarchy that way `BufferID0` is more relevant than `BufferID1` as it might carry information that was needed first (e.g. metadata, smaller variables,etc). It was never implemented at the transport layer that way for files in BP3, BP4, BP5 as it doesn't make much difference for the POSIX backend, and I'm not sure if SST can do that (maybe through different sockets). It's one of the design decisions that were influence by research ideas, more than immediate need. Hope it helps.

Thanks for the explanation. I decided to remove this (actually replace it with a bool flag for initialization) because the handling of multiple buffers of an engine cannot be controlled at the user level plus we have no engine in sight juggling separate buffers anyway.

PR #2813, opened 2021-08-04
WIP:  Alternative BlocksInfo implementation for BP5

eisenhauer:


eisenhauer:
@pnorbert This can go with or without the last commit.  The last moves MinBlocksInfo from Engine.h to VariableBase.h.  Which is your preference?  (Nothing in VariableBase uses MinBlocksInfo, it is solely produced by an Engine call and is never stored with the variable.  But perhaps conceptually it's information about a variable, so maybe VariableBase is appropriate?  I could go either way.)

pnorbert:
It sounds like then it's place is in Engine.h, I retract whatever I said in person yesterday. But what will be the case when we have file reading? Still no storing blockinfo in the variable base?

eisenhauer:
I can't say I'm certain about the file-mode-read case yet.  Lots of fogginess there.  I can say that I generally don't love having engine-specific info in the Variable* classes, although it has occurred to me that on the reader side, having engine-specific subclasses of Variable that were used when the read engines create variables might be useful.  Right now we often have to map from the Variable to engine-internal information specific to variables...  But for now I'll kill the last commit and revert to the struct in Engine.h.

PR #2828, opened 2021-08-19
added runtime control of numbers of tiers to read in mhs

JasonRuonanWang:


JasonRuonanWang:
@lwan86 This PR should allow you to stop reading upcoming tiers if the required accuracy is already satisfied.

PR #2837, opened 2021-08-20
Add to C API a function to get blockinfo:

pnorbert:
adios2_varinfo *adios2_inquire_blockinfo(adios2_engine *engine,
                                         adios2_variable *variable,
                                         const size_t step);
Caveat: Min/Max and Value are not implemented yet, except for string values.


pnorbert:
@dmitry-ganyushin See how to use it in `testing/adios2/engine/bp/TestBPWriteReadBlockInfo.cpp::835-908`. The structures are defined in `bindings/C/adios2/c/adios2_c_types.h`

PR #2852, opened 2021-09-09
EVPath: force implicit libm link

vicentebolea:
Since some compilers will implicitely link libm, the EVPath project
CMakeList.txt first tests if it needs to be explicitly linked and if
that is the case it will implicitly link libm.

Unfortunately this test seems to give false positives in some machines,
thus skipping the implicit linkage of libm results in linking errors.

Backport from a patch for ADIOS2 OpenSUSE pkg.

https://build.opensuse.org/package/view_file/home:vicentebolea/ADIOS2/evpath-link-m.patch?expand=1

@eisenhauer I believe that EVPath is your realm, this is an extract of the patch, however, probably we can find a more elegant way to resolve this

### DIFF
--- a/thirdparty/EVPath/EVPath/CMakeLists.txt
+++ b/thirdparty/EVPath/EVPath/CMakeLists.txt
@@ -131,6 +131,9 @@ set(_pkg_config_libs)
 set(_pkg_config_private_libs)
 
 include(CheckCSourceCompiles)
+
+# Avoid recent compilers to optimizing out libm function calls
+set(CMAKE_REQUIRED_FLAGS "-O0")
 set(LIBM_TEST_SOURCE "#include<math.h>\nfloat f; int main(){sqrt(f);return 0;}")
 check_c_source_compiles("${LIBM_TEST_SOURCE}" HAVE_MATH)
 if(NOT HAVE_MATH)


eisenhauer:
Yeah, we can't accept direct patches into the thirdparty/* inclusions in ADIOS, they'd just get overwritten, so the answer is to patch EVPath and then pull the new version into ADIOS.

But, I'm not sure of the patch.  This unconditionally adds libm as a dependency, yet that test exists in EVPath CMake because libm didn't exist everywhere (sometimes those functions are included in libc).  So I'd not be sure that this wouldn't break other builds.  Can you tell me more about how this returns false results?  I'd think the answer would be to fix the conditional...

vicentebolea:
@eisenhauer noted, we can prepare a patch upstream. The "quickfix" was proposed for one rpm package  but  absolutely we want to get a deeper understanding of why the test pass but later we get errors. I suspect that either:

1. The LDFLAGS /LDADD  for the test are different than the ones for the EVPATH target
2.  Somehow newer gcc has super aggressive code optimization which impede assembly  functions calls that either change state nor its return value is used.

Here is build which shows the error:
https://build.opensuse.org/package/live_build_log/home:vicentebolea/ADIOS2/openSUSE_Tumbleweed/x86_64

eisenhauer:
So, it may indeed be that the sqrt() call in the CMake test code is getting optimized away so we don't detect the need for -lm. That would call for a different test code in LIBM_TEST_SOURCE, maybe something where the result of the sqrt() is used in the return value.  Can you maybe try something like:
set(LIBM_TEST_SOURCE "#include<math.h>\nfloat f; int main(){return (sqrt(f)>1.0);}")

vicentebolea:
@eisenhauer opened PR upstream https://github.com/GTkorvo/evpath/pull/57/files

:smile:

eisenhauer:
@vicentebolea, I think we have to do this differently to make things work in ADIOS.  Rather than duplicating the patch here, we need to use the thirdparty/evpath/update.sh script to pull in directly from the EVPath repo. That makes sure we have the same commit IDs, so that future changes get pulled in the same way. (Or so I think. )

vicentebolea:
@eisenhauer I agree, consider this a one-use PR for just testing it in the CI :+1: ,  I will close this after everything pass and open another PR for updating EVPATH.

vicentebolea:
Closing this, since this has been resolved at #2861

PR #2863, opened 2021-09-13
Fix MultiBlockChangingShape test

eisenhauer:
One segment of this test previously used EndStep() as an alias for PerformGets().  In fact, in BP3/4 readers EndStep is largely implemented by calling PerformGets(), but that may sometime change and we don't want to encourage the use of EndStep without a matching BeginStep, or the use of Begin/EndStep in random access mode (which the test was running in).  This PR modifies that test to expect a logic error exception to be thrown on EndStep() and changes BP3/4/5 readers to throw that exception if EndStep() is called without a matching BeginStep().


eisenhauer:
Hierarchical test broke when we required an EndStep to occur only after a successful BeginStep, (the BeginSteps in the later loop all fail in the prior version of this test).  So, this commit merges those tests into one loop with all Inq's inside Begin/EndStep pairs.   @dmitry-ganyushin and @pnorbert, please look and make sure I did this right.

Issue #2867, opened 2021-09-14
Add functionality in Variable object to remove current Operations


williamfgc:
**Why is this feature important?**
Allows for applying operations (e.g. compression) only at certain steps depending on certain conditions.

**What is the potential impact of this feature in the community?**
User have more flexibility to storage compress data or not.

**Is your feature request related to a problem? Please describe.**
The current functionality is all (steps) or nothing for compression. 

**Describe the solution you'd like and potential required effort**
A function like `Variable<T>::RemoveOperations()` would do the work

**Describe alternatives you've considered and potential required effort**
A function that modifies the current Operations vector, but it could be more work than required

**Additional context**
Might have to be added to C++, C, Fortran, Python bindings. 
Need to check if high-level bindings might be affected (or it's a current bug).
Will do a PR to illustrate the API usage.

PR #2868, opened 2021-09-14
Remove operations

williamfgc:
Add API function in C++ `Variable<T>::RemoveOperations();` to allow for on and off compression
Add ZFP tests
Related to #2867

### DIFF
--- a/bindings/CXX11/adios2/cxx11/Variable.cpp
+++ b/bindings/CXX11/adios2/cxx11/Variable.cpp
@@ -183,6 +183,14 @@ namespace adios2
     }                                                                          \
                                                                                \
     template <>                                                                \
+    void Variable<T>::RemoveOperations()                                       \
+    {                                                                          \
+        helper::CheckForNullptr(m_Variable,                                    \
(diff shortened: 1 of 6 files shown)


pnorbert:
This sounds like a good idea, however, we need to make sure BP5 will be able to deal with this. @eisenhauer

williamfgc:
@pnorbert thanks, I agree. This sits at a higher layer (`VariableBase`), so if BP5 supports Operations the same as BP3, BP4 it should be seamless. We can integrate once BP5 have tests for compression.

eisenhauer (approved):
> This sounds like a good idea, however, we need to make sure BP5 will be able to deal with this. @eisenhauer

Since BP5 wants to put things like operator info in metameta info, not encoding it into every timestep's metadata, we'll have add a mechanism for tracking operator changes, be they additions or deletions, but I'll take care of that later.  So I have no problems with allowing operator removal.

williamfgc:
Thanks @pnorbert @eisenhauer . I will find some time to extend this API call to the other language bindings.

PR #2870, opened 2021-09-17
Simplify operator API in preparation for adding into BP5, Part I

JasonRuonanWang:


JasonRuonanWang:
The Operator class API was designed in an overly complicated way. It's unnecessary to have multiple decompress function signatures, with some of them saving redundant information in BP buffers. For example, some decompressors need block count dims while others need a single sizeOut. The sizeOut number can be easily calculated using block count dims, so they can use one single signature, and all compressors should not need to put this sizeOut into BP buffer separately since there is already the block count which is compulsory. If any decompressor's output data does not have the same size as accumulate(count)*typeSize then there is no way the data is correct, and in this case it should directly throw an exception. Saving another sizeOut variable in BP buffers will not help anything. Creating multiple function signatures for such things is creating more problems than benefits.

PR #2898, opened 2021-10-11
reorder cmake build option messages

JasonRuonanWang:
As we have more and more build options, the cmake output is getting a little confusing. This PR is to reorder the build option output in the order of ADIOS2 engines, languages/programming envs, operators, transports, and others. Items in each category are in alphabetic order.


JasonRuonanWang:
@eisenhauer Oh okay. Thank you!

PR #2901, opened 2021-10-12
Daos

lwan86:


eisenhauer:
> @eisenhauer Do you know why the first test always fails?

The "format" test?  You need to run clang-format (version 7.1.0) against your files.  
I.E. do:
clang-format -i -style=file <list of all the .h, .cpp and .tcc files in your commit>
But you must first install the right version of clang-format.  Different versions decide slightly differently on indentation, so it can mess you up if it's wrong.

lwan86:
@pnorbert @chuckatkins All the tests have passed except one of the sanitizers. I click it, but it doesn't show any details. What should I do now?

PR #2914, opened 2021-10-19
warn users when compressed buffer is larger than original data

JasonRuonanWang:


eisenhauer:
Do you want to think about limiting the warning to once (per rank maybe)?  Just thinking that if the user actually hit this enough to matter, warning each time would probably be worse for his performance than the ineffective compression...

JasonRuonanWang:
@eisenhauer Good question. But there is no MPI communicator exposed to the operator classes. The only thing I can do is to leave this work to engines.

eisenhauer:
Yeah, I was thinking more in terms of something like this:
    static bool warned = false;
    if ((sizeIn < destOffset) && !warned)
    {
        std::cerr
            << "ADIOS2 bzip2 operator warning: compression buffer size larger "
               "than original data size. Consider disabling compression."
            << std::endl;
        warned = true;
    }

It's not ideal, it doesn't tell you what variable, engine, etc., and you'll get one per rank if it happens across the whole application, but at least you won't get hundreds of thousands of them...

PR #2920, opened 2021-10-20
ZFP: enable cuda

vicentebolea:
This PR allows using ZFP [de]compressing algorithms in CUDA|OMP devices if possible.

To allow the user to chose the device where to run ZFP, this PR also  introduces a new parameter for the CompressZFP operator named `backend` with the values `cuda|omp|serial`.

Incidentally it also fixes an issue with zfp unit tests seen with newer versions of ZFP, here is the commit message of that fix:

>  This was an oddity in the ZFP unit tests since the name of the tests and
    its suite are `rate` something, but some of them hiddenly used accuracy
    instead of rate.
    Moreover, the value given to accuracy was the same value as the rate
    which is always > 1, while accuracy must be by definition less the 1,
    this that perhaps worked in previous versions of ZFP correcting these
    input values but it is definitely not that case for current versions.
    This commit resolves that and their associate errors.
    
In any case is out of the ADIOS2 scope to test all of the ZFP parameters, just testing one should be enough to make sure that we pass down those parameters to ZFP.

### DIFF
--- a/source/adios2/common/ADIOSTypes.h
+++ b/source/adios2/common/ADIOSTypes.h
@@ -305,8 +305,16 @@ namespace zfp
 namespace key
 {
 constexpr char accuracy[] = "accuracy";
+constexpr char backend[] = "backend";
 constexpr char rate[] = "rate";
 constexpr char precision[] = "precision";
+}
+
+namespace value
+{
+constexpr char backend_cuda[] = "cuda";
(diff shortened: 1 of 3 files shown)


vicentebolea:
@chuckatkins I think this is ready to go, in my machine it uses GPU correctly:
Screenshot from 2021-10-21 19-48-23

PR #2937, opened 2021-10-29
CMAKE: add check for ZFP_CUDA

vicentebolea:
In the previous PR #2920 which added among other things tests for ZFP with CUDA it does not check if the current ZFP installation supports ZFP. 

For this purpose I added a PR at the ZFP which adds a parameter to see this (https://github.com/LLNL/zfp/pull/147).

While this has been merged to keep compatibility to previous commits of ZFP I added some logic at our cmake code to figure out if we have ZFP_CUDA.

__EDITED__

I made some refactor at the compressZFP so that it moves out the include from the header to the implementation file, also I changed some of the defines provide from ZFP to more stable ones. 

I also made the code compatible for ZFP 0.5.1 (the one that we publicly support) to 0.5.5 (latest).

### DIFF
--- a/cmake/DetectOptions.cmake
+++ b/cmake/DetectOptions.cmake
@@ -95,6 +95,30 @@ elseif(ADIOS2_USE_ZFP)
 endif()
 if(ZFP_FOUND)
   set(ADIOS2_HAVE_ZFP TRUE)
+  set(ADIOS2_HAVE_ZFP_CUDA ${ZFP_CUDA})
+
+  # Older versions of ZFP
+  if(NOT ADIOS2_HAVE_ZFP_CUDA)
+    get_target_property(ZFP_INCLUDE_DIRECTORIES zfp::zfp INTERFACE_INCLUDE_DIRECTORIES)
+    set(CMAKE_REQUIRED_INCLUDES ${ZFP_INCLUDE_DIRECTORIES})
+    set(CMAKE_REQUIRED_LIBRARIES zfp::zfp)
+    include(CheckCSourceRuns)
+    check_c_source_runs("
+    #include <zfp.h>
+
+    int main()
+    {
+      zfp_stream* stream = zfp_stream_open(NULL);
+      return !zfp_stream_set_execution(stream, zfp_exec_cuda);
+    }"
+    ADIOS2_HAVE_ZFP_CUDA)
+    unset(CMAKE_REQUIRED_INCLUDES)
+    unset(CMAKE_REQUIRED_LIBRARIES)
+  endif()
+
+  if(ADIOS2_HAVE_ZFP_CUDA)
+    add_compile_definitions(ADIOS2_HAVE_ZFP_CUDA)
+  endif()
 endif()
 
 # SZ
--- a/source/adios2/common/ADIOSTypes.h
+++ b/source/adios2/common/ADIOSTypes.h
@@ -312,7 +312,9 @@ constexpr char precision[] = "precision";
 
 namespace value
 {
+#ifdef ADIOS2_HAVE_ZFP_CUDA
 constexpr char backend_cuda[] = "cuda";
+#endif
 constexpr char backend_omp[] = "omp";
 constexpr char backend_serial[] = "serial";
 }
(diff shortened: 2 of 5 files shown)


JasonRuonanWang:
Hi I have got this error if my ZFP version is too old to support cuda. But cmake still thinks the ZFP version is okay. Will this PR solve this problem?

/home/w4g/d/a2/source/adios2/operator/compress/CompressZFP.cpp: In member function ‘zfp_stream* adios2::core::compress::CompressZFP::GetZFPStream(const Dims&, adios2::DataType, const Params&) const’:
/home/w4g/d/a2/source/adios2/operator/compress/CompressZFP.cpp:287:22: error: ‘zfp_exec_cuda’ was not declared in this scope
             policy = zfp_exec_cuda;
                      ^~~~~~~~~~~~~
/home/w4g/d/a2/source/adios2/operator/compress/CompressZFP.cpp:287:22: note: suggested alternative: ‘zfp_exec_omp’
             policy = zfp_exec_cuda;
                      ^~~~~~~~~~~~~
                      zfp_exec_omp

vicentebolea:
> Hi I have got this error if my ZFP version is too old to support cuda. But cmake still thinks the ZFP version is okay. Will this PR solve this problem?
> 
> /home/w4g/d/a2/source/adios2/operator/compress/CompressZFP.cpp: In member function ‘zfp_stream* adios2::core::compress::CompressZFP::GetZFPStream(const Dims&, adios2::DataType, const Params&) const’: /home/w4g/d/a2/source/adios2/operator/compress/CompressZFP.cpp:287:22: error: ‘zfp_exec_cuda’ was not declared in this scope policy = zfp_exec_cuda; ^~~~~~~~~~~~~ /home/w4g/d/a2/source/adios2/operator/compress/CompressZFP.cpp:287:22: note: suggested alternative: ‘zfp_exec_omp’ policy = zfp_exec_cuda; ^~~~~~~~~~~~~ zfp_exec_omp

Yeah this PR attempts to solve this, but now that I see it it needs a bit more work

cmake/DetectOptions.cmake
```diff
@@ -95,6 +95,30 @@ elseif(ADIOS2_USE_ZFP)
...
+      zfp_stream* stream = zfp_stream_open(NULL);
+      return !zfp_stream_set_execution(stream, zfp_exec_cuda);
+    }"
+    ADIOS2_HAVE_ZFP_CUDA)
+    unset(CMAKE_REQUIRED_INCLUDES)
+    unset(CMAKE_REQUIRED_LIBRARIES)
+  endif()
+
+  if(ADIOS2_HAVE_ZFP_CUDA)
+    add_compile_definitions(ADIOS2_HAVE_ZFP_CUDA)
```
vicentebolea, line 120: @chuckatkins is it ok to add defines like this in ADIOS?

vicentebolea:
@JasonRuonanWang could you point me to which version of ZFP are you using?

cmake/DetectOptions.cmake
JasonRuonanWang: I actually don't know. I had this error and then I realized that it was probably a version issue, and then I just upgraded my ZFP installation and then it worked. So... haven't kept any info of the old installation. Sorry

JasonRuonanWang:
> @JasonRuonanWang could you point me to which version of ZFP are you using?

I actually don't know. I had this error and then I realized that it was probably a version issue, and then I just upgraded my ZFP installation and then it worked. So... haven't kept any info of the old installation. Sorry

vicentebolea:
@JasonRuonanWang find many changes here, I tested it for 0.5.1 to 0.5.5 and everything is good, I guess you have to take my word here until we have the CI build to test this. There is also a bit of refactor hereto encapsulate the dependencies a bit better.

JasonRuonanWang:
@vicentebolea I can live with this. I am fine as long as it compiles and allows me to work on the operators.

JasonRuonanWang:
@vicentebolea Is this good to go?

vicentebolea:
@JasonRuonanWang yes

PR #2938, opened 2021-10-30
Fix cmake cuda errors and warnings

JasonRuonanWang:


JasonRuonanWang:
This PR is to fix the following cmake error, and some other warnings about CMAKE_CUDA_ARCHITECTURES

CMake Error in source/adios2/CMakeLists.txt:
  Target "adios2_core" requires the language dialect "CUDA17" (with compiler
  extensions), but CMake does not know the compile flags to use to enable it.

PR #2939, opened 2021-10-31
temporarily disable LargeMetadata.BP5 test

JasonRuonanWang:
@eisenhauer Can we temporarily disable the LargeMetadata.BP5 test since it's not very stable? It fails at least 30% of the time when I run CI, and sometimes I need to re-run multiple times to get it passed.


eisenhauer:
I'm not opposed, but I'm curious.  I've done a lot of PRs and I've only seen this fail in the NoTestRerun case.  I'm not sure why you're having more trouble.  Since when it fails, it dies with MPI problems, can we RUN_SERIAL first?

JasonRuonanWang:
@eisenhauer Sure I can do that. I have got this error in all of my recent PRs, maybe 3 or 4, and for the CUDA one I opened yesterday I re-ran it three times to get it passed. I have also helped re-run Junmin's PR yesterday, which was also stuck at this test.

JasonRuonanWang:
@eisenhauer I don't know how to add the NON-MPI test only for BP5. There is only bp_gtest_add_tests_helper and bp3_bp4_gtest_add_tests_helper, but no bp5_gtest_add_tests_helper. Also since the purpose of this test is for *Large Metadata*, a serial run would not generate metadata that is large enough anyway. Moreover, I don't think there is really anything wrong with BP5 at all. It's just that some CI virtual machines don't like large tests. I have had enough experience with CI just randomly failing some large tests that always run successfully anywhere else on the earth. So I will just merge this as is, to make near future PRs easier. Feel free to change it in the way you want.

eisenhauer:
I didn't mean skip the mpi, but just not to run any other tests concurrently. No worries, I'll give it a shot.

PR #2943, opened 2021-11-02
Try restoring LargeMetadata for BP5

eisenhauer:


eisenhauer:
@JasonRuonanWang So, I'm not seeing any heisen-failures on BP5LargeMetata on these runs, but I'll repeat them a few times to be diligent.  And of course @chuckatkins, I can't merge because we seem to have a python version issue on macos1014_xcode103_ninja (unrelated to changes).

JasonRuonanWang:
@eisenhauer Just wondering if it's possible to make it run exclusively only on certain CI virtual machines? I guess it's the debian-sid-openmpi VM that caused most of the random failures. Other VMs don't seem to have this issue.

eisenhauer:
Oh, is it the Debian thing where you’re seeing failures?  We don’t require success on Debian in CI because it’s wildly unreliable (and I say wildly because by default we “succeed” if one out of 5 possible runs is successful.  Debian doesn’t do that reliably.  Because of that I generally just ignore the Debian status (unless it’s a compile-time failure, that’s worth fixing.)

From: Jason Wang ***@***.***>
Date: Tuesday, November 2, 2021 at 1:12 PM
To: ornladios/ADIOS2 ***@***.***>
Cc: Eisenhauer, Greg S ***@***.***>, Mention ***@***.***>
Subject: Re: [ornladios/ADIOS2] Try restoring LargeMetadata for BP5 (PR #2943)

@eisenhauer<https://github.com/eisenhauer> Just wondering if it's possible to make it run exclusively only on certain CI virtual machines? I guess it's the debian-sid-openmpi VM that caused most of the random failures. Other VMs don't seem to have this issue.

JasonRuonanWang:
@eisenhauer Yeah I know it's optional, but I still feel unsafe if something fails when I push the merge button :)

eisenhauer:
Well, for what it's worth, the 5-test-rerun thing looks like it's covering up a lot of failures.  When we do no-test-rerun, el7-gnu8-openmpi-ohpc-static and el7-gnu8-openmpi-ohpc always have 50-75 failures, mostly timeouts where it looks like MPI just hangs.  (Which may be part of why those VMs take so long to run, because we repeat those 2 minute timeouts up to 5 times.)  The InSitu engine is also a perennial trouble-maker and likely has a race condition on open.  In relative terms, I don't worry about debian much at all!

JasonRuonanWang:
Okay then I will just ignore the debian tests from now on.

PR #2946, opened 2021-11-03
Adding documentation for using Cuda buffers

anagainaru:
1. Cmake output for the cuda architecture will (hopefully) prevent users from trying to run on a different architecture than the default one (set to V100 for Summit runs) without setting the correct `CMAKE_CUDA_ARCHITECTURES`

2. Minimal text on the limitations and how to device buffers for readthedocs

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -308,6 +308,7 @@ message("    ${CMAKE_CXX_COMPILER}")
 message("")
 if(ADIOS2_HAVE_CUDA)
 	message("  Cuda Compiler : ${CMAKE_CUDA_COMPILER} ")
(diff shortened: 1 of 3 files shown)


anagainaru:
@chuckatkins the checks that failed don't seem to be related to my changes. Any advice?

eisenhauer:
@chuckatkins This is the same CI problem as PR #2943

eisenhauer:
@chuckatkins For this and another PR, I temporarily removed macos1014_xcode103_ninja as required so they could be merged.  I've restored it because I wasn't sure what you'd like to do longer term.  However, macos1014 seems to be being phased out, so maybe we just keep 1015?

PR #2948, opened 2021-11-08
Update SZ operator to support latest SZ version 2.1.12

JasonRuonanWang:


JasonRuonanWang:
1. Updated SZ Operator to support the latest SZ version, 2.1.12.
2. Fixed a bug in the original SZ Operator class which used 1 for blank dimension. Now it uses 0 as advised by SZ.
3. Allows SZ compression for 5D data, as is supported in SZ 2.1.12.
4. As this change modified the compression buffer header, a V2 SZ buffer format has been introduced and made default. V1 buffer format is still supported for legacy data.

PR #2950, opened 2021-11-09
Removed InSituMPI engine

JasonRuonanWang:


JasonRuonanWang:
Since the InSituMPI engine is no longer maintained, and SSC implemented all of its functionalities, we decided to remove the InSituMPI engine.

Changes in this PR:
1. Removed InSituMPI engine
2. Made SSC engine compulsory for MPI build as InSituMPI engine used to be
3. Transferred some InSituMPI tests to SSC

PR #2952, opened 2021-11-17
Add adiosLog in helper for better looking output

JasonRuonanWang:
For long time we have only had arbitrarily styled runtime output. Some of them are very ambiguous and uninformative. This PR is the start of an attempt to have a more unified and finer-grained output style, which at least let users know the outputs are from ADIOS2 rather than some thirdparty libraries, and have some ideas which layer of ADIOS2 they are from. 

Some basic example outputs that have been implemented are:

```
 [Wed Nov 17 02:27:02 2021] [ADIOS2 Output] [Rank 0] <Engine> <BP4Writer> <Open> : myVector_cpp.bp
 [Wed Nov 17 02:27:02 2021] [ADIOS2 Output] [Rank 0] <Engine> <BP4Writer> <PutDeferred> : bpFloats
 [Wed Nov 17 02:27:02 2021] [ADIOS2 Output] [Rank 0] <Engine> <BP4Writer> <PutDeferred> : bpInts
 [Wed Nov 17 02:27:02 2021] [ADIOS2 Output] [Rank 0] <Engine> <BP4Writer> <Close> : myVector_cpp.bp
 [Wed Nov 17 02:27:02 2021] [ADIOS2 Output] [Rank 0] <Engine> <BP4Writer> <PerformPuts> : 
```

### DIFF
--- a/source/adios2/CMakeLists.txt
+++ b/source/adios2/CMakeLists.txt
@@ -36,6 +36,7 @@ add_library(adios2_core
   helper/adiosXMLUtil.cpp
   helper/adiosYAML.cpp
   helper/adiosCUDA.cu
+  helper/adiosLog.cpp
 
 #engine derived classes
   engine/bp3/BP3Reader.cpp engine/bp3/BP3Reader.tcc
--- a/source/adios2/core/Engine.h
+++ b/source/adios2/core/Engine.h
@@ -516,9 +516,6 @@ protected:
      *  if no communicator is passed */
     helper::Comm m_Comm;
 
-    /** added to exceptions to improve debugging */
-    std::string m_EndMessage;
-
     /** keeps track of current advance status */
     StepStatus m_AdvanceStatus = StepStatus::OK;
 
--- a/source/adios2/engine/bp3/BP3Reader.cpp
+++ b/source/adios2/engine/bp3/BP3Reader.cpp
@@ -135,8 +135,7 @@ void BP3Reader::Init()
     if (m_OpenMode != Mode::Read)
     {
         throw std::invalid_argument(
-            "ERROR: BPFileReader only supports OpenMode::Read from" + m_Name +
-            " " + m_EndMessage);
+            "ERROR: BPFileReader only supports OpenMode::Read from" + m_Name);
     }
 
     // if IO was involved in reading before this flag may be true now
--- a/source/adios2/engine/bp3/BP3Writer.cpp
+++ b/source/adios2/engine/bp3/BP3Writer.cpp
@@ -32,7 +32,6 @@ BP3Writer::BP3Writer(IO &io, const std::string &name, const Mode mode,
 {
     PERFSTUBS_SCOPED_TIMER("BP3Writer::Open");
     m_IO.m_ReadStreaming = false;
-    m_EndMessage = " in call to IO Open BPFileWriter " + m_Name + "\n";
     Init();
 }
 
--- a/source/adios2/engine/bp4/BP4Reader.cpp
+++ b/source/adios2/engine/bp4/BP4Reader.cpp
@@ -30,7 +30,8 @@ BP4Reader::BP4Reader(IO &io, const std::string &name, const Mode mode,
   m_MDIndexFileManager(m_Comm), m_ActiveFlagFileManager(m_Comm)
 {
     PERFSTUBS_SCOPED_TIMER("BP4Reader::Open");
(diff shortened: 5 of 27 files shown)


pnorbert:
Great idea, thanks. However the real message starts above offset 80 offset with this superfluous format. Is the word Output and the class id <Engine> necessary here? Are all those brackets really necessary?

`[Wed Nov 17 02:27:02 2021] ADIOS2 [Rank 0] <BP4Writer> <Open> : myVector_cpp.bp`
`Wed Nov 17 02:27:02 2021 ADIOS2 Rank 0 BP4Writer Close(): myVector_cpp.bp`

Moreover, can we have the rank optional in the start of the message itself, so we can omit it from messages written only from a single rank to notify the user (when rank is irrelevant).

`[Wed Nov 17 02:27:02 2021] ADIOS2 <BP4Writer> <Open> : [Rank 0] myVector_cpp.bp`
`[Wed Nov 17 02:27:02 2021] ADIOS2 <ADIOS> <Init> : ERROR: adios2.xml does not exist`

dmitry-ganyushin:
Nice addition, but one should add an aggregation of error messages.
Otherwise, you would obtain 1000 the same error messages when running on
1000 processors. To restrict an error message to rank 0 is usually not the
best solution.

JasonRuonanWang:
> Great idea, thanks. However the real message starts above offset 80 offset with this superfluous format. Is the word Output and the class id necessary here? Are all those brackets really necessary?
> 
> `[Wed Nov 17 02:27:02 2021] ADIOS2 [Rank 0] <BP4Writer> <Open> : myVector_cpp.bp` `Wed Nov 17 02:27:02 2021 ADIOS2 Rank 0 BP4Writer Close(): myVector_cpp.bp`
> 
> Moreover, can we have the rank optional in the start of the message itself, so we can omit it from messages written only from a single rank to notify the user (when rank is irrelevant).
> 
> `[Wed Nov 17 02:27:02 2021] ADIOS2 <BP4Writer> <Open> : [Rank 0] myVector_cpp.bp` `[Wed Nov 17 02:27:02 2021] ADIOS2 <ADIOS> <Init> : ERROR: adios2.xml does not exist`

*Output* is necessary because besides output there is also warning, error and exception. As you see, removing the brackets didn't really save much space, if the message body is long enough, you can still easily get over 80 chars. By comparing the logging module of other systems, I don't see this a big problem. The outputs are only enabled if users explicitly sets Verbosity>0. Only errors and exceptions are printed regardless of Verbosity. But once an app runs into such situation, it's approaching termination anyway, and the entire run is highly likely an unsuccessful run. In this case people probably won't care how many more chars we use for printing brackets, but a quick and informative hint what's going on. Those non necessary things, in this case will make the message stand out of other messages from the OS or thirdparty libraries.

Hiding rank number will be available. I will have another much simpler API just for non-engine modules which do not even know it's rank ID and verbosity.

JasonRuonanWang:
@dmitry-ganyushin I can add it. But if an app prints 1000 errors and still runs normally, then those 1000 errors should not be errors, but outputs or warnings. If those are 'real' errors, then people probably want to see them all because that can provide information which ranks have this error, which ranks do not have. Again, if your app is terminated due to error, you won't care about the additional 1000 lines in output, but the full information what's going on. If you run on 10,000 nodes, and limit the error message only to 1000, then you are possibly going to lose the real error message that comes the 1001th, but get the first 1000 irrelevant messages. 

Also, if you want to count the total number of errors printed across ranks, you will have to use MPI communications to aggregate the info, which will be a much much bigger performance killer than just printing out all messages.

kshitij-v-mehta:
> Output is necessary because besides output there is also warning, error and exception

In that case, could you call it INFO instead of Output?

JasonRuonanWang:
> > Output is necessary because besides output there is also warning, error and exception
> 
> In that case, could you call it INFO instead of Output?

Sure.

dmitry-ganyushin:
I am just writing about complaints of users who report that logs got
overflooded at large scale with adios messages. Good lock to implement it,
although I think that such log libraries should exist.

JasonRuonanWang:
@dmitry-ganyushin That's true. It could be annoying if it prints out a lot of useless info but the run is still good. In this case the best thing we can do is probably try to change some of the naughty ERROR to INFO, which will then be controlled by the engine parameter Verbose. If it does not kill the run, it shouldn't be error, but something that users can choose to never see.

PR #2954, opened 2021-11-18
modify daos cmake file to make it find daos library more easily

JasonRuonanWang:
In the original FindDAOS.cmake file there is a condition 

if(CMAKE_VERSION VERSION_LESS 3.12)

which assumes that if cmake version is newer than 3.12 then it should be able to automatically find DAOS library. However, cmake 3.21 does not seem to find the DAOS library at all even for a DAOS installation in /usr/local. By removing this condition, any cmake version is now able to find DAOS automatically.

### DIFF
--- a/cmake/FindDAOS.cmake
+++ b/cmake/FindDAOS.cmake
@@ -25,17 +25,15 @@
 #                           variable, it will override any setting specified
 #                           as an environment variable.
 
-if(CMAKE_VERSION VERSION_LESS 3.12)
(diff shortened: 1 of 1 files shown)


JasonRuonanWang:
message(STATUS "DAOS_ROOT is is \"$ENV{DAOS_ROOT}\"")

This original line only prints DAOS_ROOT when it's copied from the environmental variable, but does not do anything with it if it's passed by cmake argument -DDAOS_ROOT. It should reflect DAOS_ROOT regardless of its source I guess.

PR #2962, opened 2021-11-30
More refactoring work on Operator

JasonRuonanWang:


JasonRuonanWang:
1. Removed CompressorFactory class. There are only two functions Compress and Decompress, which do not need any member variables, so there is no point to have a class. 
2. IsCompressionAvailable function was originally in DataMan serializer, but was recently copied to core::compress::CompressorFactory for BP5. The original one in DataMan serializer should be removed. 
3. Added m_TypeEnum in Operator class for more convenient reference for serializing and deserializing compression buffer header

PR #2972, opened 2021-12-06
Unified operator parameters, and centralized all operator logics into the Operator framework

JasonRuonanWang:
This PR refactors the Operator framework in the following aspects:

1. In the original Operator framework design, 3 maps were used to describe parameters and properties of an operator: an internal map inside the Operator class, two external maps outside the Operator class but needed to be always carried together with an Operator object in a struct wrapper. This weird combination of procedural programming and object-oriented programming caused a lot of confusions to users and even ADIOS2 developers. Nobody remembers which map is for what. In fact, the internal map has NEVER been used in any Operator implementations. So the only effect it has ever taken so far, is to confuse users. This PR removes such a confusing design, and makes the internal map the only place to configure parameters of an Operator. Thus, an Operator object becomes the only thing that developers and users need to be aware of when they deal with ADIOS2 operations.

2. In the original Operator framework design, adding a new operator needed changes in many places all over the ADIOS2 library. Common sense tells us that this is a good case for implementing a class factory to centralize all the knowledge needed to implement and use a new child class. So future developers don't need to spend a month learning the 1,000 things, but rather just spend an hour going over the Operator base class and the OperatorFactory class.

3. In the original Operator framework design, users had to first define an Operator with an ADIOS object, and then add it to a Variable object. In fact, the only place where an Operator can take effect is inside a Variable object. Even an Engine or a Serializer needs to access an operator through a Variable. There is no point to make constructing an Operator so complicated, which made it unnecessarily difficult to maintain the logic correctly. Since a Variable object is the only place where an Operator object can take effect, it should then be the only place where an Operator can be constructed too. Thus, it makes it much easier to manage the lifecycle of Operator objects, as well as separate configurations between multiple Variables and Operators. It also makes the user API logic crystal clear and extremely simple -- users can now add an Operator on-the-fly to a Variable in a single line of code, never needing to bother keeping an ADIOS object reference across their workflows. Considering the backward compatibility, the old API is still there. But we should encourage users to use the new API.

### DIFF
--- a/bindings/CXX11/adios2/cxx11/Operator.cpp
+++ b/bindings/CXX11/adios2/cxx11/Operator.cpp
@@ -46,9 +46,4 @@ Params Operator::Parameters() const
 // PRIVATE
 Operator::Operator(core::Operator *op) : m_Operator(op) {}
 
-std::string ToString(const Operator &op)
-{
-    return std::string("Operator(Type: \"" + op.Type() + "\")");
-}
-
 } // end namespace adios2
--- a/bindings/CXX11/adios2/cxx11/Operator.h
+++ b/bindings/CXX11/adios2/cxx11/Operator.h
@@ -78,8 +78,6 @@ private:
     core::Operator *m_Operator = nullptr;
 };
 
-std::string ToString(const IO &io);
-
 } // end namespace adios2
 
 #endif /* ADIOS2_BINDINGS_CXX11_CXX11_OPERATOR_H_ */
--- a/bindings/CXX11/adios2/cxx11/Variable.cpp
+++ b/bindings/CXX11/adios2/cxx11/Variable.cpp
@@ -172,6 +172,13 @@ namespace adios2
     }                                                                          \
                                                                                \
     template <>                                                                \
+    size_t Variable<T>::AddOperation(const std::string &type,                  \
+                                     const Params &parameters)                 \
+    {                                                                          \
+        return m_Variable->AddOperation(type, parameters);                     \
+    }                                                                          \
+                                                                               \
+    template <>                                                                \
     std::vector<Operator> Variable<T>::Operations() const                      \
     {                                                                          \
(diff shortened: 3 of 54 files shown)


JasonRuonanWang:
This PR refactors the Operator framework in the following aspects:

1. In the original Operator framework design, 3 maps were used to describe parameters and properties of an operator: an internal map inside the Operator class, two external maps outside the Operator class but needed to be always carried together with an operator object in a struct wrapper. This weird combination of procedural programming and object-oriented programming caused a lot of confusions to users and even ADIOS2 developers. Nobody remembers which map is for what. In fact, the internal map has NEVER been used in any operator implementations. So the only effect it has ever taken so far, is to confuse users. This PR removes such a confusing design, and makes the internal map the only place to configure parameters of an operator. Thus, an operator object becomes the only thing that developers and users need to be aware of when they deal with operations. 

2. In the original Operator framework design, adding a new operator needs changes in many places all over the ADIOS2 library. Common sense tells us that this is a good case for implementing a class factory to centralize all the knowledge needed to implement and use a new child class. So future developers don't need to spend a month remembering the 1,000 things, but just spend an hour going over the operator base class and the class factory. 

3. In the original Operator framework design, users have to first define an operator with the ADIOS object, and then add it to a Variable object. In fact, the only place that an operator can take effect is inside a Variable object. Even an Engine or a Serializer needs to access an operator through a Variable. There is no point to make constructing an operator so complicated, which makes it unnecessarily difficult to maintain the logic correctly. Since a Variable object is the only place that an Operator object can take effect, it should then be the only place where an Operator can be constructed too. Thus, it makes it much easier to manage the lifecycle of an operator object, as well as separate the configurations between multiple Variables and Operators. It also makes the user API logic crystal clear and extremely simple -- users can now add an operator on-the-fly to a variable in a single line of code, and never need to bother keeping an ADIOS object reference across their workflows. Considering the backward compatibility, the old API is still there. But we should encourage users to use the new API.

PR #2974, opened 2021-12-08
WIP: Bp5 async io

pnorbert:
Testing the CI if all my additions will pass

### DIFF
--- a/bindings/C/adios2/c/adios2_c_adios.cpp
+++ b/bindings/C/adios2/c/adios2_c_adios.cpp
@@ -233,6 +233,44 @@ adios2_error adios2_finalize(adios2_adios *adios)
     }
 }
 
+/** Inform ADIOS about entering communication-free computation block
+ * in main thread. Useful when using Async IO */
+adios2_error adios2_enter_computation_block(adios2_adios *adios)
+{
+    try
+    {
+        adios2::helper::CheckForNullptr(
+            adios,
+            "for adios2_adios, in call to adios2_enter_computation_block");
+        reinterpret_cast<adios2::core::ADIOS *>(adios)->EnterComputationBlock();
+        return adios2_error_none;
+    }
+    catch (...)
+    {
+        return static_cast<adios2_error>(
+            adios2::helper::ExceptionToError("adios2_enter_computation_block"));
+    }
+}
+
+/** Inform ADIOS about exiting communication-free computation block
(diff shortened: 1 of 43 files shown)


bindings/C/adios2/c/adios2_c_adios.cpp
```diff
@@ -233,6 +233,44 @@ adios2_error adios2_finalize(adios2_adios *adios)
     }
 }
 
+/** Inform ADIOS about entering communication-free computation block
+ * in main thread. Useful when using Async IO */
+adios2_error adios2_enter_computation_block(adios2_adios *adios)
```
williamfgc, line 238: @pnorbert great work. Quick thing, would something like `adios2_pause` and `adios2_resume` would illustrate the intention more clearly to the application? `block` is taken in the adios context (as in Variable blocks in `bpls` and BlockInfo). Also, the application can do things other than computation (e.g. moving data to GPU). Unless adios2 is still doing things under the hood when this is called.

pnorbert: You are right, maybe I should not have used block in the name. 

However, it's not a control function. The application only informs adios about what it is doing, and an adios engine may use that info to control the asynchronous I/O. We call it Guided async IO if it does, while the Naive async IO does not care about what the application is doing.

williamfgc: I see, it is communication within the app, not adios. Makes sense due to the MPI+thread policy. Thanks for clarifying. Since this is exclusive to the MPI context, reusing their semantics might be a option `adios2_enter_mpi_local` as in [here](http://www.netlib.org/utk/papers/mpi-book/node14.html)

pnorbert: Well, none of the async IO threads use MPI. But it is to inform adios that the app enters an MPI-free section of the code and thus IO will not interfere with app communication. The IO threads within the compute node however do communicate using shared memory, so the MPI local concept does not apply here.

williamfgc: I see, there is more to be considered. I can't think of something other than `adios2_start_app_nocomm` or `adios2_start_app_comm_self` (but this one might be confusing). `MPI_Probe` returns false if a certain message is not happening.

pnorbert:
This PR was just for making sure the code is still up-to-date

PR #2975, opened 2021-12-08
BP5 MinMax 

eisenhauer:


JasonRuonanWang:
@eisenhauer Do you think we can implement MinMax as an operator? So all engines / serializers can just grab and use? This has been too much repetitive work... and hard work every time

eisenhauer:
Hmm.  We'd have to think about it.  I think the principal issue would be the assumption that MinMax is in the metadata.  You've just modded operators to keep them from producing metadata, so doing the MinMax that way might be backsliding a bit.

JasonRuonanWang:
> Hmm. We'd have to think about it. I think the principal issue would be the assumption that MinMax is in the metadata. You've just modded operators to keep them from producing metadata, so doing the MinMax that way might be backsliding a bit.

That's true. Calculating itself is trivial. Metadata manipulation is the biggest part, which each engine has to do it in its own way anyway.

Issue #2979, opened 2021-12-14
Some ZFP tests fail when ZFP_CUDA is enabled


JasonRuonanWang:
These tests fail when enabling ZFP_CUDA. It only happens with a ZFP build with CUDA enabled. The same ZFP commit without CUDA enabled worked fine. 
 
```

        732 - Engine.BP.*/BPWriteReadZfpHighLevelAPI.ADIOS2BPWriteReadZfp3D/*.BP3.Serial (Subprocess aborted)
	739 - Engine.BP.*/BPWriteReadZfpHighLevelAPI.ADIOS2BPWriteReadZfp3D/*.BP3.MPI (Failed)
	746 - Engine.BP.*/BPWriteReadZfpHighLevelAPI.ADIOS2BPWriteReadZfp3D/*.BP4.Serial (Subprocess aborted)
	753 - Engine.BP.*/BPWriteReadZfpHighLevelAPI.ADIOS2BPWriteReadZfp3D/*.BP4.MPI (Failed)
	798 - Engine.BPEngineTest.ZfpComplex.Serial (Subprocess aborted)
	799 - Engine.BPEngineTest.ZfpComplex.MPI (Failed)
	913 - Engine.DataMan.DataManEngineTest.2D_Zfp.Serial (Subprocess aborted)
	1286 - Bindings.Python.BPZfpHighLevelAPI.MPI (Failed)

```
Testing env:
Ubuntu 20.04
GCC 9.3.0
CUDA 11.5
ADIOS2 master branch
ZFP develop branch built with CUDA enabled

vicentebolea:
@JasonRuonanWang you are right something broke in the last commits, I could reproduce the error. we really need the CI build that tests this

PR #2983, opened 2021-12-21
CUDA: set MAXWELL as the default architecture

vicentebolea:
Fixes: #2979

The issue was that we were only compiling for physical archs higher than our computers arch.

We must let the users to specify the CUDA arch passing `-DCMAKE_CUDA_ARCHITECUTRES="xxyyzz"` 

When it comes to decide for a good default arch, `compute_52` is the oldest non-deprecated CUDA virtual architecture, thus, is a good candidate for a base architecture to support. Setting this as the default `CMAKE_CUDA_ARCHITECTURE` arch resolves the error with the CUDA unit tests.

@JasonRuonanWang

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -167,8 +167,8 @@ if(ADIOS2_HAVE_CUDA)
     set(CMAKE_CUDA_STANDARD 11)
     set(CMAKE_CUDA_STANDARD_REQUIRED TRUE)
     if(NOT DEFINED CMAKE_CUDA_ARCHITECTURES)
-        # default 70 to match Summit V100
-        set(CMAKE_CUDA_ARCHITECTURES 70)
+        # Mininum common non-deprecated architecture
+        set(CMAKE_CUDA_ARCHITECTURES 52)
     endif()
 endif()
 


JasonRuonanWang (approved):
Fine with me. 

But I actually don't know what the best practice is for setting a default cuda architecture. @anagainaru was suggesting that we should set nothing. I slightly disagreed as in that case it would print out a lot of warning information which swipes out all the regular cmake summary. I personally think that it would cause more confusion than setting something and informing users in the cmake summary. 

In my impression with other libraries, setting a default cuda architecture seems to be a more common practice too. But I am not absolutely sure if it's the best practice. So if you have any better suggestions please feel free to suggest.

vicentebolea:
@JasonRuonanWang it appears to me that by default we must build ADIOS2 being as compatible as possible with any other GPU. Thus setting the the default virtual arch to oldest supported sounds about right.

Leaving it off it will solely build for the architecture of the machine which is building it I think (I am not sure about this_, but we do not want that since we want to use ADIOS2 binaries in multiple machines and potentially make RPM/DEB pkgs out of it which must be as universal as possible.

vicentebolea:
@JasonRuonanWang 

According to the docs, when omitting virtual arch to nvcc it defaults to the following:

```
sm_52 is used as the default value; PTX is generated for compute_52 then assembled and optimized for sm_52.
```

However not editing this (or related properties) are discouraged by CMake.

We can possibly lower the virtual version to sm_52 to mimic the default base virtual arch version of nvcc.

anagainaru:
I am fine with it also.

I thought it best to not set anything so users have to use `-DCMAKE_CUDA_ARCHITECUTRES` to avoid the warnings, if some architecture is set by default they might not set anything and maybe in some cases get weird errors or lower performance without understanding why. But I am not an expert, so I will go with whatever you suggest.

anagainaru (approved):

vicentebolea:
@anagainaru @JasonRuonanWang 

how about the virtual version to compute_52 to mimic the default base virtual arch version of nvcc.

It seems that I was wrong and `compute_52` is the lowest supported CUDA virtual arch.

anagainaru:
@vicentebolea That is fine with me

PR #3000, opened 2022-01-11
Base modifications to make BP5 GPU aware

anagainaru:
No span or operations for now.


eisenhauer (approved):

anagainaru:
@eisenhauer What is the clang-format version that I'm suppose to use? Clearly my laptop has a different one

eisenhauer:
> @eisenhauer What is the clang-format version that I'm suppose to use? Clearly my laptop has a different one
clang-format v7.1.0 

You may be able to snag a binary somewhere, rather than build all of clang from scratch.  I run 7.0 and it's usually close enough.

williamfgc:
@anagainaru CI uses v7.1.0

PR #3009, opened 2022-01-19
CMAKE: silence find_package(CUDA)

vicentebolea:
I tried to replicate error message when trying to find CUDA in the CI machines, I found this cmake message:

```
 -- Could not find nvcc, please set CUDAToolkit_ROOT
```

@pnorbert I wonder if this is the error message that you mentioned earlier, if so, this MR silence this if the user has not set `ADIOS2_ENABLE_CUDA=ON`


vicentebolea:
ping @pnorbert

PR #3016, opened 2022-01-24
CUDA: set CUDA dependencies private

vicentebolea:
The CUDA dependency must be private so that it does not propagate.

### DIFF
--- a/testing/adios2/engine/bp/operations/TestBPWriteReadZfpCuda.cpp
+++ b/testing/adios2/engine/bp/operations/TestBPWriteReadZfpCuda.cpp
@@ -4,6 +4,7 @@
  */
 
 #include "CudaRoutines.h"
+
 #include <adios2.h>
 
 #include <algorithm>
--- a/examples/cuda/CMakeLists.txt
+++ b/examples/cuda/CMakeLists.txt
@@ -6,5 +6,5 @@
 enable_language(CUDA)
 
 add_executable(GPUWriteRead_cuda cudaWriteRead.cu)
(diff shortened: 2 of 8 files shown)


anagainaru (approved):
Looks good to me (it was on my list of todo)

vicentebolea:
@anagainaru the issue was a misconfiguration so this patch was not needed. I still have a small change for the dependency type in adios_core and CUDA. Please re-review :)

anagainaru:
Looks ok

vicentebolea:
@anagainaru sorry for keep requesting reviews again and again, this is the last one I promise :innocent:

testing/adios2/engine/bp/operations/TestBPWriteReadZfpCuda.cpp
```diff
@@ -4,8 +4,11 @@
  */
 
 #include "CudaRoutines.h"
+
 #include <adios2.h>
 
+#include <cuda_runtime.h>
```
anagainaru, line 10: I would maybe leave this in the `CudaRoutines.h` file because all kernels will be here so most likely all tests using Cuda will include this file.

anagainaru (approved):
Haha no worries, I looked at the changes.
- yes to `PRIVATE`
- We are using Cuda Runtime so `cuda.h` is indeed not needed
- No strong opinion about where to include `cuda_runtime.h` for the tests

testing/adios2/engine/bp/operations/TestBPWriteReadZfpCuda.cpp
vicentebolea: Good idea

PR #3018, opened 2022-01-25
removed identity operation

JasonRuonanWang:


JasonRuonanWang:
There is apparently no longer such thing as identity operation in ADIOS2. I don't know if it ever existed before, but these codes obviously do nothing now.

williamfgc:
For completeness, this functionality allows the raw buffer to be exposed as-is (hence the identity name), not the decompressed buffer as it's done by default. If not used it's good to remove. @pnorbert a lot of the Operator functionality was done for CODAR and SIRIUS exploratory work with adios beyond 1-to-1 compression and decompression.

PR #3020, opened 2022-01-25
Bp5 append

pnorbert:

PR #3024, opened 2022-01-28
fix compile error newly introduced

JasonRuonanWang:
@pnorbert 

#3020 introduced compile errors which CI didn't catch. This PR is to fix those errors.


pnorbert (approved):
Thanks for fixing this.

PR #3030, opened 2022-01-31
PluginManager and operator plugins support

caitlinross:
Adds a `PluginManager` class that handles loading plugins (whether it be engine or operator). Also adds support for operator plugins. 

This also adds an `EncryptionOperator` that depends on [libsodium](https://doc.libsodium.org/). Right now I have this under `examples/plugins/operator`, but do we want this somewhere else? Also I created an install test for the operator example as I did for the engine example, but CI images will need to be updated to have libsodium available. Which CI images should it be tested on? All?

@chuckatkins


examples/plugins/operator/CMakeLists.txt
```diff
@@ -0,0 +1,22 @@
+#------------------------------------------------------------------------------#
+# Distributed under the OSI-approved Apache License, Version 2.0.  See
+# accompanying file Copyright.txt for details.
+#------------------------------------------------------------------------------#
+
+list(INSERT CMAKE_MODULE_PATH 0 ${CMAKE_CURRENT_SOURCE_DIR})
+find_package(Sodium 1.8 REQUIRED)
```
JasonRuonanWang, line 7: Is this compulsory or optional? If it's compulsory then we probably want to include libsodium in thirdparty, because we want users to be able to compile ADIOS2 out of box without having to manually install dependent libraries, at least for the basic BP engines.

JasonRuonanWang:
> This also adds an `EncryptionOperator` that depends on [libsodium](https://doc.libsodium.org/). Right now I have this under `examples/plugins/operator`, but do we want this somewhere else?

If we just want to showcase how to implement, load and use an operator plugin then we can put it into examples. But I personally think this encryption operator will be so useful that we or our users may want to make it a built-in feature in ADIOS2, rather than managing it as a plugin.

pnorbert:
> If we just want to showcase how to implement, load and use an operator plugin then we can put it into examples. But I personally think this encryption operator will be so useful that we or our users may want to make it a built-in feature in ADIOS2, rather than managing it as a plugin.

The idea is that if this plugin is always built and installed with ADIOS, then it will not cause any performance or convenience penalty if it stays as a plugin to serve as an example, right? 

In this case, libsodium should be required and be in third party. I think for 2.8 this can stay optional.

JasonRuonanWang:
@pnorbert Yeah. If a plugin is only needed by a single app then it's fine to let users manage the plugin code in their apps. But for common features that multiple apps may need, teaching each user how to maintain a plugin code in their apps is definitely going to be more expensive than including the feature as an ADIOS2 internal feature.

caitlinross:
@pnorbert @JasonRuonanWang FYI, libsodium is going to stay optional. Chuck and I looked into adding libsodium as a third party library, but they do not already use CMake and their build is complex. So we're leaving libsodium as an (optional) external dependency.

caitlinross:
@chuckatkins I believe I made all the changes we discussed and everything is passing. 

@pnorbert The documentation on using plugins is also added to the advanced section you set up.

Issue #3029, opened 2022-01-31
Add optional O_DIRECT flag for POSIX file transport


DominicWindisch:
We're currently using ADIOS2 to dump data streams from 40-GBit network to a local SSD RAID system. In it's current form, ADIOS does not exceed ~1.6 GB/s write speed on our RAID system capable of >6 GB/s sustainable write speed.

**Why is this feature important?**
To unlock the maximum write performance on local storage.
**What is the potential impact of this feature in the community?**
It may gather potential new users which struggle writing large amounts of data locally.
**Is your feature request related to a problem? Please describe.**
As stated, writing files without the O_DIRECT flag will drastically lower maximum performance. In our case, we need at least 3.5 GB/s sustained write speed.
**Describe the solution you'd like and potential required effort**
Adding the O_DIRECT flag when [setting up the file handles in the POSIX transport](https://github.com/ornladios/ADIOS2/blob/a382cc0ed953dcead15d6168156f9e227a10ade2/source/adios2/toolkit/transport/file/FilePOSIX.cpp#L90) should be sufficient as a first step. Potentially necessary memory alignment and padding may be left to the user. This assumes that any kind of internal buffer between the user data and writing data is properly aligned using a aligning allocator (I'm not too familiar with the inner workings of ADIOS2 regarding this).
**Describe alternatives you've considered and potential required effort**
Our current alternative is to dump data with a very primitive data dumper calling the syscall `write` after generating the file descriptor with the appropriate O_DIRECT flag. So we're left with two options: enhance ADIOS to meet our requirements on data throughput OR move away from ADIOS2 and instead implement our own meta-data handling.

**Additional context**
[O_DIRECT man page](https://man7.org/linux/man-pages/man2/open.2.html#:~:text=The%20O_DIRECT%20flag%20on%20its,used%20in%20addition%20to%20O_DIRECT.)

pnorbert:
Thank you for the observation. We never had the O_DIRECT flag as it destroys performance on Lustre/GPFS parallel file systems. They like their buffering mechanisms to be used. So, we cannot just add this flag into the POSIX transport but maybe we can add it as an option or create a specific transport for SSD or SSD RAID where this would work best.

williamfgc:
[...]

PR #3053, opened 2022-02-11
Add Direct IO support for BP5 with control parameters bool DirectIO, …

pnorbert:
…uint DirectIOAlignOffset and uint DirectIOAlignBuffer. This feature is only available on UNIX for FilePOSIX transport; it is ignored everywhere else.

Memory (re)allocations in ChunkV are aligned to the requested buffer alignment size. Chunk sizes are kept to be multiple of buffer block size.
Added a directio test where one chunk is shrunk before filling another chunk.


pnorbert:
This should address #3029

PR #3059, opened 2022-02-17
Re-add dummy dataplane

eisenhauer:


eisenhauer:
@pnorbert @vicentebolea I've added a version of the old "dummy" dataplane, tweaked so that it compiles with current interfaces.  The original dummy actually did something basic.  I haven't tested this one to see that it does.  Mainly it should be useful for the commented interfaces and as a place to start creating new data planes.

PR #3061, opened 2022-02-18
clean up extern template declarations

JasonRuonanWang:
These declarations do not seem to do anything at all.

### DIFF
--- a/source/adios2/core/Engine.h
+++ b/source/adios2/core/Engine.h
@@ -609,59 +609,6 @@ private:
                         const std::string hint) const;
 };
 
-#define declare_template_instantiation(T)                                      \
-                                                                               \
-    extern template void Engine::Put<T>(Variable<T> &, const T *, const Mode); \
-    extern template void Engine::Put<T>(const std::string &, const T *,        \
-                                        const Mode);                           \
-                                                                               \
-    extern template void Engine::Put<T>(Variable<T> &, const T &, const Mode); \
-    extern template void Engine::Put<T>(const std::string &, const T &,        \
-                                        const Mode);                           \
-                                                                               \
-    extern template void Engine::Get<T>(Variable<T> &, T *, const Mode);       \
-    extern template void Engine::Get<T>(const std::string &, T *, const Mode); \
-                                                                               \
-    extern template void Engine::Get<T>(Variable<T> &, T &, const Mode);       \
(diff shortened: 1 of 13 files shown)


JasonRuonanWang:
@eisenhauer suggested that this may have some problems with gcc 10 and later, because they changed some logics in template instantiation since gcc 10. I am going to hold on for a while and try different compilers before deciding whether to merge it.

JasonRuonanWang:
@eisenhauer I tried with gcc 6.5, 7.5, 8.5, 9.4, 10.3 and a few clang versions as well, and this branch compiles all the same as the master branch. 

gcc 11.2 does not compile though. But the master branch breaks in exactly the same way. So I guess it's safe to say this PR won't break anything that is not already broken.

eisenhauer:
Well then.   What kind of failures happen with 11.2?  Was it undefined symbols for templates instances?  Or something else?  It looks like it defaults to C++17 now with changes in template parameter matching, which could cause problems given the extensive use...

JasonRuonanWang:
> Well then. What kind of failures happen with 11.2? Was it undefined symbols for templates instances? Or something else? It looks like it defaults to C++17 now with changes in template parameter matching, which could cause problems given the extensive use...

```
[ 41%] Linking C executable ../../../bin/hello_bpWriter_c
/usr/bin/ld: /home/jason/w/adios2/lib/libadios2_core.so.2: undefined reference to `std::__exception_ptr::exception_ptr::_M_release()@CXXABI_1.3.13'
/usr/bin/ld: ../../../lib/libadios2_c.so.2.7.1: undefined reference to `std::__throw_bad_array_new_length()@GLIBCXX_3.4.29'
/usr/bin/ld: /home/jason/w/adios2/lib/libadios2_core.so.2: undefined reference to `std::__exception_ptr::exception_ptr::_M_addref()@CXXABI_1.3.13'
collect2: error: ld returned 1 exit status
make[2]: *** [examples/hello/bpWriter/CMakeFiles/hello_bpWriter_c.dir/build.make:98: bin/hello_bpWriter_c] Error 1
make[1]: *** [CMakeFiles/Makefile2:4759: examples/hello/bpWriter/CMakeFiles/hello_bpWriter_c.dir/all] Error 2
make: *** [Makefile:146: all] Error 2
```

williamfgc:
gcc 11 might be too new for the system glibc, or it is pointing at an older libstdlibc++. Thanks for cleaning this up.

PR #3062, opened 2022-02-18
BP4 GPU aware for Get calls

anagainaru:
While testing BP5 and BP4 with GPU I realized I forgot to push the changes to make the read side of BP4 CUDA aware. 

I will update the Testing function to include BP4 in the tests so we can officially test GPU buffers, but so far this code passed my offline tests.

### DIFF
--- a/source/adios2/helper/adiosMemory.h
+++ b/source/adios2/helper/adiosMemory.h
@@ -184,7 +184,8 @@ void ClipContiguousMemory(T *dest, const Dims &destStart, const Dims &destCount,
                           const Box<Dims> &intersectionBox,
                           const bool isRowMajor = true,
                           const bool reverseDimensions = false,
-                          const bool endianReverse = false);
+                          const bool endianReverse = false,
+                          const bool isGPU = false);
 
 template <class T>
 void ClipContiguousMemory(T *dest, const Dims &destStart, const Dims &destCount,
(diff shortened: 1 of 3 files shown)


eisenhauer (approved):
Generally, these look good.  My one suggestion:  Should anyone ever trigger that "Endian Reverse" error message, they are going to get a billion of them in output.  I would :

1. make it go to stderr.
2. rewrite it to be interpretable by the user.  I.E. something like : "Byte-order-incompatible data being read into a GPU buffer.  Direct byte order reversal not supported in the GPU, data will be corrupt."
3. use something like a static boolean to at least make sure this gets output at most once per node.   "if (endianReverse) { static bool first = true; if (first) { first = false; std::cerr << ... }}"

anagainaru:
Agree @eisenhauer 

I put the message there then forgot about it, I will update it. Thanks for checking.

anagainaru:
> Generally, these look good. My one suggestion: Should anyone ever trigger that "Endian Reverse" error message, they are going to get a billion of them in output. I would :
> 
> 1. make it go to stderr.
> 2. rewrite it to be interpretable by the user.  I.E. something like : "Byte-order-incompatible data being read into a GPU buffer.  Direct byte order reversal not supported in the GPU, data will be corrupt."
> 3. use something like a static boolean to at least make sure this gets output at most once per node.   "if (endianReverse) { static bool first = true; if (first) { first = false; std::cerr << ... }}"

I decided to throw an invalid argument when the two options are being used together. I think it's cleaner like this.

eisenhauer:
Agree.  I was thinking later that if we're delivering corrupt data we should throw an exception rather than just a warning...

PR #3066, opened 2022-02-22
Fix bpls when calculating get_global_array_signature() to not segfaul…

pnorbert:
…t on steps that don't have the variable.

Previously it failed to list content of file generated by test Test.Engine.BP.WriteReadAsStreamADIOS2.Serial


pnorbert:
Will be part of a new PR

PR #3084, opened 2022-03-02
Force copy DumpDeferredBlocks in PerformPuts() and hence no need for …

pnorbert:
…BufferV::CopyExternalToInternal at all, which was incorrectly implemented in ChunkV.


eisenhauer (commented):
Looks good...

PR #3090, opened 2022-03-04
Bug fixes for GPU-aware Put and Get

anagainaru:
Bug fixes include:
- Add the GPU logic in the new `NdCopy` function that broke the Get code for the BP5 engine (removed all the unused functions)
- Change the default memory space to `Detect` when CUDA is enabled and fixed the bugs caused by this
- Added examples for BP5 and BP4 using GPU buffers (and removed the generic one that was only testing one engine)


source/adios2/toolkit/format/bp5/BP5Deserializer.cpp
```diff
@@ -1248,7 +1251,8 @@ void BP5Deserializer::FinalizeGets(std::vector<ReadRequest> Requests)
 
                     helper::NdCopy(IncomingData, inStart, inCount, true, true,
                                    (char *)Req.Data, outStart, outCount, true,
-                                   true, ElementSize);
+                                   true, ElementSize, Dims(), Dims(), Dims(),
+                                   Dims(), false, Req.MemSpace);
                 }
```
anagainaru, line 1256: @eisenhauer This is the only part I don't like from my fix (having to spell out the default arguments), but I honestly didn't know how to make it better without breaking other parts of the code.

eisenhauer: @anagainaru If you want to avoid this, add MemSpace in the argument list after ElementSize rather than at the end.  (ElementSize is itself a new argument, introduced a few weeks ago when NdCopy got un-templated.  (The only way that templates were employed was in "sizeof(T)" and it was judged not worth having a dozen instances of the same code that differed only in that one constant, so ElementSize was made a parameter and the templating removed.  But yes, the default args meant that the new parameter was best introduced before all the args with default values...)

eisenhauer (approved):
Other than the suggestion about relocating the MemSpace parameter, this looks good...

source/adios2/toolkit/format/bp5/BP5Deserializer.cpp
eisenhauer: @anagainaru If you want to avoid this, add MemSpace in the argument list after ElementSize rather than at the end.  (ElementSize is itself a new argument, introduced a few weeks ago when NdCopy got un-templated.  (The only way that templates were employed was in "sizeof(T)" and it was judged not worth having a dozen instances of the same code that differed only in that one constant, so ElementSize was made a parameter and the templating removed.  But yes, the default args meant that the new parameter was best introduced before all the args with default values...)

anagainaru:
> Other than the suggestion about relocating the MemSpace parameter, this looks good...

Let's discuss next Tuesday about this, not sure if it's worth change it for this release. A great moment for the change is when I am adding other backends and will prob have to redesign the internal logic.

I will merge it into the release as soon as it passes the tests

source/adios2/toolkit/format/bp5/BP5Deserializer.cpp
JasonRuonanWang: Why not move MemSpace up to the first optional parameter? Or simply make it a compulsory parameter? Even the current looking is fine. I don't see a big problem. The ultimate goal is probably to support memory selections under all circumstances anyway, in which case the empty Dims parameters will have to be replaced with something meaningful.

anagainaru: Ok I'll leave it like this for now and will figure out a better solution when we add more backends

PR #3092, opened 2022-03-07
Refactor SSC to allow easy addition of new sub-routines

JasonRuonanWang:


pnorbert:
This should have gone into release_28 branch

chuckatkins:
> This should have gone into release_28 branch

On it

JasonRuonanWang:
> This should have gone into release_28 branch

1. Why? I am not 100% sure what this will mean to the scalability, performance and stability. If there is a problem then I will have no chance to fix it until the release. The release really means to take the stable SSC version. This is only the start of a series of changes on SSC, which is impossible to finish by the release date...

PR #3095, opened 2022-03-08
SST: Add MPI SST dataplane

vicentebolea:
This PR adds a new MPI Data Plane for the SST engine.

- It adds a SST DataPlane named `MPI_DP`
- It modifies SST tests `MPI_Init` to use `MPI_MULTIPLE_THREAD`
- It adds a new MPICH build based on AlmaLinux 8.
- It includes workarounds in the tests for the `MPI_DP` to work in Cray-MPICH implementations.
- It adds a new document detailing how to run this new DataPlane in ECP supercomputers.
- Minor fix in the selection of SST Data plane.


source/adios2/toolkit/sst/cp/cp_writer.c
```diff
@@ -1470,6 +1470,14 @@ static void CP_PeerFailCloseWSReader(WS_ReaderInfo CP_WSR_Stream,
                 CP_WSR_Stream->Connections[i].CMconn = NULL;
             }
         }
+
+        if (CP_WSR_Stream->DP_WSR_Stream)
+        {
+            CP_WSR_Stream->ParentStream->DP_Interface->destroyWriterPerReader(
```
vicentebolea, line 1476: @eisenhauer I noticed that `destroyWriterPerReader` was not called in any case, I added this here, seems to work well for the `mpi_dp` but I wonder if it would break other dp implementations

source/utils/adios_reorganize/main.cpp
```diff
@@ -20,7 +20,8 @@
 int main(int argc, char *argv[])
 {
 #if ADIOS2_USE_MPI
-    MPI_Init(&argc, &argv);
+    int provided;
```
vicentebolea, line 23: @chuckatkins I am making this change in every utility/test main functions, this is needed for the `mpi_dp`. I wonder if I should properly wrap this and add it as a function in any of the general ADIOS2 headers?

CMakeLists.txt
```diff
@@ -96,7 +96,7 @@ endif()
 set(CMAKE_CXX_STANDARD 11)
 set(CMAKE_CXX_STANDARD_REQUIRED True)
 if(NOT MSVC)
-  set(CMAKE_C_STANDARD 99)
```
vicentebolea, line 99: @chuckatkins I wonder if it would be appropriate to move to C11, among other nice changes the thread library is a big win introduced there since it makes our concurrent code compatible across any other platform.

vicentebolea: Nevermind, I have checked that the C11 threading lib requires glibc 2.28 for GCC/Clang which in turn makes ubuntu bionic (18.04) unsupported. Lets come back to this when we drop support to ubuntu 18.04.

Any other C11 feature/lib is implemented since very old glibc/gcc/clang versions. It might just be worth to keep this change.

testing/adios2/engine/ssc/TestSscBase.cpp
```diff
@@ -297,7 +297,8 @@ TEST_F(SscEngineTest, TestSscBase)
 
 int main(int argc, char **argv)
 {
-    MPI_Init(&argc, &argv);
+    int provided;
+    MPI_Init_thread(&argc, &argv, MPI_THREAD_MULTIPLE, &provided);
```
JasonRuonanWang, line 301: I am not sure if we should initialize threads in every SSC test. These tests have nothing to do with SST, and therefore they have nothing to do with the MPI DP of SST. The SSC engine is designed to ensure there is a baseline staging method that we can use everywhere that has MPI and MPMD. Some of these environments do not provide multi-thread MPI. We should aim to ensure these tests work without multi-thread MPI.

vicentebolea: @JasonRuonanWang good observation, I changed all the tests `MPI_Init` in bulk with the idea of later providing a general solution, perhaps on the form of an wrapper named `ADIOS2_MPI_Init`.

In this specific case if the system does not have THREAD_MULTIPLE` MPI it will not fail it will simply return the enabled mode in the provided argument and keep the execution.

JasonRuonanWang: @vicentebolea Yeah, I understand that if MPI does not allow multi-thread it will still successfully initialize, only returning a provided=0 or something. But still, if 'unfortunately', all of our CI virtual machines support multi-thread MPI, we will lose the chance to discover any issues that break the non-multi-thread routine.

vicentebolea: >  if 'unfortunately', all of our CI virtual machines support multi-thread MPI, we will lose the chance to discover any issues that break the non-multi-thread routine.

That is a good point.

vicentebolea:
@chuckatkins we have a new MPICH build and its green :) https://github.com/ornladios/ADIOS2/runs/6823739729?check_suite_focus=true

vicentebolea:
@suchyta1

pnorbert (approved):
Please add a short comment for MPI_Init_thread() calls in the examples that it is only needed for SST/MPI-data plane

PR #3097, opened 2022-03-10
Refactor SSC to allow easy addition of new sub-routines

chuckatkins:
Pulling #3092 into the release branch


JasonRuonanWang:
This is not for the release really. I am not 100% sure what this will mean to the scalability, performance and stability. If there is a problem then I will have no chance to fix it until the release. The release really means to take the stable SSC version. This is only the start of a series of changes on SSC, which is impossible to finish by the release date...

pnorbert:
Sorry, misunderstood. Don't merge this into the release.

PR #3099, opened 2022-03-10
BP5 index format is now record-based: 

pnorbert:
1+8 byte for recordID + record length, then content. recordID 's' for Steps, 'w' for writer map
Writer map always PRECEEDS the step record where it was recorded.


eisenhauer (commented):
Looks good...

PR #3118, opened 2022-03-21
DILL upstream

eisenhauer:
Tweak for compiling for universal binary on osx.


eisenhauer:
@chuckatkins Ugh.  Sorry, this could have and probably should have gone into the 2.8 release.  Just wasn't on the ball.  Can you pull it in?

PR #3119, opened 2022-03-22

eisenhauer:


JasonRuonanWang:
Can we make the tests more verbose? The logs currently don't show which line of code caused the segfaults...

eisenhauer:
I guess the short answer is that I don't know.  It doesn't look like any sort of stack dump is in the raw log, which is presumably why it isn't in any per-test summary.  I don't know if there are settings to CTest (or CDash) that would change this.   Generally in this circumstance I try to download the docker image associated with the failing test environment and see if I can reproduce the failure myself, which is about the only thing I have to suggest (unless @chuckatkins wants to weigh in with other options).

JasonRuonanWang:
@chuckatkins Previously when a test failed we could see the stack dump information, but now it's gone. Any chance to bring it back? Some tests only fail on CI, and it's impossible to reproduce anywhere else. So we really need CI to be more verbose to understand what's going on.

PR #3127, opened 2022-03-23
Fix memory leak in SZ operator

JasonRuonanWang:


chuckatkins:
Please remove the dockerfile changes, those are no longer used since the CI rework and wouldn't take effect anyways without re-spinning the CI images.

chuckatkins:
> Then why are those files still there? If they are still there then it means they are supposed to be somewhat useful.

The emulation images haven't been ported but everything else has.  I'm keeping the legacy CI files in place until then, at which point I'll remove them entirely.


> If they are ever going to be useful, then keeping them up to date sounds like a better idea than keeping them outdated.

Changing the docker files is only meaningful once images they create are rebuilt and deployed.  Keeping the CI images up to date in general is a complicated task that we don't have the automated infrastructure fore so at the moment they're manually rebuilt as needed.


> There are known issues in the outdated SZ versions which affect the SZ tests, so in any case, keeping them outdated is never going to be what we want.

If you check the configure output you can see they're up to date with v2.1.12.  If you need the images bumped to 2.1.12.1 let me know and I can rebuild them, but it takes nearly a day to rebuild all of the images with all of the different compilers and configurations we're using.

```
-- Found SZ: /opt/spack/opt/spack/linux-almalinux8-haswell/gcc-8.5.0/sz-2.1.12-iduyjprfejvprobyxg25pw3saiwu3m5u/lib64/libSZ.so  
```

JasonRuonanWang:
@chuckatkins Okay I will close this for now and let you know when we need to update the SZ version. It's probably going to be a future 2.1.12.2.

PR #3133, opened 2022-03-25
Bug BP5 Get for chunks in an ND array

anagainaru:
The `Get` function in `BP5` was only capable of reading continuous buffers and was failing tests when reading chunks within the array. 

With the current fix the only pattern supported is 
```
if( inIsRowMajor && outIsRowMajor && (inIsLittleEndian == outIsLittleEndian))
```
but can be easily extended if we think there will be other read patterns with GPU buffer.

I added a test for different read patterns for 2D arrays for GPU buffers for when the CI will include them.


anagainaru:
@chuckatkins For some reason I cannot create the pull request for `release_28` and it's too late to figure it out. I will take care of this in the morning, but I wanted to push it so Greg has time to review it before I wake up.

chuckatkins:
rebased and re-targeted onto release_28

eisenhauer (approved):
Great!

anagainaru:
Not sure why the readthedocs is not passing, should I trigger the testing again?

eisenhauer:
It looks like Chuck had several 2am commits to PR #3134 trying to fix readthedocs and that he hasn't succeeded yet.  I'd guess that when he gets it fixed, he'll ask for other pending PRs to be rebased and pushed again.  So you can probably just sit tight for a while...

chuckatkins:
Is this good to go?

PR #3145, opened 2022-04-01
Syntax fix in declarations for C++20 compliance

garth-wells:
Building a C++20 library against ADIOS2 fails with GCC 11.2.0. This PR allows compilation without fatal errors. An explanation of what has been tightened in the C++ standard is at https://stackoverflow.com/questions/63513984/can-class-template-constructors-have-a-redundant-template-parameter-list-in-c2.

See also https://gcc.gnu.org/bugzilla/show_bug.cgi?id=97202.

### DIFF
--- a/bindings/CXX11/adios2/cxx11/Attribute.h
+++ b/bindings/CXX11/adios2/cxx11/Attribute.h
@@ -43,8 +43,8 @@ public:
      * attributes from IO:DefineAttribute<T> or IO:InquireAttribute<T>.
      * Can be used with STL containers.
      */
-    Attribute<T>() = default;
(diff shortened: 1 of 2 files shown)


chuckatkins:
Looks good to me, thanks for the fix!  I'm starting to put togethor a 2.8.1 patch release to address the myriad of issues that have come up as people have tried to deploy 2.8.0.  I'm just going to rebase and re-target this to the release_28 branch to get it pulled into the patch release.

chuckatkins:
Fixes #3146

chuckatkins (approved):

PR #3167, opened 2022-04-17
Fix helloBPTimeWriter Python example.

manauref:
The helloBPTimeWriter python example was not working on my system. Made a few small changes and got it working, and the output file contains the expected numbers.

### DIFF
--- a/examples/hello/bpTimeWriter/helloBPTimeWriter.py
+++ b/examples/hello/bpTimeWriter/helloBPTimeWriter.py
@@ -19,28 +19,35 @@ size = comm.Get_size()
 # User data
 myArray = np.array([0, 1, 2, 3, 4, 5, 6, 7, 8, 9])
 nx = myArray.size
+time = np.array([0.0])
 
 # ADIOS
 adios = adios2.ADIOS(comm, adios2.DebugON)
 
 # IO
 bpIO = adios.DeclareIO("BPN2N")
+bpIO.SetEngine('bp3')
+
+fileID = bpIO.AddTransport('File', {'Library': 'fstream'})
 
 # Variables
-bpArray = bpIO.DefineVariable("bpArray", [size * nx], [rank * nx], [nx],
-                              adios2.ConstantDims, myArray)
-bpTimeStep = bpIO.DefineVariable("bpTimeStep")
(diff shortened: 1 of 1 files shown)


examples/hello/bpTimeWriter/helloBPTimeWriter.py
```diff
@@ -19,27 +19,34 @@
...
 myArray = np.array([0, 1, 2, 3, 4, 5, 6, 7, 8, 9])
 nx = myArray.size
+time = np.array([0.0])
 
 # ADIOS
 adios = adios2.ADIOS(comm, adios2.DebugON)
 
 # IO
 bpIO = adios.DeclareIO("BPN2N")
+bpIO.SetEngine('bp3')
```
JasonRuonanWang, line 29: Does this also work with the default BP4 engine?

examples/hello/bpTimeWriter/helloBPTimeWriter.py
```diff
@@ -19,27 +19,34 @@
...
+
+fileID = bpIO.AddTransport('File', {'Library': 'fstream'})
```
JasonRuonanWang, line 31: Also, is this a must? I guess this example is written to showcase a simple generic workflow. It might have digressed a little by specifying an outdated engine, as well as this particular transport.

examples/hello/bpTimeWriter/helloBPTimeWriter.py
manauref: If you change 'bp3' to 'bp4', yes.

examples/hello/bpTimeWriter/helloBPTimeWriter.py
manauref: Fair.
It also works if you delete these two lines. Would you prefer I do that?

JasonRuonanWang: Yeah, that would be great. Thanks!

manauref: done

JasonRuonanWang (approved):

PR #3171, opened 2022-04-21
Add template-free C++ API 

JasonRuonanWang:
Since long time, users have been struggling with the templated C++ API of ADIOS2. It makes things unnecessarily hard for users to define and use ADIOS2 variables with data types that are only determined at runtime. It makes things even harder for users who want to put variable objects of different data types into a single container. 

It's time to correct this harmful C++ fundamentalism. A C++ code does not have to use templates. Whether to use templates should be determined in a case by case manner. 

This PR adds a template-free C++ API, which should be recommended for most use cases instead of the templated API. Some parts of it is currently still built on top of the templated API, but we should aim to remove most of the templated implementations in ADIOS2 core so we will save a huge lot of type conversions and macros that cut off the debugger processes. 

This is also inline with the user defined data type work.

### DIFF
--- a/bindings/CXX11/CMakeLists.txt
+++ b/bindings/CXX11/CMakeLists.txt
@@ -15,6 +15,7 @@ add_library(adios2_cxx11
   adios2/cxx11/Types.cpp
   adios2/cxx11/Types.tcc
   adios2/cxx11/Variable.cpp
+  adios2/cxx11/VariableNT.cpp
   adios2/cxx11/fstream/ADIOS2fstream.cpp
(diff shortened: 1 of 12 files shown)


williamfgc:
@pnorbert FYI, this API will silently write corrupt data due to unsafe `reinterpret_cast` to `void*` which  could be any piece of memory. It needs to be tested and recommended to advanced users only. It's essentially replacing a compile-time safe template for an unsafe cast due to a corner case. While still good to have, adios2 shouldn't be responsible for potential damage, though.

PR #3178, opened 2022-04-26

eisenhauer:


JasonRuonanWang:
@eisenhauer The on-demand test also fails a lot in a normal PR in the same way. Do you want to disable it for a while before it's getting stable.

eisenhauer:
I don't think so, at least for now.  OnDemand is a little awkward.  I want to test, for example, that if two separate readers request steps, they get steps in the order requested.  Right now I'm doing that by delaying the readers by different times so that the requests happen in a particular order.  Unfortunately that's not fool-proof on highly-loaded CI systems where the scheduler might simply not run things in the same order they would on unloaded systems.  We could go to a lot of trouble to introduce a separate mechanism that would force the reader cohorts make requests in a specific order, despite possible scheduling changes, but I'm not sure it'd be worth the time expended.  I see the "no test reruns" thing as a way to tell us if we've introduced new instabilities in library itself.  Every failure I see on this looks like it's just a reader client not being getting scheduled when it should have been.  I.E. not a bug in ADIOS, but an instability in the testing environment, and that's less worrisome.   (If we were trying to get the no-test-rerun failures down to zero, I'd hop on it.  But we have a lot of tests like this, some of which are likely real issues in ADIOS, that we haven't gotten to yet.  One of these days...)

JasonRuonanWang:
@eisenhauer You can still have it in your develop branch. I was just suggesting that we should temporarily disable it in the master branch. It fails almost every time when a new PR is created. People will have to re-run a few times to get it passed. It fails on my local machine as well, which is completely unloaded. These re-runs won't help anything with your debugging, but create a lot of overhead to anyone that opens a PR.

eisenhauer:
I hadn’t seen a failure in PRs, just in the no-test-reruns branch.  I’ll take a look.  If it’s failing that often enough that it will regularly fail the five times in a row it needs to fail to cause a problem in a regular PR, it needs to be changed.  (And I’m not really debugging anything with that.  The OnDemand stuff is in and working as far as I can tell.  The only issue might be to the things that test it, so that they're robust under weird CI conditions.)

eisenhauer:
It doesn't look like the OnDemand test impacted in your current PR.  Has it failed with others with normal test rerunning?

JasonRuonanWang:
@eisenhauer The latest PR is the only one I didn't re-run it. Before this one I had three PRs (one of them was closed though), which I all re-ran one or two times before passing. 

I tried to run it on two different Ubuntu 20.04 machines, which didn't have any work load, and it fails almost every time. It's very rare to see it pass.

eisenhauer:
Odd.  I don't see it fail on my unloaded machine.  I'll see if I can tweak it.

JasonRuonanWang:
> Odd. I don't see it fail on my unloaded machine. I'll see if I can tweak it.

Did you try on a Linux machine? It fails more on a Linux machine than Mac.

PR #3198, opened 2022-05-03
added BlocksInfo API for VariableNT, and implementation for VariableStruct in SSC

JasonRuonanWang:
This is still based on the templated API, for apps that need to use in near future. But in the long run, we should aim to implement a non-template BlocksInfo from the API layer down to adios2::core. The current API layer is extremely inefficient in the sense that it needs to traverse every single BlockInfo element for a deep copy from adios2::core to the API objects. This causes ridiculous performance loss for apps that have a large number of blocks and need to call BlocksInfo frequently. It REALLY should be implemented so that we can use a single memcpy() to copy an entire block. Even if it means that we always need to reserve memory space for the largest possible data type, it will still be much more efficient in block intensive workflows.

### DIFF
--- a/bindings/CXX11/adios2/cxx11/Engine.cpp
+++ b/bindings/CXX11/adios2/cxx11/Engine.cpp
@@ -223,6 +223,81 @@ ADIOS2_FOREACH_STDTYPE_1ARG(declare_type)
 ADIOS2_FOREACH_PRIMITIVE_TYPE_1ARG(declare_template_instantiation)
 #undef declare_template_instantiation
 
+std::vector<VariableNT::Info> Engine::BlocksInfo(const VariableNT &variable,
+                                                 const size_t step) const
+{
+    std::vector<VariableNT::Info> ret;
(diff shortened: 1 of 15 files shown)


eisenhauer:
The "BlocksInfoMin" was meant to be a non-templated blocks info to replace the incredibly heavy BP4 blocks info stuff.  It may need to be extended for other purposes (and we've got this odd scene were we have one blocksinfo in bindings and another in core), but we should probably try to unify efforts in this direction.

JasonRuonanWang:
> The "BlocksInfoMin" was meant to be a non-templated blocks info to replace the incredibly heavy BP4 blocks info stuff. It may need to be extended for other purposes (and we've got this odd scene were we have one blocksinfo in bindings and another in core), but we should probably try to unify efforts in this direction.

Sure. The work I am doing is just some quick implementations for apps that urgently need struct data types. It's non-template and it's only a subset of the full BlocksInfo too. Once the BP5 BlocksInfo is fully ready, it should be easy to graft on it.

PR #3207, opened 2022-05-09
bpls -l for a single scalar (not over time) for a BP5 file showed 0. …

pnorbert:
…Fixed to use variable->MinMax() but also fixed in BP5Deserializer to not segfault if the scalar is not present in every step of the file.


pnorbert:
How to test:
- add BP5 engine to examples/basics/values/values_write.cpp
- `$ ./bin/values_write`
- `bpls -l values.bp Nproc`
- should be 
```
int32_t   Nproc      scalar = 1
```

PR #3217, opened 2022-05-16
fix api for gcc11

robertu94:
Instead of

```cpp
template <class T>
struct foo {
     foo<T>()=default;
}
```

do this

```cpp
template <class T>
struct foo {
     foo()=default;
}
```

`foo` doesn't need to be explicitly specialized with type T, and with GCC 11, the former doesn't even compile.


JasonRuonanWang:
Please make a new branch other than master in order to pass CI. Thanks.

robertu94:
@JasonRuonanWang Does the branch need a specific name or just not “master”

JasonRuonanWang:
> @JasonRuonanWang Does the branch need a specific name or just not “master”

Anything except master should work

robertu94:
Closing in favor of #3250

PR #3219, opened 2022-05-17
Added multithreaded reader in BP5Reader::PerformGets(). Engine parame…

pnorbert:
…ter 'Threads' added to control this.


pnorbert:
NVM, this is a bad implementation as multiple threads end up using the same posix file handle to read data at the same time. A proper implementation needs to have separate io handles in each thread.

PR #3234, opened 2022-05-27
Test writing data with multiple PerformDataWrite and read back. This …

pnorbert:
…breaks and points out the issue with BP5Reader::ReadData() in 2.8.0 release.


pnorbert:
fixes #3234

pnorbert:
@franzpoeschel Please note that this fix is needed to be able to read back the data from the BP5 file if you use PerformDataWrite.

PR #3242, opened 2022-06-06
utils: Fix bpls linking in Windows

Biswa96:
This fixes the error:
```
ld.exe: source/utils/CMakeFiles/bpls.dir/bpls/bpls.cpp.obj:bpls.cpp:(.text+0x67994): undefined reference to '__imp_PathMatchSpecA'
```

### DIFF
--- a/source/utils/CMakeLists.txt
+++ b/source/utils/CMakeLists.txt
@@ -11,7 +11,9 @@ configure_file(
 
 # BPLS
 add_executable(bpls ./bpls/bpls.cpp)
-target_link_libraries(bpls adios2_core adios2sys adios2::thirdparty::pugixml)
+target_link_libraries(bpls
+                      PUBLIC adios2_core adios2sys
+                      PRIVATE adios2::thirdparty::pugixml $<$<PLATFORM_ID:Windows>:shlwapi>)
 target_include_directories(bpls PRIVATE ${PROJECT_BINARY_DIR})
 set_property(TARGET bpls PROPERTY OUTPUT_NAME bpls${ADIOS2_EXECUTABLE_SUFFIX})
 install(TARGETS bpls EXPORT adios2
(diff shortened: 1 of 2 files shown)


source/utils/CMakeLists.txt
```diff
@@ -12,6 +12,9 @@ configure_file(
 # BPLS
 add_executable(bpls ./bpls/bpls.cpp)
 target_link_libraries(bpls adios2_core adios2sys adios2::thirdparty::pugixml)
+if(WIN32)
+  target_link_libraries(bpls shlwapi)
+endif()
```
vicentebolea, line 17:
```suggestion
target_link_libraries(bpls PRIVATE $<$<PLATFORM_ID:Windows>:shlwapi>)
```

vicentebolea (changes requested):

source/utils/CMakeLists.txt
Biswa96: Should I mention "PRIVATE" keyword? It can not be used due to this https://stackoverflow.com/a/59522268

vicentebolea: Good catch, perhaps adding `PUBLIC` for in the first call would be a workaround. @Biswa96 normally we want to make dependencies private unless otherwise.

> Also what is the problem with the current change?

There is nothing wrong, this is a code suggestion within a code review.

vicentebolea, line 17: Actually this might be even better. I do not think that pugixml needs to be a  public dep.

```suggestion
target_link_libraries(bpls
                      PUBLIC adios2_core adios2sys 
                      PRIVATE adios2::thirdparty::pugixml $<$<PLATFORM_ID:Windows>:shlwapi>)
```

vicentebolea (changes requested):

source/utils/CMakeLists.txt
Biswa96: Done.

vicentebolea (approved):

vicentebolea:
Thanks @Biswa96 for your contribution :+1:

PR #3248, opened 2022-06-10
Added default_init_allocator from http://stackoverflow.com/a/21028912…

pnorbert:
…/273767 and adiosvec<T> type in helper::adiosType.h to be able to resize a vector without initialization. Changed BufferSTL and BP3/BP4/SST using adiosvec as well as in BP5 where vector<char> was used. It is much faster than vector<char> with Release build, but it is much slower with Debug build.

Release build:
  $ ./bin/PerfVector 1024
  Size = 1073741824 std::vector = 0.145024 helper::adiosvec = 5.73e-06 Malloc = 2e-08
Debug Build:
  $ ./bin/PerfVector 1024
  Size = 1073741824 std::vector = 0.145616 helper::adiosvec = 5.48551 Malloc = 9.16e-06

### DIFF
--- a/source/adios2/engine/bp3/BP3Writer.cpp
+++ b/source/adios2/engine/bp3/BP3Writer.cpp
@@ -302,7 +302,7 @@ void BP3Writer::WriteProfilingJSONFile()
                                    transportTypes, transportProfilers) +
                                ",\n");
 
-    const std::vector<char> profilingJSON(
+    const helper::adiosvec<char> profilingJSON(
         m_BP3Serializer.AggregateProfilingJSON(lineJSON));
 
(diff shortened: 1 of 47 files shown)


pnorbert:
This whole thing is up for discussion. I don't remember right now where vector allocation was a measurable issue.

pnorbert:
It is so slow in Debug mode, that these test take a minute to run.
Engine.BP.BPBufferSizeTest.SyncDeferredIdenticalUsage.BP3.Serial
Engine.BP.BPBufferSizeTest.SyncDeferredIdenticalUsage.BP4.Serial

PR #3250, opened 2022-06-13
fix api for gcc11

robertu94:
Replaces #3217


chuckatkins:
I thought we already merged this but I think a lot of things got munged up in the 2.8.1 release cycle so it may not have propagated correctly.  Regardless, this looks good to me.

chuckatkins (approved):

PR #3266, opened 2022-06-29
Changed parameters for DAOS API functions.

dmitry-ganyushin:
DAOS 2.0 requires different parameters for DAOS functions. Code can be build on Presque now.

### DIFF
--- a/source/adios2/toolkit/transport/file/FileDaos.cpp
+++ b/source/adios2/toolkit/transport/file/FileDaos.cpp
@@ -31,7 +31,6 @@
 
 #define CheckDAOSReturnCode(r) CheckDAOSReturnCodeF((r), __FILE__, __LINE__)
 #define DefaultMaxDFSBatchSize 8589934592
-#define DAOS_VERSION 2
 namespace adios2
 {
 namespace transport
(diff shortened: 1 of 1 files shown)


dmitry-ganyushin:
I don't know for sure for which version of DAOS it worked before. Now I can
build ADIOS with libdfs 2.0 on Presque.  The doc states that 1.x  versions
of DAOS are obsolete.

eisenhauer:
I think we can safely assume that it compiled with versions of DAOS before this change, which presumably happened about six months ago when 2.0.0 was released, and likely it won't compile on the old version.  I hate to introduce discontinuity where it can be avoided.  (Not that DAOS is in serious use, but it just seems like a bad practice to break something so recently valid and which may not have been upgraded somewhere yet.)  Isn't there a #define for DAOS version or anything you can test?

dmitry-ganyushin:
Until now, I cannot find out which DAOS version was installed on Presque before upgrade. The original ADIOS2 version cannot be compiled with DAOS version 1.0.1 as well. I will modify the code in order to differentiate DAOS versions.

eisenhauer:
There are version numbers defined in daos_version.h.  Wouldn't it make more sense to use the defined DAOS_API_VERSION_MAJOR or DAOS_VERSION_MAJOR instead of #defining your own?

eisenhauer:
👍

PR #3271, opened 2022-07-01
Kokkos

germasch:
This is just a proof-of-concept for the discussion in #3254 .

### DIFF
--- a/bindings/CXX11/CMakeLists.txt
+++ b/bindings/CXX11/CMakeLists.txt
@@ -89,6 +89,7 @@ install(
         adios2/cxx11/Operator.h
         adios2/cxx11/Query.h
         adios2/cxx11/Types.h
+        adios2/cxx11/kokkos_interop.h
   DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}/adios2/cxx11
   COMPONENT adios2_cxx11-development
 )
--- a/bindings/CXX11/adios2/cxx11/Engine.h
+++ b/bindings/CXX11/adios2/cxx11/Engine.h
@@ -28,14 +28,14 @@ template <class...>
 using void_t = void;
 
 template <typename C, typename = void>
-struct is_container : std::false_type
+struct is_ndarray : std::false_type
 {
 };
 
 template <typename C>
-struct is_container<C, void_t<typename C::value_type, typename C::size_type,
-                              decltype(std::declval<C>().data()),
-                              decltype(std::declval<C>().size())>>
+struct is_ndarray<C, void_t<typename C::value_type, typename C::size_type,
+                            decltype(std::declval<C>().data()),
+                            decltype(std::declval<C>().size())>>
 : public std::true_type
 {
 };
@@ -173,13 +173,13 @@ public:
     template <
         class T, class C,
         typename = typename std::enable_if<
-            detail::is_container<C>::value &&
+            detail::is_ndarray<C>::value &&
             std::is_same<typename std::remove_cv<typename C::value_type>::type,
                          T>::value>::type>
-    void Put(Variable<T> variable, const C &container,
+    void Put(Variable<T> variable, const C &ndarray,
              const Mode launch = Mode::Deferred)
     {
-        Put(variable, const_cast<const T *>(container.data()), launch);
+        Put(variable, const_cast<const T *>(ndarray.data()), launch);
     }
 
     /**
--- a/bindings/CXX11/adios2/cxx11/Engine.h
+++ b/bindings/CXX11/adios2/cxx11/Engine.h
@@ -27,7 +27,7 @@ namespace detail
 template <class...>
 using void_t = void;
 
-}
+} // end namespace detail
 
 template <typename C, typename = void>
 struct ndarray_traits : std::false_type
@@ -38,7 +38,11 @@ template <typename C>
 struct ndarray_traits<
     C, detail::void_t<typename C::value_type, typename C::pointer,
                       typename C::size_type, decltype(std::declval<C>().data()),
-                      decltype(std::declval<C>().size())>> : std::true_type
+                      decltype(std::declval<C>().size()),
+                      // don't match std::vector since it's handled separately
+                      typename std::enable_if<!std::is_same<
+                          C, std::vector<typename C::value_type>>::type>>>
+: std::true_type
 {
     using value_type = typename C::value_type;
     using pointer = typename C::pointer;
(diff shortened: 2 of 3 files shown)


bindings/CXX11/adios2/cxx11/Engine.h
```diff
@@ -143,6 +169,21 @@ class Engine
...
              const Mode launch = Mode::Deferred);
 
+    /**
+     * Put data associated with a container-like thing, e.g., a Kokkos::View
+     */
+    template <
+        class T, class C,
+        typename = typename std::enable_if<
+            detail::is_container<C>::value &&
+            std::is_same<typename std::remove_cv<typename C::value_type>::type,
```
vicentebolea, line 179: std::is_same<typename std::remove_cv<typename C::value_type>::type,
            
            how about moving this to the is_container trait?

bindings/CXX11/adios2/cxx11/Engine.h
```diff
@@ -21,6 +21,32 @@
...
 
+namespace detail
+{
+template <typename... Ts>
+struct void_t
+{
+};
+
+template <typename C, typename = void>
+struct is_container : std::false_type
```
vicentebolea, line 32: I propose having a verbose static_assert to inform the user that the type passed does not fulfill the traits "Contract

anagainaru:
Overall I really like the implementation. It's so much cleaner than mine.
I have only two comments. 

- In the future we plan to use Kokkos View internally in ADIOS for GPU buffers to compute metadata and copy the user buffer to our internal CPU buffer. Just taking the `.data()` won't be the best solution. Also in some cases, the Kokkos View could have stride access (the View could have a `Kokkos::LayoutStride`).  

- Eventually I might need to take the memory space out of the container. Making the user set both the Kokkos view and the ADIOS memory space is not the most elegant solution.

bindings/CXX11/adios2/cxx11/Engine.h
germasch: Actually, this makes sure that the `value_type` of the container is consistent with the adios2 `Variable`, so I can't move it into `is_container`, since it's not a property of the container.

I did change this into a `static_assert`, though, so instead of a missing overload error one gets a somewhat more specific error message when trying to, e.g., write a `Kokkos::View<float*>` into a `Variable<double>`.

bindings/CXX11/adios2/cxx11/Engine.h
germasch: You can't really have a `static_assert()` in there, since this is used for SFINAE, ie., you don't want an error, you just want the non-container argument to not be considered in this case and have it try the other possible overloads.

germasch:
I update this PR to be more complete (unfortunately, as anticipated, also more complex). As a minor change, I changed `is_container` to `is_ndarray` since the former definitely has some pre-existing connotations I don't want people to confuse it with.

As I had originally said, in order to handle things like memory space, I introduced `ndarray_traits`, and that can be specialized by libraries that want to provide interoperability for their data structure with adios2, without having to adapt anything in their actual data structure. In some sense, this might eventually be something that could actually be maintained on the respective library side rather than by adios2, but since I think for now there isn't an expectation for that traits interface to be stable yet, so it's easier to add the kokkos interoperability to adios2 and have them evolve together there. 

So I added a `kokkos_interop.h` header for that purpose. It does get installed in the `cxx11` subdir, which makes for a bit of an ugly path, an application wanting to use kokkos interoperability has to `#include <adios2/cxx11/kokkos_interop.h>` as of now.

By default, the traits will set the memory space to `Detect`, and the `Put()/Get()` implementations in this case will not touch the memory space settings. However, for kokkos, the memory space will be determined from the `View` type and will be set on the adios2 side accordingly.

Layout could be handled similarly, and `Kokos::LayoutLeft/Right` could be used to automatically set RowMajor vs ColumnMajor on the adios2 side. Handling some arbitrary strided layout (`LayoutStride`) would require some substantial work on the adios2 side first, I think.

This is all lightly tested only, and I didn't deal with the various overloads for `VariableNT` or passing variable by name. This could be done, but I think there's a point in encouraging people to use the C++-y interface when using more advanced C++ libs like kokkos, anyway.

Updated example on how to use it is still here: https://github.com/germasch/kokkos-adios2

bindings/CXX11/CMakeLists.txt
```diff
@@ -66,13 +66,13 @@ install(TARGETS adios2_cxx11 ${maybe_adios2_cxx11_mpi} EXPORT adios2CXX11Exports
...
 
 install(
-  FILES adios2.h 
+  FILES adios2.h
   DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}
   COMPONENT adios2_cxx11-development
 )
 
 install(
-  FILES adios2/cxx11/fstream/ADIOS2fstream.h
```
vicentebolea, line 75: rogue trailing space

bindings/CXX11/adios2/cxx11/Engine.h
germasch: FWIW,  I did change this back later to the same behavior as the current `Put(Variable<T>, const T*, ...)`, ie., only matching types are considered an overload (kinda like `Put(Variable<T>, const Kokkos::View<T*>, ..)`, so if the View's value_type doesn't match the variable it's just not considered in overload resolution)

bindings/CXX11/CMakeLists.txt
germasch: I did rebase and I just got rid of the formatting changes -- that was just my editor removing trailing spaces, but I agree that it shouldn't be mixed in with actual changes.

germasch:
FWIW, I did remove the "Draft" status, but by no means do I think this is perfect. One thing in particular to work out, if this is going to be the approach chosen, is where to put the `kokkos_interop.h` header. Having it below `cxx11` is kinda fundamentally the right place, but having to include it with that explicit path component exposes an implementation detail that's otherwise hidden. An alternative might to have something like `#include <adios2/interop/kokkos.h>`, though there are many options.

Beyond that I mostly see future work that will make things more feature-complete (e.g., supporting Layout, more memory spaces, or adding VariableNT support), which hopefully won't break things.

anagainaru (commented):
Looks good

bindings/CXX11/adios2/cxx11/kokkos_interop.h
```diff
@@ -0,0 +1,62 @@
...
+
+template <typename T>
+struct is_kokkos_view : std::false_type
+{
+};
+
+template <typename... Ts>
+struct is_kokkos_view<Kokkos::View<Ts...>> : std::true_type
+{
+};
```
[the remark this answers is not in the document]
germasch: Thanks, using the existing one is of course better. I've updated the PR accordingly.

anagainaru:
Closed by #3320

PR #3272, opened 2022-07-01
Destroy only created vars on BeginStep and again on Reader close

eisenhauer:
Extending prior PR to BP3/4...

Note:  The test here checks to see if a variable created by reader BeginStep gets deleted in Reader::Close().  I'd like to test to make sure that non-created variables remain available, but that isn't really possible because for some engines, InqVar simply doesn't work under any circumstances for such variables.  This is because of the interwoven nature of the logic and the need to return NULL for an otherwise valid variable if the IO is in "streaming read mode" and there is no valid data for that timestep.  So for now, simply not testing that part as the tangle is too broken to fix.

However, this PR introduces Engine-class methods to RegisterCreatedVariable() and RemoveCreatedVars() so that reader engines can more easily clean up after themselves.  I've modified BP3/4, and SST readers to take advantage of these.  I'll leave the other engines up to their maintainers...

### DIFF
--- a/source/adios2/core/Engine.cpp
+++ b/source/adios2/core/Engine.cpp
@@ -195,6 +195,20 @@ void Engine::DoPutStructDeferred(VariableStruct &, const void *)
         return nullptr;                                                        \
     }
 
+void Engine::RegisterCreatedVariable(const VariableBase *var)
+{
+    m_CreatedVars.insert(var);
+}
+
+void Engine::RemoveCreatedVars()
+{
+    for (auto &VarRec : m_CreatedVars)
+    {
+        m_IO.RemoveVariable(VarRec->m_Name);
+    }
+    m_CreatedVars.clear();
+}
+
 void Engine::DoGetAbsoluteSteps(const VariableBase &variable,
                                 std::vector<size_t> &keys) const
 {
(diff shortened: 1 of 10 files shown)


eisenhauer:
Note the second commit, which fixes a bad test.  This test had code of the form below:
```
    std::string filename = "multi_put2file.bp";
    this->m_Io.SetEngine("BP3");
    auto writer = this->m_Io.Open(filename, adios2::Mode::Write);
    auto reader = this->m_Io.Open("multi_2f_input.bp", adios2::Mode::Read);
    auto var = this->m_Io.template InquireVariable<T>("var");

    MyData<T> myData(this->m_Selections);

    for (int b = 0; b < static_cast<int>(myData.NBlocks()); ++b)
    {
        var.SetSelection(myData.Selection(b));
        reader.Get(var, &myData[b][0], adios2::Mode::Sync);
        writer.Put(var, &myData[b][0], TypeParam::PutMode);
    }
    reader.Close();
    writer.Close();

```
Not that if a variable created by a reader engine has a lifetime that ends at reader.Close(), *and* none of the Put calls are performed until writer.Close(), then this code is potentially invalid because the var in the Put has become invalid before the Put might be actually performed.   This PR addresses that problem by moving the writer.Close() before the reader.Close().

PR #3275, opened 2022-07-06
Add backward compatible read with BLOSC compression to BP4 files crea…

pnorbert:
…ted pre-2.8.

Bzip2 compressed file and testing is added to show how to test for failure.


pnorbert:
The `source/adios2/toolkit/format/bp/bpBackCompatOperation/compress/BPBackCompatBlosc.cpp` file contains the decompression part from the old blosc operator, so it's not using the new compressor code. 

The file header has ADIOS version, and I added the major, minor and patch bytes to the BP4's minifooter. I just check the minor value in PreDataRead() and PostDataRead() and do the hacking for older files instead of calling the new operator infrastructure.

pnorbert:
@chuckatkins Why do we need 6 reviews for this PR?

eisenhauer:
@pnorbert I suspect the review requirement is simply to discourage casual merges to the release branches.  This is targeting the 2.8 branch, so it was subject to that review.  I've dropped it to 1, for now, but we should put it back.  (I'll also do the review.)

pnorbert:
Yeah, I was making that assumption (major = 2).

eisenhauer (approved):
👍

pnorbert:
The commit of this PR will be cherry-picked into master in a separate PR to master.

PR #3284, opened 2022-07-13
Change old embedded BP3 spec to BP4

eisenhauer:


eisenhauer:
@pnorbert When working on other things I noticed that we have some tests that work with BP3, but not with BP4.  Is the problem with the test?  Or with the engine?

pnorbert:
hmm, really don't remember, I guess it's just an ancient test never updated to BP4, and now it fails with BP4.

PR #3291, opened 2022-07-20
Fix bpls 28

pnorbert:
merge PRs #3289 and #3288 into the 2.8 release patch to fix bpls issues with local arrays


eisenhauer (approved):
👍

PR #3292, opened 2022-07-22
MPI_DP: Optimize scalable component

vicentebolea:
Multiple changes are contained in this commit:

- Adoption of BSD Queues instead of custom made linked-lists. BSD Queues are readily available in:
   - Every Linux distro.
   - macOS.
   - Windows using cygwin as we already require.
   
- Fixed minor memory leaks on StreamWSR destruction.
- Used of linked-lists in often-insertion/no-random-access scenarios.
- Removed unneeded dynamic allocated variables.
- Finer grain lock/mutex usage.
- Added multiple comments and diagram.
- Renamed most of the data structures for better consistency.
- Rearranged function order in the file to better follow its execution.

Tested in:

- [x] Crusher
- [x] Local machine with ASAN
- [x] Local machine with TSAN

### DIFF
--- a/source/adios2/toolkit/sst/dp/mpi_dp.c
+++ b/source/adios2/toolkit/sst/dp/mpi_dp.c
@@ -1,21 +1,63 @@
+/**
+ * ADIOS is freely available under the terms of the BSD license described
+ * in the COPYING file in the top level directory of this source distribution.
+ *
+ * mpi_dp.c
+ *
+ * Message Passing Interface (MPI) Data plane transport for ADIOS2 SST
+ *
+ * Data scheme of the main data structures introduced here:
+ *
+ * +-------------+     +-------------------+ +----------------+
+ * | MpiStreamWR |     | MpiStreamWPR      | | MpiStreamRD    |
+ * | (Writer)    |     | (WriterPerReader) | | (Reader)       |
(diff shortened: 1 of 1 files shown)


eisenhauer (approved):
Looks good to me.  WRT scalability, any measured performance improvements?

vicentebolea:
> Looks good to me. WRT scalability, any measured performance improvements?

Not at the moment, the scalability is expected to be improved by the choice of data structures used now, for instance we used to have an array resizing in every writer for new every cohort reader introduced, this can cause problem in the thousands number, both at connecting and disconnecting. The previous PR was merged before its final touches, here they are :art: :paintbrush: 

By the way I have found a few data races and use-uninitialized-memory in the EVpath codebase, should I submit a bug report to that project?

eisenhauer:
WRT the uninit memory reads.  When sending messages, EVPath writes the entire message structure, including compiler-added padding for field alignment.  If you just initialize the fields and don't do an memset of the entire structure, you'll get uninit memory errors.  They're meaningless, but if you want to kill them so you can run clean, memset the message.  (That rules out the init-on-declare method that I saw in at least one place I think.)  However, happy to look at any uninit memory or race conditions that don't fall into this category.

PR #3293, opened 2022-07-23
Minor fixes mpi tests

vicentebolea:
- ADD a comment explaining that multithread MPI is only needed for SST MPI_DP in every test and example.
- Removed rogue sleep(1) from priors pull-requests


vicentebolea:
@pnorbert this is just adding the same comment in about every test that do `MPI_Thread_init(...MPI_THREAD_MULTIPLE)`. I will merge this soon without review since it does not change any functional code.

PR #3299, opened 2022-07-27
Optimize processing attributes, by using a new AttributeBase.Equals()…

pnorbert:
… function instead of building a small string map. Also speed up BP5Deserializer.BreakDownVarName, and BreakDownArrayName by using strtol() calls instead of sscanf.


eisenhauer (approved):
Looks good...

PR #3308, opened 2022-08-11
ci: fix macos install test missing deps

vicentebolea:
Cmake 3.24 deprecated some features that are used in our install test. This MR fix the CMake used for MacOS build to 3.23.3.

This is a temporary fix, final fix is being tracked here: https://github.com/ornladios/ADIOS2/issues/3309


eisenhauer (approved):
Looks good...

Issue #3309, opened 2022-08-12
Install.Setup in MacOS to be compatible with CMake 3.24


vicentebolea:
Cmake 3.24 deprecated some features that are used in our install tests. When CMake 3.24 is used we get the following error in the MacOS builds:

```
nstalling: /Users/runner/work/ADIOS2/ADIOS2/macos11-xcode13_0-serial/testing/install/install/lib/cmake/adios2/adios2-config-version.cmake
-- Installing: /Users/runner/work/ADIOS2/ADIOS2/macos11-xcode13_0-serial/testing/install/install/lib/cmake/adios2/adios2-config-common.cmake
-- Installing: /Users/runner/work/ADIOS2/ADIOS2/macos11-xcode13_0-serial/testing/install/install/lib/cmake/adios2/adios2-targets.cmake
-- Installing: /Users/runner/work/ADIOS2/ADIOS2/macos11-xcode13_0-serial/testing/install/install/lib/cmake/adios2/adios2-targets-debug.cmake
-- Installing: /Users/runner/work/ADIOS2/ADIOS2/macos11-xcode13_0-serial/testing/install/install/lib/cmake/adios2/adios2-config.cmake
/var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/tmp.36ZY9sps ~/work/ADIOS2/ADIOS2/macos11-xcode13_0-serial/Testing/install
Configuring adios2-config dummy CMake project
-- The C compiler identification is AppleClang 13.0.0.13000029
-- The CXX compiler identification is AppleClang 13.0.0.13000029
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: /Applications/Xcode_13.0.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/clang - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: /Applications/Xcode_13.0.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/clang++ - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Found adios2: /Users/runner/work/ADIOS2/ADIOS2/macos11-xcode13_0-serial/testing/install/install/lib/cmake/adios2/adios2-config.cmake (found version "2.8.3") found components: C CXX 
[...]

PR #3310, opened 2022-08-12
CMAKE: Set maximum policy version project-wide

vicentebolea:
Follow-up of:

- https://github.com/ornladios/ADIOS2/issues/3309
- https://github.com/ornladios/ADIOS2/pull/3308


vicentebolea:
This does not work since the problem is not about policies but removed CMAKE variable.

PR #3313, opened 2022-08-15
Enable memory selection in ssc reader

JasonRuonanWang:


JasonRuonanWang:
@pnorbert This should enable memory selection on the reader side of SSC. Let me know if it's required on the writer side too.

PR #3325, opened 2022-08-25
CI: update macOS version for github OS builds

vicentebolea:
Github Actions deprecated the host OS macOS Catalina 10.15, here we drop CI coverage for it and add for the newly macOS 12 Monterey.

- Removed MPI CI build cmake conf files since there is not MPI in the Github macOS host machines.


scripts/ci/cmake-v2/ci-macos11-xcode13_0-mpi.cmake
```diff
@@ -1,23 +0,0 @@
-# Client maintainer: chuck.atkins@kitware.com
```
caitlinross, line 1: Was this one supposed to be deleted? It seems like there's mpi and serial versions of the other scripts. And so should there also be a `ci-macos12-xcode13_4_1-mpi.cmake`?

caitlinross (commented):
Looks good

scripts/ci/cmake-v2/ci-macos11-xcode13_0-mpi.cmake
caitlinross: Heh, nevermind. It would help if I read your earlier comment first

vicentebolea: Thanks @caitlinross can you approve this PR?

caitlinross (approved):

PR #3331, opened 2022-08-30
CI: major rework for image generations of GHA

vicentebolea:
CI: major rework for image generations of Github Actions builds.

- It adds Kokkos CUDA.
- It sets the CMAKE version to 3.22.4.
- It leverages in Spack Environments to install the dependencies needed for the builds
- In the CUDA image, we install CUDA in Spack rather than using the CUDA rockylinux image.
- It reduces the amount of intermediate images.
- It introduces pyexpand templates to easily modify packages and versions per each build in the future wihouth changing the dockerfiles and shell scrips.


scripts/ci/images-v2/build-cuda.sh
```diff
@@ -32,7 +32,10 @@ if ! build_squash \
     ornladios/adios2:ci-spack-el8-cuda-gcc8-base \
     ornladios/adios2:ci-spack-el8-cuda-serial \
     Dockerfile.ci-spack-el8-leaf \
-    "--build-arg=COMPILER_IMG_BASE=${build_conf} --build-arg=COMPILER_SPACK_ID=gcc --build-arg=CUDA_VARIANT=+cuda"
+    "--build-arg=COMPILER_IMG_BASE=${build_conf}
+     --build-arg=COMPILER_SPACK_ID=gcc
+     --build-arg=CUDA_VARIANT=+cuda
+     --build-arg=EXTRA_SPACK_PACKAGES=kokkos@3.6.01%gcc+cuda+wrapper"
```
anagainaru, line 38: I did not have time to reproduce this to test my theory, but if you are using the wrapper for kokkos, it needs to be used for adios as well. How is adios build in the CI?

anagainaru (commented):
You should be able to load kokkos without the wrapper. Did you say you were getting errors?

I haven't tested yet, but ADIOS2 and Kokkos need to both be build with the nvcc wrapper or both without.

scripts/ci/images-v2/build-cuda.sh
vicentebolea: Here you can change the flags for the CUDA build

https://github.com/ornladios/ADIOS2/blob/master/scripts/ci/cmake-v2/ci-el8-cuda-serial.cmake

vicentebolea: > You should be able to load kokkos without the wrapper. Did you say you were getting errors?

If I do a `kokkos@3.6.01%gcc+cuda+wrapper` with spack 0.18.1

source/adios2/CMakeLists.txt
```diff
@@ -120,7 +120,7 @@ if(ADIOS2_HAVE_CUDA)
     OUTPUT_NAME adios2${ADIOS2_LIBRARY_SUFFIX}_core_cuda
     )
   target_compile_features(adios2_core_cuda PRIVATE cuda_std_14)
-  target_link_libraries(adios2_core PRIVATE adios2_core_cuda CUDA::cudart CUDA::cuda_driver)
```
vicentebolea, line 123: @anagainaru I have removed this dependency as we did not use it so far and it is difficult to provide through the CI.

vicentebolea:
Merging without review since these are CI changes about the images and many iterations are expected. The images are already deployed and working

PR #3369, opened 2022-10-21
Fix race condition in OnDemand delivery

eisenhauer:
This mod excludes the timestep that is currently being processed in ProvideTimestep() from consideration for transmission in the RequestTimestep handler.  This resolves a race condition where the "in-the-queue-but-not-quite-complete" timestep might be sent out OnDemand before it's ready because it looks to the handler like it's the next one to be requested OnDemand.  Instead, the handler will queue its request and let ProvideTimestep() proceed as normal.  This obviates the need for the prior race condition patch because ProvideTimestep() will never find that the step it is working on has been sent out before it's done.

This patch also includes changes to the TestDistributionRead/Write in their OnDemand sections which were previously unused.  Those tests are now suitable for running on the command line like:  
bin/TestDistributionWrite --on_demand --num_steps 1000 SST tmp & bin/TestDistributionRead --on_demand SST tmp & bin/TestDistributionRead --on_demand SST tmp & bin/TestDistributionRead --on_demand SST tmp
And one can manually add the "Total steps received is" numbers to see if they sum to 1000.  This test is still not suitable for CI, but handy for manual testing.

### DIFF
--- a/source/adios2/toolkit/sst/cp/cp_internal.h
+++ b/source/adios2/toolkit/sst/cp/cp_internal.h
@@ -128,6 +128,7 @@ typedef struct _CPTimestepEntry
     struct _TimestepMetadataMsg *Msg;
     int MetaDataSendCount;
     int ReferenceCount;
+    int InProgressFlag;
     int Expired;
     int PreciousTimestep;
     void **DP_TimestepInfo;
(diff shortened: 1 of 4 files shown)


anagainaru (approved):
Looks good. It passed the large test using the TIL applications (with 100 whole slide images and 10 consumers and variable processing times). 

The tests look good to me. I agree that it does not work as a CI test.

PR #3374, opened 2022-10-25
Added BLOSC2 operator using the c-blosc-2 library. It works identical…

pnorbert:
…ly to the BLOSC operator that uses the old blosc library.


psychocoderHPC:
Thanks for adding blosc2. Since blosc2 support 64bit buffer with there framing API the framing API should be used instead of the self made block/patches I added to the BLOSC1 operator.

psychocoderHPC:
https://www.blosc.org/c-blosc2/format/cframe_format.html

PR #3387, opened 2022-11-23
BP5 read direct to application memory (1 dimensional case)

eisenhauer:
In general in ADIOS, writers populate portions of a virtual multidimensional variables logically decomposed across writer ranks and readers request multidimensional slices out of that space.  Each block of each writer's contribution was contiguous (or made contiguous) when it was written, and each readers completed request will end up being contiguous in reader memory, but because a reader request might pull data from multiple writer blocks or involve only a subset of a block, filling a read request often involves non-contiguous transfers.  That is, sections of a writer block that were contiguous in that block end up being non-contiguous in the reader destination, and vice versa.  Because of this BP5 was conservative, always reading data into a temporary buffer and then slicing it into the appropriate place in application memory.  It would be more efficient to read directly into application memory when we have a contiguous-to-contiguous transfer, but the general case of determining if a transfer is really contiguous-to-contiguous for multidimensional data is somewhat tricky.  However, the 1D case is easier to reason about and is addressed in this PR.  Modifications are contained entirely in BP5Serializer::GenerateReadRequests and FinalizeGets.  @anagainaru I've excluded cases where our destination is not in the Host MemorySpace (and corrected one spot where the destination memory space was not initialized in the ReadRequest.

### DIFF
--- a/source/adios2/toolkit/format/bp5/BP5Deserializer.cpp
+++ b/source/adios2/toolkit/format/bp5/BP5Deserializer.cpp
@@ -1219,6 +1219,7 @@ bool BP5Deserializer::QueueGetSingle(core::VariableBase &variable,
             Req.Count = variable.m_Count;
         }
         Req.Data = DestData;
+        Req.MemSpace = MemSpace;
         Req.Step = Step;
         PendingRequests.push_back(Req);
     }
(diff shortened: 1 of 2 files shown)


source/adios2/toolkit/format/bp5/BP5Deserializer.cpp
```diff
@@ -1219,6 +1219,7 @@ bool BP5Deserializer::QueueGetSingle(core::VariableBase &variable,
             Req.Count = variable.m_Count;
         }
         Req.Data = DestData;
+        Req.MemSpace = MemSpace;
```
anagainaru, line 1222: I don't know how I missed this :) I guess the testing I added only looks at one particular example. I double checked and I think this is the only place missing the memory space

anagainaru (approved):
Looks good. I assume we could extend `BP5Deserializer::IsContiguousTransfer` in the future to include multidimensional arrays if the read uses the same blocking scheme as the write.

One small comment, I would change the name of the PR to specify this targets only 1D arrays.

source/adios2/toolkit/format/bp5/BP5Deserializer.cpp
eisenhauer: > I don't know how I missed this :) I guess the testing I added only looks at one particular example. I double checked and I think this is the only place missing the memory space

Unfortunately we run msan only on some smallish subset of our test suite.  Partly because MPI is notoriously bad for memory abuse (so no MPI runs) and partly because msan really slows down tests and we can't have 10 hour CI runs (so some tests get eliminated).  As a result, uninit fields may slip by sometimes...

PR #3388, opened 2022-11-23
Convert Transport parameters's keys to lower case for easier lookup.

pnorbert:


eisenhauer:
So, I know I've been mapping values to a consistent case (for things like "on", "off" and "auto"), but has case been relevant with the parameter names?  If so, that does seem like a terrible thing that should be fixed throughout the code...

PR #3390, opened 2022-11-24
WIP:  Fix untyped conversion in PNG operator test

eisenhauer:
This error doesn't always seem to pop up in CI, but trying a fix regardless.


eisenhauer:
This PR fixes a problem that got a warning on visual studio and prevented the windows build from succeeding.  Windows execution seems to be broken (at least the vast majority of the tests fail), but fixing that is going to have to wait for another day.  Merging this as an improvement, with work still to be done.

PR #3392, opened 2022-11-28
Extending the BP4 engine to allow data transfers through SCR

anagainaru:
Adding over PR #3294 

Keeping most of the code changes made by @adammoody
rebased over the latest version of ADIOS2 and passing the tests. I added a test for SCR as well.

Currently only for BP4

### DIFF
--- a/source/adios2/engine/bp4/BP4Writer.cpp
+++ b/source/adios2/engine/bp4/BP4Writer.cpp
@@ -20,6 +20,10 @@
 #include <ctime>
 #include <iostream>
 
+#ifdef ADIOS2_HAVE_SCR
+#include "scr.h"
+#endif
+
 namespace adios2
 {
 namespace core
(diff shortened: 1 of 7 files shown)


eisenhauer:
Do you have a FindSCR.cmake?  Also, does the test actually test proper functioning of SCR?  That may be what the scr_complete variable represents, but it isn't actually examined in the test proper.

anagainaru:
> Do you have a FindSCR.cmake? 

Oops I forgot to add it to the first commit

> Also, does the test actually test proper functioning of SCR? That may be what the scr_complete variable represents, but it isn't actually examined in the test proper.

I am actually not sure how to examine if SCR is doing anything.
Following the code Adam has in his example, I am calling SCR functions before Put and I assume all files are handled by SCR.

```c++
SCR_Start_output(fname.c_str(), SCR_FLAG_CHECKPOINT);
// Engine Open
...
// Engine Close
SCR_Complete_output(scr_valid);
```

All the variables between the two calls are tracked with SCR inside ADIOS2. 

We might want to do some smarter things (like detect is SCR_Start_output has been called and not error in case it has not but just fall back to not using it or be able to decide for which variable we want to use SCR), but we can discuss them today.

source/adios2/engine/bp4/BP4Writer.cpp
```diff
@@ -20,6 +20,10 @@
 #include <ctime>
 #include <iostream>
 
+#ifdef ADIOS2_HAVE_SCR
+#include "scr.h"
```
williamfgc, line 24: `#include <scr.h>` would help identify it as an external header.

testing/adios2/engine/bp/TestBPWriteReadSCR.cpp
```diff
@@ -0,0 +1,1044 @@
...
+#include <numeric> //std::iota
+#include <stdexcept>
+
+#include <adios2.h>
+
+#include <gtest/gtest.h>
+
+#include "../SmallTestData.h"
+
+#include "scr.h"
```
williamfgc, line 18: Similarly here `#include <scr.h>` would help identify it as an external header. Thanks for adding a test, BTW.

williamfgc (changes requested):
Just minor suggestions. Looks great, once this PR is merged it would be good to add documentation (./docs) for the readthedocs user guide.

anagainaru:
> Just minor suggestions. Looks great, once this PR is merged it would be good to add documentation (./docs) for the readthedocs user guide.

We are looking at correctness and performance for now. Once we decide to merge it, I will add documentation for it. We are not there yet. 

Sure about the header. If we decide to merge it, I'll change it.

PR #3399, opened 2022-12-05
Add new parameters to bpls to control engine, engine params and trans…

pnorbert:
…port params

e.g. bpls -l onecell.bp  -E BP5 -P "SelectSteps=0:n:2" -T "Library=stdio" -d step


pnorbert:
These are primarily options to test new transports, engines and parameters. Right now not so useful for users until we get exotic transports for DAOS, NVME or S3, or plugin engines. But it is already useful to set "SelectSteps" for BP5 engine, or stdio for library instead of the default.

PR #3401, opened 2022-12-05
Change (at least) some sprintf() to snprintf()

eisenhauer:
sprintf() is deprecated and the new XCode compiler is throwing warnings on use.  This PR fixes changes at least some of the uses of sprintf() to snprintf() with appropriate length values.  Some uses remain and some of them demand more than the trivial mods performed in this PR.  There are additional uses of sprintf() in the thirdparty codes, but the way we're compiling those seems to keep us from getting warnings there.  I don't think I can count on the test suite exercising all of these code paths, so more eyes on this PR are welcome.

### DIFF
--- a/source/adios2/toolkit/sst/dp/daos_dp.c
+++ b/source/adios2/toolkit/sst/dp/daos_dp.c
@@ -620,7 +620,7 @@ static void *DaosReadRemoteMemory(CP_Services Svcs, DP_RS_Stream Stream_v,
     char *StringName = malloc(20);
     void *NVBlock;
     char *BaseAddr;
-    sprintf(StringName, "Timestep_%ld", Timestep);
+    snprintf(StringName, 20, "Timestep_%ld", Timestep);
     BaseAddr =
         NULL; // nvs_get_with_malloc(Stream->writer_nvs[Rank], StringName, 1);
 
@@ -714,7 +714,7 @@ static void DaosProvideTimestep(CP_Services Svcs, DP_WS_Stream Stream_v,
 
     char *StringName = malloc(20);
     void *NVBlock;
-    sprintf(StringName, "Timestep_%ld", Timestep);
+    snprintf(StringName, 20, "Timestep_%ld", Timestep);
     NVBlock = NULL; // nvs_alloc(Stream->nvs, &Data->DataSize, StringName);
     memcpy(NVBlock, Data->block, Data->DataSize);
     // nvs_snapshot_(Stream->nvs, &Stream->Rank);
--- a/bindings/C/adios2/c/adios2_c_io.cpp
+++ b/bindings/C/adios2/c/adios2_c_io.cpp
@@ -389,7 +389,7 @@ adios2_error adios2_inquire_subgroups(char ***results, const char *full_prefix,
         for (auto &m : subGrpNames)
         {
             (*results)[i] = (char *)malloc(m.size() + 1);
-            sprintf((*results)[i], "%s", m.c_str());
+            snprintf((*results)[i], m.size() + 1, "%s", m.c_str());
             (*results)[i][m.size()] = '\0';
             i++;
         }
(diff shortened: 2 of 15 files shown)


source/adios2/toolkit/sst/dp/daos_dp.c
```diff
@@ -620,7 +620,7 @@ static void *DaosReadRemoteMemory(CP_Services Svcs, DP_RS_Stream Stream_v,
     char *StringName = malloc(20);
     void *NVBlock;
     char *BaseAddr;
-    sprintf(StringName, "Timestep_%ld", Timestep);
+    snprintf(StringName, 20, "Timestep_%ld", Timestep);
     BaseAddr =
         NULL; // nvs_get_with_malloc(Stream->writer_nvs[Rank], StringName, 1);
```
anagainaru, line 626:
```
Line 701:    //    sprintf(Info->CheckString, "Daos info for timestep %ld from rank
```
I know this is a comment but if someone decided we will need this logic for DAOS and removes the comment we will get the warning again

And the same in `source/adios2/toolkit/sst/dp/evpath_dp.c`:
```
//    sprintf(Info->CheckString, "Evpath info for timestep %ld from rank
```

anagainaru (commented):
In `source/h5vol/H5VolReadWrite.c` there are 14 usages of sprintf and one commented that you missed.

Except a few commented sprintf (that are more good practice than actual issues) you got everything. 

I am surprised that the thirdparty code does not create warnings.

source/adios2/toolkit/sst/dp/daos_dp.c
eisenhauer: Well, I saw the HDF5 uses, but they are the messier ones.  For example:

```
size_t gGetNameOfNthItem(H5VL_ObjDef_t *vol, uint32_t idx, char *name)
{
 . . . 
    if (NULL != name)
    {
        sprintf(name, "%s", n);
    }
```
Here, the target of the sprintf() is passed in to the subroutine without length information, so it's a non-local change.  (Then again, why this particular use is sprintf() rather than strcat() isn't immediately obvious.)  So, I decided not to touch HDF5 here.  (I also don't run with hdf5 on my mac, so I'm not seeing those warnings either.  Maybe we can upgrade the XCode in our Mac CI at some point and we'll tackle them there.  Either that or @guj can tackle it?)

WRT code in comments, I'm always a torn about including it when making these sorts of changes.  Some folks would argue that code shouldn't be commented out to begin with, because old code is always available in version control if you want to dig for it. But given that code exists that is commented out or surrounded by untriggered #ifdef's, I think changing it in a PR in the hopes that everything will be fine if it ever gets uncommented is optimistic.  Even if it used to compile the last time it was active, anything could have rendered it invalid, including errors in this sort of change and nobody would know because it's a comment.  Compiling with a non-fatal warning is probably one of the better outcomes one could hope for, so I left those cases alone...  Happy to reconsider if someone feels strongly though.

anagainaru: I am in favor of removing all commented code, but I see your point. The PR looks good to me.

There are actually several things in the HDF5 file that I would change. If we ever need to run with this backend it might be worthwhile to look at this code.

guj: About the comments,     maybe add reason why they  are commented out along with commented code if not removing them? 

Greg (@eisenhauer) I can make the equivalent  changes for HDF5 related files  in a separate pull request.

Ana (@anagainaru) do you want to list what are the other things  you would change. for HDF5 files?

anagainaru: @guj only one comment that I actually care about. We define in the `H5VolError.h` heade file:
```c++
void *safe_calloc(size_t n, size_t s, unsigned long line);
#define SAFE_CALLOC(n, s) safe_calloc(n, s, __LINE__)
```
not in a namespace but as global names (function/macro). If a code includes this and reuses the same name it will get errors. Or if HDF5 decides to create them, since we do `#include <hdf5.h>` within this code, ADIOS2 will not compile anymore. I would prefix all of them with `ADIOS2_`

The rest of the comments are more coding style related. I like decreasing the amount of repeated code but it's not essential. There's also some commented code that I doubt you will ever need again. Nothing too critical.

guj: > 
>   I would prefix all of them with `ADIOS2_`
> 
Good point. Will include it in next PR.  Thanks @anagainaru

Issue #3398, opened 2022-12-02
SST communication between Python and C++ clients fail at because of a call to adios2::Variable.Shape()


tmondrag:
The helloSstReader.cpp example code at https://github.com/ornladios/ADIOS2/blob/master/examples/hello/sstReader/helloSstReader.cpp fails to read data written by a helloSstWriter python script because adios2::Variable.Shape() sees a null pointer.

I am attempting to set up two executables to communicate using SST over MPI and for various reasons, one has to use the Python API and the other has to use the C++ API. In order to test whether this was possible, I wrote the following Python script, trying to reproduce the HelloSstWriter example in Python
```python
from mpi4py import MPI
import numpy as np
import adios2

# MPI
comm = MPI.COMM_WORLD
rank = comm.Get_rank()
size = comm.Get_size()

# ADIOS MPI Communicator, debug mode
adios = adios2.ADIOS(comm)

# ADIOS IO
sstIO = adios.DeclareIO("myIO")
sstIO.SetEngine('Sst')
myArray = np.array([0.0, 1.0, 2.0, 3.0, 4.0, 5.0, 6.0, 7.0, 8.0, 9.0])
myArray = 10.0*rank + myArray
nx = len(myArray)
ioArray = sstIO.DefineVariable("bpFloats",myArray,[size*nx],[rank*nx],[nx],adios2.ConstantDims)
sstFileWriter = sstIO.Open("helloSst",adios2.Mode.Write)
sstFileWriter.BeginStep()
sstFileWriter.Put(ioArray,myArray,adios2.Mode.Sync)
sstFileWriter.EndStep()
sstFileWriter.Close()
```
and for the C++ SST Reader I just used the HelloSstReader from the examples.

what happens on the reader end is 
```
Incoming variable is of size Incoming variable is of size Incoming variable is of size Incoming variable is of size Invalid argument exception, STOPPING PROGRAM from rank 2
[Thu Dec  1 12:49:00 2022] [ADIOS2 EXCEPTION] <Helper> <adiosType> <CheckForNullPtr> : found null pointer in call to Variable<T>::Shape

Invalid argument exception, STOPPING PROGRAM from rank 3
[Thu Dec  1 12:49:00 2022] [ADIOS2 EXCEPTION] <Helper> <adiosType> <CheckForNullPtr> : found null pointer in call to Variable<T>::Shape

Invalid argument exception, STOPPING PROGRAM from rank 1
[...]

PR #3403, opened 2022-12-06
Add Python SstWriter example

eisenhauer:


eisenhauer:
This PR adds a python example for SST, and switching the exchanged types to double from float helps match the default (only?) floatingpoint size in Python as per issue #3398.    (There seem to be unrelated CI failures that may be the result of failures elsewhere (can't find host open.cdash.org ?  Will force-push later to retry.)

anagainaru:
Does this mean we cannot use floats in an application that is coupled with a python code? If yes, we need to document this. 

I would also add a python writer to couple it with the cpp reader (I assume the same behavior will occur).

williamfgc:
> Does this mean we cannot use floats in an application that is coupled with a python code? If yes, we need to document this.

The Float32 type needs to be passed explicitly when creating numpy arrays, the default is Float64. Not a Python, but a numpy thing: https://numpy.org/doc/stable/user/basics.types.html

williamfgc:
This is the low-level Python API, the high-level API just returns the appropriate numpy array and type if the variables is found.

Issue #3377, opened 2022-10-28
SST with backend MPI, hangs on own machine with recent MPICH 


franzpoeschel:
MPI-based streaming works (mostly) on Crusher which is where we need it, but hangs on other systems on the reader end with plain MPICH. The writer finishes cleanly when killing the reader.

**Describe the bug**

Writer SstVerbose log:
```
Sst set to use sockets as a Control Transport
Considering DataPlane "evpath" for possible use, priority is 1
Considering DataPlane "mpi" for possible use, priority is 100
Selecting DataPlane "mpi", priority 100 for use
MpiInitWriter initialized addr=0x162ba70
Opening Stream "stream"
Writer stream params are:
Param -   RegistrationMethod=File
Param -   RendezvousReaderCount=1
Param -   QueueLimit=0 (unlimited)
Param -   QueueFullPolicy=Block
Param -   StepDistributionMode=StepsAllToAll
Param -   DataTransport=mpi
Param -   ControlTransport=sockets
Param -   NetworkInterface=(default)
Param -   ControlInterface=(default to NetworkInterface if applicable)
Param -   DataInterface=(default to NetworkInterface if applicable)
Param -   CompressionMethod=None
Param -   CPCommPattern=Min
Param -   MarshalMethod=BP
Param -   FirstTimestepPrecious=False
Param -   IsRowMajor=1  (not user settable) 
Param -   OpenTimeoutSecs=60 (seconds)
Param -   SpeculativePreloadMode=Auto
Param -   SpecAutoNodeThreshold=1
Param -   ControlModule=select
Stream "stream" waiting for 1 readers
Beginning writer-side reader open protocol
MPI dataplane WriterPerReader to be initialized
Setting SpeculativePreload ON for new reader
My oldest timestep was 0, global oldest timestep was 0
Finish writer-side reader open protocol for reader 0x162b4d0, reader ready response pending
(PID 1924, TID 7f505db8e780) Waiting for Reader ready on WSR 0x162b4d0.
Reader Activate message received for Stream 0x162b4d0.  Setting state to Established.
Parent stream reader count is now 1.
Reader ready on WSR 0x162b4d0, Stream established, Starting 0 LastProvided 0.
Finish opening Stream "stream"
Reader 0 status Established has last released 4294967295, last sent 0
QueueMaintenance, smallest last released = -1, count = 1
Removing dead entries
QueueMaintenance complete
Sending TimestepMetadata for timestep 0 (ref count 1), one to each reader
[...]

PR #3407, opened 2022-12-13
CMAKE: Check usability of MPI_Port_open

vicentebolea:
This will disable the SST MPI if the MPI_Port_open function is not working or if MPICH is not defined.

Related #3377


vicentebolea:
ping @caitlinross

PR #3409, opened 2022-12-14
Rename some portions of Struct interface

eisenhauer:
More clean up of pending PRs.  This is largely cosmetic.  Where the prior struct interface called the components of a user-defined structure "items" (and had methods named appropriately), this PR now calls them "fields".  Also, it changes the name of a property of a field that I had initially misinterpreted.   The "Size" parameter to "AddItem" was used to define statically sized array fields (not to give the overall size of the field).  This has been changed to "ElementCount", and the BP5 implementation of structs changed to match this interpretation of that parameter.  Tests have also been changed to use the "offsetof()" macro rather than assuming _a priori_ knowledge of compiler-assigned field offsets.

The details of this are open to change.  Maybe "Member" instead of "Field"?  Maybe rather than a single AddField method that adds both single-element fields and static-array fields, that can be split into two methods with more explicit names?  But we should settle before the next release...

### DIFF
--- a/source/adios2/common/ADIOSTypes.cpp
+++ b/source/adios2/common/ADIOSTypes.cpp
@@ -367,4 +367,42 @@ void MinMaxStruct::Dump(DataType Type)
         break;
     }
 }
+
+int TypeElementSize(DataType adiosvartype)
+{
+    switch (adiosvartype)
+    {
+    case DataType::UInt8:
+        return 1;
+    case DataType::Int8:
+        return 1;
+    case DataType::String:
+        return -1;
+    case DataType::UInt16:
+        return 2;
+    case DataType::Int16:
+        return 2;
+    case DataType::UInt32:
+        return 4;
+    case DataType::Int32:
+        return 4;
+    case DataType::UInt64:
+        return 8;
+    case DataType::Int64:
+        return 8;
+    case DataType::Float:
+        return 4;
+    case DataType::Double:
+        return 8;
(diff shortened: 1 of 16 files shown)


source/adios2/common/ADIOSTypes.cpp
```diff
@@ -367,4 +367,42 @@ void MinMaxStruct::Dump(DataType Type)
         break;
     }
 }
+
+int TypeElementSize(DataType adiosvartype)
```
anagainaru, line 371: Is there a reason why we are hardcoding this and not using `sizeof(uint8_t)` etc?

anagainaru (commented):
Looks good, I have no preference between `Field` or `Member`

source/adios2/common/ADIOSTypes.cpp
eisenhauer: No great reason, except it was shorter to type.  ADIOS wouldn't do well if the size implications of any of the types (including float and double) ever changed.   (Which kind of generically excludes machines like the old KSR-1 where float was 8 bytes and double was 16.)  Looking at the level of type flexibility that would let us move data between machines with different underlying data representations is an interesting topic though...

anagainaru: I guess this is not an issue for everywhere we run ADIOS2 now, but it would be good to think about it before we are forced to do it :) (not part of this PR)

anagainaru (approved):

PR #3410, opened 2022-12-16
GPU buffers will never bypass the ADIOS internal buffers

anagainaru:
The BP5 tests do not pass for contiguous buffers because the `IsContiguousTransfer` function returns true regardless of the buffer memory space. The bug appears after PR #3387.
 @eisenhauer you mention that you made sure to avoid the case of GPU buffers but I cannot find any code for this. Maybe it was removed later, but in the current form, all the GPU tests with BP5 are failing.

### DIFF
--- a/source/adios2/toolkit/format/bp5/BP5Deserializer.cpp
+++ b/source/adios2/toolkit/format/bp5/BP5Deserializer.cpp
@@ -2,7 +2,7 @@
  * Distributed under the OSI-approved Apache License, Version 2.0.  See
  * accompanying file Copyright.txt for details.
  *
- * BP5Serializer.h
+ * BP5Deserializer.h
  *
  */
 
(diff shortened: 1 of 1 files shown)


eisenhauer:
Yes, in some interim version of the code I did exclude based on the memspace destination, but that must have gotten dropped.  My only concern here is that either this check shouldn't be in IsContiguousTransfer() (because it's a check on something beside contiguity), or that routine should be renamed.   I think previously I had it outside, setting the DirectToAppMemory flag to false if the destination memory was in CUDA memspace.

anagainaru:
@eisenhauer Makes sense, I moved it outside as I assume you had it before. 
Another option is to rename the function to `IsContiguousCPUTransfer` but it doesn't read too well. And if we want to use GPUdirect at any point we will have to change the function again. I'll make sure if anyone uses this function it checks the memory space first.

PR #3416, opened 2022-12-21
    SST: Add UCX SST Dataplane

sameehj:
SST: Add UCX SST Dataplane

    This commit implements a UCX SST dataplane.

    The UCX dataplane was added without support for PRELOAD functionality as discussed previously.
    Currently this commit disables PRELOAD functionality by commenting out the portion of the code. We need to decide how we want to ensure that PRELOAD
    is disabled when using the UCX dataplane.

    Another open topic is the priority of the UCX dataplane (compared to other DPs), if others are available. Currently we are hardcoding the priority to a value of 10.

    The dataplane was tested with UCX v1.11 and openmpi v4.1.4 (compiled with UCX support) with the following cases:
      1. SST ctests.
          99% tests passed, 1 tests failed out of 168
              The following tests FAILED:
              1056 - Engine.Staging.TestOnDemandMPI.ADIOS2OnDemandMPI.Sst.MPI (Timeout)
      2. Gray Scott from ADIOS2-EXAMPLES

    Signed-off-by: Sameeh Jubran <sameeh.j@gmail.com>
    Signed-off-by: Michael Laufer <mlaufer103@gmail.com>
    Signed-off-by: Erick Fredj <efredj@toganetworks.com>
    Signed-off-by: Mark Wasserman <mark.wasserman@toganetworks.com>

### DIFF
--- a/cmake/FindUCX.cmake
+++ b/cmake/FindUCX.cmake
@@ -1,0 +1,81 @@
+#------------------------------------------------------------------------------#
+# Distributed under the OSI-approved Apache License, Version 2.0.  See
+# accompanying file Copyright.txt for details.
+#------------------------------------------------------------------------------#
+#
+# FindUCX
+# -----------
+#
+# Try to find the UCX library
+#
+# This module defines the following variables:
+#
+#  UCX_FOUND - System has UCX
+#  UCX_INCLUDE_DIRS - The UCX include directories
+#  UCX_LIBRARIES - The libraries needed to use UCX
+#
+# and the following imported targets:
+#   ucx::ucx - The UCX library target
+#
+# You can also set the following variable to help guide the search:
+#   UCX_ROOT - The install prefix for UCX containing the
+#              include and lib folders
+#              Note: this can be set as a CMake variable or an
+#                    environment variable.  If specified as a CMake
+#                    variable, it will override any setting specified
+#                    as an environment variable.
+
+# This is a bit of a wierd pattern but it allows to bypass pkg-config and
+# manually specify library information
+if(NOT (PC_UCX_FOUND STREQUAL "IGNORE"))
+  find_package(PkgConfig)
+  if(PKG_CONFIG_FOUND)
+    set(_UCX_CMAKE_PREFIX_PATH ${CMAKE_PREFIX_PATH})
+    if(UCX_ROOT)
+      list(INSERT CMAKE_PREFIX_PATH 0 "${UCX_ROOT}")
+    elseif(NOT ENV{UCX_ROOT} STREQUAL "")
+      list(INSERT CMAKE_PREFIX_PATH 0 "$ENV{UCX_ROOT}")
+    endif()
+    set(PKG_CONFIG_USE_UCX_CMAKE_PREFIX_PATH ON)
+
+    pkg_check_modules(PC_UCX ucx)
+
+    set(CMAKE_PREFIX_PATH ${_UCX_CMAKE_PREFIX_PATH})
+    unset(_UCX_CMAKE_PREFIX_PATH)
+
+    if(PC_UCX_FOUND)
+      if(BUILD_SHARED_LIBS)
+        set(_PC_TYPE)
+      else()
+        set(_PC_TYPE _STATIC)
+      endif()
+      set(UCX_INCLUDE_DIRS ${PC_UCX${_PC_TYPE}_INCLUDE_DIRS})
+      set(UCX_LIBRARIES ${PC_UCX${_PC_TYPE}_LINK_LIBRARIES})
+      set(UCX_LIBRARY_DIRS ${PC_UCX${_PC_TYPE}_LIBRARY_DIRS})
+    endif()
+  endif()
+endif()
+
+include(FindPackageHandleStandardArgs)
+# handle the QUIETLY and REQUIRED arguments and set LIBXML2_FOUND to TRUE
+# if all listed variables are TRUE
+find_package_handle_standard_args(UCX DEFAULT_MSG UCX_LIBRARIES)
+
+if(UCX_FOUND)
+  message("Found UCX: ")
+  if(NOT TARGET ucx::ucx)
+    add_library(ucx::ucx INTERFACE IMPORTED)
+    if(UCX_INCLUDE_DIRS)
+      set_target_properties(ucx::ucx PROPERTIES
+        INTERFACE_INCLUDE_DIRECTORIES "${UCX_INCLUDE_DIRS}"
+      )
+      message("'${UCX_INCLUDE_DIRS}'")
+    endif()
+    if(UCX_LIBRARIES)
+      set_target_properties(ucx::ucx PROPERTIES
+        INTERFACE_LINK_LIBRARIES      "${UCX_LIBRARIES}"
+        INTERFACE_LINK_DIRECTORIES      "${UCX_LIBRARY_DIRS}"
+      )
+    endif()
+  endif()
+endif()
--- a/source/adios2/toolkit/sst/CMakeLists.txt
+++ b/source/adios2/toolkit/sst/CMakeLists.txt
@@ -28,6 +28,12 @@ if(ADIOS2_SST_HAVE_LIBFABRIC)
   endif()
 endif()
 
+if(ADIOS2_SST_HAVE_UCX)
+  target_sources(sst PRIVATE dp/ucx_dp.c)
+  target_link_libraries(sst
+                  PRIVATE ucx::ucx)
+endif()
+
 if(ADIOS2_HAVE_DAOS)
   target_sources(sst PRIVATE dp/daos_dp.c)
   target_link_libraries(sst PRIVATE DAOS::DAOS)
@@ -56,6 +62,7 @@ set_target_properties(sst PROPERTIES
 #------------------------------------------------------------------------------#
 set(SST_CONFIG_OPTS
   LIBFABRIC
+  UCX
   FI_GNI
   CRAY_DRC
   NVStream
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -397,7 +397,7 @@ foreach(opt IN LISTS ADIOS2_CONFIG_OPTS)
   endif()
 endforeach()
 
-if (ADIOS2_HAVE_SST AND ADIOS2_SST_HAVE_LIBFABRIC) 
+if (ADIOS2_HAVE_SST AND (ADIOS2_SST_HAVE_LIBFABRIC OR ADIOS2_SST_HAVE_UCX))
 message("    RDMA Transport for Staging: Available")
 else()
 message("    RDMA Transport for Staging: Unconfigured")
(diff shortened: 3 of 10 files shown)


eisenhauer:
Thanks for the PR Sameeh!  Lets see if we can get this through CI, then we can look at the details.  In order to get past the format step, we have to use clang-format --style=file on all the C++ files that are changed.  Different versions of clang-format behave slightly differently, so you might have to use what our CI does (version 9) to get it through.

source/adios2/toolkit/sst/cp/cp_writer.c
```diff
@@ -872,11 +872,19 @@ WS_ReaderInfo WriterParticipateInReaderOpen(SstStream Stream)
     CP_WSR_Stream->ReaderStatus = Opening;
     if (ReturnData->SpecPreload == SpecPreloadOn)
     {
-
-        CP_WSR_Stream->PreloadMode = SstPreloadSpeculative;
-        CP_WSR_Stream->PreloadModeActiveTimestep = 0;
-        CP_verbose(Stream, PerStepVerbose,
-                   "Setting SpeculativePreload ON for new reader\n");
+        if (strcmp(Stream->DP_Interface->DPName, "ucx") != 0)
```
eisenhauer, line 875: I think you can kill this change.  The RDMA dataplane also does not support speculative preload (only learned preload), but doesn't get handled specially.  This is reasonably in line with established ADIOS practice where unsupported engine parameters simply have no effect.  (This can be a pain, because it also means that misspelled engine parameters are silently ignored, but correcting that is another subject...)

sameehj: Thanks for your detailed explanation, dropped now.

sameehj:
> Thanks for the PR Sameeh! Lets see if we can get this through CI, then we can look at the details. In order to get past the format step, we have to use clang-format --style=file on all the C++ files that are changed. Different versions of clang-format behave slightly differently, so you might have to use what our CI does (version 7.1) to get it through.

Thank you Greg for your helpful insights. I have ran clang-format-7 for c files and updated the pull request.
Let's see how do we do now.

anagainaru:
By just looking at the code, this looks good Sameeh. Is there any system where I can test this?

eisenhauer:
Would you please rebase this PR on current master?

vicentebolea:
Many thanks @sameehj, will this work with MPICH+UCX?

eisenhauer:
> Many thanks @sameehj, will this work with MPICH+UCX?

I think we have to address two things: 

1. In what circumstances does this work (openMPI with UCX obviously, other MPIs?  With/without UCX built-in?)
2. Can we detect in CMake whether or not this will work with the MPI that's being built with and disable the transport if not?  (Also, SST builds and works when ADIOS is not built with MPI because we can have a single non-MPI process that uses SST to communicate with other MPI or non-MPI processes.  So, whether build the RDMA data plane depends upon whether or not we find libfabric, not anything about MPI.)

vicentebolea:
> In what circumstances does this work (openMPI with UCX obviously, other MPIs? With/without UCX built-in?)

Note that it will be desirable for this to work MPICH+UCX since this the standard and default build shipped in distros such as SLES. (Part of this conversation: https://github.com/ornladios/ADIOS2/issues/3377 )

> Can we detect in CMake whether or not this will work with the MPI that's being built with and disable the transport if not? 

Yes, I think that will have to come up with a more comprehensive approach to select a good default SST Dataplane, we can definitely use CMake for this as in #3407.

cmake/FindUCX.cmake
```diff
@@ -0,0 +1,79 @@
...
+    add_library(ucx::ucx INTERFACE IMPORTED)
+    if(UCX_INCLUDE_DIRS)
+      set_target_properties(ucx::ucx PROPERTIES
+        INTERFACE_INCLUDE_DIRECTORIES "${UCX_INCLUDE_DIRS}"
+      )
+      message("'${UCX_INCLUDE_DIRS}'")
+    endif()
+    if(UCX_LIBRARIES)
+      set_target_properties(ucx::ucx PROPERTIES
+        INTERFACE_LINK_LIBRARIES      "${UCX_LIBRARIES}"
```
vicentebolea, line 75: Can you also set the INTERFACE_INCLUDE_DIRECTORY, this is needed for builds done with Spack.

An example of this is here: https://github.com/GTkorvo/dill/pull/30/files

source/adios2/toolkit/sst/CMakeLists.txt
```diff
@@ -28,6 +28,13 @@ if(ADIOS2_SST_HAVE_LIBFABRIC)
   endif()
 endif()
 
+if(ADIOS2_SST_HAVE_UCX)
+  target_sources(sst PRIVATE dp/ucx_dp.c)
+  target_link_libraries(sst
+                  PRIVATE ucx::ucx)
+  set(CMAKE_REQUIRED_INCLUDES ${UCX_INCLUDE_DIRS})
```
vicentebolea, line 35: This should be instead handled by the target when using target_link_libraries

vicentebolea (dismissed):
Many thanks, some CMake changes requested

MichaelLaufer:
Hi all,
Regarding MPI, the UCX dataplane is independent of MPI so it should not be affected. Thus is should work with MPICH+UCX without issue, although we have not tested it ourselves.
Regarding UCX limitations, we will assess our usage of the UCX APIs to see if any minimum version is required and report back.
We would be happy to receive any additional feedback, in particular reports of any successful tests/deployments of this UCX DP would be useful indeed.

vicentebolea:
>  Regarding MPI, the UCX dataplane is independent of MPI so it should not be affected. Thus is should work with MPICH+UCX without issue, although we have not tested it ourselves.

It appears that MPICH with UCX devel installations does not install ucx headers/libs along it. Do you think whether there is a way to find UCX symbols in an MPICH+UCX installation? This will be just great since many modern MPICH packages in mainstream distros are built with UCX.

sameehj:
> By just looking at the code, this looks good Sameeh. Is there any system where I can test this?
@anagainaru 
Any system (LAPTOP / PC / CLUSTER) can be used to test this. All you need a Linux based system with ucx installed and ADIOS2 compiled with ucx (-DUCX_ROOT=path/to/ucx/).

Note that you can even run this using Shared Memory (xpmem / shmem) on a single node / CPU.

cmake/FindUCX.cmake
sameehj: Done

source/adios2/toolkit/sst/CMakeLists.txt
sameehj: Dropped "set(CMAKE_REQUIRED_INCLUDES ${UCX_INCLUDE_DIRS})".

sameehj:
> Many thanks @sameehj, will this work with MPICH+UCX?

@vicentebolea Yes, works with mpich. I was just done with testing it.

sameehj:
> > Many thanks @sameehj, will this work with MPICH+UCX?
> 
> I think we have to address two things:
> 
> 1. In what circumstances does this work (openMPI with UCX obviously, other MPIs?  With/without UCX built-in?)

I can confirm that it works with mpich / openmpi. I suppose that other MPI implementations should work because only ucx code was used in the SST dataplane and there is no dependency for MPI.

it doesn't matter if it has UCX built in or not as long as the ucx used to build with ADIOS2 is the same ucx that was compiled with the mpi.

I tried to use different mpich + ucx_version_a while building ADIOS2 with ucx_version_b. unfortunately that doesn't work and fails during the dynamic linking process.

In theory we should be able to make it work, but I'm not sure on the best practices when it comes to fixing it and does it bring any value in supporting this in the first place?

> 2. Can we detect in CMake whether or not this will work with the MPI that's being built with and disable the transport if not?  (Also, SST builds and works when ADIOS is not built with MPI because we can have a single non-MPI process that uses SST to communicate with other MPI or non-MPI processes.  So, whether build the RDMA data plane depends upon whether or not we find libfabric, not anything about MPI.)

Do you think this change should be in this commit?

sameehj:
> Would you please rebase this PR on current master?

rebased.

sameehj:
> > Regarding MPI, the UCX dataplane is independent of MPI so it should not be affected. Thus is should work with MPICH+UCX without issue, although we have not tested it ourselves.
> 
> It appears that MPICH with UCX devel installations does not install ucx headers/libs along it. Do you think whether there is a way to find UCX symbols in an MPICH+UCX installation? This will be just great since many modern MPICH packages in mainstream distros are built with UCX.

MPICH apparently includes ucx as a submodule.  The most convenient but also the most time consuming ship is to ship MPICH with the ucx symbols. This will probably require changes in the distros package managers as well.

either way, I think that using setting the environment variable UCX_LOG_LEVEL=data should help even without the symbols. I didn't install any debug UCX and still got the prints.

What do you think?

eisenhauer:
Unfortunately we seem to have lost the clang-format, so it's not passing CI.

sameehj:
> Unfortunately we seem to have lost the clang-format, so it's not passing CI.

Sorry my bad, fixed now.

anagainaru:
@eisenhauer Vincente is on vacation until mid January. If this cannot wait you would need to dismiss his review and tell him his changes were addressed.

eisenhauer:
> @eisenhauer Vincente is on vacation until mid January. If this cannot wait you would need to dismiss his review and tell him his changes were addressed.

I saw his note about only working a few days, but hadn't dug into how to merge without his review yet.  I'm not really working myself this week, but I'll poke at this when I get a chance.

caitlinross:
He may be able to review it next week. I just looked at his calendar and he has some days throughout January that aren't marked as out of office. He has a few days next week that I'm assuming he'll be working some based on this.

eisenhauer:
Just a status update.  I consider this ready to merge with two obstacles.  The first is Vicente's review, which I'm willing to either wait for or finagle around (as I think his issues have been addressed).  The second is that the branch is shown as out-of-date with the base branch.  I think that means it needs to be rebased on upstream/master again (as there have been subsequent merges since last time).  I tried to do a rebase and push to the PR branch, but was unable to.  Maybe that's because I did something wrong, or maybe the PR doesn't have "Allow edits from maintainers" enabled.  But there's no point in finagling to fix 1 if 2 is still blocking.  @sameehj, want to try a rebase again?

sameehj:
> Just a status update. I consider this ready to merge with two obstacles. The first is Vicente's review, which I'm willing to either wait for or finagle around (as I think his issues have been addressed). The second is that the branch is shown as out-of-date with the base branch. I think that means it needs to be rebased on upstream/master again (as there have been subsequent merges since last time). I tried to do a rebase and push to the PR branch, but was unable to. Maybe that's because I did something wrong, or maybe the PR doesn't have "Allow edits from maintainers" enabled. But there's no point in finagling to fix 1 if 2 is still blocking. @sameehj, want to try a rebase again?

No problem, rebased.

vicentebolea:
@sameehj @eisenhauer many thanks for worrying about my review, I was on vacations.

The requested changes were correctly addressed :+1:

PR #3421, opened 2022-12-22
Testing for GPU backend for correct/incorrect memory spaces

anagainaru:
Testing functions for the GPU backend (for now hardcoded for CUDA but it will extended when we include other backends).

The test that was here before was checking that ADIOS2 will produce correct results when setting the correct memory space (CUDA or Host). This PR adds two more functions to the testing:
1. When `MemorySpace::Detect` is used ADIOS2 should always be correct, this is checked by the new function `CUDADetectMemSpace`. 
2. When the memory space is not set correctly, ADIOS2 throws an error.

This PR requires a discussion about what should happen when the memory space is set incorrectly. Currently, on the Put side (for a 1D contiguous allocated array):
```c++
var.SetMemorySpace(adios2::MemorySpace::Host);
bpWriter.Put(var, gpuData)
// Segfaults

var.SetMemorySpace(adios2::MemorySpace::CUDA);
bpWriter.Put(var, cpuData)
// executes correctly (because cudamemcpy works fine if the memory is allocated on the Host regardless of the flags used: cudaMemcpyDeviceToHost or cudaMemcpyHostToDevice
```

On the Get side (for a 1D contiguous allocated array):
```c++
var.SetMemorySpace(adios2::MemorySpace::Host);
bpWriter.Get(var, gpuData, adios2::Mode::Sync)
// ADIOS2 throws std::ios_base::failure

var.SetMemorySpace(adios2::MemorySpace::CUDA);
bpWriter.Get(var, cpuData, adios2::Mode::Sync)
// ADIOS2 throws std::ios_base::failure
```
In addition, regardless of Sync or Deferred, `bpReader.EndStep();` and/or `bpReader.Close();` also throw  `std::ios_base::failure` in both Get cases while for Put they do not.

### DIFF
--- a/source/adios2/helper/adiosCUDA.h
+++ b/source/adios2/helper/adiosCUDA.h
@@ -6,8 +6,6 @@
 #ifndef ADIOS2_HELPER_ADIOSCUDA_H_
 #define ADIOS2_HELPER_ADIOSCUDA_H_
 
-#include <cstddef>
-
 namespace adios2
 {
 namespace helper
--- a/testing/adios2/engine/bp/TestBPWriteReadCuda.cpp
+++ b/testing/adios2/engine/bp/TestBPWriteReadCuda.cpp
@@ -18,6 +18,72 @@ std::string engineName; // comes from command line
 const float EPSILON = std::numeric_limits<float>::epsilon();
 const float INCREMENT = 10.0f;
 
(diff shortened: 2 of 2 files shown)


eisenhauer (approved):
Looks good...

PR #3423, opened 2022-12-27
Allowing the option of CUDA memory space only if CUDA is enabled

anagainaru:
This PR attempts to give compile time errors if a GPU application is trying to use ADIOS2 without any GPU support (I added the `ADIOS2_HAVE_GPU_SUPPORT` macro set to TRUE if there is any GPU backend: CUDA, Kokkos, HIP, etc.) 

Before this PR if an application mistakingly set the ADIOS2 memory space to CUDA or Detect when ADIOS2 was not build with CUDA enabled it would get a segmentation fault. After this PR a code like this:

```c++
        data.SetMemorySpace(adios2::MemorySpace::CUDA);
        bpReader.Get(data, gpuSimData);
```
is now giving
```
example.cpp: In function 'int BPWrite(std::string, size_t, int)':
/gpfs/alpine/csc143/proj-shared/againaru/adios/code_example_cuda/cudaADIOS2example.cpp:63:45: error: 'CUDA' is not a member of 'adios2::MemorySpace'
   63 |    data.SetMemorySpace(adios2::MemorySpace::CUDA);
      |                                             ^~~~
make[2]: *** [CMakeFiles/adiosCUDAex.dir/build.make:76: CMakeFiles/adiosCUDAex.dir/cudaADIOS2example.cpp.o] Error 1
make[1]: *** [CMakeFiles/Makefile2:83: CMakeFiles/adiosCUDAex.dir/all] Error 2
```

If the memory space it set to `MemorySpace::Detect` it will also throw a compile time error.

If the user does not choose a memory space, it defaults to `MemorySpace::Host` and the code gives a segmentation fault.

### DIFF
--- a/source/adios2/core/VariableBase.cpp
+++ b/source/adios2/core/VariableBase.cpp
@@ -50,13 +50,14 @@ size_t VariableBase::TotalSize() const noexcept
 
 MemorySpace VariableBase::DetectMemorySpace(const void *ptr)
 {
+#ifdef ADIOS2_HAVE_GPU_SUPPORT
     if (m_MemSpaceRequested != MemorySpace::Detect)
     {
         m_MemSpaceDetected = m_MemSpaceRequested;
         return m_MemSpaceRequested;
     }
+#endif
 
-    // if the user requested MemorySpace::Detect
 #ifdef ADIOS2_HAVE_CUDA
     cudaPointerAttributes attr;
     cudaPointerGetAttributes(&attr, ptr);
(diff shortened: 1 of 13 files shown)


source/adios2/core/VariableBase.cpp
```diff
@@ -72,16 +73,21 @@ MemorySpace VariableBase::DetectMemorySpace(const void *ptr)
...
+    return MemorySpace::Host;
 }
 
 void VariableBase::SetMemorySpace(const MemorySpace mem)
 {
     m_MemSpaceRequested = mem;
     // reset the detected memory space
+#ifdef ADIOS2_HAVE_GPU_SUPPORT
     m_MemSpaceDetected = MemorySpace::Detect;
+#endif
```
vicentebolea, line 90:
```suggestion
```

In my opinion we do not need to detect the `MemorySpace` for cases when the user set it to anything other than `MemorySpace::Detect`. This is for simplicity on the API. It might seem strange to the user to `SetMemorySpace` to one value but then `GetMemorySpace` to another, I think it is better to let it error if the user use the wrong parameters.

vicentebolea (dismissed):
Looks good, just a few comment on the "detecting" methods. I am signing off soon so maybe you can add Greg or Caitlin to followup this review.

source/adios2/core/VariableBase.cpp
anagainaru: I agree that it's not the cleanest way to do this. `SetMemorySpace` gives a hint to what kind of buffers to expect. 
`GetMemorySpace` returns what we are actually receiving. The `MemSpaceRequested` variable is the one requested by the user, but not necessarily the one used internally. `MemSpaceDetected` is what we use internally to decided what backend to use. This differs from what is set by the user only if the application is using `Kokkos::View`s. If a user sets the memory space to `MemorySpace::Host` but then puts a CUDA Kokkos view the detected memory space will be `MemorySpace::CUDA` but at the next loop we want to reset it to what the user requested.

I agree it is a bit of a pain, so I would be ok with giving this up and throwing an exception if the `Kokkos::View` memory space does not match what the users sets.

PR #3428, opened 2023-01-05
Reader-side struct API changes

eisenhauer:
Proposed API changes to the reader side for struct variables.

The idea is this:  Rather than the prior API where you had to do this:
`                auto varStruct = io.InquireStructVariable("particles", particleDef);`
(That is, you had to know the details of the structure definition _a priori_ in order to find the variable.)
We're going to do something like this:
```
                auto varStruct = io.InquireStructVariable("particles");
                if (varStruct && (varStruct.GetReadStructDef() == nullptr))
                {
                    auto WriteStruct = varStruct.GetWriteStructDef();
                    (void)WriteStruct; // unused here, but available if needed
                    auto particleDef =
                        io.DefineStruct("particle1", sizeof(particle));
                    particleDef.AddField("a", offsetof(particle, a),
                                          adios2::DataType::Int8);
                    particleDef.AddField("b", offsetof(particle, b),
                                          adios2::DataType::Int32, 4);
                    varStruct.SetReadStructDef(particleDef);
                }
```
That is, we're going to allow introspection of incoming types, allow a structure definition to be associated with a specific incoming variable (rather than being pulled by name from a global map), and eventually we may allow the reader side structures to not exactly match writer-side structures (to an extent to be determined later).

### DIFF
--- a/bindings/CXX11/adios2/cxx11/IO.cpp
+++ b/bindings/CXX11/adios2/cxx11/IO.cpp
@@ -238,10 +238,6 @@ VariableNT IO::InquireStructVariable(const std::string &name,
         m_IO->InquireStructVariable(name, *def.m_StructDefinition));
 }
 
-StructDefinition *GetWriteStructDef(VariableNT var) noexcept;
-StructDefinition *GetReadStructDef(VariableNT var) noexcept;
-void SetReadStructDef(VariableNT var, const StructDefinition &def);
-
 // PRIVATE
 IO::IO(core::IO *io) : m_IO(io) {}
 
--- a/bindings/CXX11/adios2/cxx11/IO.cpp
+++ b/bindings/CXX11/adios2/cxx11/IO.cpp
@@ -238,6 +238,10 @@ VariableNT IO::InquireStructVariable(const std::string &name,
         m_IO->InquireStructVariable(name, *def.m_StructDefinition));
 }
 
+StructDefinition *GetWriteStructDef(VariableNT var) noexcept;
+StructDefinition *GetReadStructDef(VariableNT var) noexcept;
+void SetReadStructDef(VariableNT var, const StructDefinition &def);
+
 // PRIVATE
 IO::IO(core::IO *io) : m_IO(io) {}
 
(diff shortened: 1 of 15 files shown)


pnorbert:
The two examples are not the same, I expected you show how to inspect the incoming types. 
For the actual example, is it required to have NO ReadStruct to allow for a separate WriteStruct? It seems you store two separate definition struct, one for reading, one for writing already.

pnorbert:
Can we have InquireVariable() work for all variables and then determine if one has a struct and call InquireStructVariable() on it?

anagainaru:
I am a bit confused

anagainaru:
I meant to send my comment, not sure how I managed to close the PR, sorry :)

My two comments:
1. I am a bit confused by the example. I assumed you are using `WriteStruct` for setting the read struct def and not have to create it again (something like `varStruct.SetReadStructDef(WriteStruct);`). Or if you create it, it would be based on info from `WriteStruct`. Am I misunderstanding the example?
2. If I have the struct to I still need to inquire the write def or can I use Jason's API?

eisenhauer:
The example is kind of the same and kind of not.  First, to be clear, the WriteStruct is the structure definition that the writer used (and which is therefore the format of the data in the file or the stream).  The ReadStruct is the structure definition that the reader wants the data in.   This wasn't really a distinction that Jason's API made, because he assumed both _a priori_ knowledge and that both were exactly the same.   The idea behind the change is that we consider that those things might be different (for any number of reasons).    

Inspection is possible in this API; this example just doesn't show it because it was adapted from Jason's example.  (And without any kind of automatic conversion built into ADIOS, an example would be kind of messy.)

WRT "can we have InquireVariable work for all variable types?", well, not really.   InquireVariable returns the Variable class, and VariableStruct class and Variable class are both derived from VariableBase, but Variable is templated.  InquireVariable returns this templated class and you can't even use it if you don't have _a priori_ knowledge of the data type of the variable we're inquiring.  (And of course those are only for the basic types templates which are built into the ADIOS api, so not possible for user-defined structures.)  IFF we had a version that returned VariableBase, we could do this.  But early templating decisions kind of cripple our flexibility here.  You *can* do the same thing you'd do if you didn't know the type for another kind of variable, use the AvailableVariables() call to get the names, use VariableType(name) to get the type associated with a variable name, then call InquireVariable<T> or InquireStructVariable (with no templating) if it's a struct type.  This is where we get all these ugly macros in our internal code....  There may be opportunities to merge some aspects VariableStruct and Variable classes, but that's a more serious discussion.

We should probably talk about all this in a call, perhaps a dedicated call rather than the generic dev call...

anagainaru (approved):

PR #3434, opened 2023-01-13
Testing for MGARD operator with GPU buffers

anagainaru:
Adding a test for MGARD with CUDA buffers.

For small buffers, the output buffer returned by MGARD has a bigger size than the input buffer so ADIOS2 seg faults. This will be addressed when we add a threshold for when a buffer is compressed (similar to Blosc). This will be done after the Kokkos PR is merged since the API changed and fixing it here will create conflicts.


eisenhauer:
Docs are good in any case, but does each operator have to handle CUDA memory independently?  I.E. reimplemented in each?   (Maybe this was clear in an earlier commit, since this is just testing, not implementation.)

anagainaru:
@eisenhauer not all operators can handle GPU buffers. I talked to Norbert about this at the ECP meeting and I will add a function that will tell us if an operator can handle the buffers or no and if no, I will allocate memory on the CPU, move the buffer and give the operator this new buffer as input (we can discuss this at the next ADIOS2 meeting).

This PR does not do this. MGARD (the same as ZFP) can handle GPU buffers so this is testing to make sure the results are correct. Maybe the documentation should be a separate PR for all operators (and include information about which can handle GPU since right now the solution I described above is not there)

eisenhauer:
Not all operators can handle GPU buffers, but whether or not it can handle a GPU buffer or not should probably be a property of the operator (in the code), and we should handle transferring from GPU memory to CPU memory in the serialization code if it's necessary for an operator that doesn't handle it.  (Like we transparently handle Cuda/non-Cuda in GetMinMax() in BP5.

JieyangChen7:
@anagainaru I'm a little bit confused. MGARD operator should support both `double` and `float`. Do you get an error using float?

Also, what do you mean by selections? Do you mean selecting a subarray of data for decompression/access such as using SetSelection? I think that functionality is provided by ADIOS. Compression operators should not handle that. Please correct me if I'm wrong.

anagainaru:
@JieyangChen7 for 2D arrays if I uncomment the line `bpWriter.Put<float>("r32", r32s.data());` from the `void MGARDAccuracy2D(const std::string tolerance)` function and I try to run the test I get:

```
[againaru@login1.summit build]$ ./bin/Test.Engine.BP.WriteReadMGARD.Serial
[==========] Running 8 tests from 1 test suite.
[----------] Global test environment set-up.
[----------] 8 tests from MGARDAccuracy/BPWriteReadMGARD
[ RUN      ] MGARDAccuracy/BPWriteReadMGARD.BPWRMGARD2D/0
[info] Using 1 devices.
[info] Estimated memory usage: 0.001149GB, Available: 16.612458GB
[info] DomainDecomposer: no decomposition used
[info] Output preallocated: 1
[info] Input previously pinned: 0
[info] Output previously pinned: 0
[time] Prepare input and output buffer time: 0.001670 s
[info] Compressing subdomain 0 with shape: 100 50
[info] Select device: Tesla V100-SXM2-16GB
[info] Use zero copy when calculating norm
[info] L_2 norm: 2886.318359
[time] Calculate norm time: 0.000091 s
[time] Decomposition time: 0.000408 s
[time] Decomposition throughput: 0.048972 GB/s
[time] Quantization time: 0.000031 s
[time] Quantization throughput: 0.652592 GB/s
[info] Outlier ratio: 0/5000 (0.000000%)
[info] Huffman block size: 20480
[info] Huffman dictionary size: 8192
[info] Huffman compress ratio (primary): 20000/632 (2025.316456)
[info] Huffman compress ratio: 20000/34488 (0.579912)
[time] Huffman compress time: 0.001441 s
[time] Low-level compression time: 0.002038 s
[time] Low-level compression throughput: 0.009812 GB/s
GPUassert: invalid argument /ccs/home/againaru/adios/mgard/MGARD/include/mgard-x/CompressionHighLevel/../Hierarchy/../RuntimeX/DeviceAdapters/DeviceAdapterCuda.h 815
```

Same for the GPU buffer. For 1D arrays float is working. If you think it's on our side I can take a look at this next week.

anagainaru:
@JieyangChen7 the same for the selection tests (read only a selection slice). If you add 
```c++
TEST_P(BPWriteReadMGARD, BPWRMGARDSel1D) { MGARDAccuracy1DSel(GetParam()); }
```
to the `testing/adios2/engine/bp/operations/TestBPWriteReadMGARD.cpp` file there is the same error.

```
GPUassert: invalid argument /ccs/home/againaru/adios/mgard/MGARD/include/mgard-x/CompressionHighLevel/../Hierarchy/../RuntimeX/DeviceAdapters/DeviceAdapterCuda.h 815
```

JieyangChen7:
@anagainaru The issue is not caused by MGARD or ADIOS. It is actually due to a characteristic of GPU compression. Unlike CPU compression, the data compressed with GPU compressor usually contains a larger metadata, so it means when the input size is too small (e.g., <100KB), the output size can be larger than the input. If ADIOS does not provide a buffer with sufficient space, MGARD reports error. This can rarely happen since I believe we don't usually compress dataset with such small size. For the propose of testing, we need to slightly increase the input data size. In addition, I found that the MGARD operator is not properly configured in these tests. Currently it is configured to bound L2 error, but the tests are actually checking L_inf error.  Also, the function of the input data (i.e., a sequence from 0 to Nx-1) need to be changed as we increase the input size, since it would be impossible to bound small absolute error (e.g., 10^-5) when the L_inf norm of the data is large (e.g., 1e6) and we use a data type that has a large precision limitation (e.g., float). 

I already made some changes to the test code locally. If it is fine for you, I can push my changes to this branch to fix everything and you can check them on your side. Just let me know.

anagainaru:
@JieyangChen7 ok that makes sense. We will discuss this on our side and come back to you. Thanks for checking.

JieyangChen7:
@anagainaru I think I don't have permission to push changes to this branch. Could you change that?

anagainaru:
@JieyangChen7 you should be able to push now

JieyangChen7:
@anagainaru It seems I still cannot push to this branch.

`ERROR: Permission to anagainaru/ADIOS2.git denied to JieyangChen7.
fatal: Could not read from remote repository.`

anagainaru:
If there are no objections I will merge this (once all tests pass) and add the threshold in MGARD in another PR. 

Currently, if the operator is not applied (in Blosc and here as well if I just add a threshold) there is a memcpy in the compression function to copy the uncompressed data in the output buffer. This memcpy does not consider memory space so for GPU buffers it seg faults.

eisenhauer (approved):

eisenhauer:
Hmm.  Was this intended to remove FindMGARD.cmake and not replace it with a lower-case version of that file?  We get a configure warning like this...

anagainaru:
Aaah, it's not included in the files !? I will add it now

PR #3438, opened 2023-01-18
Fix MGARD operator

JieyangChen7:
Removed the host-side memory copy in the MGARD operator as input/output buffer may resides on the GPU device

### DIFF
--- a/source/adios2/operator/compress/CompressMGARD.cpp
+++ b/source/adios2/operator/compress/CompressMGARD.cpp
@@ -139,28 +139,14 @@ size_t CompressMGARD::Operate(const char *dataIn, const Dims &blockStart,
         }
     }
 
-    size_t sizeOut = 0;
-    void *compressedData = nullptr;
+    // let mgard know the output buffer size
+    size_t sizeOut =
+        helper::GetTotalSize(blockCount, helper::GetDataTypeSize(type));
+    void *compressedData = bufferOut + bufferOutOffset;
     mgard_x::compress(mgardDim, mgardType, mgardCount, tolerance, s,
-                      errorBoundType, dataIn, compressedData, sizeOut, false);
+                      errorBoundType, dataIn, compressedData, sizeOut, true);
 
-    if (bufferOutOffset + sizeOut >
-        helper::GetTotalSize(blockCount, helper::GetDataTypeSize(type)))
-    {
-        CompressNull c({});
-        bufferOutOffset =
-            c.Operate(dataIn, blockStart, blockCount, type, bufferOut);
-    }
-    else
-    {
-        std::memcpy(bufferOut + bufferOutOffset, compressedData, sizeOut);
-        bufferOutOffset += sizeOut;
-    }
-
-    if (compressedData)
-    {
-        free(compressedData);
-    }
+    bufferOutOffset += sizeOut;
 
     return bufferOutOffset;
 }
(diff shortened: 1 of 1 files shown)


JieyangChen7:
@pnorbert @anagainaru Could you review my PR?

source/adios2/operator/compress/CompressMGARD.cpp
```diff
@@ -139,28 +139,14 @@ size_t CompressMGARD::Operate(const char *dataIn, const Dims &blockStart,
...
-        CompressNull c({});
-        bufferOutOffset =
-            c.Operate(dataIn, blockStart, blockCount, type, bufferOut);
-    }
-    else
-    {
-        std::memcpy(bufferOut + bufferOutOffset, compressedData, sizeOut);
-        bufferOutOffset += sizeOut;
-    }
+                      errorBoundType, dataIn, compressedData, sizeOut, true);
```
anagainaru, line 147: The `true` flag at the end is telling the `compress` function that the output is pre-allocated?

anagainaru (commented):
Looks good. I will run it on Summit today to make sure it works with CPU and Cuda buffers and update this review

anagainaru:
I get the same error as before when I try to use MGARD with GPU (note that in the `mgard_x::compress` function, `dataIn` is a GPU pointer, `dataOut` is a CPU buffer)

```
GPUassert: invalid argument /ccs/home/againaru/adios/mgard/MGARD/include/mgard-x/CompressionHighLevel/../Hierarchy/../RuntimeX/DeviceAdapters/DeviceAdapterCuda.h 938
```

source/adios2/operator/compress/CompressMGARD.cpp
JieyangChen7: Yes. This is correct. MGARD should use the buffer provided by ADIOS for output. Internally, MGARD will automatically detect where the buffer is located (CPU or GPU) and invoke the correct memory copy function.

anagainaru:
@JieyangChen7 When you say MGARD will automatically detect where the buffer is located (CPU or GPU) and invoke the correct memory copy function you mean MGARD does this for both dataIn and dataOut? 

I get a segmentation fault now. If MGARD assumes dataIn and dataOut are allocated on the same memory space, that is fine but I need to deal with this on our side. Right now I do not do anything since ZFP can deal with different memory spaces (saving a mem copy on our side).

JieyangChen7:
@anagainaru Yes. It should work for both input and output buffer. I have test code [here](https://github.com/JieyangChen7/MGARD/blob/mgard-x/examples/mgard-x/HighLevelAPIsGPUBuffer/Example.cu) that actually tests all combinations: GPU in+GPU out, CPU in+CPU out, GPU in+CPU out, CPU in+GPU out.  

FYI, if you are testing with the code you previously sent me via email, there is a bug in that code. The std::vector `decompressedR64s` (defined in line 64) needs to be pre-allocated otherwise cudaMemcpy (in line 77) would fail.

anagainaru (approved):

anagainaru:
@JieyangChen7 as I missed that, you are right ! it passes all my tests. Please rebase and then this is ready to be merged.

JieyangChen7:
@anagainaru Thanks! I just rebased my branch and it is ready to be merged.

PR #3441, opened 2023-01-25
Do not use the deprecated ctime(), use instead localtime_r() and strf…

pnorbert:
…time() as thread-safe alternatives


eisenhauer (approved):
Looks good...

PR #3446, opened 2023-01-27
Kokkos backend in ADIOS2

anagainaru:
This PR contains several things that allows for ADIOS2 to have either a CUDA or a KOKKOS backend

1. ADIOS2 will use a single GPU backend at a given time (there is currently no compiler that can support multiple so there is no point in supporting this). This means the `MemorySpace` can only be: *Detect*, *Host* or *GPU* and based on what is enabled in ADIOS2 (CUDA or Kokkos) we use specific functions. We have an environmental variable `ADIOS2_HAVE_GPU_SUPPORT` will be true if any GPU backend is activated.
 - If CUDA is enabled then either `ADIOS2_HAVE_CUDA` or `ADIOS2_HAVE_Kokkos_CUDA` are true.

2. All GPU related functions have been moved so they are all stored in `adios2::helper::adios{BACKEND}`

3. The GPU logic is moved as low as possible for MinMax (in `helper::adiosMath` and `helper::adiosMemeory`) so that we capture all configurations (RandomAccess, Spans, AOS, etc.)

4. Option to build ADIOS2 with Kokkos and use the Kokkos backend (currently supported HIP and CUDA). This requires to build ADIOS2 with `CXX_STANDARD=17` so I modified the cmake file to be able to do that. I added a script to build ADIOS2 with Kokkos on Summit here: [https://github.com/anagainaru/ADIOS2-addons/blob/main/GPUAware/ADIOS.kokkos/biuld-adios-kokkos-summit.sh](https://github.com/anagainaru/ADIOS2-addons/blob/main/GPUAware/ADIOS.kokkos/biuld-adios-kokkos-summit.sh). I'll add something similar for Crusher, I'm not sure where to put these in the ADIOS2 repo (maybe in the documentation when I add it).
- Kokkos 3.7 is required


bindings/CXX11/adios2/cxx11/KokkosView.h
```diff
@@ -17,24 +17,24 @@ struct memspace_kokkos_to_adios2<Kokkos::HostSpace>
     static constexpr adios2::MemorySpace value = adios2::MemorySpace::Host;
 };
 
-#if defined(KOKKOS_ENABLE_CUDA) && defined(ADIOS2_HAVE_CUDA)
+#if define(KOKKOS_ENABLE_CUDA) || define(ADIOS2_HAVE_CUDA)
```
eisenhauer, line 20: Is define() rather than defined() a valid variant in the preprocessor?

eisenhauer (approved):
The changes to the preprocessor conditional in KokkosView.h look a bit odd, switching from defined to define and the && to || seems not to fit in.  But the rest looks good.

anagainaru:
> The changes to the preprocessor conditional in KokkosView.h look a bit odd, switching from defined to define and the && to || seems not to fit in. But the rest looks good.

Yes, my mistake, none of the testing looked at that, I corrected the bug. The operator change (|| instead of &&) is because we cannot build ADIOS2 with both CUDA and Kokkos backend so the two macro are never true. I still need to test better the scenario when we use ADIOS2 with a Kokkos application, but CUDA applications work (I'll add an example in Kokkos-Examples) and update the documentation

caitlinross:
@anagainaru I was finally able to reproduce this in the container I'm running locally. Add this change in a commit:

```
--- a/scripts/ci/cmake-v2/ci-el8-cuda-serial.cmake
+++ b/scripts/ci/cmake-v2/ci-el8-cuda-serial.cmake
@@ -1,7 +1,7 @@
 # Client maintainer: vicente.bolea@kitware.com
 
 set(ENV{CC}  gcc)
-set(ENV{CXX} g++)
+set(ENV{CXX} nvcc_wrapper)
 set(ENV{FC}  gfortran)
 
 set(dashboard_cache "
```

That seems to be all that is needed.

bindings/CXX11/adios2/cxx11/KokkosView.h
dmitry-ganyushin: Actually #if defined is valued:
#define ABC 2
#include <iostream>

int main()
{

#if defined ABC
std::cout << "1: yes\n";
#else
std::cout << "1: no\n";
#endif
}

anagainaru:
@caitlinross Great ! Testing this now

cmake/DetectOptions.cmake
```diff
@@ -170,8 +170,27 @@ endif()
 
 set(mpi_find_components C)
 
-# Cuda
-if(ADIOS2_USE_CUDA)
+# Kokkos
+if(ADIOS2_USE_Kokkos)
+  if(ADIOS2_USE_Kokkos STREQUAL AUTO)
+    find_package(Kokkos 3.7 QUIET)
```
vicentebolea, line 176: @anagainaru we have to hold on this. Spock/Crusher does not offer Kokkos 3.7 just yet. I am in another conversation where I have just asked this, when will Kokkos 3.7 be avaiable for those machine. I will keep you in the loop.

vicentebolea:
@anagainaru should this go into the 2.9 release?

source/adios2/toolkit/format/bp5/BP5Serializer.cpp
```diff
@@ -600,14 +599,14 @@ static void GetMinMax(const void *Data, size_t ElemCount, const DataType Type,
     if (Type == DataType::Struct)
     {
     }
-#ifdef ADIOS2_HAVE_CUDA
+#ifdef ADIOS2_HAVE_GPU_SUPPORT
 #define pertype(T, N)                                                          \
-    else if (MemSpace == MemorySpace::CUDA &&                                  \
+    else if (MemSpace != MemorySpace::Host &&                                  \
```
vicentebolea, line 604:
```suggestion
    else if (MemSpace == MemorySpace::GPU &&                                  \
```

testing/adios2/engine/bp/CMakeLists.txt
```diff
@@ -4,6 +4,10 @@
 # ------------------------------------------------------------------------------#
 include(ADIOSFunctions)
 
+if(ADIOS2_HAVE_CUDA OR Kokkos_ENABLE_CUDA)
```
vicentebolea, line 7: I think that in detectoptions `Kokkos_ENABLE_CUDA` should imply `ADIOS2_HAVE_CUDA`

vicentebolea (changes requested):
Thanks @anagainaru for waiting for my review. It looks great, I like that you are making the code more extensible. This PR brings many changes, I commented some of those. In general it looks good, but still I can see that it is still WIP.

anagainaru:
> @anagainaru should this go into the 2.9 release?

Welcome back from vacation! Hopefully it will go in the release. I'll leave comments to your questions in the corresponding places.

cmake/DetectOptions.cmake
anagainaru: This is needed to be able to use the same device id for the Kokkos backend as for the application. If the CUDA application sets the device id not to the first GPU in the list, there will be errors between the backend and CUDA. It's not an issue for now that Kokkos 3.7 is not on Crusher, you can install your own until they add it.

examples/cuda/CMakeLists.txt
```diff
@@ -3,10 +3,14 @@
 # accompanying file Copyright.txt for details.
 #------------------------------------------------------------------------------#
 
+enable_language(CUDA)
+
 add_executable(CudaBP5WriteRead_cuda cudaBP5WriteRead.cu)
 target_link_libraries(CudaBP5WriteRead_cuda PUBLIC adios2::cxx11 CUDA::cudart)
 set_target_properties(CudaBP5WriteRead_cuda PROPERTIES CUDA_SEPARABLE_COMPILATION ON)
+set_target_properties(CudaBP5WriteRead_cuda PROPERTIES LINKER_LANGUAGE CUDA)
```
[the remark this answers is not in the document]
anagainaru: I will remove it

source/adios2/CMakeLists.txt
```diff
@@ -412,7 +408,7 @@ install(DIRECTORY toolkit/
 )
 
 # Library installation
-install(TARGETS adios2_core ${maybe_adios2_core_mpi} ${maybe_adios2_core_cuda} EXPORT adios2Exports
+ install(TARGETS adios2_core ${maybe_adios2_core_mpi} EXPORT adios2Exports
```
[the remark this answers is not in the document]
anagainaru: Because it was giving me compilation errors, I do not know cmake enough to understand the difference. Maybe you can help me with this.

scripts/ci/cmake-v2/ci-el8-cuda-serial.cmake
```diff
@@ -1,7 +1,7 @@
 # Client maintainer: vicente.bolea@kitware.com
 
 set(ENV{CC}  gcc)
-set(ENV{CXX} g++)
+set(ENV{CXX} nvcc_wrapper)
```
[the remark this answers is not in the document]
anagainaru: Caitlin and I did not understand the CI errors and it seemed this fixed the issues (for CUDA). If you can figure out the problem and figure out a solution that works for both Kokkos and CUDA, I will use that

source/adios2/core/ADIOS.cpp
```diff
@@ -85,6 +92,20 @@ class ADIOS::GlobalServices
...
+#ifdef ADIOS2_HAVE_KOKKOS
+    void Init_Kokkos_API()
+    {
+        if (helper::KokkosIsInitialized())
+            return;
+        if (!isKokkosInitialized)
+        {
+            helper::KokkosInit();
+            isKokkosInitialized = true;
+        }
```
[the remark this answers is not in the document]
anagainaru: Why remove the check? If I already know Kokkos has been initialized I do not want to pay the price of checking.

source/adios2/toolkit/format/bp5/BP5Serializer.cpp
anagainaru: The initial code was using multiple backend so `!=` made sense. For the current code version it's the same thing. If you have a preference for equal, I can change it

testing/adios2/engine/bp/CMakeLists.txt
anagainaru: They are not the same thing, `ADIOS2_HAVE_CUDA` is saying ADIOS2 has the CUDA backend enabled. 

`Kokkos_ENABLE_CUDA` is saying ADIOS2 has the Kokkos backend enabled and KOKKOS has CUDA enabled.

I can change the name to `ADIOS2_HAVE_{name}_BACKEND` with name=`CUDA` or 'KOKKOS`

We already have a `ADIOS2_HAVE_GPU_SUPPORT` to know if any of the backends can handle GPU pointers. On a CUDA machine this is equivalent to knowing if CUDA is enabled in ADIOS2 (either through the CUDA backend or through the Kokkos backend).

vicentebolea:
@anagainaru can you allow other developers to push to this branch, I have changes I want to propose to help with this :)

vicentebolea:
```
diff --git a/.github/workflows/everything.yml b/.github/workflows/everything.yml
index bf00d264b..3ed2e0908 100644
--- a/.github/workflows/everything.yml
+++ b/.github/workflows/everything.yml
@@ -97,7 +97,7 @@ jobs:
       image: ornladios/adios2:ci-spack-el8-${{ matrix.compiler }}-${{ matrix.parallel }}
       options: --shm-size=1g
       env:
-        GH_YML_JOBNAME: ${{ matrix.os }}-${{ matrix.compiler }}-${{ matrix.parallel }}
+        GH_YML_JOBNAME: ${{ matrix.os }}-${{ matrix.gpu_backend }}${{ matrix.compiler }}-${{ matrix.parallel }}
         GH_YML_BASE_OS: Linux
         GH_YML_MATRIX_OS: ${{ matrix.os }}
         GH_YML_MATRIX_COMPILER: ${{ matrix.compiler }}
@@ -114,6 +114,11 @@ jobs:
             compiler: cuda
             parallel: serial
             constrains: build_only
+          - os: el8
+            compiler: cuda
+            parallel: serial
+            gpu_backend: kokkos-
+            constrains: build_only
           - os: el8
             compiler: gcc10
             parallel: mpich
index 9f9c84cb7..808c8b8d2 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -149,8 +149,8 @@ adios_option(SZ         "Enable support for SZ transforms" AUTO)
 adios_option(LIBPRESSIO "Enable support for LIBPRESSIO transforms" AUTO)
 adios_option(MGARD      "Enable support for MGARD transforms" AUTO)
 adios_option(PNG        "Enable support for PNG transforms" AUTO)
-adios_option(CUDA       "Enable support for Cuda" AUTO)
-adios_option(Kokkos     "Enable support for Kokkos" AUTO)
+adios_option(CUDA       "Enable support for Cuda" OFF)
+adios_option(Kokkos     "Enable support for Kokkos" OFF)
 adios_option(MPI        "Enable support for MPI" AUTO)
 adios_option(DAOS       "Enable support for DAOS" AUTO)
 adios_option(DataMan    "Enable support for DataMan" AUTO)
diff --git a/cmake/DetectOptions.cmake b/cmake/DetectOptions.cmake
index 70a46a598..8fe512467 100644
--- a/cmake/DetectOptions.cmake
+++ b/cmake/DetectOptions.cmake
@@ -170,6 +170,10 @@ endif()
 
 set(mpi_find_components C)
 
+if(ADIOS_USE_Kokkos AND ADIOS_USE_CUDA)
+  message(FATAL_ERROR "ADIOS2_USE_Kokkos is incompatible with ADIOS_USE_CUDA")
+endif()
+
 # Kokkos
 if(ADIOS2_USE_Kokkos)
   if(ADIOS2_USE_Kokkos STREQUAL AUTO)
@@ -177,20 +181,16 @@ if(ADIOS2_USE_Kokkos)
   else()
     find_package(Kokkos REQUIRED)
   endif()
-  if(NOT Kokkos_ENABLE_CUDA AND (ADIOS2_USE_CUDA STREQUAL ON))
-    set(ADIOS2_USE_Kokkos FALSE)
-  else()
-    if(Kokkos_FOUND)
-      set(ADIOS2_HAVE_Kokkos TRUE)
-    endif()
+  if(Kokkos_FOUND)
+    set(ADIOS2_HAVE_Kokkos TRUE)
     if(Kokkos_ENABLE_CUDA OR Kokkos_ENABLE_HIP OR Kokkos_ENABLE_SYCL)
-	    set(ADIOS2_HAVE_GPU_Support TRUE)
+      set(ADIOS2_HAVE_GPU_Support TRUE)
     endif()
   endif()
 endif()
 
 # CUDA
-if(ADIOS2_USE_CUDA AND NOT Kokkos_ENABLE_CUDA)
+if(ADIOS2_USE_CUDA)
   include(CheckLanguage)
   check_language(CUDA)
   if(ADIOS2_USE_CUDA STREQUAL AUTO)
@@ -199,6 +199,7 @@ if(ADIOS2_USE_CUDA AND NOT Kokkos_ENABLE_CUDA)
     find_package(CUDAToolkit REQUIRED)
   endif()
 endif()
+
 if(CMAKE_CUDA_COMPILER AND CUDAToolkit_FOUND)
   enable_language(CUDA)
   set(ADIOS2_HAVE_CUDA TRUE)
diff --git a/scripts/ci/cmake-v2/ci-el8-kokkos-cuda-serial.cmake b/scripts/ci/cmake-v2/ci-el8-kokkos-cuda-serial.cmake
new file mode 100644
index 000000000..d80a0c72b
--- /dev/null
+++ b/scripts/ci/cmake-v2/ci-el8-kokkos-cuda-serial.cmake
@@ -0,0 +1,26 @@
+# Client maintainer: vicente.bolea@kitware.com
+
+set(ENV{CC}  gcc)
+set(ENV{CXX} g++)
+set(ENV{FC}  gfortran)
+
+set(dashboard_cache "
+ADIOS2_USE_BZip2:BOOL=ON
+ADIOS2_USE_Blosc:BOOL=ON
+ADIOS2_USE_DataMan:BOOL=ON
+ADIOS2_USE_Fortran:BOOL=ON
+ADIOS2_USE_HDF5:BOOL=ON
+ADIOS2_USE_Python:BOOL=ON
+ADIOS2_USE_SZ:BOOL=ON
+ADIOS2_USE_ZeroMQ:STRING=ON
+ADIOS2_USE_ZFP:BOOL=ON
+ADIOS2_USE_Kokkos:BOOL=ON
+ADIOS2_USE_MPI:BOOL=OFF
+CMAKE_C_FLAGS:STRING=-Wall
+CMAKE_CXX_FLAGS:STRING=-Wall
+CMAKE_Fortran_FLAGS:STRING=-Wall
+")
+
+set(CTEST_CMAKE_GENERATOR "Ninja")
+list(APPEND CTEST_UPDATE_NOTES_FILES "${CMAKE_CURRENT_LIST_FILE}")
+include(${CMAKE_CURRENT_LIST_DIR}/ci-common.cmake)
diff --git a/source/adios2/CMakeLists.txt b/source/adios2/CMakeLists.txt
index e1bbe0c26..9a82d0f69 100644
--- a/source/adios2/CMakeLists.txt
+++ b/source/adios2/CMakeLists.txt
@@ -124,9 +124,12 @@ if(ADIOS2_HAVE_CUDA)
   set(maybe_adios2_core_cuda adios2_core_cuda)
 endif()
 
+set(maybe_adios2_core_kokkos)
 if(ADIOS2_HAVE_Kokkos)
-  target_sources(adios2_core PRIVATE helper/adiosKokkos.h helper/adiosKokkos.cpp)
-  target_link_libraries(adios2_core PRIVATE Kokkos::kokkos)
+  target_sources(adios2_core_kokkos PRIVATE helper/adiosKokkos.h helper/adiosKokkos.cpp)
+  target_link_libraries(adios2_core_kokkos PRIVATE Kokkos::kokkos)
+  target_link_libraries(adios2_core PRIVATE adios2_core_kokkos)
+  set(maybe_adios2_core_kokkos adios2_core_kokkos)
 endif()
 
 target_include_directories(adios2_core
@@ -417,7 +420,7 @@ install(DIRECTORY toolkit/
 )
 
 # Library installation
-install(TARGETS adios2_core ${maybe_adios2_core_mpi} ${maybe_adios2_core_cuda} EXPORT adios2Exports
+install(TARGETS adios2_core ${maybe_adios2_core_mpi} ${maybe_adios2_core_cuda} ${maybe_adios2_core_kokkos} EXPORT adios2Exports
   RUNTIME DESTINATION ${CMAKE_INSTALL_BINDIR} COMPONENT adios2_core-runtime
   LIBRARY DESTINATION ${CMAKE_INSTALL_LIBDIR} COMPONENT adios2_core-libraries NAMELINK_COMPONENT adios2_core-development
   ARCHIVE DESTINATION ${CMAKE_INSTALL_LIBDIR} COMPONENT adios2_core-development
```

vicentebolea:
@anagainaru you can `git cherry-pick` the tip commit in #3480 that would fix your CI. If it works for you please:

- [ ] Create a single commit adding the CI stuff.
- [ ] Create another commit adding the feature. You can add me as co-author by writing the commits message `co-authored-by: vicente.bolea@kitware.com` as shown here (https://docs.github.com/en/pull-requests/committing-changes-to-your-project/creating-and-editing-commits/creating-a-commit-with-multiple-authors)
- [ ] Rebase onto the commit that has the tag `v2.0.0-rc1`
- [ ] Change the `base` branch in this PR to release_29.

I will do the backport to master later.

anagainaru:
@vicentebolea I think the CI should be a separate PR but I can add it here if you would prefer (and yes you'll co-author) Thanks for working on this.

The kokkos backend will require more than one commit since there are multiple unrelated changes but I don't think that should be a problem (I'm working now on getting this done today).

Agree with the rebase.  Please mark as resolved your change request.

vicentebolea:
> I think the CI should be a separate PR but I can add it here if you would prefer (and yes you'll co-author) Thanks for working on this.

I think we will have the a chicken-egg problem, the kokkos builds needs the kokkos support; the kokkos support needs kokkos build. You can have it here, also that would be better for the backport (1 PR instead of 2 PR).

> The kokkos backend will require more than one commit since there are multiple unrelated changes but I don't think that should be a problem (I'm working now on getting this done today).

That is ok, divide them in logically, build-able commits.

anagainaru:
@vicentebolea is it possible to update the CI?
```
  CMake Error at cmake/DetectOptions.cmake:182 (find_package):
    Could not find a configuration file for package "Kokkos" that is compatible
    with requested version "3.7".
```
Without this I cannot use the feature in the last commit.

vicentebolea:
> @vicentebolea is it possible to update the CI?

Sure, oops I thought spack would use 3.7 by default, I will update the images to include it.

vicentebolea:
@anagainaru I added the new image with Kokkos 3.7, please amend last commit and push to trigger the CI pipeline.

vicentebolea:
>@anagainaru I added the new image with Kokkos 3.7, please amend last commit and push to trigger the CI pipeline.
 
nevermind, it was re-trigered

cmake/DetectOptions.cmake
```diff
@@ -179,11 +207,15 @@ if(ADIOS2_USE_CUDA)
...
+    set(ADIOS2_HAVE_GPU_Support TRUE)
+  endif()
 endif()
-if(CMAKE_CUDA_COMPILER AND CUDAToolkit_FOUND)
-  enable_language(CUDA)
-  set(ADIOS2_HAVE_CUDA TRUE)
-  set(ADIOS2_HAVE_GPU_Support TRUE)
+
+if(ADIOS_HAVE_Kokkos AND ADIOS_HAVE_CUDA)
+    message(FATAL_ERROR "The Kokkos and CUDA backends cannot be active concurrently")
```
vicentebolea, line 218:
```suggestion
```

examples/CMakeLists.txt
```diff
@@ -24,6 +19,15 @@ if(ADIOS2_BUILD_EXAMPLES_EXPERIMENTAL)
...
   add_subdirectory(cuda)
 endif()
+
+if(ADIOS2_HAVE_Kokkos)
+  add_subdirectory(kokkos)
+else()
+  find_package(Kokkos QUIET)
+  if(Kokkos_FOUND)
+    add_subdirectory(kokkos)
+  endif()
```
vicentebolea, line 32: I am not sure, maybe I am wrong, but perhaps we do not need this code block.

vicentebolea (changes requested):

CMakeLists.txt
```diff
@@ -169,8 +173,10 @@ adios_option(AWSSDK     "Enable support for S3 compatible storage using AWS SDK'
 include(${PROJECT_SOURCE_DIR}/cmake/DetectOptions.cmake)
 
 if(ADIOS2_HAVE_CUDA)
```
vicentebolea, line 175:
```suggestion
if(ADIOS2_HAVE_CUDA OR ADIOS2_HAVE_Kokkos_CUDA)
```

This should fix the warnings

CMakeLists.txt
```diff
@@ -92,10 +92,13 @@ if(NOT CMAKE_BUILD_TYPE AND NOT CMAKE_CONFIGURATION_TYPES)
   set(CMAKE_BUILD_TYPE "Release" CACHE STRING "Choose the type of build." FORCE)
 endif()
 
+if(NOT CMAKE_CXX_STANDARD)
 # Force C++11 (No extensions) and C99
 set(CMAKE_CXX_STANDARD 11)
 set(CMAKE_CXX_STANDARD_REQUIRED True)
 set(CMAKE_CXX_EXTENSIONS OFF)
+endif()
+
```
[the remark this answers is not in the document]
anagainaru: This is typical for any project and it's not hurting since by default it will use c++11

But if someone uses `-D CMAKE_CXX_STANDARD=14` (which is something you use for many projects) or any other than 11, we do not use this input and not give a warning that we disregard the flag. I think we should leave it.

cmake/DetectOptions.cmake
anagainaru: Having this would prevent mistakes if we change the way we find different backends (like the one in PR #3485)

examples/CMakeLists.txt
anagainaru: We do, this is a Kokkos example that can be used with any backend (it could be CUDA in which case Kokkos is not found before this example)

CMakeLists.txt
anagainaru: Agree with this one, will push the change

CMakeLists.txt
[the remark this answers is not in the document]
anagainaru: I really dislike that we need to hardcode 17 in the cmake of all Kokkos files but for now we need this for your CI. As soon as Kokkos move to another standard I will immediately remove this and we will have to update the CI to support any version. When I compile on Summit and I set the standard by hand hardcoding is not required.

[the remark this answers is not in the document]
vicentebolea: > This is typical for any project

That is not either typical or a _recommended_ setting for every Cmake project, some projects might be built using different CXX standards, that is not our case project-wide in our current ADIOS2.

> it's not hurting since by default it will use c++11

I do not see the need of this since ADIOS2 is not using any C++17/20 feature and ADIOS2 needs C++11 as min c++ revision; It is not in the scope of this PR to have this change; This might allow other users/package management software to build adios with other CXX_STANDARD and trigger strange errors.

Lets keep it simple, we do not need this really. If it comes the time we can add this feature.

examples/CMakeLists.txt
vicentebolea: I see, that part of for the case where you build CUDA but you still want to build the Kokkos example code.

But that will enable the kokkos example even if you set `-DADIOS2_US_Kokkos=OFF`

vicentebolea (changes requested):

cmake/DetectOptions.cmake
anagainaru: The problem is that when that time comes we will have to debug this, why not just leave it?

CMakeLists.txt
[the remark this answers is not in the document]
vicentebolea: On another note if you need some files to be build with C++17 such as `AdiosKokkos.cpp` you can set the property CXX_STANDARD=17 for the `adios_core_kokkos` library.

examples/CMakeLists.txt
anagainaru: Yes `ADIOS2_USE_Kokkos` will be off so we use the CUDA backend underneath but if root to the Kokkos installation will exist the Kokkos will be found. I tried this on Summit and it works.

anagainaru: We want to support Kokkos applications even if they only use ADIOS2 with CUDA and not with Kokkos so why not enable the example for both cases?

vicentebolea: I think the devs and users expect the kokkos example to be disable when they  set `-DADIOS2_USE_Kokkos=OFF` what if I am in a machine with a incompatible kokkos, having this will break my build.

cmake/DetectOptions.cmake
vicentebolea: In the same line, what if the future code changes and we now support both Kokkos and CUDA, will have to remove this as well.

cmake/DetectOptions.cmake
```diff
@@ -170,7 +170,35 @@ endif()
...
+    find_package(Kokkos 3.7 QUIET)
+  else()
+    find_package(Kokkos 3.7 REQUIRED)
+  endif()
+  if(Kokkos_FOUND)
+    set(ADIOS2_HAVE_Kokkos TRUE)
+    if(Kokkos_ENABLE_CUDA OR Kokkos_ENABLE_HIP OR Kokkos_ENABLE_SYCL)
+      if(Kokkos_ENABLE_CUDA)
+        set(ADIOS2_HAVE_Kokkos_CUDA TRUE)
+        find_package(CUDAToolkit REQUIRED)
```
vicentebolea, line 189:
```suggestion
```

Maybe we do not need this anymore with the new Kokkos, I did not try it.

CMakeLists.txt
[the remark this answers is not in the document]
anagainaru: I'll remove this from here because the workaround of hardcoding 17 makes this irrelevant for this PR but I will open an issue and make another PR with this because I think it's unacceptable to not be able to set the standard. 
When the users/package management software specifies a standard they mean to use. It's unacceptable to ignore it and force C++11. I don't know of any project that does this (I looked even in cmake itself)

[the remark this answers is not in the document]
vicentebolea: For instance, VTK/ParaView/VTK-m they all are only compatible with a single standard same as ADIOS2, sure it will be nice to support different standards but we are not there.

>  I will open an issue and make another PR with this

That is a good idea

cmake/DetectOptions.cmake
anagainaru: indeed it is not needed anymore, it's removed

cmake/DetectOptions.cmake
anagainaru: We will never support both

examples/CMakeLists.txt
anagainaru: Yes I see your point.
It's a tough choice because if I only allow it with the Kokkos backend active it will suggest you can only use Kokkos applications with the Kokkos backend. 
And since there is really no such thing as incompatible Kokkos with the example (it can use any version of Kokkos) I tend to leave it like this. 

I could remove it completely and create a new PR with a Kokkos example so we can debate what's the best solution.

anagainaru: I researched it a bit and I am confident we should keep it, the example works with any version of Kokkos, there is nothing that can conflict, and users have a choice of adding -D CMAKE_DISABLE_FIND_PACKAGE_Kokkos=OFF if they want to deactivate it.

anagainaru:
@vicentebolea I think this is ready to merge, let me know if you think otherwise

examples/CMakeLists.txt
vicentebolea: @anagainaru I am still against this forced Kokkos dependency. In my opinion is unreasonable to expect the user to figure out the `CMAKE_DISABLE_FIND_PACKAGE_Kokkos` option if their build does not work, and while compatible at this day to many kokkos versions there is not guarantee in the future or what if an user has a broken Kokkos installation, he would no be able to bypass this, normally this is not a problem when you can just disable the feature, but here is being forced.

How about this, this is an useful example that uses kokkosView and does not need the Kokkos backend per se, as such that it is not represented by the ADIOS_USE_KOKKOS var. How about adding a new option/var that enables/disable this corner case. something akin to `ADIOS2_USE_EXAMPLE_KOKKOSVIEW`.

CMakeLists.txt
```diff
@@ -169,8 +173,10 @@ adios_option(AWSSDK     "Enable support for S3 compatible storage using AWS SDK'
 include(${PROJECT_SOURCE_DIR}/cmake/DetectOptions.cmake)
 
 if(ADIOS2_HAVE_CUDA)
-    set(CMAKE_CUDA_STANDARD 11)
-    set(CMAKE_CUDA_STANDARD_REQUIRED TRUE)
+    if(NOT CMAKE_CUDA_STANDARD)
```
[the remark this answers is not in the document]
vicentebolea: @anagainaru this thread was for the same issue as the CMAKE_CXX_STANDARD thread. CMAKE_CUDA_STANDARD is not much of a trouble because it is used in a few TUs but still having this would signal that we we support multiple C++ revisions for files with CUDA functions while not being true.

I'll leave it to you whether to maintain the change if you see that this is needed for the feature you are adding.

examples/CMakeLists.txt
anagainaru: Ok open an issue for this, this PR is not changing the status quo of the example, what you are describing was a problem before this PR as well so we shouldn't solve it here. But I agree we should discuss it.
I would also be ok moving this example in the `ADIOS2-Examples` repo (maybe as part of an application or benchmark)

CMakeLists.txt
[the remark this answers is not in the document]
anagainaru: Ups, I though I changed both, will update the code to remove the CUDA standard as well.

cmake/DetectOptions.cmake
vicentebolea: I still do not see the reason of writing conditional for potential future code, but you might have some changes in mind here for the future and this is a noop, Ill resolve this.

examples/CMakeLists.txt
vicentebolea: You are right that was here before this PR. On the other hand we can resolve this by adding in the elseif(ADIOS2_USE_EXAMPLE_KOKKOSVIEW) or similar for instance, in that manner we can avoid the issue, PR, and backport.

anagainaru: I would prefer not adding a new option for activating examples in this PR. I think we should discuss the best option.

vicentebolea: I think we should resolve this in this PR, this PR adds the `ADIOS_USE_KOKKOS` for Kokkos backends, thus it should refactor the code so that it remains consistent. Right now it is not consistent because if ADIOS_USE_KOKKOS=OFF it will still `find_package(Kokkos)` and build the kokkosview example.

The KokkosView example is an exception because it can use CUDA backend and KokkosView (not backend), so a solution would be to be enable it optionally. We also already set by default `ADIOS_USE_KOKKOS=OFF` and `ADIOS_USE_CUDA=OFF` what is the problem with doing the same for this example? We should keep these features optional.

vicentebolea: @anagainaru we have however been discussing this for a few days, lets move forward and address this later on. I will follow-up with a PR for this.

vicentebolea (approved):

examples/CMakeLists.txt
anagainaru: With `ADIOS_USE_CUDA=OFF` we still compile the CUDA example (that will use the Kokkos backend) so in a way this is similar.

We can talk about changing `ADIOS2_USE_CUDA` to `ADIOS2_USE_CUDA_BACKEND` (and same for Kokkos) if you think it's confusing right now but I don't like introducing something just for one example. Let's discuss Tuesday with everyone. For now I will rebase and merge this. Thanks for looking over this.

PR #3451, opened 2023-01-30
Fig bug in the IME file transport

anagainaru:
This bug was introduced in PR #3388 
here:
[https://github.com/ornladios/ADIOS2/commit/f31d2cb05ff4dd375d431d8043d2e16aee798b64](https://github.com/ornladios/ADIOS2/commit/f31d2cb05ff4dd375d431d8043d2e16aee798b64#diff-4cb33e244ce899f10daacb5669a29eed59c1227475c9825e81ae68990a8de984)

I do not understand how the CI did not catch this nor did any of my previous runs. 

@pnorbert please make sure my fix is correct. Should we include checking this in the CI?

### DIFF
--- a/source/adios2/toolkit/transport/file/FileIME.cpp
+++ b/source/adios2/toolkit/transport/file/FileIME.cpp
@@ -122,13 +122,13 @@ void FileIME::Open(const std::string &name, const Mode openMode,
     m_IsOpen = true;
 }
 
-void FileIME::SetParameters(const Params &parameters)
+void FileIME::SetParameters(const Params &params)
 {
     auto param = params.find("synctopfs");
     if (param != params.end())
     {
-        m_SyncToPFS =
-            helper::StringTo<bool>(value, " in Parameter key=SyncToPFS");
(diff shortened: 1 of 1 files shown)


pnorbert (changes requested):
Can I suggest to use a helper function for this entire code snippet? See FileAWSSDK.cpp
```
helper::SetParameterValue("synctopfs", params, m_SyncToPFS);
```

anagainaru:
> ```
> helper::SetParameterValue("synctopfs", params, m_SyncToPFS);
> ```

`syncToPFS` is a bool, the `SetParameterValue` expects a std::string as the last parameter, so in order to use the helper function I need to convert the string to a bool.

pnorbert (commented):
Sorry, I missed that it was a string. This made it more complicated than your original fix. Also, the default value (true) is lost here in case the parameter is not set by the user. Please revert back to your original fix,

anagainaru:
@pnorbert done, mark your change as resolved and I'll merge this

pnorbert (approved):

PR #3458, opened 2023-02-01
Tweak min blocks interface to avoid leak in Shape()

eisenhauer:


eisenhauer:
The msan and tsan failures on the last run were due to calling AvailableVariables() after a BeginStep that returned EndOfStream.  This is because we had deallocated the BP5 metadata buffer, but _not_ removed all the variables that the previous (successful) BeginStep had created.  (We do remove them on a successful BeginStep, but not on a failed one...  Arguably we should remove created vars on a failed BeginStep, but we don't really test for stuff like that.  I won't fix this in BP5/SST in this PR, because its not directly related to the purpose here, but we may have a lot of these sorts of situations that we're not testing.

source/adios2/toolkit/format/bp5/BP5Deserializer.cpp
```diff
@@ -2127,16 +2127,15 @@ Dims *BP5Deserializer::VarShape(const VariableBase &Var,
             (MetaArrayRec *)GetMetadataBase(VarRec, AbsStep, WriterRank);
         if (writer_meta_base && writer_meta_base->Shape)
         {
-            Dims *Shape = new Dims();
-            Shape->reserve(writer_meta_base->Dims);
+            Shape.resize(writer_meta_base->Dims);
```
anagainaru, line 2133: If Shape has a bigger size than the writer_meta_base, resize will remove the values but not update the size. Would this be a problem? This is not the case in any of the test so I assume no, but just want to point this out.

anagainaru (approved):

eisenhauer:
> The msan and tsan failures on the last run were due to calling AvailableVariables() after a BeginStep that returned EndOfStream. This is because we had deallocated the BP5 metadata buffer, but _not_ removed all the variables that the previous (successful) BeginStep had created. (We do remove them on a successful BeginStep, but not on a failed one... Arguably we should remove created vars on a failed BeginStep, but we don't really test for stuff like that. I won't fix this in BP5/SST in this PR, because its not directly related to the purpose here, but we may have a lot of these sorts of situations that we're not testing.

Nevermind.  BP5Deserializer should register its created variables, then they get removed at the appropriate time...

source/adios2/toolkit/format/bp5/BP5Deserializer.cpp
eisenhauer: I wouldn't think it would be a problem.  In all the current uses, Shape passed in by reference is an fresh Dims value with default size 0.  So, I think no worries...

PR #3466, opened 2023-02-03
Added support for Joined Arrays in the BP4 format and engine.

pnorbert:
Declaration:          auto var = outIO.DefineVariable<double>(table", {adios2::JoinedDim, Ncols}, {}, {1, Ncols});
Setting block size:   var.SetSelection({{}, {Nrows[block], Ncols}});

### DIFF
--- a/testing/adios2/engine/bp/TestBPJoinedArray.cpp
+++ b/testing/adios2/engine/bp/TestBPJoinedArray.cpp
@@ -1,0 +1,226 @@
+/*
+ * Distributed under the OSI-approved Apache License, Version 2.0.  See
+ * accompanying file Copyright.txt for details.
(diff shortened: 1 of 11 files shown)


testing/adios2/engine/bp/TestBPJoinedArray.cpp
```diff
@@ -0,0 +1,226 @@
+/*
```
eisenhauer, line 1: Put this in staging-common so that we're ready to test this in BP5, SST, etc?

pnorbert: Task for another PR. I want to push this out for Axel a.s.a.p

PR #3467, opened 2023-02-03
Warning fix when finding the MGARD library

anagainaru:
Related to #3434 after cherry picking Jieyang's commits, we no longer need the FindMGARD file.
Making small changes so we don't have a warning related to this when MGARD is not used.


eisenhauer (approved):

anagainaru:
@eisenhauer Jieyang changed the way ADIOS2 find the MGARD package to CONFIG mode so the find file is no longer needed.  I got confused this morning when I added the file back (the joys of muti-tasking :) 
I made small changes to suppress the warnings you were seeing.

PR #3469, opened 2023-02-06
Add JoinedArray test in staging-common, limit to BP4 currently

eisenhauer:


eisenhauer:
This is essentially just testing/adios/engines/bp/TestBPJoinedArray.cpp, split into reader and writer sections and dropped into the basic framework used by all the staging-common tests.

PR #3480, opened 2023-02-10
TESTING SOMETHING, no merge: Gpu reorg kokkos

vicentebolea:


source/adios2/CMakeLists.txt
```diff
@@ -124,9 +124,12 @@ if(ADIOS2_HAVE_CUDA)
   set(maybe_adios2_core_cuda adios2_core_cuda)
 endif()
 
+set(maybe_adios2_core_kokkos)
 if(ADIOS2_HAVE_Kokkos)
-  target_sources(adios2_core PRIVATE helper/adiosKokkos.h helper/adiosKokkos.cpp)
-  target_link_libraries(adios2_core PRIVATE Kokkos::kokkos)
+  add_library(adios2_core_kokkos helper/adiosKokkos.h helper/adiosKokkos.cpp)
```
anagainaru, line 130:
```
  set_target_properties(adios2_core_kokkos PROPERTIES
    INCLUDE_DIRECTORIES "$<BUILD_INTERFACE:${ADIOS2_SOURCE_DIR}/source>;$<BUILD_INTERFACE:${ADIOS2_BINARY_DIR}/source>"
    )
```
I added this so Kokkos can see the macro header function.

source/adios2/helper/adiosCUDA.h
```diff
@@ -10,27 +10,19 @@ namespace adios2
 {
 namespace helper
 {
-
-#ifdef _WIN32
-#define ADIOS2_EXPORT __declspec(dllexport)
-#else
-#define ADIOS2_EXPORT __attribute__((visibility("default")))
-#endif
-
```
anagainaru, line 19: The errors are different than what I am seeing on Summit because I think you used a older version of my commits. The `ADIOS2_EXPORT` should not be deleted from this file and it should appear in the `adiosKokkos.h` header file as well

anagainaru (commented):
Once you add the two changes I mentioned, you should see the same errors as on Summit

anagainaru:
```diff
-   target_link_libraries(adios2_core PRIVATE adios2_core_cuda CUDA::cudart)
+  target_link_libraries(adios2_core PUBLIC adios2_core_cuda PRIVATE CUDA::cudart)
```
With this change from your PR I was able to compile ADIOS2 on Summit for both Kokkos and CUDA backends, I'll propagate your updates to the other PR.

Do you want to deal with setting up the CI in a different PR?

vicentebolea:
@anagainaru I added a level of indirection in the adiosGPUfunction.h files to hedge device code with host code. This was needed since the workaround of changing the linking dependency to `adios_core_cuda/kokkos` from PRIVATE to PUBLIC in the adios_core made every file linking  adios_core lib to being pseudo-setup as a CUDA source code, we do not want this since ADIOS2 code is not fully CUDA compiler compatible.

However in order to change from PUBLIC to PRIVATE it means that we can only call functions of `adios_core_cuda/kokkos` from cpp files of `adios_core` not header. Since we call those functions in my many headers, I added generic functions in the `adiosGPUfunctions` which achieve compilation time polymorphism.

anagainaru:
@vicentebolea thanks, I'll take a look this weekend and update the other PR. 

Related to the CIs, shouldn't we set the CUDA architecture in CMAKE?
Also ADIOS with Kokkos requires the CXX compiler to point to the `nvcc_wrapper` used to compile Kokkos. This is what I used on Summit:
```cmake
cmake -D CMAKE_CXX_STANDARD=17 -D CMAKE_CXX_EXTENSIONS=OFF  -DCMAKE_C_COMPILER=gcc -DADIOS2_USE_Kokkos=ON  -DKokkos_ROOT=${Kokoks_dir}/install -D CMAKE_CUDA_ARCHITECTURES=70 -DCMAKE_CXX_COMPILER=${Kokoks_dir}/bin/nvcc_wrapper ..
```

vicentebolea:
@anagainaru I think I got it in a good state, at least the CI is passing, there is a cuda warning that is due to the docker image not providing libcuda.so. I will fix this in another PR since regenerating the CI images takes a while

anagainaru:
Closed by #3446

PR #3482, opened 2023-02-10
Make adios_iotest work with SST again: add CurrentStep() to SstWriter…

pnorbert:
… and initialize multithreaded MPI in adios_iotest.


vicentebolea (approved):

vicentebolea:
We need to backport this to master

PR #3485, opened 2023-02-14
Add kokkos build

vicentebolea:


vicentebolea:
@anagainaru I organized the commits from #3446 here. Feel free to `git reset --hard` to the tip commit of this PR and then you can split the commits as you see fit and continue development if needed

anagainaru:
Great, thanks @vicentebolea 
I got distracted by proposals today, but will finish this tomorrow.

source/adios2/CMakeLists.txt
```diff
@@ -120,10 +121,38 @@ if(ADIOS2_HAVE_CUDA)
...
+    OUTPUT_NAME adios2${ADIOS2_LIBRARY_SUFFIX}_core_kokkos
+    )
+
+  if(Kokkos_ENABLE_CUDA)
+    set_property(SOURCE helper/adiosKokkos.cpp PROPERTY LANGUAGE CUDA)
+    set_property(SOURCE helper/adiosKokkos.cpp APPEND PROPERTY COMPILE_FLAGS "--extended-lambda")
+
+    set_target_properties(adios2_core_kokkos PROPERTIES
+      CUDA_SEPARABLE_COMPILATION ON
+      CUDA_VISIBILITY_PRESET hidden
```
anagainaru, line 145: Is there a reason why we are setting this to hidden? 
Also why is the separate compilation set to ON?

cmake/DetectOptions.cmake
```diff
@@ -180,6 +206,7 @@ if(ADIOS2_USE_CUDA)
     find_package(CUDAToolkit REQUIRED)
   endif()
 endif()
+
 if(CMAKE_CUDA_COMPILER AND CUDAToolkit_FOUND)
```
anagainaru, line 210: In my PR this if is inside the `if(ADIOS2_USE_CUDA`) block which is why the Kokkos CI is failing. However your version is not correct. Right now if I build ADIOS2 with Kokkos both backends are set to true:
```
      GPU_Support: ON
      CUDA      : ON
      Kokkos    : ON
```
This branch does not compile for me on Summit when I use Kokkos. How do you configure Kokkos in the CI, are we sure we are compiling the Kokkos backend?

anagainaru (changes requested):
I tested the version in my PR and it compiles with Kokkos/CUDA on three different systems. But it does not pass the Kokkos CI. I can make it pass if in `cmake/DetectOptions.cmake` I make this changes:
```diff
diff --git a/cmake/DetectOptions.cmake b/cmake/DetectOptions.cmake
index caf454969..2bc2a8277 100644
--- a/cmake/DetectOptions.cmake
+++ b/cmake/DetectOptions.cmake
@@ -185,6 +185,7 @@ if(ADIOS2_USE_Kokkos)
     set(ADIOS2_HAVE_Kokkos TRUE)
     if(Kokkos_ENABLE_CUDA OR Kokkos_ENABLE_HIP OR Kokkos_ENABLE_SYCL)
       if(Kokkos_ENABLE_CUDA)
+        find_package(CUDAToolkit REQUIRED)
         enable_language(CUDA)
       endif()
       if(Kokkos_ENABLE_HIP)
@@ -204,11 +205,12 @@ if(ADIOS2_USE_CUDA)
   else()
     find_package(CUDAToolkit REQUIRED)
   endif()
-  if(CMAKE_CUDA_COMPILER AND CUDAToolkit_FOUND)
-    enable_language(CUDA)
-    set(ADIOS2_HAVE_CUDA TRUE)
-    set(ADIOS2_HAVE_GPU_Support TRUE)
-  endif()
+endif()
+
+if(CMAKE_CUDA_COMPILER AND CUDAToolkit_FOUND)
+  enable_language(CUDA)
+  set(ADIOS2_HAVE_CUDA TRUE)
+  set(ADIOS2_HAVE_GPU_Support TRUE)
 endif()
```

but then the code will not compile on Summit when I use Kokkos.

cmake/DetectOptions.cmake
vicentebolea: @anagainaru you are indeed right, yep in the current state it enables both of the backend as shown in the CI here:

https://open.cdash.org/build/8476271/configure

vicentebolea: For some strange reason the CI passes. I will edit this

anagainaru:
Closed by #3446

PR #3487, opened 2023-02-17
Fix Matlab build which depends on the C bindings not the C++ bindings.

pnorbert:
Needs to merged to 2.9 as well


vicentebolea:
> Needs to merged to 2.9 as well

@pnorbert can you then change the base branch to release_29 and rebase onto the commit 2.9.0-rc1?

PR #3495, opened 2023-02-21
Update the documentation for the GPU backend

anagainaru:
Adding info about the Kokkos backend and on how to send Kokkos::Views directly with Put/Get


vicentebolea (approved):
Looks good :)

PR #3497, opened 2023-02-21
kokkos,cmake: toggle building kokkosview example

vicentebolea:
This is from a a discussion in the PR #3446.

Currently and prior to #3446 , when we build adios with examples it will attempt to build the Kokkos examples (with a find_package(Kokkos). This is problematic since:

It will load kokkos package and build the example even if we do `ADIOS_USE_Kokkos=OFF`. Thus a user has no way to disable the kokkos dependency even if he has an incompatible Kokkos or a broken Kokkos installation. This is against our idea of Kokkos dependency to be optional.

This PR introduces a new option which toggles the build of this example. If `ADIOS_USE_Kokkos=ON` it will be on, if `ADIOS_USE_Kokkos=OFF`, the user can manually `ADIOS2_BUILD_EXAMPLE_KOKKOSVIEW=ON` to build the desired example. Note that similar things are already used for other examples.


anagainaru:
We can discuss during the meeting today but leaving this here: users have the option of using `CMAKE_DISABLE_FIND_PACKAGE_Kokkos=OFF` to disable the example. 

The CUDA example is still added even if `ADIOS2_HAVE_CUDA=OFF`

vicentebolea:
> The CUDA example is still added even if `ADIOS2_HAVE_CUDA=OFF`

That is true, however, the CUDA example is different it does not find_package a new thing and it is enable only if Kokkos_CUDA is enabled because when that happens we do a enable_language(CUDA) anyways.

vicentebolea:
> `CMAKE_DISABLE_FIND_PACKAGE_Kokkos=OFF`

To link from the previous discussion, the problem with this is that: is it unreasonable for the user to figure out this internal find_package flag, we will have to document this to give the user a hint and that its okay but it still does not fix the issue since it will still by default force the kokkos dependency.

PR #3499, opened 2023-02-21
Fix plugin engine example

### LINKED ISSUE #3492: Example Plugin Engine: [ADIOS2 EXCEPTION] <Helper> <adiosType> <CheckForNullPtr> : found null pointer for variable in call to Engine::Get
**Describe the bug**
On top of mater on github repository, compiled from source code on Debian Linux.
examplePluginEngine_write  succeeds
examplePluginEngine_read fails with:

variable does not exist
Engine "TestPlugin" destroyed without a prior Close().
This may have negative consequences.
Invalid argument exception, STOPPING PROGRAM
[Mon Feb 20 16:14:26 2023] [ADIOS2 EXCEPTION] <Helper> <adiosType> <CheckForNullPtr> : found null pointer for variable in call to Engine::Get

**To Reproduce**
A minimal reproducible example is preferred.
Or Steps to reproduce the behavior:
1. Built latest git master branch
2. Run examplePluginEngine_read
3. Run examplePluginEngine_write
4. See error on terminal

**Expected behavior**
Success on terminal:
data was read correctly!

**Desktop (please complete the following information):**
 - OS/Platform: Debian 11.6
 - Build 
ADIOS2 build configuration:
  ADIOS Version: 2.9.0
  C++ Compiler : GNU 10.2.1 
    /usr/bin/c++

  Fortran Compiler : GNU 10.2.1 
    /usr/bin/f95

  Installation prefix: /media/src/york.ac.uk/usr
        bin: bin
        lib: lib
    include: include
      cmake: lib/cmake/adios2
     python: lib/python3.9/site-packages

  Features:
    Library Type: shared
    Build Type:   Release
    Testing: ON
    Examples: ON
    Build Options:
      BP5       : ON
      DataMan   : OFF
      DataSpaces: OFF
      HDF5      : ON
      HDF5_VOL  : OFF
      MHS       : ON
      SST       : ON
      CUDA      : OFF
      Fortran   : ON
      MPI       : ON
      Python    : OFF
      Blosc2    : OFF
      BZip2     : ON
      LIBPRESSIO: OFF
      MGARD     : OFF
      PNG       : OFF
      SZ        : OFF
      ZFP       : OFF
      DAOS      : OFF
      IME       : OFF
      O_DIRECT  : ON
      Sodium    : OFF
      Catalyst  : OFF
      SysVShMem : ON
      UCX       : OFF
      ZeroMQ    : OFF
      Profiling : ON
      Endian_Reverse: OFF
      GPU_Support: OFF
      AWSSDK    : OFF
    RDMA Transport for Staging: Unconfigured
-- Configuring done

**Additional context**
None

**Following up**

caitlinross:
Fixes #3492. For some reason a `PerformPuts()` was removed that causes the example to break. The example is used in one of the install tests, but it wasn't returning failure states when it happens so that's why we missed in CI. That's been fixed as well. It also turns out that that has been failing for awhile due to not being able to find the plugin libs, so that should be fixed now too.

### DIFF
--- a/examples/plugins/operator/examplePluginOperator_write.cpp
+++ b/examples/plugins/operator/examplePluginOperator_write.cpp
@@ -41,6 +41,7 @@ int main(int argc, char *argv[])
     };
     const std::size_t Nx = myDoubles.size();
 
+    bool success = false;
     try
     {
         /** ADIOS class factory of IO class objects */
@@ -73,6 +74,7 @@ int main(int argc, char *argv[])
 
         /** Engine becomes unreachable after this*/
         writer.Close();
+        success = true;
     }
     catch (std::invalid_argument &e)
     {
@@ -90,5 +92,5 @@ int main(int argc, char *argv[])
         std::cout << e.what() << "\n";
     }
 
-    return 0;
+    return success ? EXIT_SUCCESS : EXIT_FAILURE;
 }
--- a/docs/user_guide/source/advanced/plugins.rst
+++ b/docs/user_guide/source/advanced/plugins.rst
@@ -131,6 +131,35 @@ If ``ADIOS2_PLUGIN_PATH`` is not set, and a path is not specified when loading y
 .. note::
     The ``ADIOS2_PLUGIN_PATH`` environment variable can contain multiple paths, which must be separated with a ``:``.
 
+
+When building on Windows, you will likely need to explicitly export the Create and Destroy symbols for your plugin, as symbols are invisible by default on Windows.
+To do this in a portable way across platforms, you can add something similar to the following lines to your CMakeLists.txt:
+
+.. code-block:: cmake
+
+    include(GenerateExportHeader)
+    generate_export_header(PluginEngineWrite BASE_NAME plugin_engine_write)
+    target_include_directories(PluginEngineWrite PUBLIC
+        $<BUILD_INTERFACE:${CMAKE_CURRENT_BINARY_DIR}>
+        $<INSTALL_INTERFACE:include>)
+
+
+Then in your plugin header, you'll need to ``#include "plugin_engine_write_export.h"``. Then edit your function defintions as follows:
+
+.. code-block:: c++
+
+    extern "C" {
+
+    PLUGIN_ENGINE_WRITE_EXPORT adios2::plugin::ExampleWritePlugin *
+        EngineCreate(adios2::core::IO &io, const std::string &name,
+                 const adios2::Mode mode, adios2::helper::Comm comm);
+
+    PLUGIN_ENGINE_WRITE_EXPORT void
+        EngineDestroy(adios2::plugin::ExampleWritePlugin * obj);
+
+    }
+
+
 ***********************************
 Using Your Plugin in an Application
 ***********************************
--- a/examples/plugins/engine/CMakeLists.txt
+++ b/examples/plugins/engine/CMakeLists.txt
@@ -3,15 +3,26 @@
 # accompanying file Copyright.txt for details.
 #------------------------------------------------------------------------------#
 
+include(GenerateExportHeader)
 add_library(PluginEngineWrite
   ExampleWritePlugin.cpp
 )
 target_link_libraries(PluginEngineWrite adios2::cxx11 adios2_core)
+generate_export_header(PluginEngineWrite BASE_NAME plugin_engine_write)
+target_include_directories(PluginEngineWrite PUBLIC
+  $<BUILD_INTERFACE:${CMAKE_CURRENT_BINARY_DIR}>
+  $<INSTALL_INTERFACE:include>
+)
 
 add_library(PluginEngineRead
   ExampleReadPlugin.cpp
 )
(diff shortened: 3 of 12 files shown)


caitlinross:
@vicentebolea will you review? Also not sure how to fix the failure on windows in this case. 

```
 723: 1: [Tue Feb 21 22:50:32 2023] [ADIOS2 WARNING] <Plugins> <PluginManager> <LoadPlugin> : The plugin PluginEngineWrite could not be loaded. Double check ADIOS2_PLUGIN_PATH or plugin library name is correct.
 723: 1: ADIOS2_PLUGIN_PATH: D:/a/ADIOS2/ADIOS2/win2019-vs2019-mpi/Testing/install/cmake/EnginePlugin
```

In `testing/install/EnginePlugin/CMakeLists.txt` I set `ADIOS2_PLUGIN_PATH` environment variable.  I'm assuming this doesn't work on windows because of the slashes, but I set it using `${CMAKE_CURRENT_BINARY_DIR}` as opposed to something hardcoded, so I don't understand why CMake wouldn't provide a valid windows path in that case. Any ideas on how to fix?

vicentebolea:
> @vicentebolea will you review? Also not sure how to fix the failure on windows in this case.
> 
> ```
>  723: 1: [Tue Feb 21 22:50:32 2023] [ADIOS2 WARNING] <Plugins> <PluginManager> <LoadPlugin> : The plugin PluginEngineWrite could not be loaded. Double check ADIOS2_PLUGIN_PATH or plugin library name is correct.
>  723: 1: ADIOS2_PLUGIN_PATH: D:/a/ADIOS2/ADIOS2/win2019-vs2019-mpi/Testing/install/cmake/EnginePlugin
> ```
> 
> In `testing/install/EnginePlugin/CMakeLists.txt` I set `ADIOS2_PLUGIN_PATH` environment variable. I'm assuming this doesn't work on windows because of the slashes, but I set it using `${CMAKE_CURRENT_BINARY_DIR}` as opposed to something hardcoded, so I don't understand why CMake wouldn't provide a valid windows path in that case. Any ideas on how to fix?

I think that Cmake is ok with forward slashes even in Windows as long as it is used internally. But that ENV var will be read by a Powershell script or the C/C++ stdlib in Windows which will not accept that path. 

Maybe this cmake function will help to transform the path to _native_, I found it in the file cmake module:

```
file(TO_NATIVE_PATH "<path>" <variable>)
```

vicentebolea (commented):
Oops I wrote the comments but somehow forgot to press the submit review. Ill have a look to see what can be done for the windows failures

examples/plugins/operator/examplePluginOperator_write.cpp
```diff
@@ -90,5 +92,5 @@ int main(int argc, char *argv[])
         std::cout << e.what() << "\n";
```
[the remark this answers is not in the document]
caitlinross: I'll just leave it as is. It's just example code (that also gets used in a couple of tests).

Thanks on the cmake function and explanation. That looks like what I need to use!

caitlinross:
@vicentebolea So that cmake command appears to have worked, but it's not finding the plugin libs in that directory. Do the built libraries get put in a different directory on Windows? I don't have a Windows system to test this out myself (and haven't really used Windows in 10+ years 😅 ).

vicentebolea:
@caitlinross I made some changes to the PR; it appears to be something with the dynamicBinary loading dll files. Not sure, lets see the CI once it finishes

vicentebolea:
@caitlinross after some changes finally the DynamicBinder can find the plugin file, however, for some reason there is an exception while loading it. Let me know if you need more help after your return from vacations  :)

source/adios2/helper/adiosDynamicBinder.cpp
```diff
@@ -46,7 +46,6 @@ DynamicBinder::DynamicBinder(std::string libName, std::string libPath)
 #endif
 
     std::vector<std::string> libSuffixes;
-    libSuffixes.emplace_back("");
```
caitlinross, line 49: So this shouldn't be removed. It's supposed to be possible that you could tell ADIOS the full library name (e.g. `libMyPlugin.so` instead of just `MyPlugin`), so without this line ADIOS would fail in this situation.

vicentebolea: Oh I see, that makes sense

caitlinross:
@vicentebolea It looks like I have the windows error fixed now. I also cleaned up the fixup commits and added some notes to the plugin docs about needing to export symbols. So I think it's ready unless you have any other changes?

caitlinross:
Actually one more question, do you want this in release?

vicentebolea:
> Actually one more question, do you want this in release?

Yes that's a good idea.

vicentebolea (dismissed):

caitlinross:
@vicentebolea since I changed the target branch to release, I need you to approve again.

vicentebolea (approved):

PR #3500, opened 2023-02-21
ci,kokkos,cuda: remove warnings

vicentebolea:
Followup #3446 

We missed this change, its to set the `CMAKE_CUDA_ARCHITECTURES` when using Kokkos_CUDA

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -169,7 +169,7 @@ adios_option(Catalyst   "Enable support for in situ visualization plugin using P
 adios_option(AWSSDK     "Enable support for S3 compatible storage using AWS SDK's S3 module" AUTO)
 include(${PROJECT_SOURCE_DIR}/cmake/DetectOptions.cmake)
 
-if(ADIOS2_HAVE_CUDA)
+if(ADIOS2_HAVE_CUDA OR ADIOS2_HAVE_Kokkos_CUDA)
     set(CMAKE_CUDA_STANDARD 11)
     set(CMAKE_CUDA_STANDARD_REQUIRED TRUE)
     if(NOT DEFINED CMAKE_CUDA_ARCHITECTURES)
--- a/docs/user_guide/source/advanced/gpu_aware.rst
+++ b/docs/user_guide/source/advanced/gpu_aware.rst
@@ -37,7 +37,7 @@ When building ADIOS2 with CUDA enabled, the user is responsible with setting the
 Building with Kokkos enabled
 --------------------------
 
-The Kokkos library can be used to enable GPU within ADIOS2. Based on how Kokkos is build, either the CUDA or HIP backend will be enabled. Building with Kokkos requires ``-DADIOS2_USE_Kokkos=ON``. The user is not responsible to set the ``CMAKE_CUDA_ARCHITECTURES`` since it will be detected from the Kokkos installation.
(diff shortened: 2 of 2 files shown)


CMakeLists.txt
```diff
@@ -169,7 +169,7 @@ adios_option(Catalyst   "Enable support for in situ visualization plugin using P
 adios_option(AWSSDK     "Enable support for S3 compatible storage using AWS SDK's S3 module" AUTO)
 include(${PROJECT_SOURCE_DIR}/cmake/DetectOptions.cmake)
 
-if(ADIOS2_HAVE_CUDA)
+if(ADIOS2_HAVE_CUDA OR ADIOS2_HAVE_Kokkos_CUDA)
```
anagainaru, line 172: When we link to Kokkos the architecture should be set to whatever was used for Kokkos. We should not try to set another one.

anagainaru (commented):
What is the warning you are seeing? There should not be any.

CMakeLists.txt
vicentebolea: Kokkos does not set the architecture for the whole project but only to the cmake targets which `links` depends on it. the enable_language(CUDA) Will still need this variable.

The warnings are in current master/release_29 as shown here:

https://open.cdash.org/index.php?project=ADIOS&filtercount=1&showfilters=1&field1=revision&compare1=63&value1=a9ad351ea9e870ffade25ae2f76059a5b0d615a3

anagainaru: You are right, I forgot we are treating the them as CUDA files. In this case, I think we should wait for this:
https://github.com/kokkos/kokkos/issues/5891

We do not want to risk setting another architecture than what was used for building Kokkos.

vicentebolea: I think that we should set for the whole adios2 project the same arch as we use for ADIOS2 CXX: C++11. For the particular components that use Kokkos we already set them to what kokkos use.

I am against having these warnings until kokkos team fixes this, if they do fix it.

vicentebolea: > For the particular components that use Kokkos we already set them to what kokkos use.

Hmm, We do this by manually setting it to C++17 but I am not sure if `kokkos_compilation()` or `target_links_library(kokkos::kokkos)` does set the CUDA_ARCHITECTURE automatically.

vicentebolea: Kokkos does seems to inject the -arch=XXYYZZ in the `INTERFACE_COMPILE_OPTIONS` `INTERFACE_LINK_OPTIONS` but the target for CUDA_ARCHITECTURE remains unset.

We need this to both set the default CMAKE_CUDA_ARCHITECTURE which is needed for the CI and also affect the cuda  example. And in the targets that use kokkos backend we need to  set it to whatever Kokkos give us.

Same thing for CMAKE_CUDA_STANDARD, however, kokkos already propagates this when you link your cmake target to kokkos::kokkos which for 3.7 seems to default to c++14.

vicentebolea: An also, we might want to use a different `CUDA_ARCHITECTURE` than kokkos, what if you building ADIOS2 with the kokkos RPM or DEB pkg. In that case the package would have been configured with a old cuda arch but you might want to use a new cuda arch.

anagainaru: Let's do the following, leave it like this for the release and I will change it once Kokkos exposes the architecture. Acceptable?

Side note. If the Kokkos package was configured with an old cuda arch, our code will not work. I'll make a note in the documentation that users should set the `CUDA_ARCHITECTURE` to the same used for configuring Kokkos.

anagainaru (dismissed):

CMakeLists.txt
vicentebolea: > Let's do the following, leave it like this for the release and I will change it once Kokkos exposes the architecture. Acceptable?

Sounds good, if kokkos address it, we can comeback to this.

PR #3514, opened 2023-02-28
Move the buffer copy logic outside operators

anagainaru:
If an operator (direct or inverse) is not applied (due to a threshold or some other reason) the `Operate` or `Decompress` functions are now returning 0 and are not responsible with copying the data to the ADIOS2 internal buffers.

This allows ADIOS2 to use the appropriate function for copying the data (taking into consideration the memory space).

Operators that use this: Blosc, MGARD, MGARDPlus

**Testing** Adding test for MGARD with small GPU buffers. 

*Note:* the original code inside operators was using `memcpy` so I am simply moving the logic outside (I am not checking for reverse endian)

Adding to #3465 to skip copy for blocks with 0 elements

### DIFF
--- a/source/adios2/core/Operator.cpp
+++ b/source/adios2/core/Operator.cpp
@@ -98,6 +98,8 @@ Dims Operator::ConvertDims(const Dims &dimensions, const DataType type,
     return ret;
 }
 
+size_t Operator::GetHeaderSize() const { return 0; }
+
 // PRIVATE
 void Operator::CheckCallbackType(const std::string type) const
 {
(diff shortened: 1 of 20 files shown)


anagainaru:
@JieyangChen7 I use 1024 as the threshold for when to apply MGARD based on my testing on Summit but you might have a better idea of what value to use.

JieyangChen7:
@anagainaru I think 1e6 might be a safer choice based on my experience.

pnorbert (approved):

williamfgc (approved):

anagainaru:
@pnorbert I will merge this after I rebase it over `release_29` and if it doesn't solve Axel's issue I will update his PR with the fix. I think it make more sense like that compared to changing this PR further.

pnorbert (approved):

PR #3515, opened 2023-03-01
Fix for the issue 3503.

dmitry-ganyushin:
Python 1D arrays of size N are presented internally with (1, N) structures. Extra case is added.

### DIFF
--- a/source/adios2/toolkit/format/dataman/DataManSerializer.tcc
+++ b/source/adios2/toolkit/format/dataman/DataManSerializer.tcc
@@ -282,7 +282,11 @@ int DataManSerializer::GetData(T *outputData, const std::string &varName,
             if (j.shape.empty() or (j.shape.size() == 1 and j.shape[0] == 1))
             {
                 std::memcpy(reinterpret_cast<char *>(outputData), input_data,
-                            sizeof(T));
+                            varCount[0] * sizeof(T));
+            }else if (j.shape.size() == 2 and j.shape[0] == 1)
+            {
+                std::memcpy(reinterpret_cast<char *>(outputData), input_data,
+                            varCount[1] * sizeof(T));
             }
         }
     }
--- a/source/adios2/toolkit/format/dataman/DataManSerializer.tcc
+++ b/source/adios2/toolkit/format/dataman/DataManSerializer.tcc
@@ -283,7 +283,8 @@ int DataManSerializer::GetData(T *outputData, const std::string &varName,
             {
                 std::memcpy(reinterpret_cast<char *>(outputData), input_data,
                             varCount[0] * sizeof(T));
-            }else if (j.shape.size() == 2 and j.shape[0] == 1)
+            }
+            else if (j.shape.size() == 2 and j.shape[0] == 1)
             {
                 std::memcpy(reinterpret_cast<char *>(outputData), input_data,
                             varCount[1] * sizeof(T));
--- a/source/adios2/toolkit/format/dataman/DataManSerializer.tcc
+++ b/source/adios2/toolkit/format/dataman/DataManSerializer.tcc
@@ -257,7 +257,7 @@ int DataManSerializer::GetData(T *outputData, const std::string &varName,
                 input_data += j.position;
             }
 
-            if (j.shape.size() > 0 and j.shape[0] > 1 and j.start.size() > 0 and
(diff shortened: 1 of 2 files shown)


eisenhauer (changes requested):
I'm just trying to understand this change, because it looks oddly specific.  I.E. it only applies to two dimensional arrays whose first dimension is one (and it's not at all clear why you change sizeof(T) to varCount[0]*sizeof(T).  Or was that just a debugging thing?  

In whole, I don't think this is a well-formed fix.  It looks like in the original code the if statement above, which ends up calling NdCopy if true, seems to be designed to handle everything except single elements.  That is, all single and multidimensional array selections, etc.   Then the next if (that you've changed here) handles one-element things.  I really wonder why there wasn't an "else" used here, because these conditionals aren't guaranteed to cover all cases together and if both fail, we simply don't move data and that's probably not right.   Your change here just adds a fallback that didn't exist in the original where if you happen to get a 2d array with the first dimension being 1 you do a simple memcpy.  Why is this the right fix?  Does NdCopy not work in this case?  If so, we need to know that because we use that elsewhere, and the fix should go there.  (Don't get me wrong, that there are two indepenent conditionals here and we might fail them both and not move data is probably itself a problem that should be fixed.  But that fix should be general, not customized to the problem reported.)  Also, when we fix bugs like this we should be adding testing that exposes the bugs and passes after they are fixed.

anagainaru:
After adding the testing required by Greg, please clean the history before we merge this.

eisenhauer:
Just opining in general here.  I'd love to see the two dozen or so lines of code that fill the read request to evolve to more of the form:
if (/* shape is global single value*/) {
     // simple memcpy of sizeof(T)
} else if (/* some conditional that selects other known working cases of N dimensional arrays */) {
    // do the NdCopy
} else {
   assert and unhandled case in Get and die
}

Now it may be that there really isn't a conditional necessary for the 1st else and everything should fall into it.  In that case we don't need the 2nd else.  But the two independent if's where something might fall through without being handled just seems wrong.  

It also seems wrong to be checking to see if all the elements (shape, count, start, etc) all have the same dimensionality.   Surely if we get this deep in serving a read request, surely we've sorted out these basics before now, haven't we?  I guess the circumstance where maybe we haven't is if what is written was a local array (and then (j.start.size() == 0)?)  Which maybe isn't supported at all by dataman, in which case we really should be throwing a shoe if we see this.  ( @pnorbert  And particularly given that we are adding a new Shape, JoinedArrays, we should have a table in the documentation that identifies what engines support what shapes. )

eisenhauer:
We're getting there.   A few more things to do.  First, we want this in the 2.9 release, so you'll have to rebase on upstream/release_29 and submit the PR to be merged into that branch.  (You may have to close and resubmit the PR for this.)  Second, I think there are some issues with the test.  It probably shouldn't be in testing/adios2/bindings/python, but instead should be in testing/adios2/engine/dataman, since it is a dataman specific test (and not a generic python bindings test).  Also, it isn't actually being run as part of CI the way you've done it.  You need to add it to the CMakeLists.txt so that it gets executed.   Lastly, I'm not entirely certain that this test exercises the code in the way that would have exposed the bug.  The problem happened with a 2-D array in which the first dimension was 1.   The reader side of this test does: `self.assertTrue(bufshape[0] == self.Nx)`, which sure looks like it's getting a 1-D array.  That's not where the problem was.  Make sure you've got a test that fails if you run it without your fix and succeeds if your fix is in place.

dmitry-ganyushin:
Actually, the original problem was for 1D numpy arrays. They are presented
internally in ADIOS as 1xN arrays.
I left the first "if" for arrays because they are more expected to
receive than single values. The last "else" should contain only logging. I
don't think we should kill the server because it got wrong data.
At least this is not common.
I am not sure if logging is fully available, although some logging was
implemented. I can put an exception.

eisenhauer:
Hmm.  Didn't your first fix contain this code:  `if (j.shape.size() == 2 and j.shape[0] == 1)`?   Or do 1D python arrays come into C++ as 2D arrays somehow?  I'm just trying to get some assurance that this is all consistent, fixes the original problem, etc...

WRT logging, no, we don't have a logging mechanism.  But this isn't "killing the server because it got the wrong data", it's that this subroutine is being called in a circumstance where it doesn't work.  An assert(FALSE) in that case would be justifiable.  Really, that's what assertions are for, and they are preferred over the silent failure you're trying to fix.

dmitry-ganyushin:
Closed in favor of PR#3519

PR #3516, opened 2023-03-01
Hdf5 vol support will require latest hdf5 1.14

guj:


source/CMakeLists.txt
```diff
@@ -8,14 +8,16 @@ add_subdirectory(utils)
 
 # HDF5 VOL requires 1.13+
 if(ADIOS2_HAVE_HDF5)
-  if(HDF5_VERSION VERSION_LESS 1.13)
+  if(HDF5_VERSION VERSION_LESS 1.14)
```
vicentebolea, line 11: This version conditional should be done in `cmake/DetectOptions.cmake:308`

guj: Oh, as discussed yesterday in the call, our VOL support will require the most  recent hdf5 version. 

Other HDF5 related features can work with older versions. 

That's why I put a version check here.

vicentebolea: ohh that is right

source/h5vol/H5Vol_dataset.c
```diff
@@ -85,20 +85,30 @@ void *H5VL_adios2_dataset_open(void *obj, const H5VL_loc_params_t *loc_params,
...
 
-    var->m_HyperSlabID = file_space_id;
-    var->m_MemSpaceID = mem_space_id;
+        var->m_HyperSlabID = file_space_id_array[i];
+        var->m_MemSpaceID = mem_space_id_array[i];
 
-    var->m_Data = buf;
-    return gADIOS2ReadVar(var);
+        var->m_Data = buf_array[i];
+        gADIOS2ReadVar(var);
```
vicentebolea, line 109:
```suggestion
        herr_t ret = ADIOS2ReadVar(var);
        
        // A herr_t variable is considered successful only when > 0
        if (ret < 0)
        {
           return ret; 
        }
```

source/h5vol/H5Vol_dataset.c
```diff
@@ -85,20 +85,30 @@ void *H5VL_adios2_dataset_open(void *obj, const H5VL_loc_params_t *loc_params,
...
-                                hid_t plist_id, void *buf, void **req)
+herr_t H5VL_adios2_dataset_read(size_t count, void *dset_array[],
+                                hid_t mem_type_id_array[],
+                                hid_t mem_space_id_array[],
+                                hid_t file_space_id_array[], hid_t dxpl_id,
+                                void *buf_array[],
+                                void **req) // last parameter is not unused
+                                            // void *dset, hid_t mem_type_id,
+// hid_t mem_space_id, hid_t file_space_id,
+// hid_t plist_id, void *buf, void **req)
```
vicentebolea, line 96:
```suggestion
```

source/h5vol/H5Vol_dataset.c
```diff
@@ -130,30 +140,40 @@ herr_t H5VL_adios2_dataset_get(void *dset, H5VL_dataset_get_args_t *args,
...
 
+herr_t H5VL_adios2_dataset_write(size_t count, void *dset_array[],
+                                 hid_t mem_type_id_array[],
+                                 hid_t mem_space_id_array[],
+                                 hid_t file_space_id_array[], hid_t dxpl_id,
+                                 const void *buf_array[], void **req)
+/*
 herr_t H5VL_adios2_dataset_write(void *dset, hid_t mem_type_id,
                                  hid_t mem_space_id, hid_t file_space_id,
                                  hid_t plist_id, const void *buf, void **req)
```
vicentebolea, line 151:
```suggestion
```

source/h5vol/H5Vol_dataset.c
```diff
@@ -130,30 +140,40 @@ herr_t H5VL_adios2_dataset_get(void *dset, H5VL_dataset_get_args_t *args,
...
-    H5VL_VarDef_t *varDef = (H5VL_VarDef_t *)(vol->m_ObjPtr);
+    for (size_t i = 0; i < count; i++)
+    {
+        REQUIRE_NOT_NULL_ERR(dset_array[0], -1);
+        H5VL_ObjDef_t *vol = (H5VL_ObjDef_t *)(dset_array[0]);
+        H5VL_VarDef_t *varDef = (H5VL_VarDef_t *)(vol->m_ObjPtr);
 
-    // H5VL_VarDef_t *varDef = (H5VL_VarDef_t *)dset;
-    varDef->m_Data = (void *)buf;
+        // H5VL_VarDef_t *varDef = (H5VL_VarDef_t *)dset;
```
vicentebolea, line 160:
```suggestion
```

source/h5vol/H5Vol_dataset.c
vicentebolea: A note for the future: according to hdf5 docs: https://docs.hdfgroup.org/hdf5/develop/_h5public_8h.html#title52 . A herr_t represents failure when <= 0, but gADIOS2ReadVar considers == 0 success. No something to change in this MR, but it will be good to address it in the future either getting rid of using herr_t in our codebase or following the hdf5 usage of it.

vicentebolea (changes requested):
Looks good; few comments removals; forwarding return code;

source/h5vol/H5Vol_dataset.c
guj: Thanks for the suggestions. I removed the comments etc and adjusted return code. To clarify, the definition of success in hdf5 is 0 as defined in H5private.h.

vicentebolea (approved):
Looks good :)

guj:
> Looks good :)

Thanks.  Not sure why merge is blocked though.

vicentebolea:
> > Looks good :)
> 
> Thanks. Not sure why merge is blocked though.

It seems to be waiting of CI to pass

PR #3517, opened 2023-03-02
Set the CMAKE GPU architectures to what is used for Kokkos

anagainaru:
Using the newly introduced `Kokkos_CUDA_ARCHITECTURES` and `Kokkos_HIP_ARCHITECTURES`

If the user sets the `-D CMAKE_CUDA_ARCHITECTURES` to something else, the user defined variable will be used (the same for HIP) so we cover the case when the used intends to use a different architecture.


vicentebolea (approved):
Looks great!

vicentebolea:
@anagainaru how about including this for the 2.9.0 rather than 2.9.1.

PR #3519, opened 2023-03-06
Fix for the issue 3503. 

dmitry-ganyushin:
Python 1D arrays of size N are presented internally as C++ (1, N) structures. That made dataman serializer failed by reading.
Clean-up for PR #3515


eisenhauer:
So, riddle me this:  You say that 1-D arrays in python show up in ADIOS as 2-D arrays with the first dimension being 1.  I'm willing to buy that.  But then in the test you have code on the reading side that looks like this:
`                bufshape = recvar.Shape()`
`                self.assertTrue(bufshape[0] == self.Nx)`
That is, you've asked ADIOS for the dimensions of the array, and you assertTrue that the first dimension is not 1, but is instead self.Nx.  Even if somehow Python presents 1D numpy arrays as 2-D to ADIOS, I'm pretty sure that there's no mechanism in the read side of the ADIOS python bindings for that "extra" first dimension with value 1 to disappear.  That is, if ADIOS C++ on the writer side is really seeing an array with shape [1, self.Nx] in the Put(), I don't think there's any way that the recvar.Shape() on the reader side would return anything other than [1, self.Nx].  This makes me wonder if you're successfully recreating the original problem.  (Maybe I'm wrong, but if so I'd love to hear how that extra dimension disappeared.)

dmitry-ganyushin:
The original problem was reported for 1D numpy arrays presented as (1, N) shape. That is accepted at least for BP4 engine but did not work for DataMan. Added tests for both (N) and (1,N) shapes.

testing/adios2/engine/dataman/CMakeLists.txt
```diff
@@ -17,6 +17,9 @@ foreach(tst IN ITEMS
   )
 endforeach()
 
+python_add_test(NAME Test.Engine.DataMan1D.Serial SCRIPT TestDataMan1D.py)
+python_add_test(NAME Test.Engine.DataMan1xN.Serial SCRIPT TestDataMan1xN.py)
+
```
caitlinross, line 22: I think you just need to put these lines in a `if(ADIOS2_HAVE_Python)` condition. If you look at the cdash output, you'll see that python isn't turned on in the nvhpc build: https://open.cdash.org/build/8527861

source/adios2/toolkit/format/dataman/DataManSerializer.tcc
```diff
@@ -257,8 +257,8 @@ int DataManSerializer::GetData(T *outputData, const std::string &varName,
                 input_data += j.position;
             }
 
-            if (j.shape.size() > 0 and j.shape[0] > 1 and j.start.size() > 0 and
-                j.start.size() == j.count.size() and
+            if (j.shape.size() > 0 and (j.shape[0] > 1 or j.shape[1] > 1) and
```
eisenhauer, line 260: Correct me if I'm wrong, but if j.shape.size() == 1, won't a hard-coded reference to j.shape[1] be a reference to unallocated memory?  That seems like a bad idea...

dmitry-ganyushin: If ```j.shape.size() == 1```  and ```j.shape[0] > 1``` the second condition ```j.shape[1] > 1``` will not be checked by a compiler. The only problem here could be single values presented as arrays like [1]

dmitry-ganyushin: ... but it goes to memcopy branch

eisenhauer: And if j.shape.size() == 1 and  j.shape[0] == 1 the second condition will be checked and will result in accessing unallocated memory, right?    But I think you're missing the point.  This code was structured badly before this PR, resulting in a silent failure to move data.  That your proposed changes are predicated on checking the first two dimensions of an N-dimensional array is a pretty solid indication that this isn't a robust fix.  How can you write this so that it's actually robust?

dmitry-ganyushin: If there should be a wrong memory access, I am wondering why CI is not showing it?

eisenhauer: Unfortunately, msan is particularly computationally expensive and if we were to run all our tests under it, they would take several hours to complete. Our tier of CI times out after an hour, so we can run only a small subset of tests in the sanitizers.  Fortunately, this case is obvious from inspection, but I do sometimes compile with msan or use valgrind to verify things that don't run in CI.

eisenhauer:
I think the code looks good now, but as per the conversation in the dev meeting, can we get this rebased on release_29 and the PR targeted to that branch too?

vicentebolea:
@dmitry-ganyushin you do not need to open new Pull-requests to have this also in release_29. You can change this PR to _target_ release_29, and once merged I can do the _backport_ onto master.

Note that you will then have to rebase onto _release_29_ branch, the git command will be: `git fetch origin; git rebase origin/release_29`

eisenhauer (commented):
My comments are addressed.  I'll let Vicente do final approval.

vicentebolea:
> My comments are addressed. I'll let Vicente do final approval.

The latest rebase brings master only changes here. I will help Dmitry tomorrow to reorganize the commits. OTher than that it looks good to me

eisenhauer (approved):

PR #3522, opened 2023-03-08
Add adios2_mode_readRandomAccess to fortran bindings

eisenhauer:
Resubmitting to the release branch.   Note that there is fragility here because the numerical values in the fortran bindings must match the numerical values in the C bindings and we have nothing that guarantees that.  Also, there are modes for different things mixed in the different lists.  Sigh.


pnorbert (approved):

vicentebolea:
@pnorbert do we need to backport this change to master?

eisenhauer:
Yes.

Issue #3503, opened 2023-02-23
Python bindings - Bug sending 1 dimensional arrays


stefaniereuter:
**Describe the bug**
Using the dataman engine to send different numpy arrays, I encountered several different behaviours:
1. If the  array is multidimensional everything seems to work
2. If trying to send an one dimensional array( e.g. [1 2 ] , the data I receive is random zeros [4.9e-324 9.9e-323]
3. Trying to reshape the one dimensional array into a multidimensional array before sending results in two different behaviours. 
- If the original array is a random array (random.rand), I receive a correct result after reshaping
- If the array is an arange array (e.g.` np.arange(1,20`))and I try to reshape I still get random zeros
4. If I try to send an one dimensional random array (`np.random.rand(1,20)`) I don't receive any data and the values are still the pre initialized data values

**To Reproduce**
```Python
 
import numpy as np
from multiprocessing import Process, Pipe
import logging
import adios2

def thread_send(name:str):

    data = name.recv() #receive data from main thread
    shape = data.shape
    count = shape
    start = (0,)*len(shape)
    print(f"data on sender side \n{data!s}")

    adios_io = adios2.ADIOS()
    wan = adios_io.DeclareIO("Server")
    wan.SetEngine("Dataman")

    wan.SetParameters(
        {
            "IPAddress": "0.0.0.0",
            "Port": "12306",
            "Timeout": "5",
            "TransportMode": "reliable",
            "RendezvousReaderCount":"1",
        }
    )
    logging.info(f"Sender: initiating sending")
    writer = wan.Open("testdata_sender", adios2.Mode.Write)
    sendbuffer = wan.DefineVariable("np_data",data, shape, start, count, adios2.ConstantDims)
    if sendbuffer:
        writer.BeginStep()
        writer.Put(sendbuffer,data,adios2.Mode.Deferred)
        writer.EndStep()
    else:
        raise ValueError("DefineVariable failed")
    
    writer.Close()
  
    logging.info(f"Sender: sending finished")

def thread_receive(name:str):

    adios_io = adios2.ADIOS()
    wan = adios_io.DeclareIO("Client")
    wan.SetEngine("Dataman")
    wan.SetParameters(
        {
            "IPAddress": "0.0.0.0",
            "Port": "12306",
            "Timeout": "5",
            "TransportMode": "reliable",
            "RendezvousReaderCount":"1",
        }
    )   
    logging.info(f" Receiver: initiating receiving ")
[...]

PR #3527, opened 2023-03-09
Fix for issue 3503 for release_29 branch

dmitry-ganyushin:
Fix for python 1D arrays of the 1xN shape. Issue #3503

### DIFF
--- a/source/adios2/toolkit/format/dataman/DataManSerializer.tcc
+++ b/source/adios2/toolkit/format/dataman/DataManSerializer.tcc
@@ -200,7 +200,7 @@ int DataManSerializer::GetData(T *outputData, const std::string &varName,
                                const size_t step, const Dims &varMemStart,
                                const Dims &varMemCount)
 {
-    PERFSTUBS_SCOPED_TIMER_FUNC();
+    PERFSTUBS_SCOPED_TIMER_FUNC()
 
     DmvVecPtr vec = nullptr;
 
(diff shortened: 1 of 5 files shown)


source/adios2/toolkit/format/dataman/DataManSerializer.tcc
```diff
@@ -200,7 +200,7 @@ int DataManSerializer::GetData(T *outputData, const std::string &varName,
                                const size_t step, const Dims &varMemStart,
                                const Dims &varMemCount)
 {
-    PERFSTUBS_SCOPED_TIMER_FUNC();
```
eisenhauer, line 203: Grepping through the code, everywhere else the invocation of the macro PERFSTUBS_SCOPED_TIMER_FUNC() has a trailing semicolon.  It's a null definition unless we build with Tau (which we don't in CI), but I worry that this change might break something.  At the very least it's unnecessary.  Revert?

dmitry-ganyushin: I think it is not needed everywhere, but I can restore it.

eisenhauer: Well, it was there for a reason, not a functionality reason but for performance diagnostics.  Those aren't maintained as reliably they they probably should be, but killing an existing monitoring point in an unrelated change is not a step in the right direction...

PR #3530, opened 2023-03-10
H5 subfile support

guj:
If H5 has subfile support, it will be the default to write H5 files. 

Use envs to control subfile parameters through the underlying  HDF5 library: 
(# of subfiles per node) H5FD_SUBFILING_IOC_PER_NODE  
H5FD_SUBFILING_STRIPE_SIZE

etc.


eisenhauer (dismissed):

anagainaru:
@guj Do you mind cleaning the history before merging so if we ever have to look back at this it will be easier to find the commits?

guj:
> @guj Do you mind cleaning the history before merging so if we ever have to look back at this it will be easier to find the commits?

Never done that. What commands do you use?

pnorbert (dismissed):

anagainaru (approved):

PR #3537, opened 2023-03-14
Enable Linux Emulation build (Power8, IBM XL)

vicentebolea:
fixes: #3551 

it removes the smpi build since there is no way of building spectrum mpi with the current base image.


caitlinross (approved):
I'm not super knowledgable about this stuff, but it all seems reasonable to me.

.github/workflows/everything.yml
```diff
@@ -145,7 +145,11 @@ jobs:
 
   linux_emu:
     needs: [format, git_checks]
-    if: false # needs.git_checks.outputs.num_code_changes > 0
+    if: needs.git_checks.outputs.num_code_changes > 0 &&
```
vicentebolea, line 148: @caitlinross this is the only diff. It means just to run in master/release* and when code changes

caitlinross (approved):

Issue #3538, opened 2023-03-15
Dataman additional bug in 1dimensional arrays


stefaniereuter:
Thanks @dmitry-ganyushin  for having a look into the python 1 dimensional arrays. I had a play with your commit and found that part of the problem (#3503) still exists. 
I can now successfully send 1 dimensional float arrays. It seems that there is still a problem if the data type is not np.float64. 

**Describe the bug**
- Sending `numpy.random.rand(3) #works`
- Sending `numpy.random.randint(2,size=3) # receives wrong data`
- Sending `numpy.arange(4) # receives wrong data`
- Sending `numpy.arange(4,dtype = numpy.float64) # works`
- Sending `numpy.arange(4,dtype = numpy.float32) #receives wrong data`
- Sending `numpy.arange(4,dtype = numpy.float128) #receives wrong data`
- Sending `numpy.arange(4,dtype=int) #receives wrong data`

**To Reproduce**
```PYTHON
import numpy as np
from multiprocessing import Process, Pipe
import logging
import adios2

def thread_send(name:str):

    data = name.recv() #receive data from main thread
    shape = data.shape
    count = shape
    start = (0,)*len(shape)
    print(f"data on sender side \n{data!s}")

    adios_io = adios2.ADIOS()
    wan = adios_io.DeclareIO("Server")
    wan.SetEngine("Dataman")

    wan.SetParameters(
        {
            "IPAddress": "0.0.0.0",
            "Port": "12306",
            "Timeout": "5",
            "TransportMode": "reliable",
            "RendezvousReaderCount":"1",
        }
    )
    logging.info(f"Sender: initiating sending")
    writer = wan.Open("testdata_sender", adios2.Mode.Write)
    sendbuffer = wan.DefineVariable("np_data",data, shape, start, count, adios2.ConstantDims)
    if sendbuffer:
        writer.BeginStep()
        writer.Put(sendbuffer,data,adios2.Mode.Deferred)
        writer.EndStep()
    else:
        raise ValueError("DefineVariable failed")
    
    writer.Close()
  
    logging.info(f"Sender: sending finished")

def thread_receive(name:str):

    adios_io = adios2.ADIOS()
    wan = adios_io.DeclareIO("Client")
    wan.SetEngine("Dataman")
    wan.SetParameters(
        {
            "IPAddress": "0.0.0.0",
            "Port": "12306",
            "Timeout": "5",
            "TransportMode": "reliable",
            "RendezvousReaderCount":"1",
[...]

PR #3544, opened 2023-03-19
Throw an exception in Python bindings on Get() buffer type mismatch

eisenhauer:
Relevant to issue #3538

### DIFF
--- a/bindings/Python/py11Engine.cpp
+++ b/bindings/Python/py11Engine.cpp
@@ -129,6 +129,12 @@ void Engine::Get(Variable variable, pybind11::array &array, const Mode launch)
 #define declare_type(T)                                                        \
     else if (type == helper::GetDataType<T>())                                 \
     {                                                                          \
+        if (!array.dtype().is(pybind11::dtype::of<T>()))                       \
+        {                                                                      \
+            throw std::invalid_argument(                                       \
+                "In ADIOS2 Get - Type mismatch between Python buffer and "     \
+                "incoming data.");                                             \
+        }                                                                      \
         m_Engine->Get(                                                         \
             *dynamic_cast<core::Variable<T> *>(variable.m_VariableBase),       \
             reinterpret_cast<T *>(const_cast<void *>(array.data())), launch);  \
--- a/testing/adios2/bindings/python/TestGetException_nompi.py
+++ b/testing/adios2/bindings/python/TestGetException_nompi.py
@@ -26,7 +26,7 @@ def writer():
     logging.info(f"        data dtypes: {data.dtype!s}")
     shape = data.shape
     count = shape
-    start = (0,)*len(shape)
+    start = (0,) * len(shape)
     logging.info(f"        data on writer side {data!s}")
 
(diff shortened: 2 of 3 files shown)


anagainaru (dismissed):
Maybe we could add a test for this

eisenhauer:
Good point on the test.   Added test code modified from issue #3538 to use default file engine instead of streaming.

anagainaru (approved):

testing/adios2/bindings/python/TestGetException_nompi.py
```diff
@@ -0,0 +1,133 @@
...
+                logging.info("        Incoming variable type is " + typ)
+                # allocate buffer for new numpy
+                data = np.ones(bufshape)
+                logging.info(f"        receive buffer  dtype: {data.dtype!s}")
+                try:
+                    reader.Get(recvar, data, adios2.Mode.Sync)
+                except Exception as e:
+                    got_exception = 1
+                    logging.info(str(e))
+                if not got_exception:
```
dmitry-ganyushin, line 76: Python has a more convenient construction for that:
try:
       # Some Code.... 
except:
       # Handling of exception (if required)
else:
       # execute if no exception

eisenhauer: Interesting...  My knowledge of Python is pretty superficial, but that would eliminate a variable...

bindings/Python/py11Engine.cpp
```diff
@@ -129,6 +129,12 @@ void Engine::Get(Variable variable, pybind11::array &array, const Mode launch)
 #define declare_type(T)                                                        \
     else if (type == helper::GetDataType<T>())                                 \
     {                                                                          \
+        if (!array.dtype().is(pybind11::dtype::of<T>()))                       \
```
dmitry-ganyushin, line 132: If the type and size of received data is known, we could just reallocate the buffer to the received type.

eisenhauer: Yes, we have other options.  We could also redefine Get() so that it simply returned a buffer of the right type instead of asking the user to provide a buffer that might potentially be of the wrong type.  (See notes in issue #3538.)  The current structure is likely more the result of directly translating the C++ bindings into Python without a lot of thought of what a redesign should look like.  A redesign might be desirable, but not for this release.  For this, I did the minimum that would have let us know about a problem that came up in that issue.  Want to create an issue for rethinking the Python APIs?

dmitry-ganyushin: Probably it should be improved rather on the core level, not in the python binding.

eisenhauer: I'm happy to hear ideas about that, but I do think that Python has specific capabilities (and flexibilities) that aren't present in C++, C, and Fortran, specifically regarding type flexibility.  There will be things that we could do in Python that simply aren't easily replicated in type-safe languages (at least not efficiently) and those are the sorts of things that could be more easily addressed in bindings, and would likely be difficult to mirror all the way into core.

PR #3557, opened 2023-03-22
Refactoring and clean-up group API

dmitry-ganyushin:


eisenhauer:
Can you summarize these changes?  (And maybe rebase on current release_29 so it's not out-of-date?)

eisenhauer:
Still curious about what's driving this.  This includes changes to the external API as well as internal implementation. The affected routines don't seem to be documented in the user guide, but what is driving their removal?  Is this preparation for some other more extensive work?  Or a more casual decision to change stuff?

eisenhauer (changes requested):
Just starting a review, requests for details/motivation are in other comments.

dmitry-ganyushin:
Was a quick initial string-based implementation. Contained some debug functions and loosely access to internal functions. Was just a bit long in the TODO list for finalizing.

eisenhauer:
So, this is cleanup from years ago?  Certainly there does seem to be some weirdness in there.   You want to put off merging this until after the release in order to avoid complicating life for @vicentebolea ?

pnorbert (approved):
This is a cleanup and can go into the release.

dmitry-ganyushin:
It is not used currently so no damage can be possibly done.

eisenhauer:
> It is not used currently so no damage can be possibly done.

It's been a public API for years, so asserting that it isn't used is hard.  But I'm not too worried about the consequences...

eisenhauer (approved):

PR #3567, opened 2023-03-29
Throw an error when unsupported parameters are used with the ZFP CUDA backend

anagainaru:
The ZFP compressor supports only the `rate` parameter when the GPU backend is used (even if the arrays are CPUs if ZFP is linked to CUDA, the `accuracy` and `precision` parameters will result in errors). This PR makes the errors easier to understand.

ZFP supported parameters:
(de)compression mode

This is something that might change soon so it will be updated (the main reason why I am not adding this in testing).


anagainaru:
A fast last addition to the release @vicentebolea

vicentebolea:
@anagainaru sounds good, we need to rebase this onto release_29 and change the base branch to release_29.

vicentebolea:
@anagainaru I think that you have to amend and repush, the CI seems to have gotten stuck

anagainaru:
@vicentebolea resent

If the checks pass please approve the PR and merge

vicentebolea (approved):

PR #3571, opened 2023-03-30
Allowing Put/Get to receive Kokkos Views with HIP backend

anagainaru:
This is not a bug, more of an oversight (we don't need this in the current release). 

I tested the Kokkos backend in ADIOS2 with HIP on Crusher and it works fine as long as we pass the pointer (i.e. `view.data()`). I forgot to add the HIP memory space in the part of ADIOS2 that allows submitting directly the `Kokkos::View`. This PR fixes that.

Tested on Crusher BP4/BP5.

### DIFF
--- a/bindings/CXX11/adios2/cxx11/KokkosView.h
+++ b/bindings/CXX11/adios2/cxx11/KokkosView.h
@@ -39,6 +39,26 @@ struct memspace_kokkos_to_adios2<Kokkos::CudaHostPinnedSpace>
 
 #endif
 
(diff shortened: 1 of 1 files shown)


anagainaru:
@vicentebolea I am using `hipcc` to compile ADIOS2 so we can use the HIP backend in Kokkos. So far it works on Crusher just fine (even with Kokkos 4.0). You might want to add your `enable_language` section in CMakeLists.txt and see if it still works.

eisenhauer (approved):

anagainaru:
@vicentebolea this will have to go in the patch for `release_29`. Should I mark it somehow so we don't forget about it?

vicentebolea:
@anagainaru backported

PR #3579, opened 2023-04-07
Build scripts for ADIOS2 with CUDA and Kokkos backend

anagainaru:
For Summit, Perlmutter and Crusher

### DIFF
--- a/scripts/build_scripts/build-adios2-cuda-perlmutter.sh
+++ b/scripts/build_scripts/build-adios2-cuda-perlmutter.sh
@@ -1,4 +1,5 @@
 #!/bin/bash
+# shellcheck disable=SC2191
 
 module load cudatoolkit/11.7
 module load gcc/11.2.0
--- a/scripts/build_scripts/build-adios2-cuda-perlmutter.sh
+++ b/scripts/build_scripts/build-adios2-cuda-perlmutter.sh
@@ -1,0 +1,35 @@
+#!/bin/bash
+
+module load cudatoolkit/11.7
+module load gcc/11.2.0
+module load cmake/3.24.3
+module refresh
+
+######## User Configurations ########
+ADIOS2_HOME=$(pwd)
+BUILD_DIR=${ADIOS2_HOME}/build-cuda-perlmutter
+INSTALL_DIR=${ADIOS2_HOME}/install-cuda-perlmutter
+
+num_build_procs=4
+
+######## ADIOS2 ########
+mkdir -p "${BUILD_DIR}"
+rm -f "${BUILD_DIR}/CMakeCache.txt"
+rm -rf "${BUILD_DIR}/CMakeFiles"
+
+ARGS_ADIOS=(
+    -D CMAKE_INSTALL_PREFIX="${INSTALL_DIR}"
+    -D BUILD_TESTING=OFF
+    -D ADIOS2_BUILD_EXAMPLES=OFF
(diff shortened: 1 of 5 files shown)


anagainaru:
I am not sure we need 5 scripts, a lot of the things are the same but I didn't know where to document how to build Kokkos and what modules to use for each system.

scripts/build_scripts/build-adios2-cuda-perlmutter.sh
```diff
@@ -0,0 +1,35 @@
+#!/bin/bash
+
+module load cudatoolkit/11.7
+module load gcc/11.2.0
+module load cmake/3.24.3
+module refresh
+
+######## User Configurations ########
+ADIOS2_HOME=`pwd`
+BUILD_DIR=${ADIOS2_HOME}/build-cuda-summit
```
vicentebolea, line 10: summit -> perlmutter

vicentebolea (changes requested):
@anagainaru they look great! I am in favor of having those as a help to deploy in those systems.

We normally pass the `shellcheck` script on shell scripts. Can you pass it and fix the potential warnings?

vicentebolea:
> I am not sure we need 5 scripts, a lot of the things are the same but I didn't know where to document how to build Kokkos and what modules to use for each system.

Maybe it is just ok for now, if we use those scripts more we can improve them iteratively

scripts/build_scripts/build-adios2-kokkos-perlmutter.sh
```diff
@@ -0,0 +1,62 @@
...
+
+module load cudatoolkit/11.7
+module load gcc/11.2.0
+module load cmake/3.24.3
+module refresh
+
+######## User Configurations ########
+Kokkos_HOME=$HOME/kokkos/3.7/kokkos-3.7.01
+ADIOS2_HOME=`pwd`
+BUILD_DIR=${ADIOS2_HOME}/build-kokkos-summit
```
pnorbert, line 11: Same here, summit/perlmutter

pnorbert (changes requested):

anagainaru:
I did not know about `shellcheck`, that's so cool. Will fix all the warnings and the build names

anagainaru:
I fixed the warnings and the naming, I testing the new versions on Summit just to make sure I did not introduce some issues and it seems good.

pnorbert (dismissed):

vicentebolea:
@anagainaru can you do a amend and repush, it seems that the CI did not start. Lets try our new shellcheck test :)

anagainaru:
@vicentebolea I tried shellcheck online and I did not get this warnings.
I am not sure I understand what I am supposed to do for most of them. What is this one saying?
```
In scripts/build_scripts/build-adios2-cuda-perlmutter.sh line 21:
    -D CMAKE_INSTALL_PREFIX="${INSTALL_DIR}"
       ^-------------------^ SC2191: The = here is literal. To assign by index, use ( [index]=value ) with no spaces. To keep as literal, quote it.
```

vicentebolea:
>  I am not sure I understand what I am supposed to do for most of them. What is this one saying?

I think that shellcheck is all confused about -D KEY=VALUE type of argument, you can either do like "-DKEY=VALUE" disable that specify check in that file. by adding `# shellcheck disable=SC2191`

vicentebolea (approved):

PR #3582, opened 2023-04-07
shellcheck: add shellcheck ci check

vicentebolea:
This PR adds shellcheck CI check for shell scripts in the `scripts` folder.

This will not affect existing shell scripts since "cleaning" them can break them and we cannot test all of them in detail.


anagainaru:
Good idea, I will re-base my PR once you merge this one

anagainaru (approved):
I'm not an expert but it looks good to me

PR #3585, opened 2023-04-11
cmake: set correct soname

vicentebolea:
In the installation tree, we need to set the soname/version of our third library same as our ADIOS soname/soversion. This is needed for package manager to deal with multiple versions of ADIOS installed in their system.

### DIFF
--- a/bindings/C/CMakeLists.txt
+++ b/bindings/C/CMakeLists.txt
@@ -46,7 +46,7 @@ set_target_properties(
   adios2_c ${maybe_adios2_c_mpi}
   PROPERTIES
   VERSION ${ADIOS2_LIBRARY_VERSION}
-  SOVERSION ${ADIOS2_VERSION_MAJOR}
+  SOVERSION ${ADIOS2_LIBRARY_SOVERSION}
 )
 
 install(TARGETS adios2_c ${maybe_adios2_c_mpi} EXPORT adios2CExports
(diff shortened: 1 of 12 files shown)


eisenhauer:
Huh.  Isn't this potentially going to mess things up internally?   For example, evpath/CMakeLists looks for versions of ffs and atl:
find_package(atl 2.2.1 REQUIRED)
find_package(ffs 2.0.0 REQUIRED)

Also, when you run the update.sh script, in those thirdparty libraries will it wipe your changes to their CMakeLists.txt?

vicentebolea:
@eisenhauer good question

>  For example, evpath/CMakeLists looks for versions of ffs and atl:

This should not be a problem since we are not change the version of those packages only the versions of their share library. This is same as if we have a share library with a different naming, nothing will change internally (or externally, if that library name change is reflected the pkg-config.cmake (or targets) file.

The original problem with package managers is that if you have lets say adios2.8.3 and adios 2.9.0 installed now since those thirdparty libraries potentially have the same name/version problems arises of who owns it.

> Also, when you run the update.sh script, in those thirdparty libraries will it wipe your changes to their CMakeLists.txt?

Yep, but I made every change outside of the folder that gets wiped (in the case of perfstubs, its cmakelists is set as an expection not to be wipe, upstream does not have a cmakelists.txt actually)

eisenhauer (approved):

Issue #3456, opened 2023-02-01
SST with MPI plane on Frontier 


suchyta1:
When I attempt to use MPI in SST for XGC — GEM coupling, I'm encountering an error on Frontier. When I have a chance, I will try to running it on Crusher, and see if that reproduces the error. 

Here is the error log, which is occurring in XGC:

> MPICH ERROR [Rank 0] [job id 1267839.0] [Tue Jan 31 15:16:32 2023] [frontier01774] - Abort(673264166) (rank 0 in comm 0): Fatal error in PMPI_Comm_connect: Invalid port, error stack:
> PMPI_Comm_connect(125).........: MPI_Comm_connect(port="tag#0$connentry#0024CC21$", MPI_INFO_NULL, root=0, MPI_COMM_SELF, newcomm=0x7fe7ffff69b8) failed
> MPID_Comm_connect(192).........:
> MPIDI_OFI_mpi_comm_connect(662):
> (unknown)(): Invalid port
> 
> aborting job:
> Fatal error in PMPI_Comm_connect: Invalid port, error stack:
> PMPI_Comm_connect(125).........: MPI_Comm_connect(port="tag#0$connentry#0024CC21$", MPI_INFO_NULL, root=0, MPI_COMM_SELF, newcomm=0x7fe7ffff69b8) failed
> MPID_Comm_connect(192).........:
> MPIDI_OFI_mpi_comm_connect(662): 
> (unknown)(): Invalid port
> srun: error: frontier01774: task 0: Exited with exit code 255 
> srun: launch/slurm: _step_signal: Terminating StepId=1267839.0
> slurmstepd: error: *** STEP 1267839.0 ON frontier01774 CANCELLED AT 2023-01-31T15:16:33 ***
> srun: error: frontier02670: tasks 48-55: Terminated
> srun: error: frontier02556: tasks 40-47: Terminated
> srun: error: frontier02428: tasks 32-39: Terminated
> srun: error: frontier01902: tasks 8-15: Terminated
> srun: error: frontier02300: tasks 24-31: Terminated
> srun: error: frontier02814: tasks 56-63: Terminated
> srun: error: frontier02031: tasks 16-23: Terminated
> srun: error: frontier01774: tasks 1-7: Terminated
> srun: Force Terminated StepId=1267839.0
> 

[...]

Issue #3378, opened 2022-11-01
MPI_Finalize() hangs on Crusher


franzpoeschel:
This is not an ADIOS2 bug, but a system bug on Crusher/Frontier that affects the recent MPI-based implementation of SST.

**Describe the bug**
When using the `MPI_Comm_Connect` feature of MPI on Crusher, the `MPI_Finalize()` call will hang.

I have been able to reproduce this issue with a minimal MPI example. What's a bit weird is that in SST it's always the writer that hangs for me, here it's the reader.

**To Reproduce**
Use the following two MPI programs:

Server/Reader:
```c
#include <mpi.h>

#include <stddef.h>
#include <stdio.h>
#include <stdlib.h>

int main(int argsc, char **argsv)
{
    MPI_Init_thread(NULL, NULL, MPI_THREAD_MULTIPLE, NULL);

    char port_name[MPI_MAX_PORT_NAME];
    MPI_Open_port(MPI_INFO_NULL, port_name);
    printf("Port name:\n%s\n", port_name);
    MPI_Comm dynamic_comm;
    int err = MPI_Comm_accept(
        port_name, MPI_INFO_NULL, 0, MPI_COMM_SELF, &dynamic_comm);
    if (err)
    {
        printf("Error: %d\n", err);
        exit(1);
    }

    int size, rank;
    MPI_Comm_size(dynamic_comm, &size);
    MPI_Comm_rank(dynamic_comm, &rank);
    printf("Size: %d\tRank: %d\n", size, rank);

    int recv;
    MPI_Recv(
        &recv, 1, MPI_INT, MPI_ANY_SOURCE, 0, dynamic_comm, MPI_STATUS_IGNORE);
    printf("Received: %d\n", recv);

    MPI_Finalize();
}
```

Client/Writer:
```c
#include <mpi.h>

#include <stddef.h>
#include <stdio.h>
#include <stdlib.h>

int main(int argsc, char **argsv)
{
    if (argsc != 2)
    {
        printf("Need exactly one argument, the MPI port.\n");
        exit(1);
    }
    MPI_Init_thread(NULL, NULL, MPI_THREAD_MULTIPLE, NULL);
    MPI_Comm dynamic_comm;
    int err = MPI_Comm_connect(
        argsv[1], MPI_INFO_NULL, 0, MPI_COMM_SELF, &dynamic_comm);
    if (err)
    {
        printf("Error: %d\n", err);
        exit(1);
    }

    int size, rank;
[...]

PR #3588, opened 2023-04-14
Fix MPI Data plane cohort handling

vicentebolea:
We have been observing scalability issues with the MPI DP. I have been able to narrow down to potential root causes and I was able to propose possible solutions to it in this PR.

- Task id collision between different ranks at scale. This was expected to happens since we used the process id (PID) as the rank id (not a good idea). This has been replaced by an aggregate ID of the process PID and the machine hostid.

- Currently in the event of a reader rank failure the MPI Dataplane write simply frees and disconnect to every resource pertaining to its cohort.  Since at large scale the failure of readers nodes (or timeout event) are not that rare, in this PR I make the Write to skip releasing resources pertaining to the reader unless all the peers of the cohorts have exited. To make happy MPI, I do perform a `MPI_Port_disconnect` since otherwise the "failing" node might get stuck (in case of a non critical error). 

Possibly fixes: #3456
Possibly fixes: #3439
Possibly fixes: #3378

@franzpoeschel I was able to run the given reproducer in Crusher without getting any error with this changes.

### DIFF
--- a/source/adios2/toolkit/sst/dp/mpi_dp.c
+++ b/source/adios2/toolkit/sst/dp/mpi_dp.c
@@ -30,6 +30,7 @@
 
 #include <mpi.h>
 
+#include <stdint.h>
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
@@ -54,7 +55,7 @@ typedef struct _MpiWriterContactInfo
 {
     char ContactString[MPI_DP_CONTACT_STRING_LEN];
     void *StreamWPR;
-    int PID;
+    long taskID;
 } *MpiWriterContactInfo;
 
 /* Base Stream class, used implicitly */
@@ -62,7 +63,7 @@ typedef struct _MpiStream
 {
     void *CP_Stream;
     int Rank;
-    int PID;
+    long taskID;
 } MpiStream;
 
 /* Link Stream class, used implicitly */
@@ -218,7 +219,7 @@ static FMField MpiWriterContactList[] = {
     {"ContactString", "char[" MACRO_TO_STR(MPI_DP_CONTACT_STRING_LEN) "]", sizeof(char),
      FMOffset(MpiWriterContactInfo, ContactString)},
     {"writer_ID", "integer", sizeof(void *), FMOffset(MpiWriterContactInfo, StreamWPR)},
-    {"PID", "integer", sizeof(int), FMOffset(MpiWriterContactInfo, PID)},
+    {"taskID", "integer", sizeof(long), FMOffset(MpiWriterContactInfo, taskID)},
     {NULL, NULL, 0, 0}};
 
 static FMStructDescRec MpiWriterContactStructs[] = {
@@ -226,6 +227,16 @@ static FMStructDescRec MpiWriterContactStructs[] = {
     {NULL, NULL, 0, NULL}};
 
 /*****Internal functions*****************************************************/
+
+/**
+ * Return an unique process ID (Task ID) for the current process. We do this by
+ * combining the PID of the process and the hostid (as return the same output
+ * as `hostid` or the content of /etc/machine-id in modern UNIX-like systems).
+ */
+static uint64_t GetUniqueTaskId()
+{
+    return ((uint32_t)getpid() * (1ll << 32ll)) | (uint32_t)gethostid();
+}
 
 static void MpiReadReplyHandler(CManager cm, CMConnection conn, void *msg_v, void *client_Data,
                                 attr_list attrs);
@@ -242,9 +253,7 @@ static void MpiReadRequestHandler(CManager cm, CMConnection conn, void *msg_v, v
  * the reader side.  It should do whatever is necessary to initialize a new
  * reader-side data plane.  A pointer to per-reader-rank contact information
  * should be placed in *ReaderContactInfoPtr.  The structure of that
- * information should be described by DPInterface.ReaderContactFormats.  (This
- * is an FFS format description.  See
- * https://www.cc.gatech.edu/systems/projects/FFS/.)
+ * information should be described by DPInterface.ReaderContactFormats.
  */
 static DP_RS_Stream MpiInitReader(CP_Services Svcs, void *CP_Stream, void **ReaderContactInfoPtr,
                                   struct _SstParams *Params, attr_list WriterContact,
@@ -256,7 +265,7 @@ static DP_RS_Stream MpiInitReader(CP_Services Svcs, void *CP_Stream, void **Read
     CMFormat F;
 
     Stream->Stream.CP_Stream = CP_Stream;
-    Stream->Stream.PID = getpid();
+    Stream->Stream.taskID = GetUniqueTaskId();
     Stream->Link.Stats = Stats;
 
     SMPI_Comm_rank(comm, &Stream->Stream.Rank);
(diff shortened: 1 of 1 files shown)


eisenhauer (commented):
I assume not quite ready yet since you still have printf's in it, but generally looks good.

franzpoeschel:
Hey Vicente, this is great to hear! 
I understand correctly that this is already in a usable state so I can test this out, using our actual simulation code?

vicentebolea:
> Hey Vicente, this is great to hear! I understand correctly that this is already in a usable state so I can test this out in the next days, using our actual simulation code?

Yes, please, if you can test this out that will be great!

suchyta1:
How scalable is this expected to be? Or SST in general?

I have a test where I start from using and XGC restart file, to be sent to a reader. Each writing process reads one block of data from the BP file. By block, I mean an ADIOS block, where one block is from one of the original XGC writer process' send dispatch. To mimic the way XGC runs, there are 8 senders per node, and for now, I was just doing one reader, but I can turn that up too. The run worked at 64 sender nodes, and took about 7 seconds a single reader to receive the data. (I get around needing multiple readers by using MPMD.) But when I double to 128 sender nodes, it hangs. That's about 176 GB of data over 1024 blocks. Increasing the number of readers doesn't prevent the hang. I've tried adjusting some of the engine parameters too, but they don't seem to have impact relative to hanging.

eisenhauer:
> How scalable is this expected to be? Or SST in general?
> 
In general, the expectation is "highly scalable".  As always the answer in practice depends upon a lot of things, what is scalable on one machine may not be on another, resource availability can be an issue at all levels, etc.  A hang may mean a bug like a race condition, it may mean that an operation failed silently leaving SST waiting for a result that'll never come, or pretty much anything else.  If possible, the first step is to turn on verbosity to figure out what is hanging.  I can't speak to what level of verbosity is in the MPI data plane, but setting the environment variable "SstVerbose" to a numerical value between 1 and 5 is a place to start.  5 is the highest verbosity.

suchyta1:
I don't know how to parse it very well, but I attached a lot with 100 sender nodes and 1 reader.

[slurm-1431185.out.log](https://github.com/ornladios/ADIOS2/files/12537069/slurm-1431185.out.log)

suchyta1:
This doesn't say a ton, but there is a request that didn't report completed:

> [esuchyta@login15.frontier BlockTest-03]$ grep -i "memory read to rank" slurm-1431185.out | grep "Waiting" | wc -l
2311
[esuchyta@login15.frontier BlockTest-03]$ grep -i "memory read to rank" slurm-1431185.out | grep "completed" | wc -l
2310

eisenhauer:
Interesting.  I'll leave this for @vicentebolea to analyse because it is using the MPI data plane, but an incomplete request could be the cause of deadlock (or the symptom of it).  At a quick glance, it looks like the reader didn't finish with the first timestep, but the writer created several more...

suchyta1:
Yes, that's correct. I can try simplifying to a single time step first. 

Looking through the file, I noticed that there's three passes through the "memory read to rank" kind of output, where the first two go through the 800 ranks for the writers that the reader is connected to, but the third time gets between starting and completing rank 710 and then nothing more is reported.

suchyta1:
I attached a single time step one, since it's a bit clearer what I mentioned above on the third pass through with cutting off, here at rank 653.

[slurm-1431250.out.log](https://github.com/ornladios/ADIOS2/files/12537841/slurm-1431250.out.log)

suchyta1:
I looked through the code a little, and did a rerun with `CMVerbose=1` and `CMLowLevelVerbose=1`, so the log attached now has more information.

[slurm-1432345.out.log.zip](https://github.com/ornladios/ADIOS2/files/12549077/slurm-1432345.out.log.zip)

franzpoeschel:
I reverted the workarounds that I used so far and ran an example on Frontier at 128 nodes. 
With the rolled-back workarounds, there is now one instance of N:1 and one instance of 1:M communication. Seems to work without trouble with this :)

vicentebolea:
@suchyta1 thanks for reporting issues with the MPI DP. I Will take a look to it. Quick question: are you using the tip commit of this PR MR?

@franzpoeschel I am happy that this fix your reproducer app.

suchyta1:
Yes, that's what I was using.

suchyta1:
I added some extra debugs, and as far as I can tell, there's a lock operation that is never returned from.
[slurm-1432539.out.zip](https://github.com/ornladios/ADIOS2/files/12587819/slurm-1432539.out.zip)

eisenhauer:
Yes, that tracks.   The interesting bit here is the bit from the reader-side data plane on rank 0 (which is the "DP Reader 0" line).  
```
DP Reader 0 (0x1004380): Waiting for completion of memory read to rank 652, condition 3855,timestep=0, is_local=0
Writev success
DP Reader 0 (0x1004380): Remote read; CMCondition_wait
Before unlockCManager Unlock at "/ccs/home/esuchyta/software/build/frontier/adios2-gcc12.2.0-test/thirdparty/EVPath/EVPath/cm_interface.c" line 632
After unlockBefore lockCManager Lock at "/ccs/home/esuchyta/software/build/frontier/adios2-gcc12.2.0-test/thirdparty/EVPath/EVPath/cm_interface.c" line 84

```
From the first line above, we know we're in the MPI data plane, in MpiWaitForCompletion(), and it has fallen into the CMCondition_wait() at line 571.   That we never fall out of this means that the corresponding CMCondition_signal (at line 708 at the end of the MPIReadReplyHandler() subroutine) never happens.    So, what's going on?  I don't know, but it's certainly something at the dataplane level.  Just what isn't clear and it's probably going to require the attention of the author @vicentebolea .

vicentebolea:
@eisenhauer it appears that its waiting for the rank=652 (writer) to reply its message however the rank=652 has long been either dead or unresponsive.

Few questions I have noticed that this rank has the same ID has the rank=748, is this odd? See below:

```
DP Writer 652 (0x102c3a0): MpiReadRequestHandler: Replying reader=0 with MPI port name=tag#0$connentry#0140AE23$
DP Writer 748 (0x102c3a0): MpiReadRequestHandler: Replying reader=0 with MPI port name=tag#0$connentry#01408A24$
```

Similarly it seems that this setup of a single reader and 800 writers might cause to have 800 MPI connections on the reader side simultaneously, maybe hitting some limit of the mpi client/server implementations. I am working on a reproducer to further debug this. I will also investigate how to add such as test in the CI.

eisenhauer:
> Few questions I have noticed that this rank has the same ID has the rank=748, is this odd? See below:

The ID printed after the rank is the base address of the writer-side stream structure.  If an application has more than one output stream open, this differentiates them in the output.  But in your case it just means that two ranks happened to allocate the same base address for the stream structure.  They're different processes, so this shouldn't matter.  (They'd probably be the same more often if most OS' didn't use memory randomization as a security measure.)

Anyhow, no help in this situation I'm afraid...

franzpoeschel:
Using this, I now had a full-scale run on Frontier from `9126(nodes) x 8(GPUs)` writers to `9126(nodes)` readers. Data exchange happened only node-internally, 6Gb were exchanged per step and node (not much, but I was focusing on scaling the application to a large number of writers/readers for now).
I have not (yet) done a full evaluation of performance, but by squinting at the numbers, I/O time looks stable.
The workaround described in https://github.com/ornladios/ADIOS2/issues/3846 was necessary to scale above 7000 nodes, but that has nothing to do with this PR.

vicentebolea:
@eisenhauer do you think that we should have this merged before waiting the workaround of @franzpoeschel to be included (in a more generic bugfix form) in master?

eisenhauer:
> @eisenhauer do you think that we should have this merged before waiting the workaround of @franzpoeschel to be included (in a more generic bugfix form) in master?

I would go ahead and merge.  @franzpoeschel workaround is an orthogonal issue that just happens to also come up when we scale up.

eisenhauer (approved):

PR #3593, opened 2023-04-25
Heat Transfer Example: MPI Datatype

ax3l:
Replace the `MPI_REAL8` datatype, which is _optional_ in the MPI standard and not always available, with `MPI_DOUBLE`. The latter is always available and corresponds to the data sent around.

First seen as an OpenMPI compile error on a user's machine:
https://github.com/openPMD/openPMD-api/issues/1424#issuecomment-1516916044

I found no further examples that use the same MPI type and would need an update.

### DIFF
--- a/examples/heatTransfer/write/HeatTransfer.cpp
+++ b/examples/heatTransfer/write/HeatTransfer.cpp
@@ -136,7 +136,7 @@ void HeatTransfer::exchange(MPI_Comm comm)
 {
     // Build a custom MPI type for the column vector to allow strided access
     MPI_Datatype tColumnVector;
-    MPI_Type_vector(m_s.ndx + 2, 1, m_s.ndy + 2, MPI_REAL8, &tColumnVector);
+    MPI_Type_vector(m_s.ndx + 2, 1, m_s.ndy + 2, MPI_DOUBLE, &tColumnVector);
     MPI_Type_commit(&tColumnVector);
 
     // Exchange ghost cells, in the order left-right-up-down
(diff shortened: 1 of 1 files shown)


eisenhauer:
@vicentebolea What is your preferred protocol for maintaining 2.9 and master branches currently?  I've submitted some things (like the DAOS engine) to master because we clearly don't want them in a release any time soon.  Others, like PR #3591 are clear bug fixes related to new things in the release (default to BP5) and probably need to be in the release (and backported to master).   Is this PR similar?  Or more generally should most everything go to release_29 and be tagged for backporting to master?  I.E. every kind of "normal" fix that isn't a feature that we don't want in 2.9 (which I guess is most every PR).

vicentebolea:
> What is your preferred protocol for maintaining 2.9 and master branches currently?

We should aim to keep the release_XY branches API/ABI compatible with very counted exceptions, this is, only bugfixes should go to the release branch and they should be carefully chosen so that they dont change either the API (ABI). The rationale of being strict with it is that downstream they expect it to be like this. They expect not to recompile when changing the patch version of adios2.

This PR, for instance looks fine, change the function call of mpi procedures no change in our API/ABI.

#3591 is a bigger change a deeper analysis is needed. There is a high chance of failure doing this manually, so I think that in the future we would want to add automated test for checking that API/ABI compatibility maybe something like running/building ADIOS2 tests from the X.Y.0 version using the source code of the new PR (API) or just linking with the new binaries (ABI) of the pull-request.

On the other hand we will have one more release before the end of September.

Vicente

vicentebolea:
Also this branch can be rebased to release_29 for easier backporting.

eisenhauer:
> #3591 is a bigger change a deeper analysis is needed. There is a high chance of failure doing this manually, so I think that in the future we would want to add automated test for checking that API/ABI compatibility maybe something like running/building ADIOS2 tests from the X.Y.0 version using the source code of the new PR (API) or just linking with the new binaries (ABI) of the pull-request.

Thanks Vicente.  I'll followup in #3591.

anagainaru (commented):
I assume this was not approved because it's not rebased on `release_29`?

eisenhauer:
> I assume this was not approved because it's not rebased on `release_29`?

Yeah, at least on my end.  @ax3l, can you rebase and retarget to the release_29 branch?

vicentebolea:
Alternatively, we can merge to master this and I can backport it later to release. @ax3l is a good long time contributor but not a core developer, so we can give him a hand on this :). @anagainaru if you approve Ill follow up the backport.

anagainaru (approved):

ax3l:
Thanks for the help!

Sorry for being a bit slow, extended proposal season and deadlines kept me busy.

Thanks for the help backporting. Otherwise, if a simple rebase helps, please do not hesitate to rebase my PRs and force push on it :)

PR #3595, opened 2023-05-01
Adding missing HIP specific functionality

anagainaru:
The functionality needed to set the device id and check if the buffer is allocated on the GPU was missing for the HIP backend.


anagainaru:
Replication of #3605

PR #3605, opened 2023-05-02
Adding missing HIP specific functionality

anagainaru:
We are missing functionality to set the device id and to check if a buffer is allocated on the GPU for the HIP backend. The tests are passing because the memory space is set manually. This needs to be included in the patch for the latest release.


eisenhauer (approved):

vicentebolea:
@anagainaru We will add a nightly build that runs in Crusher soon. How about merging this in master, mark it as 2.9.1 backport, and after we pass the build in Crusher with GPU we backport this. The reason is that we might want to change more things in this area again once we test  this in Crusher so we can skip the trouble of double backporting then.

anagainaru:
@vicentebolea sorry I saw this too late. I will do that for my next PR (unless it's after you add the nightly build)

vicentebolea:
> @vicentebolea sorry I saw this too late. I will do that for my next PR (unless it's after you add the nightly build)

No problem, these things happens. We will have the crusher build soon :tada:

Issue #3523, opened 2023-03-08
IBM XL compiler: BP5 aborts


pnorbert:
Summit with XL compiler: writer aborts due to something going wrong with the buffered data. 
Using `bin/globalArray_write`:

```
$ gdb  adios.xl/bin/globalArray_write core.71311

(gdb) bt
#0  0x000000004139d300 in ?? ()
#1  0x0000200002a2c370 in adios2::core::engine::BP5Writer::WriteMyOwnData (this=0x413979a0, Data=0x4139c630)
    at /ccs/proj/e2e/pnorbert/ADIOS/ADIOS2/source/adios2/engine/bp5/BP5Writer_TwoLevelShm.cpp:157
#2  0x0000200002a2ea48 in adios2::core::engine::BP5Writer::WriteData_TwoLevelShm (this=0x413979a0, Data=0x4139c630)
    at /ccs/proj/e2e/pnorbert/ADIOS/ADIOS2/source/adios2/engine/bp5/BP5Writer_TwoLevelShm.cpp:114
#3  0x0000200002a08b18 in adios2::core::engine::BP5Writer::WriteData (this=0x413979a0, Data=0x4139c630)
    at /ccs/proj/e2e/pnorbert/ADIOS/ADIOS2/source/adios2/engine/bp5/BP5Writer.cpp:266
#4  0x0000200002a172f8 in adios2::core::engine::BP5Writer::EndStep (this=0x413979a0)
    at /ccs/proj/e2e/pnorbert/ADIOS/ADIOS2/source/adios2/engine/bp5/BP5Writer.cpp:555
#5  0x000020000044562c in adios2::Engine::EndStep (this=0x7fffc4f68890)
    at /ccs/proj/e2e/pnorbert/ADIOS/ADIOS2/bindings/CXX11/adios2/cxx11/Engine.cpp:117
#6  0x000000001000a4e4 in main (argc=1, argv=0x7fffc4f68e98)
    at /ccs/proj/e2e/pnorbert/ADIOS/ADIOS2/examples/basics/globalArray/globalArray_write.cpp:111
(gdb) up
#1  0x0000200002a2c370 in adios2::core::engine::BP5Writer::WriteMyOwnData (this=0x413979a0, Data=0x4139c630)
    at /ccs/proj/e2e/pnorbert/ADIOS/ADIOS2/source/adios2/engine/bp5/BP5Writer_TwoLevelShm.cpp:157
[...]

Issue #3602, opened 2023-05-02
Enable CI builds in OLCF systems


vicentebolea:
Supersede #3486 

We want to implements the builds in orange and purple boxes

![ADIOS2 BUILD CI Goal(2)](https://github.com/ornladios/ADIOS2/assets/939798/d7d7dbc3-59c9-4639-9a79-ce6278d0dc5f)

- [x] #3603 
- [x] #3546 
- [x] #3598 
- [x] #3601 
- [x] #3599

PR #3610, opened 2023-05-10
ci,ascent: enable ascent builds

vicentebolea:
fixes: #3601
fixes: #3598
fixes: #3551 
fixes: #3599 
fixes: #3546
fixes: #3523
fixes: #3603
related: #3602

This PR adds the following power9 builds:

- Ascent IBM XL (SST evpath)
- Ascent NVHPC (SST evpath)
- Ascent CUDA/GCC (SST libfabric)
- Ascent Kokkos_CUDA/GCC (SST libfabric)

The builds are accessible from the Github PR page and it is shown as one more check.

Screenshot from 2023-05-11 13-29-14

Builds are here: https://code.ornl.gov/ecpcitest/adios2/-/pipelines


source/adios2/toolkit/format/bp5/BP5Serializer.cpp
```diff
@@ -1333,19 +1333,27 @@ BP5Serializer::TimestepInfo BP5Serializer::CloseTimestep(int timestep,
     MBase->BitField = tmp;
     NewAttribute = false;
 
-    struct TimestepInfo Ret
-    {
-        Formats, Metadata, AttrData, CurDataBuffer
-    };
+    struct TimestepInfo Ret;
+    Ret.NewMetaMetaBlocks = Formats;
```
vicentebolea, line 1337: @pnorbert I figured the issue. IBM XL does not do the RVO optimization meaning that returning the struct will call the destructoror of TimeStepInfo. This resulted in this the DataBuffer being freed before being used.

eisenhauer: Unfortunate.  That'll presumably require an internal API change to fix.  If nobody else does it, I'll look tomorrow.

vicentebolea: Find the commit that adds the fix to it in: https://github.com/ornladios/ADIOS2/pull/3610/commits/501278282a04712b490eccc7519e23f71f20a3a7

vicentebolea: That would be hard to backport since we change member attributes of a public class, we can do a backport of the fix by removing the delete DataBuffer in the destructor of TimestepInfo

eisenhauer: Not a public class, at least not in the sense that it is exposed to users.  If it is changed, only an ADIOS-internal ABI changes, not the user-facing ABI.

vicentebolea: @eisenhauer we cannot be sure, this is a tricky one, it is a ABI breaking change for users that potentially interact with more internal classes of ADIOS2 (Which probably doesnt exist). The problem is here: https://github.com/ornladios/ADIOS2/blob/8ce4ff9bd43091f76de5f12dac1a19d7ddd49be1/source/adios2/toolkit/format/bp5/BP5Serializer.h#LL110C1-L110C17

This method is publicly available meaning that an user that an existing built app that uses this method when using the new 2.9.1 ADIOS2 libraries ld.so would not be able runlink due to changes in the memory size of the return type of that method.

Probably the best course of action is to adapt this to be ABI compatible when backporting to release_29. Not that it would realistically break users adios2 installations but for avoiding problems when adding automated API/ABI checks is best to be strict with it.

How about having this merged to master and later backport this to the release_29 in its own PR.

scripts/ci/gitlab-ci/run.sh
```diff
@@ -1,48 +1,63 @@
...
+  exit 2
+fi
 
-  echo "**********module avail Begin************"
-  module avail
-  echo "**********module avail End**************"
+if [ "$STEP" = "update" ]
+then
+  echo "**********Env Begin**********"
+  env | sort
```
scottwittenburg, line 29: Are you thinking to leave this in?  I didn't scan carefully through the printed environment variables, but is it possible some sensitive information is exposed there?

scottwittenburg (approved):
I'm just reviewing the changes related to CI here, and leaving the changes in `source/` to someone else.  I had one question about printing out the all environment variables, but otherwise I didn't notice any gotchas.  It's great you got this working @vicentebolea!

scripts/ci/gitlab-ci/run.sh
vicentebolea: That is correct, this was inherited from existing code, if we ever need this we can re-add it later.

scottwittenburg (approved):

vicentebolea:
@eisenhauer can you review the source code changes?

eisenhauer (approved):
I think the C++ changes are OK...

PR #3621, opened 2023-05-13
Adding SYCL support to the Kokkos backend 

anagainaru:
Adding the SYCL backend in ADIOS2 (tested on ALCF Polaris with oneapi and clang++)

This was also tested with the new build system that Vicente is pushing in #3623 

One change that affects the other backends is that `long double` will no longer be supported for metadata computation (min/max) since it is not supported by several configurations I tried running on.


anagainaru:
This is ready for review

@masterleinad let me know if you have any comments on the SYCL specific functions

source/adios2/helper/adiosKokkos.cpp
```diff
@@ -43,6 +43,10 @@ void KokkosMinMaxImpl(const char * /*values*/, const size_t /*size*/,
                       char & /*min*/, char & /*max*/)
 {
 }
+void KokkosMinMaxImpl(const long double * /*values*/, const size_t /*size*/,
+                      long double & /*min*/, long double & /*max*/)
+{
+}
```
masterleinad, line 49: You could possibly use `Kokkos::abort` for all of these.

source/adios2/helper/adiosKokkos.cpp
```diff
@@ -98,6 +92,15 @@ bool IsGPUbuffer(const void *ptr)
     {
         return true;
     }
+#endif
+#ifdef ADIOS2_HAVE_KOKKOS_SYCL
+    auto ret =
+        sycl::address_space_cast<sycl::access::address_space::local_space,
+                                 sycl::access::decorated::no>(ptr);
```
masterleinad, line 99: That should probably be
```suggestion
        sycl::address_space_cast<sycl::access::address_space::global_space,
                                 sycl::access::decorated::no>(ptr);
```
assuming that you expect the allocations to come from sycl::malloc. You would use `local_space` for scratch memory. If you care about testing for host accessibility you would refine to `ext_intel_global_device_space` resp. `ext_intel_global_host_space`, see https://github.com/intel/llvm/blob/f363bb272b5a85849059fff40c3b321910f87781/sycl/doc/extensions/supported/sycl_ext_intel_usm_address_spaces.asciidoc#sycl-device-memory-model.

source/adios2/helper/adiosKokkos.cpp
anagainaru: I am not aborting because for now I don't want to terminate the program, just not compute the metadata on the GPU.

source/adios2/helper/adiosKokkos.cpp
anagainaru: We are using `std::copy` to cache the data in an internal ADIOS buffer outside this function. I want to use `address_space_cast` to know when there is no way to use `std::copy` in which case I fall back to Kokkos. I was assuming `global_space` would still allow the Host to access it.

vicentebolea:
@anagainaru this looks good, but in principle I would skip it for the patch release since this is not a bugfix. Is there a reason of why would you need this in the next patch release?

vicentebolea:
> Should I wait until you merge your PR https://github.com/ornladios/ADIOS2/pull/3623?

That makes sense to wait for the rebase, the CI should pass in less then 1h.

vicentebolea:
@anagainaru #3623 was merged to master

source/adios2/helper/adiosKokkos.cpp
anagainaru: Makes sense, thanks. Changed.

anagainaru:
Rebased and ready to be reviewed one last time

masterleinad (approved):
Looks reasonable to me.

vicentebolea (approved):
Looks good, good refactor there

PR #3623, opened 2023-05-16
ci: use nvcc_wrapper in adiosKokkos

### LINKED ISSUE #3566: ADIOS2 GPU-aware breaks with Kokkos 4.0
**Describe the bug**
When building agains Kokkos 4.0, GPU applications using ADIOS2 give runtime errors. 

**To Reproduce**
1. Build Kokkos 4.0 with the following flags
```
    -D CMAKE_BUILD_TYPE=RelWithDebInfo
    -D CMAKE_CXX_COMPILER=/path/to/kokkos/bin/nvcc_wrapper
    -D Kokkos_ENABLE_CUDA=ON
    -D Kokkos_ENABLE_CUDA_LAMBDA=ON
    -D CMAKE_CXX_STANDARD=17
    -D CMAKE_CXX_EXTENSIONS=OFF
    -D CMAKE_POSITION_INDEPENDENT_CODE=TRUE
    -D BUILD_SHARED_LIBS=ON

    -D Kokkos_ARCH_AMPERE80=ON     # if building on Perlmutter
    -D Kokkos_ARCH_VOLTA70=ON      # if building on Summit
```

2. Build ADIOS2 with Kokkos 4.0
```
    -DADIOS2_USE_Kokkos=ON 
    -DCMAKE_CXX_COMPILER=g++ 
    -DCMAKE_C_COMPILER=gcc 
    -DCMAKE_CUDA_FLAGS="-fPIC" 
    -DKokkos_ROOT=/path/to/kokkos/install
```

3. Run the CUDA example
```bash
$ ./bin/CudaBPWriteRead_cuda
Using engine BP5
terminate called after throwing an instance of 'std::runtime_error'
  what():  Kokkos::Impl::ParallelReduce< Cuda > requested too much L0 scratch memory
Aborted
```

**Expected behavior**
```bash
$ ./bin/CudaBPWriteRead_cuda
Using engine BP5
Simualation step 0 : 6000 elements: 0
Simualation step 1 : 6000 elements: 10
Simualation step 2 : 6000 elements: 20
Simualation step 3 : 6000 elements: 30
Simualation step 4 : 6000 elements: 40
Simualation step 5 : 6000 elements: 50
Simualation step 6 : 6000 elements: 60
Simualation step 7 : 6000 elements: 70
Simualation step 8 : 6000 elements: 80
Simualation step 9 : 6000 elements: 90
```

**Desktop (please complete the following information):**
Tested on Perlmutter
```
gcc/11.2.0
cudatoolkit/11.7
cmake/3.20.4
```
Tested on Summit
```
module load gcc/10.2
module load cuda/11.5
module load cmake/3.23
```

**Additional context**

Removing from `source/adios2/CMakeLists.txt` the following lines:
```cmake
set_property(SOURCE helper/adiosKokkos.cpp PROPERTY LANGUAGE CUDA)
set_property(SOURCE helper/adiosKokkos.cpp APPEND PROPERTY COMPILE_FLAGS "--extended-lambda")
```
and building ADIOS2 with `-D CMAKE_CXX_COMPILER=/path/to/kokkos/bin/nvcc_wrapper` make the code run correctly on Perlmutter and Summit.

vicentebolea:
fixes #3566

Taken from the comment added in the changes

```
  # Kokkos imposes us to set our CMAKE_CXX_COMPILER to Kokkos_CXX_COMPILER.
  # The problem is that we do not want this for the whole project and with
  # CMake we cannot set the CXX_COMPILER for a single target. The solution is
  # to move the adios2 module that uses Kokkos to its independent subdir and
  # set there CMAKE_CXX_COMPILER, which is possible (and scoped to that subdir)
  # in cmake.
```


vicentebolea:
@anagainaru

CODEOWNERS
```diff
@@ -20,6 +20,6 @@ scripts/ @vicentebolea
 source/adios2/toolkit/sst/dp/mpi_dp.c @vicentebolea
 
 # GPU-aware specific files
-source/adios2/helper/adiosKokkos.* @anagainaru
+source/adios2/helper/kokkos/adiosKokkos.* @anagainaru
```
anagainaru, line 23: If we make a separate folder I might as well cover all files that will be here. The query logic might require some Kokkos components as well.
```suggestion
source/adios2/helper/kokkos/* @anagainaru
```

cmake/DetectOptions.cmake
```diff
@@ -177,22 +177,30 @@ endif()
...
   endif()
   if(Kokkos_FOUND)
-    set(ADIOS2_HAVE_Kokkos TRUE)
-    if(Kokkos_ENABLE_CUDA OR Kokkos_ENABLE_HIP OR Kokkos_ENABLE_SYCL)
-      if(Kokkos_ENABLE_CUDA)
-        set(ADIOS2_HAVE_Kokkos_CUDA TRUE)
-        enable_language(CUDA)
+    # There is no easy way in CMake find_package with a range of multiple minor versions.
+    if (Kokkos_VERSION VERSION_LESS 3.7)
+      if(ADIOS2_USE_Kokkos STREQUAL ON)
```
anagainaru, line 187:
```suggestion
      if(ADIOS2_USE_Kokkos)
```

cmake/DetectOptions.cmake
```diff
@@ -177,22 +177,30 @@ endif()
...
-    find_package(Kokkos 3.7...<4.0 REQUIRED)
+    find_package(Kokkos REQUIRED)
   endif()
   if(Kokkos_FOUND)
-    set(ADIOS2_HAVE_Kokkos TRUE)
-    if(Kokkos_ENABLE_CUDA OR Kokkos_ENABLE_HIP OR Kokkos_ENABLE_SYCL)
-      if(Kokkos_ENABLE_CUDA)
-        set(ADIOS2_HAVE_Kokkos_CUDA TRUE)
-        enable_language(CUDA)
+    # There is no easy way in CMake find_package with a range of multiple minor versions.
```
anagainaru, line 185: Why would `find_package(Kokkos 3.7 REQUIRED)` not work? Why do we need to separate this.

source/adios2/helper/kokkos/CMakeLists.txt
```diff
@@ -0,0 +1,22 @@
...
+
+add_library(adios2_core_kokkos adiosKokkos.h adiosKokkos.cpp)
+
+set_target_properties(adios2_core_kokkos PROPERTIES
+  VISIBILITY_INLINES_HIDDEN ON
+  INCLUDE_DIRECTORIES "$<BUILD_INTERFACE:${ADIOS2_SOURCE_DIR}/source>;$<BUILD_INTERFACE:${ADIOS2_BINARY_DIR}/source>"
+  EXPORT_NAME core_kokkos
+  OUTPUT_NAME adios2${ADIOS2_LIBRARY_SUFFIX}_core_kokkos
+  )
+
```
anagainaru, line 20: We still need to set the CXX standard to 17

anagainaru (commented):
Thanks Vicente for working on this. I will test this on Summit/Perlmutter/Crusher and come back to update the review

cmake/DetectOptions.cmake
vicentebolea: 3.7 will work with only 3.7, we want to support kokkos >=3.7.

On the other hand the range 3.7...4.0 does not really work as we expected it only works  between patch releases 3.7.0...3.7.4; for diff minor releases the ...< is possible but still for only symbolizing until next minor release. We want to support 3.7 and 4.0 and possibly newer versions

cmake/DetectOptions.cmake
vicentebolea: I want to discriminate here against the AUTO value too.

source/adios2/helper/kokkos/CMakeLists.txt
vicentebolea: added

cmake/DetectOptions.cmake
anagainaru: I am building ADIOS with Kokkos 4.0 with `find_package(Kokkos 3.7 REQUIRED)` and it works fine. In our case the 3.7 is the minimum requirement because Kokkos is forward compatible:
https://github.com/kokkos/kokkos/blob/62d2b6c879b74b6ae7bd06eb3e5e80139c4708e6/cmake/kokkos_install.cmake#L20-L22

anagainaru:
Screenshot 2023-05-17 at 1 22 30 AM

I tested the new script on the three systems and it seems to work everywhere ! Perlmutter with gcc 11.2 and cuda 11.7 gives some errors for Kokkos 3.7 but works starting with the patch on (it was giving the same error before this PR, I just didn't test with the vanilla 3.7 until now). Once we converge in the `find_package` we can merge this.

vicentebolea:
@anagainaru fantastic work there checking where it builds/runs :+1: :+1: :+1:

anagainaru:
I realized that now we can always use `g++` to build ADIOS2 and the Kokkos parts will use whatever we use to build Kokkos. I updated the HIP script.

PR #3625, opened 2023-05-19
ci,cmake: cleanup scripts subdirectory

vicentebolea:
- [x] Get rid of CUDA builds (we test this in Ascent)
- [X] Get rid of EMU Power8 builds (we test this in Ascent)
- [x] Removed appveyor, travis, azure files/
- [x] Removed old and unused dashboard files for cori/aargh and perlmutter.
- [x] Removed dockerfiles and other sh scripts to generate docker images for builds that we do not have anymore.

related: #3696


scripts/ci/images-v2/oneAPI.repo
scottwittenburg, line 1: This file is still used, it's needed by `scripts/ci/images/Dockerfile.ci-el8-intel`.

scripts/ci/images/build-el8.sh
```diff
@@ -1,5 +1,4 @@
-
-#~/bin/bash
+#!/bin/bash
```
scottwittenburg, line 1: Whoops, obviously I made a typo, and the fact that I had shellcheck ignore it, allowed it to sneak through.  Thanks for fixing.

scottwittenburg (changes requested):
Looks good @vicentebolea, thanks!  I only found one thing to fix, and then we should be good to merge.

scottwittenburg (approved):

PR #3632, opened 2023-05-24
Update ADIOS2 HDF5 VOL with basic set of capability flags

jhendersonHDF:
This PR updates the ADIOS2 HDF5 VOL connector to return a basic set of capability flags so that it can be tested with the [HDF5 VOL tests repository](https://github.com/HDFGroup/vol-tests/). Otherwise, the tests would fail when trying to call the H5Pget_vol_cap_flags API routine to retrieve the connector's capability flags.

Defining the new H5VL_adios2_introspect_get_cap_flags callback function allows HDF5 to properly resolve situations where the connector may have other connectors stacked on top of it that pass operations down and need to check if the connectors below it support certain functionality.

### DIFF
--- a/source/h5vol/H5VolReadWrite.h
+++ b/source/h5vol/H5VolReadWrite.h
@@ -32,6 +32,8 @@ extern herr_t H5VL_adios2_beginstep(const char *engine_name, adios2_step_mode m)
 
 extern herr_t H5VL_adios2_endstep(const char *engine_nane);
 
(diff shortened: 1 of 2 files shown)


guj (approved):
Thanks for adding the flags.

vicentebolea:
@jhendersonHDF I rebased/solved conflict so that it can be merged

guj (approved):
Thanks!

jhendersonHDF:
Thanks for rebasing this @vicentebolea!

vicentebolea:
This is approved, should it be merged?

PR #3637, opened 2023-05-25
Ascent,CI: do not report sync script build

vicentebolea:


vicentebolea:

PR #3639, opened 2023-05-26
Update documentation for the GPU-aware functionality

anagainaru:
Including the SYCL backend and the way users can specify they are using GPU buffers.


eisenhauer (approved):

anagainaru:
@vicentebolea do you know why some tests will not start running? I am not changing anything in the code so we don't actually need the test anything. Is there anyway to bypass them?

vicentebolea:
Yeah this is hanging on the contract tests, but those will not run when no source code is changed. Is this ready to merge?

anagainaru:
@vicentebolea yes

PR #3640, opened 2023-05-26
replaced uint with size_t 

guj:
as reported in: 
https://github.com/ornladios/ADIOS2/issues/3638

### DIFF
--- a/source/h5vol/H5VolUtil.c
+++ b/source/h5vol/H5VolUtil.c
@@ -56,16 +56,16 @@ void *safe_ralloc(void *ptr, size_t newsize, unsigned long line)
     return p;
 }
 
-void gUtilConvert(hsize_t *fromH5, size_t *to, uint ndims)
+void gUtilConvert(hsize_t *fromH5, size_t *to, size_t ndims)
 {
-    uint i = 0;
+    size_t i = 0;
     for (i = 0; i < ndims; i++)
     {
         to[i] = fromH5[i];
     }
 }
 
-int gUtilADIOS2GetShape(hid_t space_id, size_t *shape, uint ndims)
+int gUtilADIOS2GetShape(hid_t space_id, size_t *shape, size_t ndims)
 {
     if (gUtilADIOS2IsScalar(space_id))
     {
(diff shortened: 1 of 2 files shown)


eisenhauer:
@guj I'd see this as a bug, so maybe target to release_29 rather than master(so that it goes into the 2.9.1 release) and tag it as backport to master (and @vicentebolea will manage that bit) ?

guj:
> @guj I'd see this as a bug, so maybe target to release_29 rather than master(so that it goes into the 2.9.1 release) and tag it as backport to master (and @vicentebolea will manage that bit) ?

Fine with me.  Double checking with @vicentebolea.  Shall I merge to release_29 instead of master?

vicentebolea:
master, this MR changes the signature of public headers functions.

guj:
> master, this PR changes the signature of public headers functions.

OK Thanks!

vicentebolea:
Do we need this for the next patch release?

guj:
> Do we need this for the next patch release?

Guess not urgently.

eisenhauer:
Actually, I think I need to push back on the public headers argument.  I don't see that this is in any way a "public" header.  It doesn't get installed, so it's not going to be referenced outside the ADIOS build.  The two functions affected are not consumed outside of the h5vol build.  Is there a conceivable way that putting this into a release would do more than let ADIOS compile in an environment in which it is currently broken?

vicentebolea:
> . I don't see that this is in any way a "public" header.

@eisenhauer you are right, my bad, I thought that this header gets installed but it is not the case. I am not too familiar about the H5Vol component, but I can see now that it is intended to be used as a plugin.

@guj This is change is in an internal function that does not affect the ABI of the plugin thus it should be good to go :). Please merge to master and I will backported with other MRs to release_29.

guj:
@eisenhauer  @vicentebolea 
Please approve so I can merge :)

vicentebolea (approved):

ax3l:
> master, this PR changes the signature of public headers functions.

Just FYI, I will patch this into any Windows-related package manager I found for 2.9.0 - so you can also add it to master @vicentebolea :)

PR #3651, opened 2023-06-02
Open file with in-memory metadata

pnorbert:
* Added a new functionality to an Engine, to serialize metadata in memory after Open(ReadRandomAccess), to send over to another programs that can then use this metadata to open the same dataset with an accelerated function (using metadata from memory instead of retrieving from disk). BP4 and BP5 engine supports this new function.

  Program 1:  reader = io.Open(fname, adios2::Mode::ReadRandomAccess);
              char *md;
              size_t mdsize;
              reader.GetMetadata(&md, &mdsize);
  Program 2:  reader = io.Open(fname, md, mdsize);

* Added Transport::CurrentPos() to get the current seek position

### DIFF
--- a/bindings/C/adios2/c/adios2_c_engine.cpp
+++ b/bindings/C/adios2/c/adios2_c_engine.cpp
@@ -199,6 +199,27 @@ adios2_error adios2_engine_openmode(adios2_mode *mode,
     }
 }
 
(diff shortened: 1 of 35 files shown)


eisenhauer (approved):
All this looks reasonable.

Issue #3649, opened 2023-06-01
2.9.0: H5Vol w/ HDF5 1.14.0 Fails Compile on Win


ax3l:
**Describe the bug**
HDF5 with h5Vol does not compile on Windows with:
- ADIOS 2.9.0
- HDF5 1.14.0

**To Reproduce**
See https://github.com/conda-forge/adios2-feedstock/pull/71

Errors:
```
D:\bld\adios2-split_1685654832661\work\source\h5vol\H5VolUtil.h(40): error C2146: syntax error: missing ')' before identifier 'ndims'
D:\bld\adios2-split_1685654832661\work\source\h5vol\H5VolUtil.h(40): error C2081: 'uint': name in formal parameter list illegal
D:\bld\adios2-split_1685654832661\work\source\h5vol\H5VolUtil.h(40): error C2061: syntax error: identifier 'ndims'
D:\bld\adios2-split_1685654832661\work\source\h5vol\H5VolUtil.h(40): error C2059: syntax error: ';'
D:\bld\adios2-split_1685654832661\work\source\h5vol\H5VolUtil.h(40): error C2059: syntax error: ')'
D:\bld\adios2-split_1685654832661\work\source\h5vol\H5VolUtil.h(42): error C2146: syntax error: missing ')' before identifier 'ndims'
D:\bld\adios2-split_1685654832661\work\source\h5vol\H5VolUtil.h(42): error C2081: 'uint': name in formal parameter list illegal
D:\bld\adios2-split_1685654832661\work\source\h5vol\H5VolUtil.h(42): error C2061: syntax error: identifier 'ndims'
D:\bld\adios2-split_1685654832661\work\source\h5vol\H5VolUtil.h(42): error C2059: syntax error: ';'
D:\bld\adios2-split_1685654832661\work\source\h5vol\H5VolUtil.h(42): error C2059: syntax error: ')'
D:\bld\adios2-split_1685654832661\work\source\h5vol\H5VolReadWrite.c(19): fatal error C1083: Cannot open include file: 'unistd.h': No such file or directory
NMAKE : fatal error U1077: 'D:\bld\adios2-split_1685654832661\_build_env\Library\bin\cmake.exe' : return code '0x2'
```

**Expected behavior**
Compile.

**Desktop (please complete the following information):**
 - OS/Platform: Windows
 - Build: MSVC 19.29.30148.0 w/ latest CMake and NMake

**Additional context**
[...]

PR #3652, opened 2023-06-02
work around  with include <unistd.h> on  windows

guj:
to fix the issue reported here:  
https://github.com/ornladios/ADIOS2/issues/3649#issuecomment-1572840107

### DIFF
--- a/source/h5vol/CMakeLists.txt
+++ b/source/h5vol/CMakeLists.txt
@@ -9,6 +9,21 @@ if(CMAKE_C_COMPILER_ID MATCHES "^(GNU)$")
   string(APPEND CMAKE_C_FLAGS " -Wno-format-zero-length")
 endif()
 
+include(CheckIncludeFiles)
+check_include_files(unistd.h HAVE_UNISTD_H)
+check_include_files(windows.h HAVE_WINDOWS_H)
+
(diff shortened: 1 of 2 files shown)


source/h5vol/CMakeLists.txt
```diff
@@ -9,6 +9,21 @@ if(CMAKE_C_COMPILER_ID MATCHES "^(GNU)$")
   string(APPEND CMAKE_C_FLAGS " -Wno-format-zero-length")
 endif()
 
+include(CheckIncludeFiles)
+check_include_files(unistd.h HAVE_UNISTD_H)
+check_include_files(windows.h HAVE_WINDOWS_H)
+
+message("- HAVE_UNISTD?  ${HAVE_UNISTD_H} ")
```
eisenhauer, line 16: I'd probably kill these "message" lines.  The check_include_files will include verbosity about checking and whether or not it was found.  No reason to add yet another line to the two lines of output that are already generated by that...

eisenhauer (changes requested):

source/h5vol/CMakeLists.txt
guj: Okay!

eisenhauer (approved):

ax3l:
@vicentebolea can you please backport this to `release_29`? :)

PR #3660, opened 2023-06-13
An effort to make profiler more user friendly

guj:
renamed profiler entries in BP5

added bytes account for posix  transport


eisenhauer (approved):
Looks reasonable...

Issue #3646, opened 2023-06-01
2.9.0: OSX Python Pickle Issue


ax3l:
**Describe the bug**
This is a Python regression in 2.9.0, first seen by @liangwang0734 in https://github.com/conda-forge/adios2-feedstock/pull/71

This seems to be partly a regression from https://github.com/ornladios/ADIOS2/commit/a0d458676f7e6c6296a3b4176c2f08004efe2905 possibly due to the [change](https://docs.python.org/3/library/multiprocessing.html#contexts-and-start-methods) in multiprocessing on OS X since Python 3.8. Searching the error `TypeError: cannot pickle '_io.TextIOWrapper' object` gives a bunch of discussions.

cc @vicentebolea - thank you for offering to help with this

**To Reproduce**
Build and test ADIOS2 as in https://github.com/conda-forge/adios2-feedstock/pull/71

**Desktop (please complete the following information):**
 - OS/Platform: OSX x86_64 for CPython 3.8-3.11 (all)
 - Build: **LLVM** Clang 14, CMake (latest), shared build

**Additional context**
This bug prevents us from releasing a new ADIOS2 release on conda-forge.

ax3l:
ping @dmitry-ganyushin 

Issue is in https://github.com/ornladios/ADIOS2/blob/master/testing/adios2/engine/dataman/TestDataMan1D.py with macOS

ax3l:
@dmitry-ganyushin we suspect this is from a0d458676f7e6c6296a3b4176c2f08004efe2905

dmitry-ganyushin:
I am looking into that. It is only OSX specific? I made a fast test with python 3.9.16 ubuntu, everything is fine.

dmitry-ganyushin:
Python 3.11.3 + Ubuntu is fine

pnorbert:
@vicentebolea Can we please have ZeroMQ, DataMan, HDF5 and some compression library on for the macos tests?

dmitry-ganyushin:
Fixed

ax3l:
Thanks a lot, @dmitry-ganyushin!

PR #3665, opened 2023-06-14
Fix for the issue #3646

dmitry-ganyushin:
It looks like the default for macos and python 3.8+ "spawn" method does not work on macos, switch to "fork" for all situations. https://bugs.python.org/issue33725


eisenhauer (commented):
I'm curious why you came to the conclusion that spawn doesn't work on macos and you should use fork.  At various points in the issue that you link to, and in a followup one at https://github.com/python/cpython/issues/84287, they recommend switching the Python default on macos to Spawn and away from fork because using fork can lead to application crashes.  Do we have more recent recommendations than these?

dmitry-ganyushin:
Discussed in person. A hand test on MacOS was successful; please merge.

eisenhauer:
I thought the outcome of the discussion was that the older sources (2018-2020 issues) were inconsistent.  I.E. Python moved away from 'fork' previously because it didn't work in all situations and recommended spawn, but now we're moving away from spawn because it doesn't work and we're doing back to fork?  My impression was that we were going to look for some more recent information on the proper approach to portable use of multiprocessing (or some other similar package) to accomplish the simple thing that this test is trying to do (just create a subprocess to be the other party to communication.)

anagainaru (commented):
I tend to agree with Greg, the little I read on the internet right now about spawn and fork, it seems spawn is recommended for performance.

I am also confused how we fix the bug Axel reported in #3646 by changing an example code. Is it your recommendation that they always use fork and not spawn? Shouldn't we make sure our library works regardless of what the users choose to use considering we don't control their codes?

pnorbert (dismissed):

dmitry-ganyushin:
Continues in #3687

PR #3667, opened 2023-06-16
ci: remove power8 emulated builds

vicentebolea:
We already test on powerpc machines though our builds in OLCF Ascent.


scottwittenburg (approved):
Looks good, nice to remove a bunch of unused code, now that these are handled on real power arch.

PR #3672, opened 2023-06-20
Adapt libfabric dataplane of SST to Cray CXI provider

franzpoeschel:
This PR is a first step towards trying to make use of libfabric+CXI on Frontier for SST streaming.

It successfully connects to the provider in `init_fabric()` in my tests, and the hello_sst examples finishes without error – but for some reason without loading any data. I will provide a detailed run log later (I lost my earlier logs somehow and a new job is currently enqueued), but essentially `fi_cq_sread` returns one finished task from the queue that reports a successful remote read of zero bytes.

When changing the data size in the helloSst examples, different errors occur (permission violations or (if I remember correctly) error code 90 - message too long) depending on the chosen data size.

As you can see in the diff, the configuration of libfabric is somewhat different from the one used in SST so far.
The most important differences are:

* Use of FI_MR_ENDPOINT is mandatory. I have adapted all calls to `fi_mr_reg()`, but maybe this flag has more implications.
* ~~The old implementation uses FI_MR_BASIC which implies FI_MR_LOCAL. I just noticed that I did not set this flag for some reason. I don't remember if this is necessary or if I just forgot this. I will need to try this out.~~ Setting FI_MR_LOCAL does not make a difference, I pushed a commit containing it.
* Use of FI_PROGRESS_MANUAL is mandatory (contradicting the libfabric documentation that says that AUTO should always be supported). I don't really know the implications of this. From the libfabric documentation of this parameter that I found, this reads more like a performance implication, turning `fi_cq_sread()` into a blocking/synchronous operation (it is called without a timeout in ADIOS2).

~~Note that I did not test the latest two commits (cleanup and documentation) yet, but their changes are not major.~~

Most things in this PR are currently hardcoded specifically for the requirements of the CXI provider.

Once I get a job on the system again, I will try to be a bit more specific in some things and answer some questions above, and provide a logfile.

cc @pnorbert @eisenhauer 

I used this submission script in my tests:

```bash
#!/bin/bash
#SBATCH -N 2
#SBATCH --ntasks-per-node=1
#SBATCH --cpus-per-task=1
#SBATCH -t 1:00
#SBATCH -A <project id>
#SBATCH -o out.txt
#SBATCH -e err.txt
#SBATCH --network=single_node_vni,job_vni

export FABRIC_IFACE=cxi2
export SstVerbose=5
echo Allocating jobs
srun --network=single_node_vni,job_vni -N 1 --ntasks=1 ../build/bin/hello_sstWriter > sst1.txt 2>&1 &
echo Allocated job 1
srun --network=single_node_vni,job_vni -N 1 --ntasks=1 ../build/bin/hello_sstReader > sst2.txt 2>&1 &
echo Allocated job 2
wait
echo Completed
```

Current diff: https://github.com/ornladios/ADIOS2/compare/release_29...franzpoeschel:ADIOS2:libfabric-cray

### DIFF
--- a/source/adios2/toolkit/sst/dp/rdma_dp.c
+++ b/source/adios2/toolkit/sst/dp/rdma_dp.c
@@ -20,8 +20,9 @@
 #include <rdma/fi_rma.h>
 
 #ifdef SST_HAVE_CRAY_CXI
-// Needs to be included before rdma/fi_cxi_ext.h
 #include <stdbool.h>
+// This comment prevents clang-format from reordering these includes.
+// The CXI extension header requires the bool header, but does not include it on its own.
 #include <rdma/fi_cxi_ext.h>
 #endif
 
--- a/source/adios2/toolkit/sst/dp/rdma_dp.c
+++ b/source/adios2/toolkit/sst/dp/rdma_dp.c
@@ -19,6 +19,12 @@
 #include <rdma/fi_endpoint.h>
 #include <rdma/fi_rma.h>
 
+#ifdef SST_HAVE_CRAY_CXI
+// Needs to be included before rdma/fi_cxi_ext.h
+#include <stdbool.h>
+#include <rdma/fi_cxi_ext.h>
+#endif
+
 #if defined(__has_feature)
 #if __has_feature(thread_sanitizer)
 #define NO_SANITIZE_THREAD __attribute__((no_sanitize("thread")))
@@ -130,6 +136,9 @@ struct fabric_state
     struct fid_cq *cq_signal;
     struct fid_av *av;
     pthread_t listener;
+#ifdef SST_HAVE_CRAY_CXI
+    struct cxi_auth_key *cxi_auth_key;
+#endif
 #ifdef SST_HAVE_CRAY_DRC
     drc_info_handle_t drc_info;
     uint32_t credential;
@@ -173,13 +182,24 @@ struct fabric_state
  *   plane would replace one or both of these with RDMA functionality.
  */
 
+static char const *get_preferred_domain(struct _SstParams *Params)
+{
+    if (Params->DataInterface)
+    {
+        return Params->DataInterface;
+    }
+    else
+    {
+        return getenv("FABRIC_IFACE");
+    }
+}
+
 static void init_fabric(struct fabric_state *fabric, struct _SstParams *Params, CP_Services Svcs,
-                        void *CP_Stream)
+                        void *CP_Stream, char const *ifname)
 {
     struct fi_info *hints, *info, *originfo, *useinfo;
     struct fi_av_attr av_attr = {FI_AV_UNSPEC};
     struct fi_cq_attr cq_attr = {0};
-    char *ifname;
     int result;
 
     hints = fi_allocinfo();
@@ -187,24 +207,68 @@ static void init_fabric(struct fabric_state *fabric, struct _SstParams *Params, 
         FI_MSG | FI_SEND | FI_RECV | FI_REMOTE_READ | FI_REMOTE_WRITE | FI_RMA | FI_READ | FI_WRITE;
     hints->mode =
         FI_CONTEXT | FI_LOCAL_MR | FI_CONTEXT2 | FI_MSG_PREFIX | FI_ASYNC_IOV | FI_RX_CQ_DATA;
-    hints->domain_attr->mr_mode = FI_MR_BASIC;
-    hints->domain_attr->control_progress = FI_PROGRESS_AUTO;
-    hints->domain_attr->data_progress = FI_PROGRESS_AUTO;
     hints->ep_attr->type = FI_EP_RDM;
 
-    if (Params->DataInterface)
+    uint32_t fi_version;
+#ifdef SST_HAVE_CRAY_CXI
+    if (fabric->cxi_auth_key)
     {
-        ifname = Params->DataInterface;
+        fi_version = FI_VERSION(1, 11);
+
+        hints->domain_attr->mr_mode = FI_MR_ENDPOINT;
+        hints->domain_attr->control_progress = FI_PROGRESS_MANUAL;
+        hints->domain_attr->data_progress = FI_PROGRESS_MANUAL;
+
+        // Authentication is needed
+        // TODO: the first ID in SLINGSHOT_SVC_IDS is chosen, but we should
+        // rather choose the one corresponding with the FABRIC_IFACE
+        // example:
+        // SLINGSHOT_SVC_IDS=5,5,5,5
+        // SLINGSHOT_VNIS=1310,1271
+        // SLINGSHOT_DEVICES=cxi0,cxi1,cxi2,cxi3
+        // FABRIC_IFACE=cxi2 (user specified)
+
+        hints->ep_attr->auth_key = malloc(sizeof(struct cxi_auth_key));
+        memcpy(hints->ep_attr->auth_key, fabric->cxi_auth_key, sizeof(struct cxi_auth_key));
+        hints->ep_attr->auth_key_size = sizeof(struct cxi_auth_key);
+
+        hints->domain_attr->auth_key = malloc(sizeof(struct cxi_auth_key));
+        memcpy(hints->domain_attr->auth_key, fabric->cxi_auth_key, sizeof(struct cxi_auth_key));
+        hints->domain_attr->auth_key_size = sizeof(struct cxi_auth_key);
     }
     else
     {
-        ifname = getenv("FABRIC_IFACE");
+        fi_version = FI_VERSION(1, 5);
+
+        hints->domain_attr->mr_mode = FI_MR_BASIC;
+        hints->domain_attr->control_progress = FI_PROGRESS_AUTO;
+        hints->domain_attr->data_progress = FI_PROGRESS_AUTO;
+    }
+#else
+    fi_version = FI_VERSION(1, 5);
+
+    hints->domain_attr->mr_mode = FI_MR_BASIC;
+    hints->domain_attr->control_progress = FI_PROGRESS_AUTO;
+    hints->domain_attr->data_progress = FI_PROGRESS_AUTO;
+#endif
+
+    /*
+     * ifname is passed as a function parameter of init_fabric() if
+     * a provider-specific key was configured and sent to the reader.
+     * Since the key is generally domain-specific, we must use that one in this
+     * case.
+     * The preferred domain is already considered upon key configuration,
+     * so this is fine.
+     */
+    if (!ifname)
+    {
+        ifname = get_preferred_domain(Params);
     }
 
     fabric->info = NULL;
 
     pthread_mutex_lock(&fabric_mutex);
-    fi_getinfo(FI_VERSION(1, 5), NULL, NULL, 0, hints, &info);
+    fi_getinfo(fi_version, NULL, NULL, 0, hints, &info);
     pthread_mutex_unlock(&fabric_mutex);
     if (!info)
     {
@@ -228,7 +292,8 @@ static void init_fabric(struct fabric_state *fabric, struct _SstParams *Params, 
             break;
         }
         if ((((strcmp(prov_name, "verbs") == 0) && info->src_addr) ||
-             (strcmp(prov_name, "gni") == 0) || (strcmp(prov_name, "psm2") == 0)) &&
+             (strcmp(prov_name, "gni") == 0) || (strcmp(prov_name, "psm2") == 0) ||
+             (strcmp(prov_name, "cxi") == 0)) &&
             (!useinfo || !ifname || (strcmp(useinfo->domain_attr->name, ifname) != 0)))
         {
             Svcs->verbose(CP_Stream, DPTraceVerbose,
@@ -238,7 +303,7 @@ static void init_fabric(struct fabric_state *fabric, struct _SstParams *Params, 
             useinfo = info;
         }
         else if (((strstr(prov_name, "verbs") && info->src_addr) || strstr(prov_name, "gni") ||
-                  strstr(prov_name, "psm2")) &&
+                  strstr(prov_name, "psm2") || strstr(prov_name, "cxi")) &&
                  !useinfo)
         {
             Svcs->verbose(CP_Stream, DPTraceVerbose,
@@ -489,6 +554,12 @@ static void fini_fabric(struct fabric_state *fabric, CP_Services Svcs, void *CP_
         free(fabric->ctx);
     }
 
+#ifdef SST_HAVE_CRAY_CXI
+    if (fabric->cxi_auth_key)
+    {
+        free(fabric->cxi_auth_key);
+    }
+#endif
 #ifdef SST_HAVE_CRAY_DRC
     if (Fabric->auth_key)
     {
@@ -677,6 +748,191 @@ static TimestepList GetStep(Rdma_WS_Stream Stream, long Timestep)
     return (Step);
 }
 
+#ifdef SST_HAVE_CRAY_CXI
+static int get_cxi_auth_key_from_env(CP_Services Svcs, void *CP_Stream, struct _SstParams *Params,
+                                     struct cxi_auth_key *key, char **used_device)
+{
+    int vni, first_vni, second_vni, svc_id;
+
+    // Just some safety against faulty strings in string processing.
+    size_t const no_infinite_loops = 10000;
+
+    // struct cxi_auth_key {
+    //     /* The CXI service assigned to the Domain and Endpoints. A CXI
+    //     service
+    //     * is associated with a set of local resource limits, VNIs, and
+    //     Traffic
+    //     * Classes.
+    //     *
+    //     * The svc_id used by an OFI Domain must match all Endpoints belonging
+    //     * to the Domain.
+    //     */
+    //     uint32_t svc_id;
+
+    //     /* The Virtual Network ID (VNI) assigned to the Endpoint. Two
+    //     Endpoints
+    //     * must use the same VNI in order to communicate.
+    //     *
+    //     * Note that while the CXI service may define one or more VNIs which a
+    //     * process can access, an Endpoint is assigned to only one.
+    //     */
+    //     uint16_t vni;
+    // };
+
+    // typical value SLINGSHOT_DEVICES=cxi0,cxi1,cxi2,cxi3
+    char const *slingshot_devices = getenv("SLINGSHOT_DEVICES");
+    char const *preferred_device = get_preferred_domain(Params);
+
+    /*
+     * In the following loop, find out if the preferred_device is found within
+     * the slingshot_devices.
+     * If the preferred_device is NULL, just pick the first.
+     * Upon success, modifies the output parameter used_device and stores
+     * the retrieved device index.
+     */
+    size_t device_index = 0;
+    for (size_t no_infinite_loop_counter = 0;; ++device_index, ++no_infinite_loop_counter)
+    {
+        if (no_infinite_loop_counter > no_infinite_loops)
+        {
+            return EXIT_FAILURE;
+        }
+
+        // Are we at the end of the environment variable?
+        int found_end = 0;
+
+        // Find out the length of the current item in slingshot_devices.
+        size_t find_end_of_current_string = 0;
+        for (size_t no_infinite_loop_inner_counter = 0;;
+             ++find_end_of_current_string, ++no_infinite_loop_inner_counter)
+        {
+            if (no_infinite_loop_inner_counter > no_infinite_loops)
+            {
+                return EXIT_FAILURE;
+            }
+
+            switch (slingshot_devices[find_end_of_current_string])
+            {
+            case '\0':
+                found_end = 1;
+                goto break_first_loop;
+            case ',':
+                goto break_first_loop;
+            default:
+                break;
+            }
+        }
+    break_first_loop:;
+        int use_this_device = !preferred_device || (strncmp(preferred_device, slingshot_devices,
+                                                            find_end_of_current_string) == 0);
+        if (use_this_device)
+        {
+            char *construct_used_device = malloc(find_end_of_current_string + 1);
+            memcpy(construct_used_device, slingshot_devices, find_end_of_current_string);
+            construct_used_device[find_end_of_current_string] = '\0';
+            *used_device = construct_used_device;
+            break;
+        }
+        else if (found_end)
+        {
+            return EXIT_FAILURE;
+        }
+        else
+        {
+            // go to next iteration
+            slingshot_devices += find_end_of_current_string + 1;
+        }
+    }
+
+    Svcs->verbose(CP_Stream, DPTraceVerbose, "Found device %s at index %zu\n", *used_device,
+                  device_index);
+
+    // typical value SLINGSHOT_VNIS=4576,4530
+    char const *vni_env_str = getenv("SLINGSHOT_VNIS");
+    if (!vni_env_str)
+    {
+        return EXIT_FAILURE;
+    }
+
+    // typical value SLINGSHOT_SVC_IDS=5,5,5,5
+    char const *svc_ids_env_str = getenv("SLINGSHOT_SVC_IDS");
+    if (!svc_ids_env_str)
+    {
+        return EXIT_FAILURE;
+    }
+
+    {
+        int num_vnis = sscanf(vni_env_str, "%d,%d", &first_vni, &second_vni);
+        switch (num_vnis)
+        {
+        // first VNI is the subjob's VNI
+        case 1:
+            Svcs->verbose(CP_Stream, DPTraceVerbose, "Using first vni.\n");
+            vni = first_vni;
+            break;
+        // if present, the second VNI is the containing job's VNI
+        // the first VNI belongs to the subjob
+        case 2:
+            Svcs->verbose(CP_Stream, DPTraceVerbose, "Using second vni.\n");
+            vni = second_vni;
+            break;
+        default:
+            return EXIT_FAILURE;
+        }
+    }
+
+    {
+        // Pick the service ID according to the device_index found above.
+        for (size_t svc_id_index = 0; svc_id_index < device_index; ++svc_id_index)
+        {
+            for (size_t no_infinite_loop_counter = 0;; ++no_infinite_loop_counter)
+            {
+                if (no_infinite_loop_counter > no_infinite_loops)
+                {
+                    return EXIT_FAILURE;
+                }
+
+                switch (*(svc_ids_env_str++))
+                {
+                case ',':
+                    goto break_second_loop;
+                case '\0':
+                    return EXIT_FAILURE;
+                default:
+                    continue;
+                }
+            }
+        break_second_loop:;
+        }
+
+        int num_svc_ids = sscanf(svc_ids_env_str, "%d", &svc_id);
+        switch (num_svc_ids)
+        {
+        case 1:
+            break;
+        default:
+            return EXIT_FAILURE;
+        }
+    }
+
+    key->vni = vni;
+    key->svc_id = svc_id;
+
+    return EXIT_SUCCESS;
+}
+
+static int get_cxi_auth_key_from_writer(struct cxi_auth_key *key, attr_list WriterContact)
+{
+    long vni;
+    if (!get_long_attr(WriterContact, attr_atom_from_string("vni"), &vni))
+    {
+        return EXIT_FAILURE;
+    }
+    key->vni = (uint16_t)vni;
+    return EXIT_SUCCESS;
+}
+#endif
+
 static DP_RS_Stream RdmaInitReader(CP_Services Svcs, void *CP_Stream, void **ReaderContactInfoPtr,
                                    struct _SstParams *Params, attr_list WriterContact,
                                    SstStats Stats)
@@ -723,6 +979,38 @@ static DP_RS_Stream RdmaInitReader(CP_Services Svcs, void *CP_Stream, void **Rea
         Stream->PreloadAvail = 0;
     }
 
+    char *required_device = NULL;
+#ifdef SST_HAVE_CRAY_CXI
+    struct
+    {
+        struct cxi_auth_key key;
+        int valid;
+    } tagged_key;
+
+    /*
+     * The svc_id of the key must match the device that this particular reader
+     * connects with.
+     * The vni (virtual network ID) must be the same across all communicating
+     * instances (get this from the writer).
+     */
+
+    tagged_key.valid =
+        get_cxi_auth_key_from_env(Svcs, CP_Stream, Params, &tagged_key.key, &required_device);
+
+    if (tagged_key.valid == EXIT_SUCCESS &&
+        get_cxi_auth_key_from_writer(&tagged_key.key, WriterContact) == EXIT_SUCCESS)
+    {
+        Svcs->verbose(CP_Stream, DPSummaryVerbose, "Reader found CXI auth key: %d %d\n",
+                      tagged_key.key.vni, tagged_key.key.svc_id);
+        Stream->Fabric->cxi_auth_key = calloc(1, sizeof(struct cxi_auth_key));
+        memcpy(Stream->Fabric->cxi_auth_key, &tagged_key.key, sizeof(struct cxi_auth_key));
+    }
+    else
+    {
+        Svcs->verbose(CP_Stream, DPSummaryVerbose, "Reader found no CXI auth key\n");
+    }
+#endif
+
 #ifdef SST_HAVE_CRAY_DRC
     int attr_cred, try_left, rc;
     if (!get_int_attr(WriterContact, attr_atom_from_string("RDMA_DRC_KEY"), &attr_cred))
@@ -753,7 +1041,11 @@ static DP_RS_Stream RdmaInitReader(CP_Services Svcs, void *CP_Stream, void **Rea
 
 #endif /* SST_HAVE_CRAY_DRC */
 
-    init_fabric(Stream->Fabric, Stream->Params, Svcs, CP_Stream);
+    init_fabric(Stream->Fabric, Stream->Params, Svcs, CP_Stream, required_device);
+    if (required_device)
+    {
+        free(required_device);
+    }
     if (!Fabric->info)
     {
         Svcs->verbose(CP_Stream, DPCriticalVerbose, "Could not find a valid transport fabric.\n");
@@ -858,6 +1150,42 @@ static DP_WS_Stream RdmaInitWriter(CP_Services Svcs, void *CP_Stream, struct _Ss
 
     Stream->Fabric = calloc(1, sizeof(struct fabric_state));
     Fabric = Stream->Fabric;
+
+    char *required_device = NULL;
+#ifdef SST_HAVE_CRAY_CXI
+    struct
+    {
+        struct cxi_auth_key key;
+        int valid;
+    } tagged_key;
+
+    /*
+     * The svc_id of the key must match the device that this particular writer
+     * connects with.
+     * The vni (virtual network ID) must be the same across all communicating
+     * instances (use the one seen by rank 0).
+     */
+    tagged_key.valid =
+        get_cxi_auth_key_from_env(Svcs, CP_Stream, Params, &tagged_key.key, &required_device);
+
+    // Ensure that all writers use the same virtual network ID
+    SMPI_Bcast(&tagged_key.key.vni, sizeof(tagged_key.key.vni), SMPI_BYTE, 0, comm);
+
+    if (tagged_key.valid == EXIT_SUCCESS)
+    {
+        Svcs->verbose(CP_Stream, DPSummaryVerbose, "Writer found CXI auth key: %d %d\n",
+                      tagged_key.key.vni, tagged_key.key.svc_id);
+
+        set_long_attr(DPAttrs, attr_atom_from_string("vni"), tagged_key.key.vni);
+        Stream->Fabric->cxi_auth_key = calloc(1, sizeof(struct cxi_auth_key));
+        memcpy(Stream->Fabric->cxi_auth_key, &tagged_key.key, sizeof(struct cxi_auth_key));
+    }
+    else
+    {
+        Svcs->verbose(CP_Stream, DPSummaryVerbose, "Writer found no CXI auth key");
+    }
+#endif
+
 #ifdef SST_HAVE_CRAY_DRC
     int try_left, rc;
     if (Stream->Rank == 0)
@@ -901,7 +1229,11 @@ static DP_WS_Stream RdmaInitWriter(CP_Services Svcs, void *CP_Stream, struct _Ss
     set_long_attr(DPAttrs, attr_atom_from_string("RDMA_DRC_CRED"), attr_cred);
 #endif /* SST_HAVE_CRAY_DRC */
 
-    init_fabric(Stream->Fabric, Params, Svcs, CP_Stream);
+    init_fabric(Stream->Fabric, Params, Svcs, CP_Stream, required_device);
+    if (required_device)
+    {
+        free(required_device);
+    }
     Fabric = Stream->Fabric;
     if (!Fabric->info)
     {
@@ -1735,7 +2067,7 @@ static struct _CP_DP_Interface RdmaDPInterface = {0};
 static int RdmaGetPriority(CP_Services Svcs, void *CP_Stream, struct _SstParams *Params)
 {
     struct fi_info *hints, *info, *originfo;
-    char *ifname;
+    char const *ifname;
     char *forkunsafe;
     int Ret = -1;
 
@@ -1744,19 +2076,40 @@ static int RdmaGetPriority(CP_Services Svcs, void *CP_Stream, struct _SstParams 
         FI_MSG | FI_SEND | FI_RECV | FI_REMOTE_READ | FI_REMOTE_WRITE | FI_RMA | FI_READ | FI_WRITE;
     hints->mode =
         FI_CONTEXT | FI_LOCAL_MR | FI_CONTEXT2 | FI_MSG_PREFIX | FI_ASYNC_IOV | FI_RX_CQ_DATA;
-    hints->domain_attr->mr_mode = FI_MR_BASIC;
-    hints->domain_attr->control_progress = FI_PROGRESS_AUTO;
-    hints->domain_attr->data_progress = FI_PROGRESS_AUTO;
     hints->ep_attr->type = FI_EP_RDM;
 
-    if (Params->DataInterface)
+    char const *vni_env_str = getenv("SLINGSHOT_VNIS");
+
+    uint32_t fi_version;
+    if (vni_env_str)
     {
-        ifname = Params->DataInterface;
+        // try fishing for the CXI provider
+        Svcs->verbose(CP_Stream, DPSummaryVerbose,
+                      "RDMA Dataplane trying to check for an available CXI "
+                      "provider since environment variable SLINGSHOT_VNIS is "
+                      "defined (value: '%s').\n",
+                      vni_env_str);
+        fi_version = FI_VERSION(1, 11);
+
+        hints->domain_attr->mr_mode = FI_MR_ENDPOINT;
+        hints->domain_attr->control_progress = FI_PROGRESS_MANUAL;
+        hints->domain_attr->data_progress = FI_PROGRESS_MANUAL;
     }
     else
     {
-        ifname = getenv("FABRIC_IFACE");
+        Svcs->verbose(CP_Stream, DPSummaryVerbose,
+                      "RDMA Dataplane trying to check for an available non-CXI "
+                      "provider since environment variable SLINGSHOT_VNIS is "
+                      "not defined.\n");
+
+        fi_version = FI_VERSION(1, 5);
+
+        hints->domain_attr->mr_mode = FI_MR_BASIC;
+        hints->domain_attr->control_progress = FI_PROGRESS_AUTO;
+        hints->domain_attr->data_progress = FI_PROGRESS_AUTO;
     }
+
+    ifname = get_preferred_domain(Params);
 
     forkunsafe = getenv("FI_FORK_UNSAFE");
     if (!forkunsafe)
@@ -1765,7 +2118,7 @@ static int RdmaGetPriority(CP_Services Svcs, void *CP_Stream, struct _SstParams 
     }
 
     pthread_mutex_lock(&fabric_mutex);
-    fi_getinfo(FI_VERSION(1, 5), NULL, NULL, 0, hints, &info);
+    fi_getinfo(fi_version, NULL, NULL, 0, hints, &info);
     pthread_mutex_unlock(&fabric_mutex);
     fi_freeinfo(hints);
 
@@ -1783,6 +2136,10 @@ static int RdmaGetPriority(CP_Services Svcs, void *CP_Stream, struct _SstParams 
 
         prov_name = info->fabric_attr->prov_name;
         domain_name = info->domain_attr->name;
+        Svcs->verbose(CP_Stream, DPPerStepVerbose,
+                      "[RdmaGetPriority] Seeing and evaluating fabric with "
+                      "provider: '%s', domain: '%s'\n",
+                      prov_name, domain_name);
         if (ifname && strcmp(ifname, domain_name) == 0)
         {
             Svcs->verbose(CP_Stream, DPPerStepVerbose,
@@ -1793,7 +2150,7 @@ static int RdmaGetPriority(CP_Services Svcs, void *CP_Stream, struct _SstParams 
             break;
         }
         if ((strstr(prov_name, "verbs") && info->src_addr) || strstr(prov_name, "gni") ||
-            strstr(prov_name, "psm2"))
+            strstr(prov_name, "psm2") || strstr(prov_name, "cxi"))
         {
 
             Svcs->verbose(CP_Stream, DPPerStepVerbose,
--- a/source/adios2/toolkit/sst/dp/rdma_dp.c
+++ b/source/adios2/toolkit/sst/dp/rdma_dp.c
@@ -120,6 +120,7 @@ struct fabric_state
     struct fi_info *info;
     // struct fi_info *linfo;
     int local_mr_req;
+    int mr_virt_addr; /* Stores if the mr_mode includes FI_MR_VIRT_ADDR */
     int rx_cq_data;
     size_t addr_len;
     size_t msg_prefix_size;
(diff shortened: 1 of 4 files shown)


eisenhauer:
Thanks Franz, happy to help look at this too.  Might not have time until later this week, but I'll be watching in case you get more info before then.

franzpoeschel:
Thanks Greg. I am also willing to put time into this, but this concerns several systems that I don't really know about, so I'm rather limited in what I can do, and I'm currently somewhat stuck on this issue of "loading zero data".

The output of the writer is the following:

```
srun: warning: can't run 1 processes on 2 nodes, setting nnodes to 1
Writer 0 (0x3078d0): Sst set to use sockets as a Control Transport
Provider: 'cxi', domain: 'cxi0'
Provider: 'cxi', domain: 'cxi1'
Provider: 'cxi', domain: 'cxi2'
DP Writer 0 (0x3078d0): RDMA Dataplane found the requested interface cxi2, provider type cxi.
DP Writer 0 (0x3078d0): RDMA Dataplane evaluating viability, returning priority 100
DP Writer 0 (0x3078d0): Considering DataPlane "evpath" for possible use, priority is 1
DP Writer 0 (0x3078d0): Considering DataPlane "rdma" for possible use, priority is 100
DP Writer 0 (0x3078d0): Selecting DataPlane "rdma", priority 100 for use
DP Writer 0 (0x3078d0): ignoring fabric cxi because it's not of a supported type. It may work to force this fabric to be used by setting FABRIC_IFACE to cxi0, but it may not be stable or performant.
DP Writer 0 (0x3078d0): ignoring fabric cxi because it's not of a supported type. It may work to force this fabric to be used by setting FABRIC_IFACE to cxi1, but it may not be stable or performant.
DP Writer 0 (0x3078d0): using interface set by FABRIC_IFACE.
DP Writer 0 (0x3078d0): Fabric parameters to use at fabric initialization: fi_info:
    caps: [ FI_MSG, FI_RMA, FI_READ, FI_WRITE, FI_RECV, FI_SEND, FI_REMOTE_READ, FI_REMOTE_WRITE, FI_MULTI_RECV, FI_TRIGGER, FI_SHARED_AV ]
    mode: [  ]
    addr_format: FI_ADDR_CXI_COMPAT
    src_addrlen: 4
    dest_addrlen: 0
    src_addr: fi_addr_cxi://0x002205ff
    dest_addr: (null)
    handle: (nil)
    fi_tx_attr:
        caps: [ FI_MSG, FI_RMA, FI_READ, FI_WRITE, FI_SEND, FI_TRIGGER, FI_SHARED_AV ]
        mode: [  ]
        op_flags: [  ]
        msg_order: [  ]
        comp_order: [ FI_ORDER_NONE ]
        inject_size: 192
        size: 256
        iov_limit: 1
        rma_iov_limit: 1
        tclass: 0x0
    fi_rx_attr:
        caps: [ FI_MSG, FI_RMA, FI_RECV, FI_REMOTE_READ, FI_REMOTE_WRITE, FI_MULTI_RECV, FI_TRIGGER, FI_SHARED_AV ]
        mode: [  ]
        op_flags: [  ]
        msg_order: [ FI_ORDER_WAW, FI_ORDER_SAS, FI_ORDER_RMA_WAW, FI_ORDER_ATOMIC_RAR, FI_ORDER_ATOMIC_RAW, FI_ORDER_ATOMIC_WAR, FI_ORDER_ATOMIC_WAW ]
        comp_order: [ FI_ORDER_NONE ]
        total_buffered_recv: 0
        size: 1024
        iov_limit: 1
    fi_ep_attr:
        type: FI_EP_RDM
        protocol: FI_PROTO_CXI_COMPAT
        protocol_version: 1
        max_msg_size: 1073741824
        msg_prefix_size: 0
        max_order_raw_size: 18446744073709551615
        max_order_war_size: 18446744073709551615
        max_order_waw_size: 18446744073709551615
        mem_tag_format: 0x0000aaaaaaaaaaaa
        tx_ctx_cnt: 0
        rx_ctx_cnt: 0
        auth_key_size: 8
    fi_domain_attr:
        domain: 0x0
        name: cxi2
        threading: FI_THREAD_SAFE
        control_progress: FI_PROGRESS_MANUAL
        data_progress: FI_PROGRESS_MANUAL
        resource_mgmt: FI_RM_ENABLED
        av_type: FI_AV_UNSPEC
        mr_mode: [ FI_MR_VIRT_ADDR, FI_MR_ALLOCATED, FI_MR_PROV_KEY, FI_MR_ENDPOINT ]
        mr_key_size: 4
        cq_data_size: 8
        cq_cnt: 32
        ep_cnt: 128
        tx_ctx_cnt: 256
        rx_ctx_cnt: 256
        max_ep_tx_ctx: 256
        max_ep_rx_ctx: 256
        max_ep_stx_ctx: 0
        max_ep_srx_ctx: 0
        cntr_cnt: 16
        mr_iov_limit: 1
        caps: [  ]
        mode: [  ]
        auth_key_size: 8
        max_err_data: 0
        mr_cnt: 100
        tclass: 0x0
    fi_fabric_attr:
        name: cxi
        prov_name: cxi
        prov_version: 0.0
        api_version: 1.11
    nic:
        fi_device_attr:
            name: cxi2
            device_id: 0x501
            device_version: 2
            vendor_id: 0x17db
            driver: cxi_core
            firmware: (null)
        fi_bus_attr:
            bus_type: FI_BUS_PCI
            fi_pci_attr:
                domain_id: 0
                bus_id: 213
                device_id: 0
                function_id: 0
        fi_link_attr:
            address: 0x1102
            mtu: 2112
            speed: 200000000000
            state: FI_LINK_UP
            network_type: HPC Ethernet

Using second vni.
Writer found CXI auth key: 10059 5
DP Writer 0 (0x3078d0): Fabric Parameters:
fi_info:
    caps: [ FI_MSG, FI_RMA, FI_READ, FI_WRITE, FI_RECV, FI_SEND, FI_REMOTE_READ, FI_REMOTE_WRITE, FI_MULTI_RECV, FI_TRIGGER, FI_SHARED_AV ]
    mode: [  ]
    addr_format: FI_ADDR_CXI_COMPAT
    src_addrlen: 4
    dest_addrlen: 0
    src_addr: fi_addr_cxi://0x002205ff
    dest_addr: (null)
    handle: (nil)
    fi_tx_attr:
        caps: [ FI_MSG, FI_RMA, FI_READ, FI_WRITE, FI_SEND, FI_TRIGGER, FI_SHARED_AV ]
        mode: [  ]
        op_flags: [  ]
        msg_order: [  ]
        comp_order: [ FI_ORDER_NONE ]
        inject_size: 192
        size: 256
        iov_limit: 1
        rma_iov_limit: 1
        tclass: 0x0
    fi_rx_attr:
        caps: [ FI_MSG, FI_RMA, FI_RECV, FI_REMOTE_READ, FI_REMOTE_WRITE, FI_MULTI_RECV, FI_TRIGGER, FI_SHARED_AV ]
        mode: [  ]
        op_flags: [  ]
        msg_order: [ FI_ORDER_WAW, FI_ORDER_SAS, FI_ORDER_RMA_WAW, FI_ORDER_ATOMIC_RAR, FI_ORDER_ATOMIC_RAW, FI_ORDER_ATOMIC_WAR, FI_ORDER_ATOMIC_WAW ]
        comp_order: [ FI_ORDER_NONE ]
        total_buffered_recv: 0
        size: 1024
        iov_limit: 1
    fi_ep_attr:
        type: FI_EP_RDM
        protocol: FI_PROTO_CXI_COMPAT
        protocol_version: 1
        max_msg_size: 1073741824
        msg_prefix_size: 0
        max_order_raw_size: 18446744073709551615
        max_order_war_size: 18446744073709551615
        max_order_waw_size: 18446744073709551615
        mem_tag_format: 0x0000aaaaaaaaaaaa
        tx_ctx_cnt: 0
        rx_ctx_cnt: 0
        auth_key_size: 8
    fi_domain_attr:
        domain: 0x0
        name: cxi2
        threading: FI_THREAD_SAFE
        control_progress: FI_PROGRESS_MANUAL
        data_progress: FI_PROGRESS_MANUAL
        resource_mgmt: FI_RM_ENABLED
        av_type: FI_AV_UNSPEC
        mr_mode: [ FI_MR_VIRT_ADDR, FI_MR_ALLOCATED, FI_MR_PROV_KEY, FI_MR_ENDPOINT ]
        mr_key_size: 4
        cq_data_size: 8
        cq_cnt: 32
        ep_cnt: 128
        tx_ctx_cnt: 256
        rx_ctx_cnt: 256
        max_ep_tx_ctx: 256
        max_ep_rx_ctx: 256
        max_ep_stx_ctx: 0
        max_ep_srx_ctx: 0
        cntr_cnt: 16
        mr_iov_limit: 1
        caps: [  ]
        mode: [  ]
        auth_key_size: 8
        max_err_data: 0
        mr_cnt: 100
        tclass: 0x0
    fi_fabric_attr:
        name: cxi
        prov_name: cxi
        prov_version: 0.0
        api_version: 1.11
    nic:
        fi_device_attr:
            name: cxi2
            device_id: 0x501
            device_version: 2
            vendor_id: 0x17db
            driver: cxi_core
            firmware: (null)
        fi_bus_attr:
            bus_type: FI_BUS_PCI
            fi_pci_attr:
                domain_id: 0
                bus_id: 213
                device_id: 0
                function_id: 0
        fi_link_attr:
            address: 0x1102
            mtu: 2112
            speed: 200000000000
            state: FI_LINK_UP
            network_type: HPC Ethernet

Writer 0 (0x3078d0): Opening Stream "helloSst"
Writer 0 (0x3078d0): Writer stream params are:
Param -   RegistrationMethod=File
Param -   RendezvousReaderCount=1
Param -   QueueLimit=0 (unlimited)
Param -   QueueFullPolicy=Block
Param -   StepDistributionMode=StepsAllToAll
Param -   DataTransport=rdma
Param -   ControlTransport=sockets
Param -   NetworkInterface=(default)
Param -   ControlInterface=(default to NetworkInterface if applicable)
Param -   DataInterface=(default to NetworkInterface if applicable)
Param -   CompressionMethod=None
Param -   CPCommPattern=Min
Param -   MarshalMethod=BP5
Param -   FirstTimestepPrecious=False
Param -   IsRowMajor=1  (not user settable) 
Param -   OpenTimeoutSecs=60 (seconds)
Param -   SpeculativePreloadMode=Auto
Param -   SpecAutoNodeThreshold=1
Param -   ControlModule=select
Writer 0 (0x3078d0): Stream "helloSst" waiting for 1 readers
Writer 0 (0x3078d0): Beginning writer-side reader open protocol
DP Writer 0 (0x3078d0): Received contact info for RS_Stream 0x3adaa0, WSR Rank 0
Writer 0 (0x3078d0): Setting SpeculativePreload ON for new reader
Writer 0 (0x3078d0): My oldest timestep was 0, global oldest timestep was 0
Writer 0 (0x3078d0): Finish writer-side reader open protocol for reader 0x3f9a70, reader ready response pending
Writer 0 (0x3078d0): (PID abb4, TID 7fffd3e6af80) Waiting for Reader ready on WSR 0x3f9a70.
Writer 0 (0x3078d0): Reader Activate message received for Stream 0x3f9a70.  Setting state to Established.
Writer 0 (0x3078d0): Parent stream reader count is now 1.
Writer 0 (0x3078d0): Reader ready on WSR 0x3f9a70, Stream established, Starting 0 LastProvided 0.
Writer 0 (0x3078d0): Finish opening Stream "helloSst"
DP Writer 0 (0x3078d0): Providing timestep data with block 0x446550 and access key -50386303
Writer 0 (0x3078d0): Reader 0 status Established has last released 4294967295, last sent 0
Writer 0 (0x3078d0): QueueMaintenance, smallest last released = -1, count = 1
Writer 0 (0x3078d0): Removing dead entries
Writer 0 (0x3078d0): QueueMaintenance complete
Writer 0 (0x3078d0): Sending TimestepMetadata for timestep 0 (ref count 1), one to each reader
Writer 0 (0x3078d0): Sent timestep 0 to reader cohort 0
Writer 0 (0x3078d0): ADDING timestep 0 to sent list for reader cohort 0, READER 0x3f9a70, reference count is now 2
Writer 0 (0x3078d0): PRELOADMODE for timestep 0 non-default for reader , active at timestep 0, mode 1
Writer 0 (0x3078d0): Sending a message to reader 0 (0x307490)
Writer 0 (0x3078d0): SubRef : Writer-side Timestep 0 now has reference count 1, expired 0, precious 0
Writer 0 (0x3078d0): Reader 0 status Established has last released 4294967295, last sent 0
Writer 0 (0x3078d0): QueueMaintenance, smallest last released = -1, count = 1
Writer 0 (0x3078d0): Removing dead entries
Writer 0 (0x3078d0): QueueMaintenance complete
Writer 0 (0x3078d0): SstWriterClose, Sending Close at Timestep 0, one to each reader
Writer 0 (0x3078d0): Working on reader cohort 0
Writer 0 (0x3078d0): Sending a message to reader 0 (0x307490)
Writer 0 (0x3078d0): Reader 0 status Established has last released 4294967295, last sent 0
Writer 0 (0x3078d0): QueueMaintenance, smallest last released = -1, count = 1
Writer 0 (0x3078d0): Removing dead entries
Writer 0 (0x3078d0): QueueMaintenance complete
Writer 0 (0x3078d0): Waiting for timesteps to be released in WriterClose
Writer 0 (0x3078d0): IN TS WAIT, ENTRIES are Timestep 0 (exp 0, Prec 0, Ref 1), Count now 1
Writer 0 (0x3078d0): The timesteps still queued are: 0 
Writer 0 (0x3078d0): Reader Count is 1
Writer 0 (0x3078d0): Reader [0] status is Established
Writer 0 (0x3078d0): Received a release timestep message for timestep 0 from reader cohort 0
Writer 0 (0x3078d0): Got the lock in release timestep
Writer 0 (0x3078d0): Doing dereference sent
Writer 0 (0x3078d0): Reader sent timestep list 0x446330, trying to release 0
Writer 0 (0x3078d0): Reader considering sent timestep 0,trying to release 0
Writer 0 (0x3078d0): SubRef : Writer-side Timestep 0 now has reference count 0, expired 0, precious 0
Writer 0 (0x3078d0): Doing QueueMaint
Writer 0 (0x3078d0): Reader 0 status Established has last released 0, last sent 0
Writer 0 (0x3078d0): QueueMaintenance, smallest last released = 0, count = 1
Writer 0 (0x3078d0): Writer tagging timestep 0 as expired
DP Writer 0 (0x3078d0): Releasing timestep 0
Writer 0 (0x3078d0): Removing dead entries
Writer 0 (0x3078d0): Remove queue Entries removing Timestep 0 (exp 1, Prec 0, Ref 0), Count now 0
Writer 0 (0x3078d0): QueueMaintenance complete
Writer 0 (0x3078d0): Releasing the lock in release timestep
Writer 0 (0x3078d0): Reader Close message received for stream 0x3f9a70.  Setting state to PeerClosed and releasing timesteps.
Writer 0 (0x3078d0): In PeerFailCloseWSReader, releasing sent timesteps
Writer 0 (0x3078d0): Dereferencing all timesteps sent to reader 0x3f9a70
Writer 0 (0x3078d0): DONE DEREFERENCING
Writer 0 (0x3078d0): Moving Reader stream 0x3f9a70 to status PeerClosed
Writer 0 (0x3078d0): Reader 0 status PeerClosed has last released 0, last sent 0
Writer 0 (0x3078d0): QueueMaintenance, smallest last released = LONG_MAX, count = 0
Writer 0 (0x3078d0): Removing dead entries
Writer 0 (0x3078d0): QueueMaintenance complete
Writer 0 (0x3078d0): 
Stream "helloSst" (0x3078d0) summary info:
Writer 0 (0x3078d0): 	Duration (secs) = 0.100589
Writer 0 (0x3078d0): 	Timesteps Created = 1
Writer 0 (0x3078d0): 	Timesteps Delivered = 1
Writer 0 (0x3078d0): 
Writer 0 (0x3078d0): All timesteps are released in WriterClose
Writer 0 (0x3078d0): Destroying stream 0x3078d0, name helloSst
DP Writer 0 (0x3078d0): Releasing reader-specific state for remaining readers.
DP Writer 0 (0x3078d0): Releasing remaining timesteps.
DP Writer 0 (0x3078d0): Tearing down RDMA state on writer.
Writer 0 (0x3078d0): Reference count now zero, Destroying process SST info cache
Writer 0 (0x3078d0): Freeing LastCallList
Writer 0 (0x7fffffff2f18): SstStreamDestroy successful, returning
```

The reader:

```
srun: warning: can't run 1 processes on 2 nodes, setting nnodes to 1
Reader 0 (0x307490): Sst set to use sockets as a Control Transport
Reader 0 (0x307490): Looking for writer contact in file helloSst.sst, with timeout 60 secs
Reader 0 (0x307490): Waiting for writer DPResponse message in SstReadOpen("helloSst")
Reader 0 (0x307490): finished wait writer DPresponse message in read_open, WRITER is using "rdma" DataPlane
Provider: 'cxi', domain: 'cxi0'
Provider: 'cxi', domain: 'cxi1'
Provider: 'cxi', domain: 'cxi2'
DP Reader 0 (0x307490): RDMA Dataplane found the requested interface cxi2, provider type cxi.
DP Reader 0 (0x307490): RDMA Dataplane evaluating viability, returning priority 100
DP Reader 0 (0x307490): Prefered dataplane name is "rdma"
DP Reader 0 (0x307490): Considering DataPlane "evpath" for possible use, priority is 1
DP Reader 0 (0x307490): Considering DataPlane "rdma" for possible use, priority is 100
DP Reader 0 (0x307490): Selecting DataPlane "rdma" (preferred) for use
DP Reader 0 (0x307490): ignoring fabric cxi because it's not of a supported type. It may work to force this fabric to be used by setting FABRIC_IFACE to cxi0, but it may not be stable or performant.
DP Reader 0 (0x307490): ignoring fabric cxi because it's not of a supported type. It may work to force this fabric to be used by setting FABRIC_IFACE to cxi1, but it may not be stable or performant.
DP Reader 0 (0x307490): using interface set by FABRIC_IFACE.
DP Reader 0 (0x307490): Fabric parameters to use at fabric initialization: fi_info:
    caps: [ FI_MSG, FI_RMA, FI_READ, FI_WRITE, FI_RECV, FI_SEND, FI_REMOTE_READ, FI_REMOTE_WRITE, FI_MULTI_RECV, FI_TRIGGER, FI_SHARED_AV ]
    mode: [  ]
    addr_format: FI_ADDR_CXI_COMPAT
    src_addrlen: 4
    dest_addrlen: 0
    src_addr: fi_addr_cxi://0x002325ff
    dest_addr: (null)
    handle: (nil)
    fi_tx_attr:
        caps: [ FI_MSG, FI_RMA, FI_READ, FI_WRITE, FI_SEND, FI_TRIGGER, FI_SHARED_AV ]
        mode: [  ]
        op_flags: [  ]
        msg_order: [  ]
        comp_order: [ FI_ORDER_NONE ]
        inject_size: 192
        size: 256
        iov_limit: 1
        rma_iov_limit: 1
        tclass: 0x0
    fi_rx_attr:
        caps: [ FI_MSG, FI_RMA, FI_RECV, FI_REMOTE_READ, FI_REMOTE_WRITE, FI_MULTI_RECV, FI_TRIGGER, FI_SHARED_AV ]
        mode: [  ]
        op_flags: [  ]
        msg_order: [ FI_ORDER_WAW, FI_ORDER_SAS, FI_ORDER_RMA_WAW, FI_ORDER_ATOMIC_RAR, FI_ORDER_ATOMIC_RAW, FI_ORDER_ATOMIC_WAR, FI_ORDER_ATOMIC_WAW ]
        comp_order: [ FI_ORDER_NONE ]
        total_buffered_recv: 0
        size: 1024
        iov_limit: 1
    fi_ep_attr:
        type: FI_EP_RDM
        protocol: FI_PROTO_CXI_COMPAT
        protocol_version: 1
        max_msg_size: 1073741824
        msg_prefix_size: 0
        max_order_raw_size: 18446744073709551615
        max_order_war_size: 18446744073709551615
        max_order_waw_size: 18446744073709551615
        mem_tag_format: 0x0000aaaaaaaaaaaa
        tx_ctx_cnt: 0
        rx_ctx_cnt: 0
        auth_key_size: 8
    fi_domain_attr:
        domain: 0x0
        name: cxi2
        threading: FI_THREAD_SAFE
        control_progress: FI_PROGRESS_MANUAL
        data_progress: FI_PROGRESS_MANUAL
        resource_mgmt: FI_RM_ENABLED
        av_type: FI_AV_UNSPEC
        mr_mode: [ FI_MR_VIRT_ADDR, FI_MR_ALLOCATED, FI_MR_PROV_KEY, FI_MR_ENDPOINT ]
        mr_key_size: 4
        cq_data_size: 8
        cq_cnt: 32
        ep_cnt: 128
        tx_ctx_cnt: 256
        rx_ctx_cnt: 256
        max_ep_tx_ctx: 256
        max_ep_rx_ctx: 256
        max_ep_stx_ctx: 0
        max_ep_srx_ctx: 0
        cntr_cnt: 16
        mr_iov_limit: 1
        caps: [  ]
        mode: [  ]
        auth_key_size: 8
        max_err_data: 0
        mr_cnt: 100
        tclass: 0x0
    fi_fabric_attr:
        name: cxi
        prov_name: cxi
        prov_version: 0.0
        api_version: 1.11
    nic:
        fi_device_attr:
            name: cxi2
            device_id: 0x501
            device_version: 2
            vendor_id: 0x17db
            driver: cxi_core
            firmware: (null)
        fi_bus_attr:
            bus_type: FI_BUS_PCI
            fi_pci_attr:
                domain_id: 0
                bus_id: 213
                device_id: 0
                function_id: 0
        fi_link_attr:
            address: 0x1192
            mtu: 2112
            speed: 200000000000
            state: FI_LINK_UP
            network_type: HPC Ethernet

Reader found CXI auth key: 10059 5
Reader 0 (0x307490): Waiting for writer response message in SstReadOpen("helloSst")
Reader 0 (0x307490): finished wait writer response message in read_open
Reader 0 (0x307490): Opening Reader Stream.
Writer stream params are:
Param -   RegistrationMethod=File
Param -   RendezvousReaderCount=1
Param -   QueueLimit=0 (unlimited)
Param -   QueueFullPolicy=Block
Param -   StepDistributionMode=StepsAllToAll
Param -   DataTransport=rdma
Param -   ControlTransport=sockets
Param -   NetworkInterface=(default)
Param -   ControlInterface=(default to NetworkInterface if applicable)
Param -   DataInterface=(default to NetworkInterface if applicable)
Param -   CompressionMethod=None
Param -   CPCommPattern=Min
Param -   MarshalMethod=BP5
Param -   FirstTimestepPrecious=False
Param -   IsRowMajor=1  (not user settable) 
Param -   OpenTimeoutSecs=60 (seconds)
Param -   SpeculativePreloadMode=Auto
Param -   SpecAutoNodeThreshold=1
Param -   ControlModule=select
Reader 0 (0x307490): Reader stream params are:
Param -   RegistrationMethod=File
Param -   DataTransport=rdma
Param -   ControlTransport=sockets
Param -   NetworkInterface=(default)
Param -   ControlInterface=(default to NetworkInterface if applicable)
Param -   DataInterface=(default to NetworkInterface if applicable)
Param -   AlwaysProvideLatestTimestep=False
Param -   OpenTimeoutSecs=60 (seconds)
Param -   SpeculativePreloadMode=Auto
Param -   SpecAutoNodeThreshold=1
Param -   ControlModule=select
Reader 0 (0x307490): Writer is using Minimum Connection Communication pattern (min)
DP Reader 0 (0x307490): Received contact info for WS_stream 0x3faef0, WSR Rank 0
Reader 0 (0x307490): Sending Reader Activate messages to writer
Reader 0 (0x307490): Finish opening Stream "helloSst", starting with Step number 0
Reader 0 (0x307490): Wait for next metadata after last timestep -1
Reader 0 (0x307490): Waiting for metadata for a Timestep later than TS -1
Reader 0 (0x307490): (PID 1205a, TID 7fffd3e6af80) Stream status is Established
Reader 0 (0x307490): Received a Timestep metadata message for timestep 0, signaling condition
Reader 0 (0x307490): Received a writer close message. Timestep 0 was the final timestep.
Reader 0 (0x307490): Examining metadata for Timestep 0
Reader 0 (0x307490): Returning metadata for Timestep 0
Reader 0 (0x307490): Setting TSmsg to Rootentry value
DP Reader 0 (0x307490): RdmaTimestepArrived with Timestep = 0, PreloadMode = 1
Reader 0 (0x307490): SstAdvanceStep returning Success on timestep 0
DP Reader 0 (0x307490): Performing remote read of Writer Rank 0 at step 0
DP Reader 0 (0x307490): Block address is 0x446550, with a key of -50386303
DP Reader 0 (0x307490): Remote read target is Rank 0 (Offset = 0, Length = 40)
DP Reader 0 (0x307490): Posted RDMA get for Writer Rank 0 for handle 0x441590
DP Reader 0 (0x307490): Rank 0, RdmaWaitForCompletion
DP Reader 0 (0x307490): got completion for request with handle 0x441590 (flags 260).
Incoming variable is of size 10
Reader rank 0 reading 10 floats starting at element 0
Reader 0 (0x307490): Sending ReleaseTimestep message for timestep 0, one to each writer
Reader 0 (0x307490): 
Stream "helloSst" (0x307490) summary info:
Reader 0 (0x307490): 	Duration (secs) = 0.001023
Reader 0 (0x307490): 	Timestep Metadata Received = 1
Reader 0 (0x307490): 	Timesteps Consumed = 1
Reader 0 (0x307490): 	MetadataBytesReceived = 176 (176 bytes)
Reader 0 (0x307490): 	DataBytesReceived = 0 (0 bytes)
Reader 0 (0x307490): 	PreloadBytesReceived = 0 (0 bytes)
Reader 0 (0x307490): 	PreloadTimestepsReceived = 0
Reader 0 (0x307490): 	AverageReadRankFanIn = 1.0
Reader 0 (0x307490): 
Reader 0 (0x307490): Destroying stream 0x307490, name helloSst
DP Reader 0 (0x307490): Tearing down RDMA state on reader.
Reader 0 (0x307490): Reader-side close handler invoked
Reader 0 (0x307490): Reference count now zero, Destroying process SST info cache
Reader 0 (0x307490): Freeing LastCallList
Reader 0 (0x7fffffff2f08): SstStreamDestroy successful, returning
```

eisenhauer:
Are you just concerned about the zero bytes in the "summary info" at the end?  If so, that's more of an informational thing and it appears that it wasn't actually implemented in the original RDMA data plane, so that value never gets updated and you can ignore it.  (I can sort actually implementing it, but perhaps not until Friday because I have ORNL visitors here today and tomorrow.). So the question is whether or not you're getting the right data in the read buffer.

franzpoeschel:
> Are you just concerned about the zero bytes in the "summary info" at the end?

It's not only that, no. In the call to `WaitForAnyPull` that corresponds with the data load request, the `CQEntry` (to my understanding) reports zero loaded data:

```c
static int WaitForAnyPull(CP_Services Svcs, Rdma_RS_Stream Stream)
{
    FabricState Fabric = Stream->Fabric;
    RdmaCompletionHandle Handle_t;
    struct fi_cq_data_entry CQEntry = {0};

    ssize_t rc;
    rc = fi_cq_sread(Fabric->cq_signal, (void *)(&CQEntry), 1, NULL, -1);
    if (rc < 1)
    {
        struct fi_cq_err_entry error;
        fi_cq_readerr(Fabric->cq_signal, &error, 0);
        Svcs->verbose(Stream->CP_Stream, DPCriticalVerbose,
                      "failure while waiting for completions WaitForAnyPull "
                      "(%d (%s - %s)).\n",
                      rc, fi_strerror(error.err),
                      fi_cq_strerror(Fabric->cq_signal, error.err,
                                     error.err_data, NULL, error.len));
        return 0;
    }
    else
    {
        Svcs->verbose(
            Stream->CP_Stream, DPTraceVerbose,
            "got completion for request with handle %p (flags %li).\n",
            CQEntry.op_context, CQEntry.flags);
        Handle_t = (RdmaCompletionHandle)CQEntry.op_context;
        Handle_t->Pending--;
        Stream->PendingReads--;

        // TODO: maybe reuse this memory registration
        if (Fabric->local_mr_req)
        {
            fi_close((struct fid *)Handle_t->LocalMR);
        }
    }
    return 1;
}
```

Latching onto this with GDB:

```
(gdb) p CQEntry
$2 = {op_context = 0x441d40, flags = 260, len = 0, buf = 0x0, data = 0}
(gdb) p *Handle_t
$4 = {LocalMR = 0x0, CPStream = 0x3adaf0, Buffer = 0x440a50, Length = 40, Rank = 0, Pending = 1, PreloadBuffer = 0x0}
(gdb) p *(float*)Handle_t->Buffer
$6 = 0
```

`Length = 40` should be the load call that attempted to read 10 floating point values. 260 is `FI_READ | FI_RMA`.

eisenhauer:
Ah, OK, puzzling.  I have limited time today and tomorrow, but I'll look when I can and let you know if I have some kind of insight...

eisenhauer:
OK, I'm trying to run this, but I'm getting somewhat different output on the reader (sst2.txt):
DP Reader 0 (0x2327a0): Posted RDMA get for Writer Rank 0 for handle 0x8ccfe0
DP Reader 0 (0x2327a0): Rank 0, RdmaWaitForCompletion
DP Reader 0 (0x2327a0): failure while waiting for completions WaitForAnyPull (-259 (Input/output error - PERM_VIOLATION)).
Incoming variable is of size 10

That there's a failure in WaitForAnyPull is consistent with not getting any data, but I haven't sorted out a possible cause yet.  Also trying to sort FI_LOG_LEVEL to get internal info from libfabric.   I thought I had output from that, but then I switched to starting from your script and I don't seem to be getting anything.  Will keep poking.

franzpoeschel:
Thank you for trying this out

> OK, I'm trying to run this, but I'm getting somewhat different output on the reader (sst2.txt): DP Reader 0 (0x2327a0): Posted RDMA get for Writer Rank 0 for handle 0x8ccfe0 DP Reader 0 (0x2327a0): Rank 0, RdmaWaitForCompletion DP Reader 0 (0x2327a0): failure while waiting for completions WaitForAnyPull (-259 (Input/output error - PERM_VIOLATION)). Incoming variable is of size 10

I did get this PERM_VIOLATION error in two different situations:

1. Without using `fi_mr_bind()` and `fi_mr_enable()` (as required by the `FI_MR_ENDPOINT` flag, I always got this error
2. With the implementation as it currently is, I get the error when increasing the vector size from 10 to something like 10000.

My guess: That this runs without an apparent error at a low vector length for me is probably a fluke?

> 
> That there's a failure in WaitForAnyPull is consistent with not getting any data, but I haven't sorted out a possible cause yet. Also trying to sort FI_LOG_LEVEL to get internal info from libfabric. I thought I had output from that, but then I switched to starting from your script and I don't seem to be getting anything. Will keep poking.

In my tests, the FI_LOG_LEVEL=Debug output was pretty much useless except when connecting to the network. Maybe there is some secret Cray debugging flag that I don't know, but their implementation seems to mostly ignore the FI_LOG_LEVEL..

franzpoeschel:
What just came to my mind: The MPI dataplane only works when launching at least two processes each for writer and reader. I will try out next week if this might make a difference here.

Also, I generally played a lot with `LD_PRELOAD` and seeing what MPI does on the system in order to figure things out. What I noticed is that MPICH does not seem to use `fi_mr_reg()` at all, but something else. I should maybe take a look at the libfabric implementation in MPICH next week again and see if I can figure out anything from there (though this seems more like a fool's hope..).

franzpoeschel:
Using two processes per writer and per reader did not make a difference.
Also I was hoping that there might be some `fi_*()` call returning an error status that was not checked. I added some error checks (last commit) but none of them reports an error.

eisenhauer:
Well, worth a shot. Unfortunately I'm on travel this week and not likely to get back to this until I return.

franzpoeschel:
I have isolated a minimal example of libfabric as it is used in SST here: https://codebase.helmholtz.cloud/poesch58/libfabric-minimal-example
I will try adding the CXI adaptations there as well. i expect that this will either result in the same issue, giving us a good base for getting this back to the ORNL and HPE support, or that this might even work and give us an idea what's going wrong in the "big" codebase.

franzpoeschel:
Alright, the minimal example does reproduce the exact same issue that I see in this PR. 

I have two tags in the Git repo:

1. Tag [sockets-working](https://codebase.helmholtz.cloud/poesch58/libfabric-minimal-example/-/tree/sockets-working?ref_type=tags): Communication via the Sockets provider works. On my local machine:
	```
	> export FABRIC_IFACE=eno1
	> ./write &
	...
	> ./read
	...
	Received 128 bytes of data at address 0x7fffffff17a0
	Received remote message:
	Lorem ipsum dolor sit amet, consetetur sadipscing elitr, sed diam nonumy eirmod tempor invidunt ut lLorem ipsum dolor sit amet,
	```

2. Tag [cxi-empty-message](https://codebase.helmholtz.cloud/poesch58/libfabric-minimal-example/-/tree/cxi-empty-message?ref_type=tags). The adaptations made for the CXI provider are seen in [this diff](https://codebase.helmholtz.cloud/poesch58/libfabric-minimal-example/-/commit/3fd76c60554acd60f0db1eef6775c17603b9f780).
	```	
	> salloc -N2 -n2 -c1 --ntasks-per-node=1 -ACSC380 -t 1:00:00 --network=single_node_vni,job_vni
	> export FABRIC_IFACE=cxi2
	> srun -N1 -n1 ./writer &
	...
	> srun -N1 -n1 ./reader
	...
	Received 0 bytes of data at address (nil)
	Received remote message:
	w?
	```

franzpoeschel:
I think I have good news: I received feedback from HPE on the minimal reproducer and it seems like it was not too far from a working implementation.
They suggested that the CXI provider does not support `FI_MR_VIRT_ADDR` which ADIOS2 currently uses. Adapting to this is quite easy since it implies that the virtual memory base address of read requests needs to be replaced with a zero.

A first attempt shows this result from the minimal example on Crusher:

```
Using second vni.
Address len: 4l
Received 0 bytes of data at address (nil)
Received remote message:
Lorem ipsum dolor sit amet, consetetur sadipscing elitr, sed diam nonumy eirmod tempor invidunt ut lLorem ipsum dolor sit amet,
```

(The length of received data is apparently just not reported in this call)

eisenhauer:
Well, at least somewhat promising.  Next steps?  I can poke at this at some point, but things are getting complex with next week being a holiday week.

franzpoeschel:
> Well, at least somewhat promising. Next steps? I can poke at this at some point, but things are getting complex with next week being a holiday week.

I tried to integrate the minimal example thing back into ADIOS2 just now and had the first functioning libfabric-based SST run on Frontier (I used Frontier since the Crusher queue is currently full). So it seems like this is the right approach and the fix suggested by Cray/HPE was the right one.

I pushed my changes (and also rebased this PR onto the `release_29` branch since that contains the most recent fixes for SST. Can rebase it back onto master after the release).

>  Next steps?

* Scaling tests, I only did a serial test from one node to another so far
* Implement the fix for Speculative Preload Mode, I skipped this for now since I wanted to see if this works at all and the preload logic is more involved than the normal one
* Remove the hardcoded changes to bring this into a mergeable state
* Remove other instances of "shortcut" logic (skipped "free()", selecting the right keys from the environment (they are currently all the same, but this might change))
* As a follow-up: Maybe integrate the minimal example for libfabric into ADIOS2, @pnorbert once mentioned that there is potential interest in a tool that can be used to determine which libfabric provider configuration might work for SST. I think that this minimal example is a good start for adding sth like that.
* Try this on other Cray systems such as Perlmutter

> but things are getting complex with next week being a holiday week.

This is fine. I will continue working on this PR and on the points mentioned above, but this is not super urgent for me and the code is currently far from being mergeable anyway.

franzpoeschel:
> > Well, at least somewhat promising. Next steps? I can poke at this at some point, but things are getting complex with next week being a holiday week.
> 
> I tried to integrate the minimal example thing back into ADIOS2 just now and had the first functioning libfabric-based SST run on Frontier (I used Frontier since the Crusher queue is currently full). So it seems like this is the right approach and the fix suggested by Cray/HPE was the right one.
> 
> I pushed my changes (and also rebased this PR onto the `release_29` branch since that contains the most recent fixes for SST. Can rebase it back onto master after the release).
> 
> > Next steps?
> 
>     * Scaling tests, I only did a serial test from one node to another so far
> 
>     * Implement the fix for Speculative Preload Mode, I skipped this for now since I wanted to see if this works at all and the preload logic is more involved than the normal one
> 
>     * Remove the hardcoded changes to bring this into a mergeable state
> 
>     * Remove other instances of "shortcut" logic (skipped "free()", selecting the right keys from the environment (they are currently all the same, but this might change))
> 
>     * As a follow-up: Maybe integrate the minimal example for libfabric into ADIOS2, @pnorbert once mentioned that there is potential interest in a tool that can be used to determine which libfabric provider configuration might work for SST. I think that this minimal example is a good start for adding sth like that.
> 
>     * Try this on other Cray systems such as Perlmutter
> 
> 
> > but things are getting complex with next week being a holiday week.
> 
> This is fine. I will continue working on this PR and on the points mentioned above, but this is not super urgent for me and the code is currently far from being mergeable anyway.
The important thing for now is that SST+libfabric seems to be (minimally) functional with this on Frontier.

vicentebolea:
@franzpoeschel thanks for your contribution, I will review this next week (running out of cycles for adios2 this week). In the meanwhile please fix the git conflicts.

franzpoeschel:
> @franzpoeschel thanks for your contribution, I will review this next week (running out of cycles for adios2 this week). In the meanwhile please fix the git conflicts.

This PR is not yet really in a reviewable state. 
The current update is that SST+libfabric now has basic functionality with the Cray CXI provider, but the CXI provider is currently essentially hardcoded. There is still some way to go before merging this.

eisenhauer:
@franzpoeschel Happy to lend a hand with integration too (after the US holiday).

franzpoeschel:
I have pushed a commit that ideally adds those fixes for speculative preload mode as well.
Has anyone recently tested if speculative preload mode still works? Setting `sstIO.SetParameter("SpeculativePreloadMode", "OFF");` in `helloSstReader.cpp` has no effect in my tests.

eisenhauer:
> I have pushed a commit that ideally adds those fixes for speculative preload mode as well. Has anyone recently tested if speculative preload mode still works? Setting `sstIO.SetParameter("SpeculativePreloadMode", "OFF");` in `helloSstReader.cpp` has no effect in my tests.

While the Preload modes had confirmed positive effects on the sockets-based data plane (big improvements on simple bandwidth tests), I was never as clear that they would be a win in an RDMA situation.  But as long as the RDMA parts work, we can work on Preload at our leisure...

franzpoeschel:
> > I have pushed a commit that ideally adds those fixes for speculative preload mode as well. Has anyone recently tested if speculative preload mode still works? Setting `sstIO.SetParameter("SpeculativePreloadMode", "OFF");` in `helloSstReader.cpp` has no effect in my tests.
> 
> While the Preload modes had confirmed positive effects on the sockets-based data plane (big improvements on simple bandwidth tests), I was never as clear that they would be a win in an RDMA situation. But as long as the RDMA parts work, we can work on Preload at our leisure...

I was just wondering since the RDMA/libfabric implementation has a rather complex preload logic which I was not actually able to activate. 
But I'm fine with focusing on pull-based reading for now.

franzpoeschel:
I successfully ran a 128-node setup with this using PIConGPU on Frontier, so this is beginning to look like a breakthrough.

Do you know if there is any logic in SST to evenly distribute the network cards found on a node among jobs? On Frontier, there are generally four per node (cxi0, cxi1, cxi2, cxi3) and it would probably be good to use them evenly.

eisenhauer:
So far as I know, no, there is no SST logic to select network interfaces.  We have relied upon the ability to set the FI_INTERFACE variable when necessary, but that is only sufficient to specify a functional interface, not to distribute load between multiple.  So if that's necessary, we'll have to add something.

franzpoeschel:
Ok, that's what I suspected. Choices that I see for this:

* Just continue using environment variables. Users can write a little wrapper script `srun select_interface.sh bin/code_with_adios_stream`. Pro: It's flexible, users can use their knowledge on how the application is scheduled to make a good distribution of network cards. Con: No one is actually going to do it.
* If the FABRIC_IFACE is not set, just pick one at random. Pro: Still better than all instances using the same card, can be used in conjunction with the above approach (as a fallback). Con: Might break code that relies on the legacy fabric selection logic, so maybe implement this only for CXI?
* Collectively gather all hostnames, figure out what MPI ranks are running on the same host as the current rank, select a card depending on that. Pro: Good performance by default. Con: Still not the best performance (ignores internal layout of nodes), most complex logic.

I think that the more complex options can be done in a later PR while this PR should focus on getting CXI runnable.

franzpoeschel:
I successfully ran a benchmark at 4096 nodes now. The performance of libfabric-based SST seems to be slightly better than that of MPI-based SST.

What I'm measuring is the "perceived" throughput, i.e. the throughput based on the time from load request to load completion on the reader site. This figure is skewed by communication overhead.
With libfabric I get roughly 3.6~4.7 GiB per second and node (14~19 TiB per second in total). With MPI, that's 2.7~3.7GiB per second and node (11~15TiB per second in total).

eisenhauer:
Fantastic!  I was paging through the changes to rdma_dp and it shouldn't be hard to integrate in such a way that we support the prior providers as well.  Quick question, I notice that you're using manual progress rather than auto (which I recall was an earlier problem with cxi).  Do you have to have an provision to check the completion queue in the background so that transfers continue to progress even if ADIOS is busy doing something other than rdma calls?

eisenhauer:
Yes, SST does have a blocking thread (actually on both sides waiting for incoming socket messages for the control plane), but that blocking thread is blocking on select()/poll() and won't be doing fi_cq_read() asynchronously unless we can associate a CQ with an FD.  This was not possible with early versions of the CXI provider, but that might have changed more recently.  Without the ability to wait on an FD, the alternative is periodic polling of the CQ or just letting things take their natural course where libfabric code gets control whenever a libfabric call happens.  The consequences of each may be weirder.  If we need to service a completion queue on the writer as part of a handshake with the reader and we only write every few minutes, that handshake may be delayed until the next write.  This might not be noticable at all in something like a bandwidth test, but could cause odd latencies or have performance impacts in other circumstances.

Anyhow, not saying that I know this will be a problem and the results sound great so far, but it may be something to keep an eye out for as this will be our only "manual progress" provider.  I'm tied up until next week, but hopefully we can cooperate to get this into master quickly.  (And the performance notes may mean that we need to drop the priority on the MPI data plane too.  I'm not sure we've compared its performance with the RDMA data plane on the other targets, but we obviously want to select the fastest and need to make sure the DP priority assigned reflects that. @vicentebolea, just taking note. )

franzpoeschel:
> won't be doing fi_cq_read() asynchronously unless we can associate a CQ with an FD

Ah, so essentially what we can expect is a serialization of I/O because progress is only made for one request?
As long as we're not running into deadlocks, I'd suggest keeping the features of this PR as they are now and focusing on bringing this into master first.
My test case is probably not really affected from this serialization behavior since it entails a low number of big operations that are in the end bound by what the network can do.



> (And the performance notes may mean that we need to drop the priority on the MPI data plane too. I'm not sure we've compared its performance with the RDMA data plane on the other targets, but we obviously want to select the fastest and need to make sure the DP priority assigned reflects that. @vicentebolea, just taking note. )

At the same time, I don't think that this makes the MPI DP obsolete:

* Given how shaky SST has been on Frontier so far, it's good to have two redundant implementations
* Ideally on a system with a well-configured MPI, the MPI DP is probably easier to install and use 
* Being able to compare the performance of two implementations can point us at performance bugs in one of them that we would otherwise not notice
* Above discussion on FI_PROGRESS_MANUAL, the performance of the libfabric implementation might not be ideal yet

But since the libfabric implementation is connection-less and hand-shake between writer and reader works fully via the control plane, while at the same time MPI has to do the MPI_Open_Port()/MPI_Comm_Connect() dance, the achievable performance of the direct libfabric implementation is likely better. My guess is that this is a large part of what I'm measuring above.

franzpoeschel:
Just tested: This PR also works with the CXI provider on Perlmutter

eisenhauer:
> Ah, so essentially what we can expect is a serialization of I/O because progress is only made for one request? As long as we're not running into deadlocks, I'd suggest keeping the features of this PR as they are now and focusing on bringing this into master first. My test case is probably not really affected from this serialization behavior since it entails a low number of big operations that are in the end bound by what the network can do.
The exact side of not servicing the CQ regularly likely depend upon the implementation.  I'd be on the lookout for things like the reader not making progress until the writer hits EndStep() on the next timestep (I.E. when the writer gets back to doing libfabric calls where CQ service might happen as a side effect).

franzpoeschel:
The old libfabric implementation (before this PR) had some Cray-specific implementation, activated by `cmake/DetectOptions.cmake`:

```cmake
    find_package(CrayDRC)
    if(CrayDRC_FOUND)
      set(ADIOS2_SST_HAVE_CRAY_DRC TRUE)
    endif()
```

Should the old Cray-specific implementation be removed or does it have any use still?

I did not find a good solution for an automatic compile-time configuration of the CXI provider (which is needed due to imports from `#include <rdma/fi_cxi_ext.h>` being used). I found no CMake package that can be queried like CrayDRC above, and while the compiler on Frontier defines `#define __cray__ 1`, the compiler on Perlmutter does not.

Suggestion:

* Make this a CMake option that users can manually set
* Use `try_compile()` in CMake to check if `__cray__` is defined and turn this option on manually if yes.

source/adios2/toolkit/sst/dp/rdma_dp.c
```diff
@@ -617,6 +776,189 @@ static TimestepList GetStep(Rdma_WS_Stream Stream, long Timestep)
...
+
+    //     /* The Virtual Network ID (VNI) assigned to the Endpoint. Two
+    //     Endpoints
+    //     * must use the same VNI in order to communicate.
+    //     *
+    //     * Note that while the CXI service may define one or more VNIs which a
+    //     * process can access, an Endpoint is assigned to only one.
+    //     */
+    //     uint16_t vni;
+    // };
```
franzpoeschel, line 810: Might need to remove this comment due to licensing reasons? It's just a copy-pasted documentation from the CXI header.

source/adios2/toolkit/sst/dp/rdma_dp.c
```diff
@@ -617,6 +776,189 @@ static TimestepList GetStep(Rdma_WS_Stream Stream, long Timestep)
...
+                return EXIT_FAILURE;
+            }
+
+            switch (slingshot_devices[find_end_of_current_string])
+            {
+            case '\0':
+                found_end = 1;
+                goto break_first_loop;
+            case ',':
+                goto break_first_loop;
```
franzpoeschel, line 841: Don't know how strict you guys are with `goto`s. This one here is just to circumvent that you can't break out of a loop while inside a `switch` statement. IMO cleaner and easier to read than the alternatives.

source/adios2/toolkit/sst/dp/rdma_dp.c
```diff
@@ -916,8 +1348,15 @@ static DP_WSR_Stream RdmaInitWriterPerReader(CP_Services Svcs,
...
-                     &WSR_Stream->ReaderAddr[i], 0, NULL);
+        if (fi_av_insert(Fabric->av, providedReaderInfo[i]->Address, 1,
+                         &WSR_Stream->ReaderAddr[i], 0, NULL) < 1)
+        {
+
+            Svcs->verbose(WS_Stream->CP_Stream, DPCriticalVerbose,
+                          "[RdmaInitWRiterPerReader] Failed inserting address "
+                          "into vector\n");
+            return NULL;
+        }
```
franzpoeschel, line 1359: While trying to figure out where things were going south, I added some additional checks to the libfabric dataplane. Not strictly related to this PR, but I think that it's still worthwhile to keep them now that they're written anyway?

franzpoeschel (commented):
This is now enabled automatically via CMake by checking if the CXI header can be included.
I've done a small test on my local computer, this now works with the sockets provider as well as with the CXI provider.

Remaining todo:

* Rebase back onto master
* Test on Summit, maybe do another somewhat larger test on Frontier if this is still working properly (though I don't see why not)
* Remove changes from SST example again?


Before I rebase this back onto master, this diff is more precise than the one shown in the PR: https://github.com/ornladios/ADIOS2/compare/release_29...franzpoeschel:ADIOS2:libfabric-cray

Todo after this PR:

* Some form of polling for better supporting `FI_PROGRESS_MANUAL`
* Maybe add an environment variable that lets you select the provider directly. Testing SST against the sockets provider was a bit annoying, since the `FABRIC_IFACE=eno1` was matched by multiple providers and the tcp provider (selected by the dataplane) did not work. I can do that in a new PR, it should be a simple enough change.

source/adios2/toolkit/sst/dp/rdma_dp.c
eisenhauer: I think you could do it the way you are, or you could set a flag inside the case and use it to control breaking out of the loop outside.  That might be less offensive to any anti-goto adherent.

eisenhauer:
Looks like clang-format wants to reorder your includes:

 #ifdef SST_HAVE_CRAY_CXI
 // Needs to be included before rdma/fi_cxi_ext.h
-#include <stdbool.h>
 #include <rdma/fi_cxi_ext.h>
+#include <stdbool.h>
 #endif

If stdbool.h is required for rdma/fi_cxi_ext.h (and so must come first), you can either move it before the #ifdef (shouldn't hurt for other includes), or put a blank line between the two (so clang-format won't mess with it).

source/adios2/toolkit/sst/dp/rdma_dp.c
franzpoeschel: That would be the most obvious alternative, yes. But that is more error-prone and more cumbersome to read than the `goto` in my opinion. I think that using `goto` to emulate a labeled break or labeled continue (this exists e.g. in Java) is considered one of the "legitimate" use cases.

Since you're the maintainer, I'm fine with either way. Should I change it?

franzpoeschel:
> If stdbool.h is required for rdma/fi_cxi_ext.h (and so must come first), you can either move it before the #ifdef (shouldn't hurt for other includes), or put a blank line between the two (so clang-format won't mess with it).

It is required for that header, yes. I'll add a little comment in between, that usually keeps clang-format from touching the include order.

eisenhauer:
OK, @vicentebolea  might have opinions about the CMake modes, but if they work, they're a start.  @ax3l , you've tested on Frontier and Perlmutter?  @pnorbert, before merging we probably need to test there everywhere we can that uses RDMA to make sure we don't have a regression.  I can do summit, but I don't have as much access to other platforms as you probably do...

franzpoeschel:
There does seem to be a regression that affects Summit. I'll try to figure out where it goes wrong.

eisenhauer:
@franzpoeschel If you can fix the formatting and rebase this on current master, we can probably merge this.

franzpoeschel:
> @franzpoeschel If you can fix the formatting and rebase this on current master, we can probably merge this.

done

source/adios2/toolkit/sst/dp/rdma_dp.c
eisenhauer: It's fine...

eisenhauer (approved):

franzpoeschel:
This seems to hang on full-scale Frontier. My current suspicion is that this might have to do with the manual data progress; maybe #3964 can help there.
The MPI dataplane works fine at full scale Frontier.

PR #3673, opened 2023-06-22
ci,ascent: disable install tests

vicentebolea:
The DFS in Ascent is not playing well with our install tests. Those install tests in AScent are second priority since we normally install in DoE systems using Spack


vicentebolea:

Issue #3680, opened 2023-06-30
ADIOS2: 2.9.0 MSVC w/ C-Blosc 2.9.0


ax3l:
ADIOS2 v2.9.0 does not compile with C-Blosc 2.9.0 on Windows:
```
...
    BPBackCompatBlosc.cpp
  C:\Program Files (x86)\blosc\include\blosc2.h(1056,5): error C7555: use of designated initializers requires at least '/std:c++20' [D:\a\openPMD-api\openPMD-api\src\build-adios2\source\adios2\adios2_core.vcxproj]
  D:\a\openPMD-api\openPMD-api\src\dep-adios2\ADIOS2-2.9.0\source\adios2\toolkit\format\bp\bpBackCompatOperation\compress\BPBackCompatBlosc.cpp(145,1): error C2589: '(': illegal token on right side of '::' [D:\a\openPMD-api\openPMD-api\src\build-adios2\source\adios2\adios2_core.vcxproj]
  D:\a\openPMD-api\openPMD-api\src\dep-adios2\ADIOS2-2.9.0\source\adios2\toolkit\format\bp\bpBackCompatOperation\compress\BPBackCompatBlosc.cpp(146): error C2062: type 'unknown-type' unexpected [D:\a\openPMD-api\openPMD-api\src\build-adios2\source\adios2\adios2_core.vcxproj]
  D:\a\openPMD-api\openPMD-api\src\dep-adios2\ADIOS2-2.9.0\source\adios2\toolkit\format\bp\bpBackCompatOperation\compress\BPBackCompatBlosc.cpp(146,1): error C2059: syntax error: ')' [D:\a\openPMD-api\openPMD-api\src\build-adios2\source\adios2\adios2_core.vcxproj]
```
and
```
...
    CompressBlosc.cpp
  C:\Program Files (x86)\blosc\include\blosc2.h(1056,5): error C7555: use of designated initializers requires at least '/std:c++20' [D:\a\openPMD-api\openPMD-api\src\build-adios2\source\adios2\adios2_core.vcxproj]
  D:\a\openPMD-api\openPMD-api\src\dep-adios2\ADIOS2-2.9.0\source\adios2\operator\compress\CompressBlosc.cpp(155,53): warning C4267: 'initializing': conversion from 'size_t' to 'int32_t', possible loss of data [D:\a\openPMD-api\openPMD-api\src\build-adios2\source\adios2\adios2_core.vcxproj]
[...]

PR #3681, opened 2023-07-01
Fix: std::min w/ windows.h in C-Blosc2

ax3l:
Explictly specialize to avoid macro resolution when windows.h is included. What a world to live in.

X-ref: https://github.com/ornladios/ADIOS2/issues/3680#issuecomment-1615308336

@vicentebolea did I do this right? This PR should go in `master` and `release_29` alike.


williamfgc:
> What a world to live in

Indeed

anagainaru (approved):
Looks good

vicentebolea:
Backported to master

PR #3690, opened 2023-07-10
Merge pull request #3689 from vicentebolea/fix-evpath-modules

vicentebolea:
cmake,evpath: namespace modules install dir

(cherry picked from commit 73846790079480575db32d8f99ee9989b423f4e3)


vicentebolea:
@eisenhauer while this changes the path of the Module file in the installation tree, these evpath modules are private and not for the user to directly use them.

eisenhauer (approved):
Sounds good.

PR #3698, opened 2023-07-18
added gitattributes file

vicentebolea:
This is needed to avoid having other than LF EOL in the repo. Having CRLF returns makes difficult impossible to properly patch files.

This was needed for backporting patches to the Conda pkg


vicentebolea:

Note that the CI is red since this whitespace change cannot apply in this PR since the PR tests the merge-head from master. We have to go ahead and merge to see this change.

PR #3699, opened 2023-07-18
Update files with new EOL

vicentebolea:
This is needed to complete #3698


vicentebolea:

PR #3705, opened 2023-07-20
Use ccache in github actions builds

scottwittenburg:
Add an [action](https://github.com/hendrikmuhs/ccache-action) available on the marketplace to GithubActions builds, which simplifies using `ccache` to speed up compilation.  The table below shows the speed up for some of the affected builds (note: times are build times only, and do not include any of the other steps).  On the first run, the cache was empty, so there was no benefit at all.  On the second run, the cache hit rates were all around 100%, since I didn't change any source files in between.

```
                  Build Name                  |         Time To Build
                                              |   first run  |  second run
----------------------------------------------|-----------------------------
  linux_ubuntu (ubuntu20.04, gcc8, serial)    |     7m 28s   |      44s
  linux_ubuntu (ubuntu20.04, gcc8, ompi)      |    10m 30s   |   1m 36s
  linux_ubuntu (ubuntu20.04, gcc9, ompi)      |     9m 27s   |   1m 36s
  linux_ubuntu (ubuntu20.04, gcc10, ompi)     |    12m 36s   |   1m 29s
  linux_ubuntu (ubuntu20.04, gcc10, mpich)    |     9m  3s   |   1m 20s
  linux_ubuntu (ubuntu20.04, gcc11, ompi)     |     9m 13s   |   1m 25s
  linux_ubuntu (ubuntu20.04, clang6, serial)  |     4m 38s   |      40s
  linux_ubuntu (ubuntu20.04, clang6, ompi)    |     6m 58s   |   1m 31s
  linux_ubuntu (ubuntu20.04, clang10, ompi)   |     7m 22s   |   1m 20s
```

I pre-installed `ccache` in builds where we control the image being used, but `ccache-action` installs it if it doesn't exist.

As a follow-on, or even in this PR, we could consider using dependabot to keep this action (and any others we use, eg `checkout`, `upload-artifact`, etc) up to date.  See [here](https://docs.github.com/en/code-security/dependabot/working-with-dependabot/keeping-your-actions-up-to-date-with-dependabot#enabling-dependabot-version-updates-for-actions).


scottwittenburg:
Looking at one of the windows builds, I see this [issue](https://github.com/ornladios/ADIOS2/actions/runs/5616513933/job/15218987283#step:16:7), where it isn't saving any cache files after the build.

It seems to be a known issue, see [here](https://github.com/hendrikmuhs/ccache-action/issues/112).

vicentebolea:
> Looking at one of the windows builds, I see this [issue](https://github.com/ornladios/ADIOS2/actions/runs/5616513933/job/15218987283#step:16:7), where it isn't saving any cache files after the build.
> 
> It seems to be a known issue, see [here](https://github.com/hendrikmuhs/ccache-action/issues/112).

If this will halt this feature, we can come back to add ccache to windows builds later.

vicentebolea:
@scottwittenburg let me know when you need review, I was asked to review automatically since I was listed in the codeowner.

vicentebolea:
> Do you have a preference? 

Disable it in Windows, using ccache we always incur in a overhead, cache miss + build TU is worse than build TU.

.github/workflows/everything.yml
```diff
@@ -131,6 +131,10 @@ jobs:
         with:
           ref: ${{ github.event.pull_request.head.sha }}
           path: source
+      - name: ccache
```
vicentebolea, line 134: @scottwittenburg We already provide ccache in the docker images, why downloading it again with this ccache action. I can see that this also setup some ccache settings. 
Concretely:
```
    await execBash(`ccache --set-config=cache_dir='${path.join(ghWorkSpace, '.ccache')}'`);
    await execBash(`ccache --set-config=max_size='${maxSize}'`);
    await execBash(`ccache --set-config=compression=true`);
```

It might be better doing this ourselves and avoid this dependency. Note that in ccache all these settings can be also setup with a env variable CCACHE_...`

vicentebolea (changes requested):
Great that this is working! Thanks a lot! lets try to get rid of that gh action since we could replace it by a few env var exports.

.github/workflows/everything.yml
scottwittenburg: > why downloading it again with this ccache action?

This action only installs `ccache` if it doesn't exist already, see [here](https://github.com/hendrikmuhs/ccache-action/blob/53e53642769c4b03a7881a7283314d1f364134ec/src/restore.ts#L169).  So it's handling it for us on the mac builds (where we'll need to manage it ourselves in some script otherwise), but not on the linux ones.

> It might be better doing this ourselves and avoid this dependency

We already depend on a few other actions, I'm not sure I understand the reasoning to avoid this one, since it makes things pretty easy.  See the [actions.yml](https://github.com/hendrikmuhs/ccache-action/blob/53e53642769c4b03a7881a7283314d1f364134ec/action.yml#L11) for the configuration options it exposes, we could set the max size if you think it should be different.  It seems we cannot change the cache dir or compression, but I don't know if we would want to change those anyway.

Having said that, I'll leave the decision up to you.

vicentebolea: >  It seems we cannot change the cache dir or compression, but I don't know if we would want to change those anyway.

Yes not need to change it a the moment.

The reasoning is for getting rid of this, is that using ccache without this action is not really harder than using this github-action. Also configuring it directly can let us in the future to further change options if needed. You can try to setup the options that it sets through env vars and it should yield the same result. Actually compression is true by default so you can skip it.

scottwittenburg: Ok, let me try it without the action.

.github/workflows/everything.yml
```diff
@@ -227,7 +255,6 @@ jobs:
       - name: Test
```
vicentebolea, line 255: Before this we can print the ccache statistics with `ccache -s` to see our hit ratio

.github/workflows/everything.yml
```diff
@@ -105,6 +105,10 @@ jobs:
         GH_YML_MATRIX_OS: ${{ matrix.os }}
         GH_YML_MATRIX_COMPILER: ${{ matrix.compiler }}
         GH_YML_MATRIX_PARALLEL: ${{ matrix.parallel }}
+        CCACHE_BASEDIR: "${GITHUB_WORKSPACE}"
+        CCACHE_DIR: "${GITHUB_WORKSPACE}/.ccache"
+        CCACHE_COMPRESS: true
```
vicentebolea, line 110:
```suggestion
        CCACHE_COMPRESS: true
```

Ccache compress by default

.github/workflows/everything.yml
```diff
@@ -131,6 +135,16 @@ jobs:
         with:
           ref: ${{ github.event.pull_request.head.sha }}
           path: source
+      - name: Restore cache
+        uses: actions/cache@v3
+        with:
+          path: .ccache
+          key: ccache-${{ matrix.os }}-${{ matrix.compiler }}-${{ matrix.parallel }}-${{ github.ref }}-${{ github.sha }}
```
vicentebolea, line 142: if we add the sha/ref then new commits will not be able to find the cache. MPI/Serial builds share a lot of cacheable units. I would as far as `ccache-${{ matrix.os }}-${{ matrix.compiler }}`

vicentebolea (changes requested):

.github/workflows/everything.yml
scottwittenburg: What's the suggestion here?  Yours looks identical to mine.

.github/workflows/everything.yml
scottwittenburg: The sha/ref is so that each build can write to a unique cache key.  According to [this](https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows#using-the-cache-action) section of the github docs, new commits *will* be able to find the cache thanks to the `restore-keys`.

If I remove the mpi/serial bit from the key, I think it could reduce cache hit rate for ompi builds, since sometimes a serial build will write the cache without any of the mpi object files.

Maybe let's discuss key selection strategy when we sync up this afternoon?

vicentebolea: > If I remove the mpi/serial bit from the key, I think it could reduce cache hit rate for ompi builds, since sometimes a serial build will write the cache without any of the mpi object files.

ccache will deal with this, in fact with ccache we want most builds in the same platform to share the cache directory (key). Ccache generates an unique file for each .o file where its filename is a hash of its path and its compiler invokation with its arguments, and the source file has. There is no chance of overrwriting objects files from one build to the other.

vicentebolea: > Maybe let's discuss key selection strategy when we sync up this afternoon?

Sure

scottwittenburg: I just found this as I was reading through the docs again:  Since cache is read-only (meaning we can't change existing cache entries, only create new ones), they provided a tip on how to update cache, [here](https://github.com/actions/cache/blob/main/tips-and-workarounds.md#update-a-cache).

That could be why the pattern of using a unique key for each build seems so frequently used.  

Looking at our cache [usage](https://github.com/ornladios/ADIOS2/actions/caches) now, with only this PR contributing, we're already over the 10GB limit.  Adding up the different cache sizes that were restored in one workflow run, it amounted to around 3GB.  So it seems to me we could easily end up with PR caches pushing out the more important ones from the default branch.

So I think I'm already leaning toward saving caches only on workflows that run on `master`, but restore caches on all branches.  Using the tip linked above, we'd "update" the cache on every master run, then PRs can just use it and hopefully get fairly high hit rates.

But let me know what you think.

vicentebolea: > I just found this as I was reading through the docs again: Since cache is read-only (meaning we can't change existing cache entries, only create new ones), they provided a tip on how to update cache, [here](https://github.com/actions/cache/blob/main/tips-and-workarounds.md#update-a-cache).

I see, then no need to swim against the stream here.

> So I think I'm already leaning toward saving caches only on workflows that run on master, but restore caches on all branches. Using the tip linked above, we'd "update" the cache on every master run, then PRs can just use it and hopefully get fairly high hit rates.

That is a sound plan.

scottwittenburg:
@vicentebolea Please take a look when you get a chance.  If you don't have any more suggestions, I will just rewrite the history into two commits, and then we can merge.

.github/workflows/everything.yml
```diff
@@ -218,16 +279,34 @@ jobs:
...
         run: gha/scripts/ci/gh-actions/run.sh update
       - name: Configure
         run: gha/scripts/ci/gh-actions/run.sh configure
       - name: Build
         run: gha/scripts/ci/gh-actions/run.sh build
+      - name: Print ccache statistics
+        run: ccache -s
+      - name: Save cache
+        uses: actions/cache/save@v3
+        if: github.ref_name == 'master' && steps.restore-cache.outputs.cache-hit == 'false'
```
vicentebolea, line 302: Does this mean only save if there was no cache hit?

vicentebolea:
@scottwittenburg I just rebased your branch so that it containes the latest commit that "fixes" the tests failures.

.github/workflows/everything.yml
scottwittenburg: Yes, if there was an "exact match" on the cache key (which is defined using the sha of the merge commit getting tested), then attempting to write the cache would produce the exact same contents.  Since github caches are not editable (updatable), it would be an error to write it anyway, so I'm avoiding.

vicentebolea (approved):

scottwittenburg:
Cache was not saved on any of the runs on `master`, see [here](https://github.com/ornladios/ADIOS2/actions/runs/5729147765/job/15525310630) for example.

In my PR testing, I exercised half of this [conditional](https://github.com/ornladios/ADIOS2/blob/15374f2d2dea7c0f11bda9a6d6f3d9441ae48237/.github/workflows/everything.yml#L171) before adding the additional criteria `github.ref_name == 'master'`, by re-running a workflow on the same `sha` and making sure that cache was not saved.

So I'm not sure yet why it didn't save cache.  [Here ](https://docs.github.com/en/actions/learn-github-actions/expressions#about-expressions)it says that the `${{ }}` is not required in `if:` statements, but then it proceeds to use them everywhere in that documentation anyway.  Maybe I need to add those?

vicentebolea:
> `}}` is

That is odd, I guess that we need to merge it to check it

PR #3715, opened 2023-07-26
Fix: Static Builds Against C-Blosc2

### LINKED ISSUE #3714: Blosc2: Config.cmake Package in 2.10.1+
Hi,

As you know as one of our activities, we build an all-static version of ADIOS2 for Python wheels in openPMD-api.

Struggling with some of the dependencies in C-Blosc2, I decided to contribute a `blosc2Config.cmake` package in:
  https://github.com/Blosc/c-blosc2/pull/537

@vicentebolea you might be able to bump the ADIOS2 C-Blosc2 dependency once this is merged and can remove the `FindBLOSC2.cmake` file, which had trouble picking up transitive dependencies in static builds.

ax3l:
This uses the new C-Blosc2 `Blosc2Config.cmake` packages, so we can do purely static builds.

Close #3714

- [x] depends on upstream https://github.com/Blosc/c-blosc2/pull/537
  - [x] part of the next release, C-Blosc2 `v2.10.1`

### DIFF
--- a/cmake/DetectOptions.cmake
+++ b/cmake/DetectOptions.cmake
@@ -69,12 +69,31 @@ endfunction()
 
 # Blosc2
 if(ADIOS2_USE_Blosc2 STREQUAL AUTO)
-  find_package(Blosc2 2.4)
+  # Prefect CONFIG mode
+  find_package(Blosc2 2.4 CONFIG QUIET)
+  if(NOT Blosc2_FOUND)
+    find_package(Blosc2 2.4 MODULE QUIET)
+  endif()
 elseif(ADIOS2_USE_Blosc2)
-  find_package(Blosc2 2.4 REQUIRED)
+  # Prefect CONFIG mode
+  find_package(Blosc2 2.4 CONFIG REQUIRED)
+  if(NOT Blosc2_FOUND)
+    find_package(Blosc2 2.4 MODULE REQUIRED)
+  endif()
 endif()
-if(BLOSC2_FOUND)
+if(Blosc2_FOUND)
   set(ADIOS2_HAVE_Blosc2 TRUE)
+  if(TARGET Blosc2::blosc2_shared)
+    set(Blosc2_shlib_available ON)
+  endif()
+
+  set(adios2_blosc2_tgt Blosc2::Blosc2)
+  if (Blosc2_VERSION VERSION_GREATER_EQUAL 2.10.1)
+    set(adios2_blosc2_tgt Blosc2::blosc2_$<IF:$<BOOL:${ADIOS2_Blosc2_PREFER_SHARED}>,shared,static>)
+  endif()
+
+  add_library(adios2_blosc2 INTERFACE)
+  target_link_libraries(adios2_blosc2 INTERFACE ${adios2_blosc2_tgt})
 endif()
 
 # BZip2
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -197,6 +197,8 @@ if(ADIOS2_HAVE_MPI)
   add_definitions(-DOMPI_SKIP_MPICXX -DMPICH_SKIP_MPICXX)
 endif()
 
+cmake_dependent_option(ADIOS2_Blosc2_PREFER_SHARED "Prefer shared Blosc2 libraries"
+  ON "Blosc2_shlib_available" OFF)
 
 #------------------------------------------------------------------------------#
 # POSIX O_DIRECT is only working for Unix in adios for now
--- a/scripts/ci/images/Dockerfile.ci-spack-ubuntu20.04-base
+++ b/scripts/ci/images/Dockerfile.ci-spack-ubuntu20.04-base
@@ -52,4 +52,21 @@ RUN apt-get update && apt-get install -y \
     update-alternatives --install /usr/bin/g++ g++ /usr/bin/g++-11 100 && \
     update-alternatives --install /usr/bin/gfortran gfortran /usr/bin/gfortran-11 100
 
+# NONSPACK dependency
+WORKDIR /opt/blosc2
+RUN curl -L https://github.com/Blosc/c-blosc2/archive/refs/tags/v2.10.1.tar.gz | tar -xvz && \
+    mkdir build && \
+    cd build && \
+    . /spack/share/spack/setup-env.sh && \
+    spack env activate adios2-ci-serial && \
+    cmake -DBUILD_SHARED_LIBS=ON -DCMAKE_BUILD_TYPE=Release -DCMAKE_INSTALL_PREFIX=/opt/blosc2/2.10.1 ../c-blosc2-2.10.1  && \
+    cmake --build . -j$(grep -c '^processor' /proc/cpuinfo) && \
+    cmake --install . && \
+    cd .. && \
+    rm -rf c-blosc2-2.10.1 build
+
+ENV PATH /opt/cmake/bin:/opt/blosc2/2.10.1/bin:${PATH}
+ENV LD_LIBRARY_PATH /opt/blosc2/2.10.1/lib64:${LD_LIBRARY_PATH}
+ENV CMAKE_PREFIX_PATH /opt/blosc2/2.10.1:${CMAKE_PREFIX_PATH}
+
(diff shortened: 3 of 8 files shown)


ax3l:
C-Blosc2 v2.10.1 contains the fix! :tada: 
https://github.com/Blosc/c-blosc2/releases/tag/v2.10.1

cmake/DetectOptions.cmake
```diff
@@ -68,12 +68,13 @@ function(lists_get_prefix listVars outVar)
 endfunction()
 
 # Blosc2
+option(ADIOS2_PREFER_SHARED_Blosc2 "prefer sharred Blosc2 libraries" ON)
 if(ADIOS2_USE_Blosc2 STREQUAL AUTO)
-  find_package(Blosc2 2.4)
+  find_package(Blosc2 2.10.1 CONFIG)
```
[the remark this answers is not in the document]
ax3l: Understand - but I think there is a preference of `find_package` to not pick up the config files if a `Find*.cmake` script is present - can we try to prefer Config files if present, e.g., if 2.10.1+ is there? Otherwise static builds won't work.
https://cmake.org/cmake/help/latest/command/find_package.html#search-modes

vicentebolea:
@scottwittenburg I am adding a blosc2 dependency to the spack builds. Note that Blosc2 does not have a Spack pkg and the reason to install this in our Spack images is that they are our standard images that we test most of the configurations. I am not too strong about it, we can move them to the nonspack build. Let me know.

On another note I changed the tag naming while building the builds so that we can avoid having to retag and we can build a leaf image without rebuilding the base image (if no changes in base).

scottwittenburg:
@vicentebolea 

Regarding building a non-spack blosc in the spack image: does blosc share any dependencies with adios2 which might be picked up from the system when building and possibly conflict later?   Maybe it's nothing to worry about, I'm just thinking out loud.

Regarding your change to remove the explicit tagging step in favor of just using the right tag when building: sounds good to me.

Thanks!

vicentebolea:
> @vicentebolea
> 
> Regarding building a non-spack blosc in the spack image: does blosc share any dependencies with adios2 which might be picked up from the system when building and possibly conflict later? Maybe it's nothing to worry about, I'm just thinking out loud.

Good question. I do not see this possibly happening we are building a very minimal blosc2

scottwittenburg:
> Good question. I do not see this possibly happening we are building a very minimal blosc2

I see by default it turns on `zlib`, `zstd`, and `lz4`, and also by default it uses it's own internal "thirdparty" versions of those.  

c-blosc2 internal dependency versions:

```
lz4-1.9.4
zlib-1.2.13
zstd-1.5.5
```

adios2 spack image contains:

```
lz4@1.9.4
zlib@1.2.13
zstd@1.5.2
```

So there's a pretty close match between what we have in the spack image, and what c-blosc2 needs, so if problems arise from doing it this way we could try to get the c-blosc2 build to use our spack-installed dependencies.  Or we could also just move c-blosc2 to the non-spack build.

By the way, our image also contains `c-blosc@1.21.1`, but I guess that's something different.  Maybe there's a c-blosc2 spack package in the works?

Actually, there *is* a c-blosc2 package in spack, and it has the version mentioned above.  At that's what it looks like from this merged PR: https://github.com/spack/spack/pull/39297

scottwittenburg:
However, it does seem that the adios2 spack package only knows about `c-blosc`, nothing about `c-blosc2`, so maybe that's why we can't use the spack package.

vicentebolea:
> However, it does seem that the adios2 spack package only knows about c-blosc, nothing about c-blosc2, so maybe that's why we can't use the spack package.

@scottwittenburg thanks for pointing out this, that is indeed the Spack package for blosc2, I did made a search using spack search to [bB]losc2 without luck but indeed the name is c-blosc2. Will have to update our adios2 pkg to use c-blocs2 rather than c-blocs.

scottwittenburg:
> Will have to update our adios2 pkg to use c-blocs2 rather than c-blocs.

Yeah.  But I think it makes sense to go ahead with the current approach here for now, as that will take some time to get merged and appear in the E4S image we're using as a base.

vicentebolea:
> Yeah. But I think it makes sense to go ahead with the current approach here for now, as that will take some time to get merged and appear in the E4S image we're using as a base.

I am adding c-blocs2 to our spec.yml in the meanwhile.

scripts/ci/images/specs.yaml
```diff
@@ -4,3 +4,5 @@ spack:
     checksum: false
   concretizer:
     unify: false
+  specs:
+    - c-blosc2
```
vicentebolea, line 8: @scottwittenburg I added here the dep

scottwittenburg: Makes sense.  Hopefully just adding that in separately didn't change the cache hit rate when building the images?  Also hopefully it results in the adios2 configure finding it, but it seems it may need to be turned on instead of `Blosc`, eg [here](https://github.com/ornladios/ADIOS2/blob/afbe6d63feefa9085a3e2e6e7e84621c5124e338/scripts/ci/cmake/ci-ubuntu20.04-gcc8-serial.cmake#L12).

vicentebolea: Interestingly the ADIOS2 concretization override the `specs` packages (such as c-blosc2)

vicentebolea:
@scottwittenburg we need to install this manually for the time being since the blosc2 in the E4S spack is older than what we req. I am building it with the adios2-ci-serial environment activated

vicentebolea:
I tested this locally with an older blosc2 build and with a blosc2 2.10.1 static build and it seems to build and run tests.

scottwittenburg:
I see the cmakecache.txt now reflect that blosc2 is enabled and found in your manual install directory.  So this looks good to me, once you remove the `tmp` from the image names.

vicentebolea:
> I see the cmakecache.txt now reflect that blosc2 is enabled and found in your manual install directory. So this looks good to me, once you remove the `tmp` from the image names.

Done

scottwittenburg (approved):

vicentebolea (approved):

cmake/DetectOptions.cmake
```diff
@@ -69,12 +69,31 @@ endfunction()
...
-if(BLOSC2_FOUND)
+if(Blosc2_FOUND)
   set(ADIOS2_HAVE_Blosc2 TRUE)
+  if(TARGET Blosc2::blosc2_shared)
+    set(Blosc2_shlib_available ON)
+  endif()
+
+  set(adios2_blosc2_tgt Blosc2::Blosc2)
+  if (Blosc2_VERSION VERSION_GREATER_EQUAL 2.10.1)
+    set(adios2_blosc2_tgt Blosc2::blosc2_$<IF:$<BOOL:${ADIOS2_Blosc2_PREFER_SHARED}>,shared,static>)
```
ax3l, line 92: @vicentebolea there is a bug here:
`ADIOS2_Blosc2_PREFER_SHARED` should not assume there is always a shared c-blosc2 library available.

C-Blosc2 builds allow to control static and shared builds independently, thus this will cause issues with `ADIOS2_Blosc2_PREFER_SHARED=ON` (default) for purely static builds (prefer != require shared).

vicentebolea: Thanks for pointing this out.

>  ADIOS2_Blosc2_PREFER_SHARED should not assume there is always a shared c-blosc2 library available.

We do not assume this, a couple of lines above we decide this. However, the prefer is not respected since if the user force ADIOS2_Blosc2_PREFER_SHARED=ON it will fail with an static lib

ax3l:
@vicentebolea @scottwittenburg just FYI, for transitive, external dependencies in C-Blosc2 you might want to use v2.10.2+
https://github.com/ornladios/ADIOS2/issues/3714#issuecomment-1684121372

cmake/DetectOptions.cmake
ax3l: Yes, that's what I mean, thx! Do you want to patch this or track it for later patching?

vicentebolea: A possible solution at #3767

PR #3722, opened 2023-07-28
format,ci: update clang-format to clang16

vicentebolea:


eisenhauer (approved):

vicentebolea:
- Disabled WriteRead*MissingPerformGets` tests since they were disabled before this big refactoring but after running clang-format they got re-enabled as the gtest TEST_F does not work if its expanded in a multi-line invocation as it was before this PR. This needs to be resolved.

eisenhauer (approved):

PR #3734, opened 2023-08-02
ci: add libcurl4 to all ubuntu images

### LINKED ISSUE #3694: Add Libcurl in some of our Github builds

scottwittenburg:
Fixes #3694

### DIFF
--- a/scripts/ci/images/Dockerfile.ci-spack-ubuntu20.04-base
+++ b/scripts/ci/images/Dockerfile.ci-spack-ubuntu20.04-base
@@ -40,7 +40,8 @@ RUN . /spack/share/spack/setup-env.sh && \
 
 ### Other missing packages (compared to el8 base):
 RUN apt-get update && apt-get install -y \
-    ccache && \
+    ccache \
+    libcurl4-gnutls-dev && \
     apt-get remove -y \
     gcc-9 \
     g++-9 \


vicentebolea:
The BP5 test error seems to be a race condition: https://open.cdash.org/test/1161780125. This is out of the scope of this PRs.

@eisenhauer

vicentebolea (approved):
The error in the win build is unrelated to this. Note that the OLCF builds have been not working for a couple of hours (outage). Now they are again working but it will take a while to get through the backlog of commits to test.

scottwittenburg:
Ok, I have built and pushed the images resulting from this change.  Feel free to merge if you think it's ok.  I'd do it, but I don't have permission.

vicentebolea:
You should be able to merge now, I removed the requirement for the branch to be uptodate

eisenhauer:
> The BP5 test error seems to be a race condition: https://open.cdash.org/test/1161780125. This is out of the scope of this PRs.
> 
> @eisenhauer

@pnorbert This test seems to have every MPI rank trying to write a file with the same name (and presumably the same contents), then they all try to read it.  Sometimes this works, but sometimes it doesn't, but generally it seems like a bad idea from a making-assumptions-about-how-the-filesystem-will-handle-this-sort-of-thing point of view.  My question:  Is this really necessary?   There's another test here, TestBPWriteReadAsStreamADIOS2_Threads.cpp, that behaves similarly, running with 12 MPI ranks all writing the same file and then maybe successfully reading it.  I'd like to make them stop that, but I want to make sure that I don't stop it in such a way that would abrogate the intention of the test, but I'm not really sure what the intention of the test is...  Can you advise?

PR #3738, opened 2023-08-03
ci: update release instructions

vicentebolea:


vicentebolea:
Merging without review since it is a small change in the docs and I need it to be merged in order be tested. I take full responsibility of its impact :angel:

PR #3743, opened 2023-08-07
readthedocs: solely use requirements.txt

vicentebolea:
Readthedocs is phasing out system packages. I am also disabling conda environment in the readthdocs build since it is redundant and we do alreay install those deps with pip. Also we do not need to install adios2 from conda to build the docs


vicentebolea:
> The "Details" link for the the readthedocs build takes me to a page that doesn't look right. It's kind of like the generated html without any images or stylesheets, where I was expecting a link to the details of running the readthedocs commands. I don't know much about rtd, so I'm not sure if this is expected, or a temporary situation, or some issue caused by this PR.

That is very odd, I Can see it displayed well. Thanks for pointing this out, which browser are you using?

scottwittenburg:
> which browser are you using?

I'm using firefox on linux

vicentebolea:
hmm same as I, could you post an screenshot, I cannot reproduce it

scottwittenburg:
readthedocs_link

vicentebolea:
@scottwittenburg I cannot reproduce this, I open it in the phone/chrome/firefox but I get the correct theme:

https://adios2--3743.org.readthedocs.build/en/3743/

Maybe something with your browser

vicentebolea:
Can you open this correctly: https://adios2.readthedocs.io/en/v2.9.1/

vicentebolea:
Screenshot from 2023-08-07 14-51-51

williamfgc:
FYI, we used conda at that time 'cause the breathe, sphinx rtd packages needed a adios2 build to translate the pybind11 doxygen md annotated code in [pyglue.cpp](https://github.com/ornladios/ADIOS2/blob/master/bindings/Python/py11glue.cpp) to html.

vicentebolea:
> FYI, we used conda at that time 'cause the breathe, sphinx rtd packages needed a adios2 build to translate the pybind11 doxygen md annotated code in [pyglue.cpp](https://github.com/ornladios/ADIOS2/blob/master/bindings/Python/py11glue.cpp) to html.

Many thanks @williamfgc for pointing this out. So I understand that we need an ADIOS2 build in order to generate the Python binding full api page with pybind11, if that is the case I think that we should then stick to conda then. A problem that I can see is that we will be always pulling the latest release from Conda rather than master branch thus changes in these py11glue files will not get reflected right away, but :shrug: I cannot see a reasonable way to resolve this.

williamfgc:
> I cannot see a reasonable way to resolve this.

Thanks @vicentebolea for revisiting this. This is far from perfect and seemed like a chicken-and-egg problem. Perhaps an adios2 wheels would solve this? Mostly asking as I am not that familiar.

vicentebolea:
@williamfgc I tried to build a minimal ADIOS2 build inside readthedocs and it seems to be sufficient for building the readthedocs site.

vicentebolea:
@scottwittenburg I was able to build a minimal adios in readthedocs without much trouble/time

.github/workflows/everything.yml
```diff
@@ -104,7 +104,7 @@ jobs:
 
   linux_ubuntu:
     needs: [format, git_checks]
-    if: needs.git_checks.outputs.num_code_changes > 0
+    if: false
```
scottwittenburg, line 107: Can you summarize why we need to run all these builds now, even when there are no code changes?

vicentebolea: I have to remove these changes before merging I tmp disabled them to avoid being run.

scripts/ci/gh-actions/get-changed-files.sh consider my changes as code changes since I made changes to the readthedocs.yml

scottwittenburg (approved):
Looks good, thanks @vicentebolea!

PR #3749, opened 2023-08-10
CI: add two static builds

vicentebolea:
Closes #1828


.github/workflows/everything.yml
```diff
@@ -137,7 +138,14 @@ jobs:
...
             compiler: clang6
             parallel: serial
-
+          - os: ubuntu20.04
+            compiler: gcc8
+            shared: -static
+            parallel: ompi
+          - os: ubuntu20.04
+            compiler: clang6
+            shared: -static
```
vicentebolea, line 147:
```suggestion
            shared: static
```

.github/workflows/everything.yml
```diff
@@ -137,7 +138,14 @@ jobs:
           - os: ubuntu20.04
             compiler: clang6
             parallel: serial
-
+          - os: ubuntu20.04
+            compiler: gcc8
+            shared: -static
```
vicentebolea, line 143:
```suggestion
            shared: static
```

vicentebolea:
This PR enables two static builds for ADIOS2 however they do not build the adios2 unittests, the reason is that each of our unit test is built in a single file, meaning that when enabling all the common options we build over 200 binaries. When we do a static build this is a problem since each of this binary will have the same size as the sum of all its dependencies. This result in Github actions returning a out-of-space error.

In order to enable tests in the static build we must group the tests in a single (or a few) binaries with the cmake function `create_test_sourcelist` https://cmake.org/cmake/help/book/mastering-cmake/chapter/Testing%20With%20CMake%20and%20CTest.html

vicentebolea:
@scottwittenburg can you review this? I am not sure if this is the best approach. I am adding two builds for linux GHA which results in a total of 13 builds as compared to the maximum of 16 simultaneous CI jobs.

scottwittenburg:
> I am not sure if this is the best approach. I am adding two builds for linux GHA which results in a total of 13 builds as compared to the maximum of 16 simultaneous CI jobs.

What other option is there?  Other than removing two of the existing builds to keep the number fixed?  This seems ok to me, so I'm just approving.  But let me know if I missed the point.

scottwittenburg (approved):

PR #3750, opened 2023-08-11
Organize the processes into groups so that the two steps of metadata …

pnorbert:
…aggregation has more or less the same number or participants. This replaces in-node aggregation in first step. The new strategy balances the size of metadata gathered in the two steps.

### DIFF
--- a/source/adios2/engine/bp5/BP5Writer.cpp
+++ b/source/adios2/engine/bp5/BP5Writer.cpp
@@ -574,14 +574,14 @@ void BP5Writer::EndStep()
         TSInfo.NewMetaMetaBlocks, {m}, {a}, {m_ThisTimestepDataSize},
         {m_StartDataPos});
 
-    if (m_Aggregator->m_Comm.Size() > 1)
+    if (m_AggregatorMetadata.m_Comm.Size() > 1)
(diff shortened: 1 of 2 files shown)


source/adios2/engine/bp5/BP5Writer.cpp
```diff
@@ -860,8 +860,7 @@ uint64_t BP5Writer::CountStepsInMetadataIndex(format::BufferSTL &bufferSTL)
 
         switch (recordID)
         {
-        case IndexRecord::WriterMapRecord:
-        {
```
eisenhauer, line 864: Wait, this is just a formatting change...  Is format checking still broken?

pnorbert: I assume that is clang-format 16 vs 7, but it is accepted. The column limit in release_29 is still 80 characters and this is still the case here.

eisenhauer: Ah, I forgot this was 2.9.  Format checking may still be broken if that fix came with the 100 character limit and so was only merged with master.

eisenhauer (approved):

PR #3753, opened 2023-08-11
ci: add mgard dependency to spack builds

### LINKED ISSUE #3752: Add MGARD to CI

vicentebolea:
Closes #3752 

- It adds MGARD as a dependency.
- Since it requires a recent (but not too recent) MGARD it uses a newer E4S release.
- It fixes a static build issue with duplicated symbols from our pugixml dependency.
- It adds a way to add non-adios2 dependencies to our spack builds.
- It disable Serial tests in MPI builds that have an equivalent dedicated Serial build.
- It removed some OMPI tweaks that breaks newer OMPI.


scripts/ci/images/Dockerfile.ci-spack-ubuntu20.04-base
```diff
@@ -35,6 +35,19 @@ RUN . /spack/share/spack/setup-env.sh && \
       --fail-fast \
       --only dependencies \
       -j$(grep -c '^processor' /proc/cpuinfo) && \
+    spack env create adios2-ci-deps /tmp/specs.yaml && \
```
scottwittenburg, line 38: It's fine if this is working, but I would have thought you'd want to process the dependencies environment first, then when you do the other environments, the `install` step for each one would make sure the dependency specs were installed in the environment as well.  So for each env, you'd add whatever adios2 spec, then add the dependency specs, then concretize, and finally install.  This way, you'd only have to `concretize` each adios2 environment once instead of twice.  Also, I don't think you'd need `concretize` or `install` steps for the dependencies environment in that case either, but maybe the `concretize` can't be avoided.

scripts/ci/images/Dockerfile.ci-spack-ubuntu20.04-base
```diff
@@ -1,43 +1,60 @@
...
+# Install deps
+RUN pip install shyaml
+COPY specs.yaml /tmp/specs.yaml
+RUN . /spack/share/spack/setup-env.sh && \
+    spack env create adios2-ci-deps /tmp/specs.yaml && \
+    spack -e adios2-ci-deps install \
+      --no-check-signature \
+      --include-build-deps \
+      --fail-fast && \
+    spack -e adios2-ci-serial add $(shyaml get-values spack.specs < /tmp/specs.yaml) && \
```
scottwittenburg, line 50: I don't know `shyaml`, but I infer that now you add only the root specs listed in `specs.yaml`, rather than using `spack find` like you did before, which likely added not only root specs, but also all dependencies.  Did you find that this way is working the same or better?

vicentebolea: Newer Spack (the one with this new E4s release) does not accept the `{/hash}` when doing `spack -e env add ...` 

It works the same since during concretization it brings the correct dependencies.

scottwittenburg (approved):
This all looks good to me.  Now that you are the first one to change the E4S base image for the ubuntu builds, can you comment on:

- Did you build with `--progress=plain` or similar to see how the buildcache hit rate was affected?
- Is the time to build and push all the images affected much?
- Did the size of the resulting images change much?

You'll have to push again to get rid of the `tmp` in the image tags everywhere.  I forget if, in this project, that will dismiss my review.  Just ping me if you need me to re-approve after pushing.

vicentebolea:
> This all looks good to me. Now that you are the first one to change the E4S base image for the ubuntu builds, can you comment on:
> 
>     * Did you build with `--progress=plain` or similar to see how the buildcache hit rate was affected?
> 
>     * Is the time to build and push all the images affected much?
> 
>     * Did the size of the resulting images change much?
> 
> 
> You'll have to push again to get rid of the `tmp` in the image tags everywhere. I forget if, in this project, that will dismiss my review. Just ping me if you need me to re-approve after pushing.

Oh, I had the auto-merge enabled for this and it merged before renaming the tags.

I will follow up in another MR and report the information asked by you

vicentebolea:
followup #3768

Issue #3742, opened 2023-08-07
ADIOS2 2.8.3 undefined reference


albenstein:
**Describe the bug**
When I try to compile on HPC system, I get error message like this:

`/beegfs/apps/openfoam-v2306/OpenFOAM-v2306/ThirdParty/build/linux64Gcc/ADIOS2-2.8.3/lib64/libadios2_ffs.so.1: undefined reference to 'setup_for_string_parse'`
`/beegfs/apps/openfoam-v2306/OpenFOAM-v2306/ThirdParty/build/linux64Gcc/ADIOS2-2.8.3/lib64/libadios2_ffs.so.1: undefined reference to 'reset_types_table'`
`/beegfs/apps/openfoam-v2306/OpenFOAM-v2306/ThirdParty/build/linux64Gcc/ADIOS2-2.8.3/lib64/libadios2_ffs.so.1: undefined reference to 'terminate_string_parse'`
`collect2: error: ld returned 1 exit status`
`make[2]: *** [bin/bpls] Error 1`
`make[1]: *** [source/utils/CMakeFiles/bpls.dir/all] Error 2`
`make: *** [all] Error 2`
`Error building: ADIOS2-2.8.3`

**To Reproduce**
A minimal reproducible example is preferred.
Or Steps to reproduce the behavior:
1. Download OpenFOAM and ThirdParty from 'Source: https://dl.openfoam.com/source/v2306/OpenFOAM-v2306.tgz
ThirdParty: https://dl.openfoam.com/source/v2306/ThirdParty-v2306.tgz'
2. Follow build doc on 'https://develop.openfoam.com/Development/openfoam/-/blob/master/doc/Build.md'
3. Try to makeAdios2 with command './makeAdios2'

**Desktop (please complete the following information):**
 - OS/Platform: [CentOS 7.9]
 - Build [compiler version gcc 9.5.0, cmake 3.20.2 ]

eisenhauer:
[...]

PR #3755, opened 2023-08-14
Upstream ffs

eisenhauer:
Mod CMakeLists.txt to kill warning (update min version) and modify cod.l to eliminate old code and possibly address #3742.


vicentebolea (approved):
CI passing

PR #3760, opened 2023-08-15
ci: Create static minimal build

vicentebolea:


vicentebolea:
We are hitting an error here https://open.cdash.org/test/1176036167: The test expect to always use BP5. Does it make sense to have a build BP5=OFF?

@pnorbert

vicentebolea:
@eisenhauer can you review this? Thanks :)

eisenhauer (approved):
Looks good as far as I can see.

PR #3804, opened 2023-09-13
Set AWS version requirement to 1.10.15 and also turn it OFF by defaul…

### LINKED ISSUE #3797: gcc13 compilation problem
```
In file included from /work2/00434/eijkhout/adios2/adios2-git/source/adios2/toolkit/transportman/TransportMan.cpp:33:
/work2/00434/eijkhout/adios2/adios2-git/source/adios2/toolkit/transport/file/FileAWSSDK.h:86:14: error: ‘S3ClientConfiguration’ in namespace ‘Aws::S3’ does not name a type
   86 |     Aws::S3::S3ClientConfiguration *s3ClientConfig = nullptr;
```

Logs attached.
[adios2_logs.zip](https://github.com/ornladios/ADIOS2/files/12579549/adios2_logs.zip)

pnorbert:
…t as it is not a stable feature of ADIOS just yet.

fixes #3797


cmake/DetectOptions.cmake
```diff
@@ -502,9 +502,9 @@ endif()
 
 # AWS S3
 if(ADIOS2_USE_AWSSDK STREQUAL AUTO)
-    find_package(AWSSDK QUIET COMPONENTS s3)
+    find_package(AWSSDK 1.10.15 COMPONENTS s3)
```
vicentebolea, line 505: Unless there is an specif reason to use from 1.10.15, Maybe 1.10.0 is best here since it is less specific and includes every patch release of the >= 1.10.X.

vicentebolea (changes requested):
Looks good, Only a quick question

cmake/DetectOptions.cmake
eisenhauer: I suspect that Norbert took the easy way out and specified "something newer than what I developed with", as opposed to spending time digging into what the real minimum might be.   Maybe 1.10.0 works, if they did the right thing with the path releases, but did they?  Given that this isn't a stable, documented part of ADIOS that we want people to use, I wouldn't worry about this too much.  The purpose here is not to get more issues about AWSSDK until if and when we're ready for it.

eisenhauer (approved):

cmake/DetectOptions.cmake
vicentebolea: Fair enough. Approving too :)

PR #3812, opened 2023-09-18
Fix abs/rel step in BP5 DoCount

eisenhauer:


eisenhauer:
@pnorbert Generally, I think we should be more explicit about when we are dealing with a relative step number or an absolute step number.  There's a ways to go to get there, but variable naming might help.

pnorbert (approved):

PR #3814, opened 2023-09-19
FFS upstream

eisenhauer:
This fixes an OSX 14 problem.

### DIFF
--- a/thirdparty/ffs/ffs/cod/cod.l
+++ b/thirdparty/ffs/ffs/cod/cod.l
@@ -185,7 +185,7 @@ return			{RETURN(RETURN_TOKEN);}
 [A-Za-z_][A-Za-z0-9_]*	{
 				yylval.info.string = create_string_from_yytext();
 				if (is_defined_type(yylval.info.string)) {
-				    RETURN(type_id);
+				    RETURN(type_identifier);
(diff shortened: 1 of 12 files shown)


vicentebolea (approved):
yay!

eisenhauer:
@anagainaru Would you please check this one too to see if it works on OSX 14?

anagainaru:
> @anagainaru Would you please check this one too to see if it works on OSX 14?

I still see the issue with the version of ffs:
```
  The following configuration files were considered but not accepted:

    /ADIOS2/build/thirdparty/ffs/ffs/ffs-config.cmake, version: 3.0.0

```

eisenhauer:
There may be some unpleasantness because I force-pushed the prior version, which might have left some stuff sitting around when you pulled the fresh version.   (I think this because this change doesn't have the version 3.0.0 stuff in it.)  It might be necessary to delete this branch in your current repo and refetch it, and/or blow away the build directory and rebuild.

anagainaru:
@eisenhauer yes you are right, I did that and now it built just fine. So far it looks good, it's at 20% compiling and no complaints. I'll approve this PR when it finishes

anagainaru (approved):

Issue #3806, opened 2023-09-13
Move ADIOS2-examples inside the main ADIOS2 repository


vicentebolea:
- [ ] We want to move the ADIOS2-example source code to the example directory in this repository.
- [ ] We want to get rid of examples that are not uptodate 
- [ ] We want to move the ADIOS2-example readthedocs to the main ADIOS2 (this repo) readthedocs.

Issue #3795, opened 2023-09-09
running tests against installed ADIOS2 library


drew-parsons:
**edit**
The original error reported here (deleted function ‘adios2::ADIOS::ADIOS(void*)’) was apparently a misconfiguration.  The issue title has been changed to indicate the actual goal: running tests against an installed ADIOS2 library for linux distribution CI testing

**Describe the bug**

I'm trying to test an MPI build of adios2 by building the examples against the installed library.
bpFWriteCRead from examples/hello fails with the error:
```
[ 25%] Building CXX object CMakeFiles/CppWriter.dir/CppWriter.o
/usr/bin/mpicxx    -MD -MT CMakeFiles/CppWriter.dir/CppWriter.o -MF CMakeFiles/CppWriter.dir/CppWriter.o.d -o CMakeFiles/CppWriter.dir/CppWriter.o -c /projects/mathlibs/build/adios2/examples/hello/bpFWriteCRead/CppWriter.cpp
/projects/mathlibs/build/adios2/examples/hello/bpFWriteCRead/CppWriter.cpp: In function ‘int main(int, char**)’:
/projects/mathlibs/build/adios2/examples/hello/bpFWriteCRead/CppWriter.cpp:35:39: error: use of deleted function ‘adios2::ADIOS::ADIOS(void*)’
   35 |     adios2::ADIOS adios(MPI_COMM_WORLD);
      |                                       ^
In file included from /usr/include/adios2.h:27,
                 from /projects/mathlibs/build/adios2/examples/hello/bpFWriteCRead/CppWriter.cpp:11:
/usr/include/adios2/cxx11/ADIOS.h:87:5: note: declared here
   87 |     ADIOS(void *define_ADIOS2_USE_MPI_to_use_MPI_constructor) = delete;
      |     ^~~~~
make[2]: *** [CMakeFiles/CppWriter.dir/build.make:76: CMakeFiles/CppWriter.dir/CppWriter.o] Error 1
make[2]: Leaving directory '/projects/mathlibs/build/adios2/examples/hello/bpFWriteCRead/build'
```

 So CppWriter.cpp is trying to use `adios2::ADIOS::ADIOS(void*)`, which is no longer the correct API.

**To Reproduce**
1. Go to 'examples/hello/bpFWriteCRead'
2. Create a new subdir `build` for the build
3.  `cd build; cmake -DCMAKE_CXX_COMPILER=mpicxx ..`
4.  `make`
5. See error

**Expected behavior**

[...]

PR #3820, opened 2023-09-20
Organize and add examples

spyridon97:
This PR performs the following tasks:

1. organizes existing examples
2. adds (most of) the examples from ADIOS2-Examples
3. disables by default BUILD_TESTING, and ADIOS2_BUILD_EXAMPLES
4. add readme.md for each example directory and provides a suggested order of checking the examples

This resolves https://github.com/ornladios/ADIOS2/issues/3806 and (partially?) https://github.com/ornladios/ADIOS2/issues/3795

Fixes: #3806


vicentebolea:
Good work here, few things:

- You can reformat the code in bulk with `sudo ./scripts/developer/run-clang-format.sh`
- As we have much more examples than before how about we disable them by default with `aDIOS_BUILD_EXAMPLES` option defaulting to off but we enable it in our CI builds.

anagainaru:
> Good work here, few things:
> 
> * You can reformat the code in bulk with `sudo ./scripts/developer/run-clang-format.sh`
> * As we have much more examples than before how about we disable them by default with `aDIOS_BUILD_EXAMPLES` option defaulting to off but we enable it in our CI builds.

I would honestly do the same with the tests if possible (`-DADIOS2_BUILD_TESTING=OFF`). Users don't really need them.

spyridon97:
@vicentebolea @anagainaru This PR is ready for review. I tried to reformat, but it does not work for python files for some reason locally, i don't even get the same output as the CI machine

```
~/Programming/Cpp/ADIOS2 organize-and-add-examples ?2 > ./scripts/developer/run-clang-format.sh
[sudo] password for spiros.tsalikis: 
---------- Begin ENV ----------
DISTTAG=f38container
FBR=f38
FGC=f38
HOME=/root
HOSTNAME=f33d9af01d56
OLDPWD=/
PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
PWD=/root/adios2
SHLVL=1
TERM=xterm
_=/usr/bin/env
----------  End ENV  ----------
~/Programming/Cpp/ADIOS2 organize-and-add-examples ?2 > 
```

vicentebolea:
@spyridon97

>  but it does not work for python files for some reason locally, i don't even get the same output as the CI machine

Yep, the script for python is in `scripts/ci/scripts/run-flake8.sh` and it does not edit the source code only complains. You will have to edit those files, perhaps a `black ./examples` will do it, otherwise you might have to go one-by-one and and make sure that they pass.

testing/install/EnginePlugin/CMakeLists.txt
```diff
@@ -43,7 +43,7 @@ endif()
 
 # add write test
 add_executable(adios_plugin_engine_write_test
-  ../../../examples/plugins/engine/examplePluginEngine_write.cpp
+  ../../../examples/plugins/engine/examplePluginEngineWrite.cpp
```
vicentebolea, line 46: Why the rename?

testing/install/EncryptionOperator/CMakeLists.txt
```diff
@@ -17,14 +17,14 @@ set(ENV{ADIOS2_PLUGIN_PATH} "${adios2_DIR}/../../")
 
 # add write test
 add_executable(adios_plugin_operator_write_test
-  ../../../examples/plugins/operator/examplePluginOperator_write.cpp
+  ../../../examples/plugins/operator/examplePluginOperatorWrite.cpp
```
vicentebolea, line 20: Why the rename?

examples/useCases/insituGlobalArrays/CMakeLists.txt
```diff
@@ -3,16 +3,42 @@
...
-target_link_libraries(insituGlobalArraysWriter adios2::cxx11)
+cmake_minimum_required(VERSION 3.12)
+project(ADIOS2UseCasesInSituGlobalArraysExample)
 
-add_executable(insituGlobalArraysReaderNxN insituGlobalArraysReaderNxN.cpp)
-target_link_libraries(insituGlobalArraysReaderNxN adios2::cxx11)
+if(NOT TARGET adios2::cxx11)
+  set(_components C CXX)
+
+  find_package(MPI COMPONENTS ${_components})
```
vicentebolea, line 12:
```suggestion
  find_package(MPI COMPONENTS C)
```
I think that we only need the MPI C components.

examples/simulations/lorenz_ode/CMakeLists.txt
```diff
@@ -0,0 +1,27 @@
...
+#------------------------------------------------------------------------------#
+
+cmake_minimum_required(VERSION 3.12)
+project(ADIOS2SimulationsLorenzOdeExample)
+
+if(NOT TARGET adios2::cxx11)
+  set(_components C CXX)
+
+  find_package(ADIOS2 2.7.1 REQUIRED COMPONENTS ${_components})
+  if(NOT ADIOS2_FOUND)
```
vicentebolea, line 13: This is unneeded:
`find_package(ADIOS2 2.7.1 REQUIRED COMPONENTS ${_components})`

will exit with error if adios is not found

examples/simulations/heatTransfer/write/IO_adios2.cpp
```diff
@@ -12,7 +12,7 @@
 
 #include <string>
 
-#include <adios2.h>
+#include "adios2.h"
```
vicentebolea, line 15: <> should be fine.

examples/plugins/engine/CMakeLists.txt
```diff
@@ -3,33 +3,43 @@
...
 #------------------------------------------------------------------------------#
 
+cmake_minimum_required(VERSION 3.12)
+project(ADIOS2PluginsEngineExample)
+
+if(NOT TARGET adios2::cxx11 AND NOT TARGET adios2_core)
+  set(_components C CXX)
+
+  find_package(ADIOS2 2.7.1 REQUIRED COMPONENTS ${_components})
+  if(NOT ADIOS2_FOUND)
```
vicentebolea, line 13: We dont need this

examples/plugins/engine/CMakeLists.txt
```diff
@@ -3,33 +3,43 @@
 # accompanying file Copyright.txt for details.
 #------------------------------------------------------------------------------#
 
+cmake_minimum_required(VERSION 3.12)
+project(ADIOS2PluginsEngineExample)
+
+if(NOT TARGET adios2::cxx11 AND NOT TARGET adios2_core)
+  set(_components C CXX)
```
vicentebolea, line 10: C should be enough

examples/plugins/operator/CMakeLists.txt
```diff
@@ -2,12 +2,28 @@
...
-
-add_executable(exampleOperatorPlugin_read
-  examplePluginOperator_read.cpp
-)
-target_link_libraries(exampleOperatorPlugin_read adios2::cxx11)
+
+cmake_minimum_required(VERSION 3.12)
+project(ADIOS2PluginsOperatorExample)
+
+if(NOT TARGET adios2::cxx11)
```
vicentebolea, line 9: better target would be `adios2_core`

examples/plugins/CMakeLists.txt
```diff
@@ -4,4 +4,6 @@
 #------------------------------------------------------------------------------#
 
 add_subdirectory(engine)
-add_subdirectory(operator)
+if(ADIOS2_HAVE_Sodium)
+  add_subdirectory(operator)
+endif()
```
vicentebolea, line 9: We already test for this in the operator cmakelists

vicentebolea (changes requested):
Great job here, this is looking good!

The review has taken me a while since it is 406 files changes, I could not go onto details in the added parts, but that is something for the CI. I left some notes. Other notes:

- We will need to reformat the shell script too to pass the linters, After you fix the python formatting you will see the errors.
- We might have to disable testing in the static builds if they fail since we hit the space limit in github actions runners.
- Once you get the smallest and most succint cmake partern for testing that whether the example is part of the adios worktree apply it to all the tests.

Once again thanks for taking care of this, this is indeed a very laborious effort, but it is looking great :+1: I will be tomorrow online for afew hours before I leave for vacations I will have another look then

anagainaru (changes requested):
Nice work, this was a lot of reorganizing ! I tried to look over most of the changes (source files and not cmake) but there were a lot of files modified so I am sure I missed things. 

I built ADIOS2 with your branch on my laptop and the Tests were built so the default is not `OFF`. This needs to be fixed or if you want to leave them, the build script need to have `-D BUILD_TESTING=OFF` (you removed this).

One additional comment:
- In some cases our examples use `#if ADIOS2_USE_MPI` and in other we check in cmake and do not generate the example if MPI is not present. Is there a reason a code uses one or the other?

This is more a discussion with @vicentebolea / @pnorbert / @eisenhauer, but didn't we decide we want to keep the complex examples in ADIOS2-Examples and import them in the docs from there? I really don't think we need all of these examples in the ADIOS2 repo (Kokkos-{backend} for each backend is a bit much, it makes sense in a repo dedicated to ADIOS2 examples but not in the main library). I am assuming this was discussed when I wasn't there but I really don't think we should bulk up the library with more than basic examples.

vicentebolea:
> This is more a discussion with @vicentebolea / @pnorbert / @eisenhauer, but didn't we decide we want to keep the complex examples in ADIOS2-Examples and import them in the docs from there? I really don't think we need all of these examples in the ADIOS2 repo (Kokkos-{backend} for each backend is a bit much, it makes sense in a repo dedicated to ADIOS2 examples but not in the main library). I am assuming this was discussed when I wasn't there but I really don't think we should bulk up the library with more than basic example

I understood from our last few conversations that the consensus is to move all the adios2-examples examples to our main repo and merge then with our main repo examples. What I originally proposed was to simply move the readthedocs for adios2-examples our main repo but in the meeting the consensus was to also move the examples itself. The reason was that it is confusing as it is having two sets of examples. At the end of the day i do not consider it a big deal eitherway. Two things to watch out:

- We might want to trim down the number of examples.
- The examples source coding standards should not be held to the same level as the aDIOS2 source code, this is unnecessary effort and it will slow down the maintenance of them.

Connecting to the previous point, we might want the examples to be a `git submodule` so that:

- We do not have duplication of code and the adios2-examples (criteria 1)
- We can have less strict coding standards (Criteria 2)
- As it is part of the adios2 main repo too, it will be kept uptodate. (Criteria 3)

The drawback is that we would have to bump the adios2-example commit version time to time, we can add an automatized rule that it does it whenever something is commited in master in adios2-examples.

What do you all think about it?

examples/useCases/insituGlobalArrays/CMakeLists.txt
spyridon97: In some tests we need the fortran also, so use the ``_components`` arguments to pass the same component parts to both MPI and ADIOS2 for simplicity.

examples/simulations/lorenz_ode/CMakeLists.txt
spyridon97: We will exit with an error, but if we compile this test as an individual project find_package is needed to make use of the ADIOS2_DIR, otherwise ADIOS2_DIR is useless.

examples/plugins/CMakeLists.txt
spyridon97: We do, but if we don't build the operator the example will be built, we can not use the example.

testing/install/EnginePlugin/CMakeLists.txt
spyridon97: For naming consistency

examples/ReadMe.md
```diff
@@ -0,0 +1,17 @@
...
+
+1. [hello](hello): The _hello_ examples are meant to introduce you to fundamental capabilities of ADIOS2, such as
+   reading and writing.
+2. [basics](basics): The _basics_ examples are meant to introduce you to basic concepts of ADIOS2, such as
+   global/joined/local arrays, values, and variables shapes.
+3. [useCases](useCases): The _useCases_ examples are meant to demonstrate how to use ADIOS2 in different scenarios,
+   such as in situ visualization, and fides schema.
+4. [plugins](plugins): The _plugins_ examples are meant to introduce you to the plugin capabilities of ADIOS2, such as
+   how to develop your own engine or operators.
+5. [simulations](simulations): The _simulations_ examples are meant to demonstrate how to integrate ADIOS2 within your
```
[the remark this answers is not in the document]
spyridon97: I improved the description of 1, hopefully that addresses your concerns.

spyridon97:
> I built ADIOS2 with your branch on my laptop and the Tests were built so the default is not `OFF`

I fixed that now.

spyridon97:
> * In some cases our examples use `#if ADIOS2_USE_MPI` and in other we check in cmake and do not generate the example if MPI is not present. Is there a reason a code uses one or the other?

Yes, some examples can compile with both mpi on and off, and some can compile only with mpi on.

anagainaru:
> Yes, some examples can compile with both mpi on and off, and some can compile only with mpi on.

I was mainly asking if there is a reason for choosing one or the other? It's not a huge deal, we can deal with this some other time.

Norbert convinced me to go with this version of examples in the main repo so I will remove my request for changes.

anagainaru (approved):

spyridon97:
@pnorbert I would appreciate checking that all the simulation examples works as expected along with their scripts.

vicentebolea:
@spyridon97 my changes are ready

vicentebolea (changes requested):
Last round of changes, everything else looks good. I have confirmed that we build the examples in every job in the CI. We might have to adjust this later to disable examples in some already slow builds (static builds probably). Make the discussed changes, rebase to the last master and we can have this merged.

Fantastic work @spyridon97 :tada:

vicentebolea:
I have made the changes to speedup the process

spyridon97:
> Last round of changes, everything else looks good. I have confirmed that we build the examples in every job in the CI. We might have to adjust this later to disable examples in some already slow builds (static builds probably). Make the discussed changes, rebase to the last master and we can have this merged.
> 
> Fantastic work @spyridon97 🎉

What discussed changes do you refer to that i have not already done?

CMakeLists.txt
```diff
@@ -293,11 +292,15 @@ if(MSVC AND BUILD_SHARED_LIBS AND ADIOS2_HAVE_HDF5)
   add_definitions(-DH5_BUILT_AS_DYNAMIC_LIB=1)
 endif()
 
+# Ctest creates BUILD_TESTING option and sets it to true by default
+# Here we disable BUILD_TESTING option by default
+set(BUILD_TESTING OFF CACHE BOOL "Build testing")
+include(CTest)
+mark_as_advanced(BUILD_TESTING)
```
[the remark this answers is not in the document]
vicentebolea: @spyridon97 this one, but I have just made the change

vicentebolea (dismissed):

vicentebolea:
@spyridon97 I have also remove the  mark_advance(BUILD_TESTING) since it is a very common CMAKE variable that the user might expect it to see it when doing `cmake -L` or `ccmake`.

vicentebolea:
If you approve the latest changes please merge after the CI passes.

vicentebolea (approved):

PR #3824, opened 2023-09-23
Added support to query local arrays

guj:
1) removed unused code sections
2) added support for Local Arrays
3) added test for local arrays in python
4) result can be returned as vector of start/count (global array only),  or vector of block ids


testing/adios2/performance/query/TestBPQuery.cpp
```diff
@@ -31,10 +31,10 @@ void WriteXmlQuery1D(const std::string &queryFile, const std::string &ioName,
     file << "   <var name=\"" << varName << "\">" << std::endl;
     file << "      <boundingbox  start=\"5\" count=\"80\"/>" << std::endl;
     file << "       <op value=\"OR\">" << std::endl;
-    file << "         <range  compare=\"GT\" value=\"6.6\"/>" << std::endl;
+    file << "         <range  compare=\"GT\" value=\"100.6\"/>" << std::endl;
     file << "         <range  compare=\"LT\" value=\"-0.17\"/>" << std::endl;
     file << "         <op value=\"AND\">" << std::endl;
-    file << "            <range  compare=\"LT\" value=\"2.9\"/>" << std::endl;
+    file << "            <range  compare=\"LT\" value=\"11.9\"/>" << std::endl;
     file << "            <range  compare=\"GT\" value=\"2.8\"/>" << std::endl;
```
anagainaru, line 38: I would not use the heat transfer data for the `Testing` files. In the examples you could use whatever you want, but when testing a function I would use some data that is easy to check correctness for. It is not clear to me why the correct answer for BP4 is `{2,1,1}`.

anagainaru (commented):
The changes look good at a first glance, I will try running it later today and see how it works with my code. My concern is with the testing code, we can discuss tomorrow at the meeting.

testing/adios2/performance/query/TestBPQuery.cpp
eisenhauer: @guj , can you address @anagainaru 's concerns?

guj: Sure, I didn't see them

guj: This test does not use heat transfer data. Rather it first writes a file, and then query it.

examples/query/test.cpp
```diff
@@ -10,7 +10,29 @@
 #include <string>
 #include <vector>
 
-// #include "adios2/toolkit/query/Worker.h"
+// returns block ids
```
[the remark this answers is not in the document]
guj: changed the comment

eisenhauer (dismissed):

guj:
Too  many commits after rebase. 
Start a new pull request instead: https://github.com/ornladios/ADIOS2/pull/3868

PR #3826, opened 2023-09-26
MDR Operator

pnorbert:
Refactor operator using MGARD-x MDR feature. 
Useless in its current form and does not have testing yet but it can refactor float/double arrays.
Reconstruction is full due to lack of support for reader side operator arguments, but MDR itself supports accuracy.


eisenhauer (commented):
Really, my only question would be whether or not we want to embed this particular definition of Accuracy as an ADIOS type (including fixing it in external bindings, etc.).  I'm assuming that other lossy compressors might have other ways to express accuracy as well (or at least their own nomenclatures).  Should we consider leaving the accuracy expression in string form instead?  Prefix these types with MGD or include them in specific namespaces so that if/when we have future lossy compressors we're inclusive of their definitions of accuracy?

pnorbert:
> Really, my only question would be whether or not we want to embed this particular definition of Accuracy as an ADIOS type (including fixing it in external bindings, etc.). 
Excellent question. I was aiming to create a definition that should encompass all compression/refactoring, instead of a simple error value. However, there is the danger that we don't get this right. 

BTW, MDR does not seem to work with norms other than inf (Linf norm) and 0.0 (L2 norm) but this may change, and only with absolute error but we aim to provide both relative and absolute errors on user request. So it's not clear how a user requesting `a1` accuracy should deal with the `a2` accuracy returned by the operator after the read. 

So, I did not want a simple double value, but also not a some hierarchy of classes for accuracy when defining the interface and came up with this triplet as a probably good enough first try. A string instead would be the safest but ugliest first try in my opinion, wanted something better than that.

eisenhauer:
Sounds reasonable.  We could always add other overloaded get/put operations with string or attribute if we got in a bind with this version...

eisenhauer (approved):

vicentebolea:
This PR added some memory leaks that were caught by our newly working (albeit not completely) asan build as seen here: https://app.circleci.com/pipelines/github/ornladios/ADIOS2/7912/workflows/9dd388ff-4437-4450-80aa-1c4e5cfeda6a/jobs/58636

pnorbert:
I think these leaks are in other branches as well and doesn't seem to be
related to this code, even though the new test fails

PR #3828, opened 2023-09-26
WIP: ci: Improve performance of mpich tests in CI

scottwittenburg:
Use the version of mpich built into the ecpe4s/ubuntu20.04:23.05 image instead of building a new one.  This replaces an mpich that used ch3/tcp with one that uses ch4/ofi.

### DIFF
--- a/scripts/ci/images/Dockerfile.ci-spack-ubuntu20.04-base
+++ b/scripts/ci/images/Dockerfile.ci-spack-ubuntu20.04-base
@@ -11,6 +11,14 @@ RUN . /spack/share/spack/setup-env.sh && \
     spack config add "config:checksum:false" && \
     spack config add "config:build_jobs:$(nproc)" && \
     spack config add "concretizer:unify:false" && \
+    spack env create adios2-ci-mpich && \
+    spack -e adios2-ci-mpich add ${ADIOS_SPEC_MPICH} && \
+    spack -e adios2-ci-mpich concretize && \
+    spack -e adios2-ci-mpich install \
+      --include-build-deps \
+      --no-check-signature \
+      --fail-fast \
+      --only dependencies && \
     spack env create adios2-ci-serial && \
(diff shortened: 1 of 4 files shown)


vicentebolea:
@scottwittenburg this looks great, I will give a better review in the coming days

scripts/ci/images/Dockerfile.ci-spack-ubuntu20.04-base
```diff
@@ -11,6 +11,14 @@ RUN . /spack/share/spack/setup-env.sh && \
     spack config add "config:checksum:false" && \
     spack config add "config:build_jobs:$(nproc)" && \
     spack config add "concretizer:unify:false" && \
+    spack env create adios2-ci-mpich && \
```
vicentebolea, line 14: Why has this to be moved up?

vicentebolea:
@scottwittenburg does this improved the perf of mpich tests? Maybe we have to run the CI a few times to check this.

Note that previously ADIOS2+MPICH+ch4/ofi was very flaky, maybe is not the case anymore with a newer MPICH version.

scottwittenburg:
> @scottwittenburg does this improved the perf of mpich tests?

No, this didn't help at all, and since it took about an hour (like the previous mpich tests), I'm not sure running more times will help.  

I may close it, or else use this PR as a way to test other approaches.

scripts/ci/images/Dockerfile.ci-spack-ubuntu20.04-base
scottwittenburg: Because it failed when it was done at the end.  I have no idea why it works when done first, and didn't take the time to investigate.

vicentebolea:
@scottwittenburg I guess that this was superseded by #3883, shall we close?

scottwittenburg:
Yes, thanks.

PR #3835, opened 2023-10-03
wip: temporarily disable mpi dataplane as a test

scottwittenburg:


scottwittenburg:
@eisenhauer The mpich tests alone still took over 51 minutes with the mpi dataplane commented out, as you could see [here](https://github.com/ornladios/ADIOS2/actions/runs/6397192456/job/17364643003?pr=3835) if you want.

So it would seem something else must be causing the tests to take twice as long when adios2 is built against mpich instead of openmpi.

eisenhauer:
I haven't looked at this in detail yet, but I'm thinking that we may need to drill down a level to identify individual tests where we have big timing differences between mpich and ompi.  That may help sort out what's going on.  (I may be able to dig later today, but tied up this morning at least.)

scottwittenburg:
> we may need to drill down a level to identify individual tests where we have big timing differences between mpich and ompi

I was thinking the same thing, so I'm going to try and generate something like that from the CDash results.  Don't worry about digging into this yet.

scottwittenburg:
> I'm going to try and generate something like that from the CDash results.

@eisenhauer @vicentebolea I've now done this for a recent set of tests run against master, and posted the results [here](https://github.com/ornladios/ADIOS2/issues/3616#issuecomment-1766898953).

scottwittenburg:
This was just a test, closing.

PR #3848, opened 2023-10-17
MPI: add timeout for conf test for MPI DataPlane

### LINKED ISSUE #3836: make ADIOS2_HAVE_MPI_CLIENT_SERVER as an option
ranjansv@gatech.edu 
is configuring ADIOS on Cambridge nodes with their gcc and mpich.

It so turns out the test ADIOS2_HAVE_MPI_CLIENT_SERVER
hang during cmake.

Is it possible  to have an option to disable this  or give it a timeout?

vicentebolea:
Fixes #3836 

This adds a c++ timer to the configure tests that determines whether the MPI DataPlane can be used.

@guj

### DIFF
--- a/cmake/DetectOptions.cmake
+++ b/cmake/DetectOptions.cmake
@@ -445,9 +445,11 @@ if(ADIOS2_USE_SST AND NOT WIN32)
     endif()
   endif()
   if(ADIOS2_HAVE_MPI)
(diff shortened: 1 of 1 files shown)


eisenhauer (approved):
Looks reasonable.  Thanks!

Issue #3855, opened 2023-10-18
Example fails to build in Crusher


vicentebolea:
Apparently Crusher CI was not available yesterday (it could be that they  normally perform maintenance on Tuesdays) and we missed this example build error in #3820.

https://code.olcf.ornl.gov/ci/csc303_crusher/dev/adios2/-/jobs/29807

vicentebolea:
@spyridon97 you can debug this by opening a new PR since Crusher CI is now back online.

vicentebolea:
@spyridon97 we still have the error in current master as shown here: https://code.olcf.ornl.gov/ci/csc303_crusher/dev/adios2/-/jobs/30050

Note that the Crusher CI takes about 30mins-2h to report. Please make sure that you have the Crusher CI status passing before merging

vicentebolea:
@spyridon97 was this resolved?

Issue #3840, opened 2023-10-10
Update examples


spyridon97:
1. Writers to use BeginStep(), EndStep()
3. Make sure readers call BeginStep() before using AvailableVariables()/InquireVariables()
4. Convert BP3/BP4/BP5 to BPFile

spyridon97:
@anagainaru @eisenhauer @pnorbert if you have in mind anything else to updates write it here.

PR #3856, opened 2023-10-19
Update examples

spyridon97:
1. Examples: Add BeginStep/EndStep wherever it was missing
2. Examples: Use BPFile instead of BP3/4/5 for future-proof
3. Update the bpWriterReadHip example's cmake to run on crusher

This PR fixes https://github.com/ornladios/ADIOS2/issues/3855, https://github.com/ornladios/ADIOS2/issues/3840.

### DIFF
--- a/examples/hello/bpReader/bpReader.cpp
+++ b/examples/hello/bpReader/bpReader.cpp
@@ -41,6 +41,7 @@ int main(int argc, char *argv[])
         /** Engine derived class, spawned to start IO operations */
         adios2::Engine bpReader = bpIO.Open(filename, adios2::Mode::Read);
 
+        bpReader.BeginStep();
         const std::map<std::string, adios2::Params> variables = bpIO.AvailableVariables();
 
         for (const auto &variablePair : variables)
@@ -96,6 +97,7 @@ int main(int argc, char *argv[])
                 std::cout << "\n";
             }
         }
+        bpReader.EndStep();
 
         /** Close bp file, engine becomes unreachable after this*/
         bpReader.Close();
--- a/examples/hello/bpReader/bpReaderHeatMap2D.cpp
+++ b/examples/hello/bpReader/bpReaderHeatMap2D.cpp
@@ -75,7 +75,9 @@ int main(int argc, char *argv[])
         /** Will create HeatMap.bp */
         adios2::Engine bpWriter = putHeatMap.Open("HeatMap2D.bp", adios2::Mode::Write);
 
+        bpWriter.BeginStep();
         bpWriter.Put(outTemperature, temperatures.data());
+        bpWriter.EndStep();
         bpWriter.Close();
 
         // ************************** READ
@@ -85,6 +87,7 @@ int main(int argc, char *argv[])
             adios2::Engine bpReader =
                 getHeatMap.Open("HeatMap2D.bp", adios2::Mode::Read, MPI_COMM_SELF);
 
+            bpReader.BeginStep();
             // this just discovers in the metadata file that the variable exists
             adios2::Variable<unsigned int> inTemperature =
                 getHeatMap.InquireVariable<unsigned int>("temperature");
@@ -111,6 +114,7 @@ int main(int argc, char *argv[])
                 }
                 std::cout << "\n";
             }
+            bpReader.EndStep();
 
             bpReader.Close();
         }
--- a/examples/hello/bpWriteReadHip/bpWriteReadHip.cpp
+++ b/examples/hello/bpWriteReadHip/bpWriteReadHip.cpp
@@ -120,7 +120,7 @@ int main(int argc, char **argv)
         std::cout << "[BPWrite] error: " << hipGetErrorString(hipExit) << std::endl;
         return 1;
     }
-    const std::vector<std::string> list_of_engines = {"BP4", "BP5"};
+    const std::vector<std::string> list_of_engines = {"BPFile"};
     const size_t N = 6000;
     int nSteps = 2, ret = 0;
 
--- a/examples/hello/bpWriteReadKokkos/bpWriteReadKokkos.cpp
+++ b/examples/hello/bpWriteReadKokkos/bpWriteReadKokkos.cpp
@@ -98,7 +98,7 @@ int BPRead(const std::string fname, const size_t N, int nSteps, const std::strin
 
 int main(int argc, char **argv)
 {
-    const std::vector<std::string> list_of_engines = {"BP4", "BP5"};
+    const std::vector<std::string> list_of_engines = {"BPFile"};
     const size_t N = 6000;
     int nSteps = 2, ret = 0;
 
--- a/examples/plugins/engine/examplePluginEngineRead.cpp
+++ b/examples/plugins/engine/examplePluginEngineRead.cpp
@@ -16,47 +16,12 @@
 
 #include "adios2.h"
 
-void testStreaming(adios2::Engine &reader, std::vector<float> &myFloats,
-                   adios2::Variable<float> &var)
-{
-    for (int i = 0; i < 2; i++)
-    {
-        if (i == 1)
-        {
-            for (auto &num : myFloats)
-            {
-                num *= 2;
-            }
-        }
-
-        reader.BeginStep();
-        std::vector<float> readFloats;
-        reader.Get(var, readFloats);
-        reader.EndStep();
-
-        if (readFloats == myFloats)
-        {
-            std::cout << "data was read correctly!" << std::endl;
-        }
-        else
-        {
-            std::cout << "data was not read correctly!" << std::endl;
-        }
-    }
-}
-
 int main(int argc, char *argv[])
 {
     std::string config;
     if (argc > 1)
     {
         config = std::string(argv[1]);
-    }
-
-    bool streaming = false;
-    if (argc > 2)
-    {
-        streaming = std::atoi(argv[2]) == 1;
     }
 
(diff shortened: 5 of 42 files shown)


examples/hello/bpWriteReadHip/bpWriteReadHip.cpp
```diff
@@ -120,7 +120,7 @@ int main(int argc, char **argv)
         std::cout << "[BPWrite] error: " << hipGetErrorString(hipExit) << std::endl;
         return 1;
     }
-    const std::vector<std::string> list_of_engines = {"BP4", "BP5"};
+    const std::vector<std::string> list_of_engines = {"BPFile"};
```
caitlinross, line 123: I'm wondering if this should be left as it was. It appears the original author may have wanted this to test out for both BP4 and BP5 since they loop through `list_of_engines`.

examples/hello/bpWriteReadKokkos/bpWriteReadKokkos.cpp
```diff
@@ -98,7 +98,7 @@ int BPRead(const std::string fname, const size_t N, int nSteps, const std::strin
 
 int main(int argc, char **argv)
 {
-    const std::vector<std::string> list_of_engines = {"BP4", "BP5"};
+    const std::vector<std::string> list_of_engines = {"BPFile"};
```
caitlinross, line 101: Same comment here as for `bpWriteReadHip.cpp`

examples/plugins/engine/examplePluginEngineRead.cpp
```diff
@@ -82,30 +89,34 @@ int main(int argc, char *argv[])
...
+                if (readFloats == myFloats)
+                {
+                    std::cout << "data was read correctly!" << std::endl;
+                }
+                else
+                {
+                    std::cout << "data was not read correctly!" << std::endl;
+                }
             }
+            reader.EndStep();
```
caitlinross, line 119: Maybe this part should be removed? I think when I originally wrote it my thinking was to show using it with Begin/EndStep (in the testStreaming function) and then to show that it works with just using PerformGets and not using Begin/EndStep, but it should probably just be simplified to only do what's in the testStreaming function.

examples/plugins/engine/examplePluginEngineWrite.cpp
```diff
@@ -84,8 +84,10 @@ int main(int argc, char *argv[])
         }
         else
         {
+            writer.BeginStep();
             writer.Put<float>(var, myFloats.data());
             writer.PerformPuts();
+            writer.EndStep();
```
caitlinross, line 90: Same comment as for `examplePluginEngineRead.cpp`

examples/hello/bpWriteReadHip/bpWriteReadHip.cpp
spyridon97: I discussed that with @eisenhauer and he said that we should just remove it.

examples/hello/bpWriteReadKokkos/bpWriteReadKokkos.cpp
spyridon97: I discussed that with @eisenhauer and he said that we should just remove it.

examples/hello/bpReader/bpReaderHeatMap2D.cpp
```diff
@@ -75,7 +75,9 @@ int main(int argc, char *argv[])
         /** Will create HeatMap.bp */
         adios2::Engine bpWriter = putHeatMap.Open("HeatMap2D.bp", adios2::Mode::Write);
 
+        bpWriter.BeginStep();
         bpWriter.Put(outTemperature, temperatures.data());
+        bpWriter.EndStep();
         bpWriter.Close();
 
         // ************************** READ
```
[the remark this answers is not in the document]
spyridon97: Fixed

examples/hello/bpReader/bpReader.cpp
```diff
@@ -41,6 +41,7 @@ int main(int argc, char *argv[])
         /** Engine derived class, spawned to start IO operations */
         adios2::Engine bpReader = bpIO.Open(filename, adios2::Mode::Read);
 
+        bpReader.BeginStep();
```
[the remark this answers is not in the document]
spyridon97: Fixed.

examples/plugins/engine/examplePluginEngineRead.cpp
spyridon97: Fixed

examples/plugins/engine/examplePluginEngineWrite.cpp
spyridon97: Fixed

spyridon97:
@caitlinross can you understand why this test fails? https://open.cdash.org/test/1266322909

I have beginStep and endStep.

williamfgc:
@spyridon97 could be a race condition in which reader is faster than the writer. The error message says the metadata is empty, but the exception should be fixed in BP5 or just rely on checking on `var`.

caitlinross:
I don't think it's a race condition. The read test doesn't run until the write test completes (I checked - there's a dependency set in the relevant CMakeLists.txt). But the write test is segfaulting (not sure why since), so that's why the read test is failing because the bp file written in the write test is corrupted for some reason. Did you try running that test locally to see if it fails? Looks like I had it set to use the BP4 engine before, so I'm not sure why it would fail when switching to BP5 engine.

caitlinross:
I'm getting something weird when trying to run locally. To run: `ctest -V -R Install.Cmake.Encryption`

So what happens results in 2 steps:
1) `Install.Setup` where the adios library is built. This part completes fine locally and looks like it passes in CI as well.
2) `Install.CMake.EncryptionOperator` This actually has a CMake file that is its own project separate from the other CMake files for ADIOS. So this finds the adios2 package (built in step 1), then it builds the write and read examples, and then runs those 2 tests. If those 2 tests pass, then this step 2 passes (as in the outer step 2, Install.CMake.EncryptionOperator). If I change the engine back to BP4, there's no issue. But when I change it to BP5, it seems to hang at the part where it's building the write/read examples. I have no idea how that change could cause it to hang at that point...

caitlinross:
Okay, I think the hanging/time out is a red herring. It seems that the tests are being built and run, but because step 2 above contains its own tests, something about how those are failing is causing it to look like it's hanging when trying to run with CTest. 

Anyway running the example outside of CTest I investigated the segfault. It's failing in BP5Serializer.cpp:736. `Offsets` is null. 

@eisenhauer the example code in question is `adios2/examples/plugins/operator/examplePluginOperatorWrite.cpp`. A local variable is created and then an operator (the EncryptionOperator plugin) is added to it. If I switch the engine back to BP4, it runs successfully. If I set the engine to BP5 and remove the operator, it's fine too. It seems to be an issue with having an operator on a local variable in BP5. Any known issues for this? Perhaps I am doing something incorrectly in the plugin that was fine before BP5?

eisenhauer:
BP5 does a lot of things differently than BP4, including keeping more information in the engine (and not in the variable, which might be shared with other engines).  I'll take a look at this, but it'll probably be tomorrow before I get to it.

eisenhauer:
Actually did get a chance for a quick look.  BP3/4 operator stuff was based on the BPInfo struct which was largely eliminated in BP5 because it was responsible for significant overhead.  So operator handling was relocated and yes, we must not have any tests for operators on local variables because that case was missed.  Can you try it with this patch?
```
diff --git a/source/adios2/toolkit/format/bp5/BP5Serializer.cpp b/source/adios2/toolkit/format/bp5/BP5Serializer.cpp
index 8b10c30b4..017f23854 100644
--- a/source/adios2/toolkit/format/bp5/BP5Serializer.cpp
+++ b/source/adios2/toolkit/format/bp5/BP5Serializer.cpp
@@ -733,7 +733,8 @@ void BP5Serializer::Marshal(void *Variable, const char *Name, const DataType Typ
             for (size_t i = 0; i < DimCount; i++)
             {
                 tmpCount.push_back(Count[i]);
-                tmpOffsets.push_back(Offsets[i]);
+		if (Offsets)
+		    tmpOffsets.push_back(Offsets[i]);
             }
             size_t AllocSize = ElemCount * ElemSize + 100;
             BufferV::BufferPos pos = CurDataBuffer->Allocate(AllocSize, ElemSize);
```
If that works we can format it properly and get it in a PR.

caitlinross:
Yup that fixes it.

pnorbert:
Thank you all for cleaning up the examples. Since BP5 is the default engine
(i.e. BPFile is BP5), do you feel if we just should drop examples for BP3
and BP4? All new users should use BP5 unless advised to use something else.
There is no need to teach the peculiarities of the other engines, only the
proper use of BP5.

eisenhauer:
> Yup that fixes it.

Sweet!  I gotta run now, but I'll do a PR later, or you guys can, whichever.  (And I guess we might want to think about a test for that case?)

caitlinross:
@eisenhauer I can go ahead and take care of it. And it'd be pretty easy to write a test for this case (basically the same thing this encryption operator write test is doing, except use a built in compression operator to simplify the test), so I can add that as well.

williamfgc:
@pnorbert agree to drop BP3/BP4. BP5 has the advantage of all the learning from BP3 and BP4 and breaks too much API. I wouldn't even call it BP5, just BPFile for the end user moving forward.

caitlinross:
@spyridon97 if you rebase on master, that operator test should be fixed now.

spyridon97:
@caitlinross can you approve this PR?

caitlinross (approved):

PR #3862, opened 2023-10-20
Remote Server Improvements

eisenhauer:
- Add idle timeout to server.  Unless the -no_timeout flag is specified, the server will exit after 10 minutes of idle time
- Specify the server port with a single constant in remote_common.h, rather than several explicit constants.


eisenhauer:
> One small comment, isn't CManager_close(cm) called at the destruction of the Singleton? Can we call this multiple times?

CManager does have a reference count, so we can do it multiple times.  It might not be quite right in this case, but this is really just a backstop to prevent an unfortunate situation.  Sometimes when you use ssh port forwarding to run the server and kill ssh with ^C, the server process remains running forever on whatever random login node had been selected by the ssh.  The timeout is just to make sure that those things go away in relatively short order.  A perfectly clean shutdown isn't necessary isn't the first priority.

PR #3875, opened 2023-10-27
Fix data length calculation for hash

eisenhauer:
Turns out that master already has this fix because it migrated into FFS master in July.  However, we haven't pulled in fresh FFS to release because there were significant API changes and we didn't want those in a patch release.  So this is a one-line change that fixes our hash issues (which were not really hash issues, but instead "calculating the length over which to do a hash" issues).


anagainaru (approved):

vicentebolea (approved):
It seems that the issue with crusher is harder than what we thought. Lets ignore for the time being.

PR #3876, opened 2023-10-28
WIP: Ci docker use external mpich

scottwittenburg:
Build a non-spack mpich using `ch3:sock:tcp`, which spack doesn't support, and work it to spack environment in the ci image as an external.  The `ch3:sock:tcp` build should handle the oversubscription in CI much better, resulting in faster tests when building with mpich.


scottwittenburg:
It's odd that using the `ch3:sock:tcp` version sped things up so much on my desktop, but didn't seem to make a difference at all here.  I'll run again locally now that's it's built into the test image.

scripts/ci/images/Dockerfile.ci-spack-ubuntu20.04-base
```diff
@@ -1,13 +1,17 @@
...
+FROM ecpe4s/ubuntu20.04-runner-amd64-gcc-11.4:2023.08.01
+
+ARG MPICH_VERSION=4.1.1
+ARG E4S_VERSION=23.08
 
 COPY packages.yaml /etc/spack/packages.yaml
 
 # Install Base specs
-RUN . /spack/share/spack/setup-env.sh && \
+RUN git clone --depth 1 --single-branch --branch e4s-${E4S_VERSION} https://github.com/spack/spack && \
```
vicentebolea, line 10: spack should be already installed from the ecpe4s image, is there any reason for this clone?

scripts/ci/images/Dockerfile.ci-spack-ubuntu20.04-base
```diff
@@ -26,7 +30,52 @@ RUN . /spack/share/spack/setup-env.sh && \
...
-      --only dependencies && \
+      --only dependencies
+
+# # Build external mpich (using ch3:sock:tcp is faster in CI and not supported
+# # by spack), while trying to use any dependencies from the spack builds above.
+# #
+# #   configure with "--enable-g=dbg,log" for runtime debug output
+# #
+RUN . /spack/share/spack/setup-env.sh && \
+    export HWLOC_LOCATION="$(spack location --install /$(spack find --format {hash:7} hwloc))" && \
```
vicentebolea, line 41: Why not `spack load hwloc` before this cmd?

vicentebolea (commented):
Looks good to me, great work there :+1: 

Few questions to understand the changes

scripts/ci/images/Dockerfile.ci-spack-ubuntu20.04-base
scottwittenburg: I'll try that and see if it allows me to remove the specification of the `hwloc` path in the configure args.

scripts/ci/images/Dockerfile.ci-spack-ubuntu20.04-base
scottwittenburg: Spack is not already installed with the new image I used as a base.

scottwittenburg: The new base image is used in spack ci as one of the runner images, and is intended to have spack cloned before each run.

scottwittenburg:
I made a mistake when building the images, forgetting to switch the `FROM` statement in the images that build on the base.  This is probably why mpich tests didn't complete any faster in the last run :facepalm: 

I'm rebuilding/pushing now, and we'll see if it's any better.

scripts/ci/images/Dockerfile.ci-spack-ubuntu20.04-base
vicentebolea: I see, any reason to use the runners image rather than the minimal spack image as we used before?

scottwittenburg: Several reasons:

- the spack in the image we used before failed to concretize the stack with the external mpich
- according to the maintainer of the E4S images, the runner images are:
  - better: modern gnu make, patchelf, cmake, ccache, ninja exposed in default PATH
  - smaller: `1.91GB` uncompressed

vicentebolea: Oh that makes sense

vicentebolea:
@scottwittenburg I guess that this was superseded by #3883, shall we close?

scottwittenburg:
Yes, thanks for the reminder.

PR #3877, opened 2023-10-29
Replace LookupWriterRec's linear search on RecList with an unordered_…

pnorbert:
…0k variables, time goes from 21sec to 0.1sec in WSL.


eisenhauer (approved):
Great.  We should likely look for more of these.  Much of FFS-based serialization was a quick throw-together and less thought was given to scaling when it got converted into BP5...

Issue #3792, opened 2023-09-05
Add unit tests for zero-sized reads


vicentebolea:
We want to have a test that performs a zero-sized read of a block in variable with and without compression. We can do this based on some of our simplest read/write tests.

Taken from:

              @pnorbert @vicentebolea @anagainaru generally, shall we add a general test to ADIOS2 for zero-write blocks with and w/o compression? This is a pretty common corner case for us and can easily break :)
              
_Originally posted by @ax3l in https://github.com/ornladios/ADIOS2/issues/3459#issuecomment-1696047949_

vicentebolea:
Related: https://github.com/ornladios/ADIOS2/issues/3460

ax3l:
Related: #3542

ping @pnorbert @anagainaru https://github.com/ornladios/ADIOS2/issues/3460#issuecomment-1696052751

anagainaru:
Sorry @ax3l I somehow did not see this until now.
Just so that it's clear what exactly this corner case is, the test should write data where some blocks are empty (using and not using compression)?

vicentebolea:
@spyridon97 was originally assigned to this task as it was a good introduction for him how to write a "Simple" test. However, he is now busy with other tasks. I do not have cycles this week for adios2. @anagainaru feel free to re-assign you if you want to take it.

anagainaru:
I'll finish this tomorrow, for some reason the test I made does not fail before the fix :) so I need to double check what I'm doing wrong. If I can't finish I'll re-assign this back.

Issue #3460, opened 2023-02-02
Bug: Zero-Sized (NULL) Blocs and Operators


ax3l:
On Summit as of ADIOS 2.8.1, writes crash when trying to write zero-sized/null blocks and having c-blosc compression enabled.

Crash in `void adios2::format::BPSerializer::PutOperationPayloadInBuffer<unsigned long>(adios2::core::Variable<unsigned long> const&, adios2::core::Variable<unsigned long>::BPInfo const&)+0xd8`:

```
 3: /sw/summit/spack-envs/base/opt/linux-rhel8-ppc64le/gcc-9.3.0/adios2-2.8.1-r37wqwbpanpyyelbijmm5qj27v6xe4x5/lib64/../lib64/libadios2_core.so.2(_ZN6adios26format12BPSerializer27PutOperationPayloadInBufferImEEvRKNS_4core8VariableIT_EERKNS6_6BPInfoE+0xd8) [0x2000378725d8]
    ?? ??:0

 4: /sw/summit/spack-envs/base/opt/linux-rhel8-ppc64le/gcc-9.3.0/adios2-2.8.1-r37wqwbpanpyyelbijmm5qj27v6xe4x5/lib64/../lib64/libadios2_core.so.2(_ZN6adios24core6engine9BP4Writer13PutSyncCommonImEEvRNS0_8VariableIT_EERKNS6_6BPInfoEb+0x130) [0x2000377e3820]
    ?? ??:0

 5: /sw/summit/spack-envs/base/opt/linux-rhel8-ppc64le/gcc-9.3.0/adios2-2.8.1-r37wqwbpanpyyelbijmm5qj27v6xe4x5/lib64/../lib64/libadios2_core.so.2(_ZN6adios24core6engine9BP4Writer16PerformPutCommonImEEvRNS0_8VariableIT_EE+0xec) [0x2000377eaafc]
    ?? ??:0

 6: /sw/summit/spack-envs/base/opt/linux-rhel8-ppc64le/gcc-9.3.0/adios2-2.8.1-r37wqwbpanpyyelbijmm5qj27v6xe4x5/lib64/../lib64/libadios2_core.so.2(_ZN6adios24core6engine9BP4Writer11PerformPutsEv+0x81c) [0x2000377dae8c]
    ?? ??:0

[...]

Issue #3459, opened 2023-02-02
Bug: Zero-Sized Reads Should Never Fail


ax3l:
In ADIOS 2.8.3, requesting to read a zero-size of data from a NULL block segfaults.

```
#6  0x00002002126f1a78 in void adios2::format::BP4Deserializer::PostDataRead<double>(adios2::core::Variable<double>&, adios2::core::Variable<double>::Info&, adios2::helper::SubStreamBoxInfo const&, bool, unsigned long) () from /autofs/nccs-svm1_home1/huebl/sw/venvs/warpx/lib/python3.8/site-packages/openpmd_api/[openpmd_api_cxx.cpython-38-powerpc64le-linux-gnu.so](http://openpmd_api_cxx.cpython-38-powerpc64le-linux-gnu.so/)
#7  0x000020021252bfe8 in void adios2::core::engine::BP4Reader::ReadVariableBlocks<double>(adios2::core::Variable<double>&) ()
   from /autofs/nccs-svm1_home1/huebl/sw/venvs/warpx/lib/python3.8/site-packages/openpmd_api/[openpmd_api_cxx.cpython-38-powerpc64le-linux-gnu.so](http://openpmd_api_cxx.cpython-38-powerpc64le-linux-gnu.so/)
#8  0x000020021252624c in adios2::core::engine::BP4Reader::PerformGets() ()
   from /autofs/nccs-svm1_home1/huebl/sw/venvs/warpx/lib/python3.8/site-packages/openpmd_api/[openpmd_api_cxx.cpython-38-powerpc64le-linux-gnu.so](http://openpmd_api_cxx.cpython-38-powerpc64le-linux-gnu.so/)
#9  0x000020021235d3d4 in adios2::Engine::PerformGets() () from /autofs/nccs-svm1_home1/huebl/sw/venvs/warpx/lib/python3.8/site-packages/openpmd_api/[openpmd_api_cxx.cpython-38-powerpc64le-linux-gnu.so](http://openpmd_api_cxx.cpython-38-powerpc64le-linux-gnu.so/)
#10 0x00002002122afa98 in openPMD::detail::BufferedActions::flush(openPMD::FlushLevel, bool) ()
[...]

PR #3878, opened 2023-10-30
Testing writing null blocks without affecting the read

anagainaru:
- Without compression
- With Blosc-2
- With MGARD


pnorbert (dismissed):

spyridon97:
Assuming that my understanding is correct that the files have the null blocks, then everything looks good to me.

anagainaru:
> Assuming that my understanding is correct that the files have the null blocks, then everything looks good to me.

If you bpls -D you will see null blocks in the files created. I think this is all we needed to test, but I wanted to make sure you understood the same (and Axel did not want a specific corner case in mind)

spyridon97:
Shall this PR close all of the following issues?

https://github.com/ornladios/ADIOS2/issues/3792, https://github.com/ornladios/ADIOS2/issues/3460, https://github.com/ornladios/ADIOS2/issues/3459 or just https://github.com/ornladios/ADIOS2/issues/3792?

anagainaru:
@spyridon97 you are right, once this PR passes we neee to close #3792 and #3460 
Not sure about #3459 (I will look at this once the PR passes). I will update the description.

vicentebolea:
> @spyridon97 you are right, once this PR passes we neee to close #3792 and #3460 Not sure about #3459 (I will look at this once the PR passes). I will update the description.

#3459 too since @ax3l described this at the followup steps for his problem, the latest related error that he is experiencing might be an opepmd thing unrelated to ADIOS2.

vicentebolea:
@anagainaru CI is passing, should we have this merged?

anagainaru:
@vicentebolea I think so. I decided not to have a separate file for MGARDNullBlocks and Blosc2NullBlocks but rather not trigger the test for BP3 inside the function. This was when we give up supporting BP3 it will be an easy fix and we won't have to deal with multiple MGARD and Blosc-2 testing files. If this is ok we can merge.

pnorbert (approved):

PR #3890, opened 2023-11-01
Http connector

dmitry-ganyushin:
This is a transport  which should work with a remote HTTP server supporting partial file downloading by specifying  HTTP range requests. 
"Range: bytes=0-100" 
https://developer.mozilla.org/en-US/docs/Web/HTTP/Range_requests.
Server port is currently hardwired to 9999.
Potentially this connector should connect to a proxy server that works with HTTP with ADIOS application and transforms HTTP requests from ADIOS application to other protocols, for instance sftp.


source/adios2/toolkit/transport/file/FileHTTP.cpp
```diff
@@ -0,0 +1,343 @@
...
+
+}
+
+size_t FileHTTP::GetSize()
+{
+    char request_template[] = "GET %s HTTP/1.1\r\nHost: %s\r\nContent-Length: bytes\r\n\r\n";
+    enum CONSTEXPR { MAX_REQUEST_LEN = 1024, BUF_SIZE = 128};
+    char request[MAX_REQUEST_LEN];
+    char buffer[BUF_SIZE] ={'\0'};
+    int request_len =
```
eisenhauer, line 244: Pretty sure that this code is wrong. snprintf() won't return more than MAX_REQUEST_LEN here.  (Or ever more than the limit it was given.). Instead, it'll return negative indicating an error.   Not sure why we'd call exit() on error anyway rather than throwing an exception.

eisenhauer (commented):
I'm surprised that the socket is closed after every request.  HTTP has defaulted to persistent connections since sometime in the '90s because keeping the socket connection alive for repeated requests is such a basic thing to do.  If this is meant for communicating with a local agent, then not doing that is crazy.  I guess the question is whether or not this is destined for serious development or abandonment.

source/adios2/engine/bp4/BP4Reader.cpp
```diff
@@ -495,6 +495,13 @@ void BP4Reader::InitBuffer(const TimePoint &timeoutInstant, const Seconds &pollS
                         "the writer is creating the new files.");
             }
         }
```
eisenhauer, line 497: This looks unrelated to the other changes.

source/adios2/toolkit/transportman/TransportMan.cpp
```diff
@@ -40,6 +40,8 @@
 #include "adios2/toolkit/transport/file/FileFStream.h"
 #include "adios2/toolkit/transport/file/FileStdio.h"
 #include "adios2/toolkit/transport/null/NullTransport.h"
+//TODODG make if
```
eisenhauer, line 43: Probably can kill the comment

source/adios2/toolkit/transport/file/FileHTTP.cpp
dmitry-ganyushin: I made a small test program:
```#include <stdio.h>
int main()
{
        char buf[10];
        char template[] = "%s0123456789";
        int len = snprintf(buf, 10, template, "0000" );
        printf("%d\n", len);
        printf("%s\n", buf);
}
```
The output is 
```
14
000001234
```
It looks like sprint returns the number of characters that would have been written on the buffer, if buffer size had been sufficiently large.

source/adios2/engine/bp4/BP4Reader.cpp
dmitry-ganyushin: Removed.

source/adios2/toolkit/transportman/TransportMan.cpp
dmitry-ganyushin: Removed

source/adios2/toolkit/transport/file/FileHTTP.cpp
eisenhauer: That behavior is more standard now, but hasn't always been and still isn't in some situations (VS has two versions of snprintf(), one of which keeps the old behavior (-1 return on hitting max) and one of which is new.  But this is all kind of beside the point.  Why would you use snprintf() in C++ anyway?  Why is the request_template a non-constant member of the class **_and_** a stack variable in a method?  Why on earth are the constants declared as enums?  Why is the code to do an http request essentially duplicated in the GetSize and Read methods?  Why doesn't the read() call in the while() loop have it's buffer and length values adjusted (like the write() does in the loop above)?  (Both read() and write() might result in partial completion requiring you do it in a while loop until it completes.  The write loop looks OK, but the read loop won't work if the read() doesn't complete in one read() (additional reads would overwrite the values read earlier rather than being added behind them).

pnorbert:
Please update your branch, that hopefully fixes the script formatting errors.

pnorbert (approved):

PR #3893, opened 2023-11-02
cmake: find threads package first

vicentebolea:
This is needed since some builds fails (oneapi in Spack) since the Threads::Threads targets is used before find_package(Threads) is called.


vicentebolea:
https://github.com/spack/spack/pull/40832

anagainaru (approved):

PR #3896, opened 2023-11-06
examples,hip: use find_package(hip) once in proj

### LINKED ISSUE #3895: find_package in examples/hello
> Not to be addressed here, Note for another PR. 

For consistency across the CMake code we should not use find_package outside of detectoptions if within the project(adios2)

_Originally posted by @vicentebolea in https://github.com/ornladios/ADIOS2/pull/3880#discussion_r1383708366_

vicentebolea:
Fixes #3895


examples/hello/bpStepsWriteReadHip/CMakeLists.txt
```diff
@@ -7,19 +7,15 @@ cmake_minimum_required(VERSION 3.12)
...
-  find_package(hip QUIET)
-  if(hip_FOUND)
-    enable_language(HIP)
-  endif()
+enable_language(HIP)
 
-  find_package(ADIOS2 REQUIRED COMPONENTS ${_components})
-endif()
+# Needed for the hip cmake targets
+find_package(hip REQUIRED)
```
anagainaru, line 16: You don't need this is Kokkos is build with HIP backend, but I guess in this example we expect a HIP application that is using ADIOS2 with Kokkos so no need to link the application to Kokkos

anagainaru (approved):
I assume this is my HIP example from the former ADIOS2-Examples repo. I am confused why we are only building it if we use ADIOS2. This example

examples/hello/bpStepsWriteReadHip/CMakeLists.txt
vicentebolea: We build it in our Hip build to add coverage to the example so we that can both test that the example is uptodate and that our adios2 for kokkos api does not break.

> 

You don't need this is Kokkos is build with HIP backend, but I guess in this example we expect a HIP application that is using ADIOS2 with Kokkos so no need to link the application to Kokkos

Yep the example is only hip, no kokkos

anagainaru: Agree

PR #3902, opened 2023-11-07
ci,crusher: increase supported num branches

vicentebolea:
This is needed for the generate_pipelines.py to support processing more than 100 branches in the Gitlab CI (The max per_page items that Gitlab CI allow).

I have also change the functional programming style form map/filter to comprehensions list to make the code more pythonic.

### DIFF
--- a/.gitlab/config/generate_pipelines.py
+++ b/.gitlab/config/generate_pipelines.py
@@ -18,9 +18,43 @@ import urllib3
 urllib3.disable_warnings(urllib3.exceptions.InsecureRequestWarning)
 
 
-def request_as_dict(url):
-    r = requests.get(url + '?per_page=100', verify=False)
-    return r.json()
+class skip_after_n_successes:
+    def __init__(self, default_value, n):
+        self.runs_max = n
+        self.runs_current = 0
+        self.default_value = default_value
+
+    def __call__(self, fn, *args, **kwargs):
+        if self.runs_current >= self.runs_max:
+            return self.default_value
+
+        ret = fn(*args, **kwargs)
+        if ret:
+            self.runs_current += 1
+        return ret
+
+
+def http_get_request(*args, **kwargs):
+    kwargs['verify'] = False
+    return requests.get(*args, **kwargs)
+
+
+def request_as_list(url, *args, **kwargs):
+    current_url = url
+    body_json = []
+    while current_url:
+        response = http_get_request(current_url, *args, **kwargs)
+        body_json += response.json()
+
+        header = response.headers
+        current_url = None
+        if 'link' in header:
+            links = re.search(
+                r'(?<=\<)([\S]*)(?=>; rel="next")', header['link'], flags=re.IGNORECASE)
+            if links is not None:
+                current_url = links.group(0)
+
+    return body_json
 
 
 def add_timestamp(branch):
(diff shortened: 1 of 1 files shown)


vicentebolea:
@scottwittenburg note that we cannot test this until it gets merged since Crusher runs master's generate_pipelines.py

.gitlab/config/generate_pipelines.py
```diff
@@ -88,14 +127,13 @@ def has_no_status(branch):
...
-
-    # Select the arg.max most least recent branches
+    branches = request_as_list(gl_url + '/repository/branches')
+    branches = [add_timestamp(branch) for branch in branches]
+    branches = [b for b in branches if is_recent(b)]
     branches = sorted(branches, key=lambda x: x['dt'])
-    branches = itertools.islice(branches, args.max)
+
+    run_n_times = skip_after_n_successes(default_value=False, n=args.max)
+    branches = [b for b in branches if run_n_times(has_no_status, b)]
```
vicentebolea, line 136: This is needed to skip making the rest api after we find our requested n number of branches. Gitlab/GitHub rest api have strong limits on how many request we can do per min.

scottwittenburg, line 136: It has taken me some time to figure what the heck this is doing, so I think a descriptive comment would make sense here.  The way I'm reading this code it's intent is to "schedule jobs for the next `args.max` (i.e. 2) branches for which github doesn't have an associated status".  Is that correct though?

vicentebolea: Sorry I should have made a better introduction :angel: . Yep

.gitlab/config/generate_pipelines.py
```diff
@@ -44,7 +78,12 @@ def has_no_status(branch):
         gh_commit_sha = branch['commit']['parent_ids'][1]
 
     # Query GitHub for the status of this commit
-    commit = request_as_dict(gh_url + '/commits/' + gh_commit_sha + '/status')
+    response = http_get_request(
+        gh_url + '/commits/' + gh_commit_sha + '/status')
+    if int(response.headers['x-ratelimit-remaining']) <= 0:
+        raise ConnectionError(response.json())
```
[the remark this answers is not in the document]
vicentebolea: Yes, I want to show: hey this error is due to us hitting the github ratelimit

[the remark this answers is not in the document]
vicentebolea: And perhaps lets the future developer to program this by catching this exception

.gitlab/config/generate_pipelines.py
scottwittenburg: Well, I was suggesting putting a comment in the code so future developers (including me a month from now) get a hint about what's happening.  But I guess that's not strictly necessary.

scottwittenburg (dismissed):

.gitlab/config/generate_pipelines.py
vicentebolea: My bad, find the comment added

.gitlab/config/generate_pipelines.py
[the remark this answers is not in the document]
scottwittenburg: Ok, just curious.

scottwittenburg (approved):

PR #3905, opened 2023-11-08
Change install directory of Gray scott files

spyridon97:
Change install directory of Gray scott files

### DIFF
--- a/examples/simulations/gray-scott-struct/CMakeLists.txt
+++ b/examples/simulations/gray-scott-struct/CMakeLists.txt
@@ -43,11 +43,11 @@ if(ADIOS2_HAVE_MPI)
     "simulation/settings-inline.json"
     "plot/decomp.py" "plot/gsplot.py" "plot/pdfplot.py"
     "ReadMe.md"
-    DESTINATION ${CMAKE_INSTALL_PREFIX}/share/adios2/examples/simulations/gray-scott-struct)
+    DESTINATION ${CMAKE_INSTALL_PREFIX}/adios2/gray-scott-struct)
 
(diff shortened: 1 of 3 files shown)


anagainaru (commented):
Good, could we also export the adios2.xml and settings json files in the same folder as executables?

spyridon97:
> Good, could we also export the adios2.xml and settings json files in the same folder as executables?

@pnorbert suggested this change and did not like the bin directory idea. I am open to anything really, your call.

anagainaru:
> @pnorbert suggested this change and did not like the bin directory idea. I am open to anything really, your call.

It's Norbert's call so I am ok with this. I am ok with just adding in the documentation a mention about where they can find the needed files.

spyridon97:
@pnorbert can I get an approval?

vicentebolea:
Note that normally we install examples and its data/conf files in the `share/packagename` this is what the FHS expects packages to do. If we go ahead with this change we will have to patch our adios2 package across our different platforms to use `share/packagename` since in Unix-like systems we cannot create the directory `/usr/adios`.

spyridon97:
> Note that normally we install examples and its data/conf files in the `share/packagename` this is what the FHS expects packages to do. If we go ahead with this change we will have to patch our adios2 package across our different platforms to use `share/packagename` since in Unix-like systems we cannot create the directory `/usr/adios`.

@pnorbert are you fine with that?

vicentebolea:
I am for reverting this, this change does not make sense. The example binaries are installed at `prefix`/bin and the data/conf files are installed at `prefix`/adios2. The `prefix`/adios2. This makes adios2 example impossible to be installed in any unix-like system.

Issue #3897, opened 2023-11-06
BPChangingShapeWithinStep.Multiblock.BP3 and BP4 tests are flaky


vicentebolea:
It fails in Many builds as shown here:

https://open.cdash.org/queryTests.php?project=ADIOS&begin=2023-10-26&end=2023-11-06&filtercount=1&showfilters=1&field1=testname&compare1=63&value1=Engine.BP.*%2FBPChangingShapeWithinStep.MultiBlock%2F*.BP3.MPI

![Screenshot from 2023-11-06 18-02-46](https://github.com/ornladios/ADIOS2/assets/939798/be1ea8ef-8ce0-4d20-8787-22e7cacc7e30)

vicentebolea:
@eisenhauer do you know what can be causing this issue?

eisenhauer:
In a word, no, I have no idea.  This test is kinda new.  @pnorbert created it in April.  That it's only failing BP3/4 is odd because those engines aren't really undergoing development currently.  The tests are timing out, which is odd behavior for such a simple test, BUT they run differently on different platforms depending upon whether BLOSC, MGARD and ZFP are available.  I.E. it'll run the same simple subtest with "none" and then with each compressor that is available.  Maybe there's a new problem with one of those compressors?

In a separate question, we no longer seem to be building testing by default.  I.E. BUILD_TESTING, which used to default to on, now defaults to off.  Was that deliberate and maybe I just missed the discussion?

vicentebolea:
> In a separate question, we no longer seem to be building testing by default. I.E. BUILD_TESTING, which used to default to on, now defaults to off. Was that deliberate and maybe I just missed the discussion?

We talked this in one of the adios2 weekly meetings a three weeks ago. The change was added in https://github.com/ornladios/ADIOS2/pull/3820. The rationale was that most of the users do not need/want to run the tests.

vicentebolea:
> Maybe there's a new problem with one of those compressors

Maybe, another hint is that we do not see the problem with BP5

eisenhauer:
Looking at the no-test-repeats CDash output, I'd suspect that MGARD is the problem...
https://open.cdash.org/viewTest.php?onlyfailed&buildid=9130458

PR #3908, opened 2023-11-08
ci: Set openmpi and openmp parameters to make an efficient use of the CI

vicentebolea:
fixes: #3897

### DIFF
--- a/scripts/ci/cmake/ci-el8-icc-ompi.cmake
+++ b/scripts/ci/cmake/ci-el8-icc-ompi.cmake
@@ -1,7 +1,3 @@
-include(ProcessorCount)
-ProcessorCount(NCPUS)
-math(EXPR N2CPUS "${NCPUS}*2")
-
 set(ENV{CC}  icc)
 set(ENV{CXX} icpc)
 set(ENV{FC}  ifort)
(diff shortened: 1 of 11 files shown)


anagainaru:
Is this going to make the testing (and thus the CI) slower? Do we really want to keep the BP3 testing? Could we maybe separate it to only test BP3 in the nightly builds?

eisenhauer:
I'm curious if this really fixes it and why.  This isn't big data.  If one of the compressors is taking more than 2 minutes to do something with it, that seems like a non-viable compressor.

vicentebolea:
I found that this problem occurs on my laptop when connected to a ethernet network but not when connected to the wifi network. I could figure out that this is due to openmpi picking certain network interfaces, a easy workaround is to use the lo interface

eisenhauer:
> I found that this problem occurs on my laptop when connected to a ethernet network but not when connected to the wifi network. I could figure out that this is due to openmpi picking certain network interfaces, a easy workaround is to use the lo interface

Well, maybe...  But why only this test?  MGard doesn't use MPI, but it does use OpenMP.  I build MGard manually on my OSX laptop and these tests have no problems.  I'm wondering if on whatever platform this is running OpenMP isn't playing well with MPI, resulting in the issues that you're seeing.  (I could use the docker image and dig further, but I'm off at a conference, so little spare time for the moment.)

vicentebolea:
> Well, maybe... But why only this test? MGard doesn't use MPI, but it does use OpenMP. I build MGard manually on my OSX laptop and these tests have no problems. I'm wondering if on whatever platform this is running OpenMP isn't playing well with MPI, resulting in the issues that you're seeing. (I could use the docker image and dig further, but I'm off at a conference, so little spare time for the moment.)

hmm, so two things happened here:

- The real culcript was the that we were using n*4) number of mpi tasks and each of the task was using n*2 number of threads (n=2) this was just too much for the weak VPS that github gives us.
- The interface thing was affecting time to time different mpi test in random fashion

vicentebolea:
@scottwittenburg if this makes sense to you please approve so that it gets merged and we get a green dashboard :sunglasses:

vicentebolea:
Yay, it's all green again :)

eisenhauer (approved):
Looks good.

scottwittenburg:
I've rebased on this change, and now my ci reproducer builds are happening in serial, which seems like an unintended consequence.

Issue #3903, opened 2023-11-08
2.9.2 w/ C-Blosc2 v2.11.1 Static


ax3l:
**Describe the bug**
I am building an all-static build of ADIOS2 v2.9.2 with C-Blosc2 v2.11.1 for Wheels in https://github.com/openPMD/openPMD-api/pull/1554

While installing ADIOS2 works, using it downstream fails with
```
2023-11-07T03:44:31.2185681Z   -- Configuring done (8.1s)
2023-11-07T03:44:31.3333449Z   CMake Error at C:/Program Files (x86)/ADIOS2/lib/cmake/adios2/adios2-targets.cmake:86 (set_target_properties):
2023-11-07T03:44:31.3337501Z     The link interface of target "adios2::adios2_blosc2" contains:
2023-11-07T03:44:31.3340764Z 
2023-11-07T03:44:31.3344814Z       Blosc2::blosc2_static
2023-11-07T03:44:31.3348057Z 
2023-11-07T03:44:31.3352143Z     but the target was not found.  Possible reasons include:
2023-11-07T03:44:31.3355109Z 
2023-11-07T03:44:31.3358878Z       * There is a typo in the target name.
2023-11-07T03:44:31.3362359Z       * A find_package call is missing for an IMPORTED target.
2023-11-07T03:44:31.3365646Z       * An ALIAS target is missing.
2023-11-07T03:44:31.3368954Z 
2023-11-07T03:44:31.3372324Z   Call Stack (most recent call first):
2023-11-07T03:44:31.3376402Z     C:/Program Files (x86)/ADIOS2/lib/cmake/adios2/adios2-config-common.cmake:197 (include)
2023-11-07T03:44:31.3379473Z     C:/Program Files (x86)/ADIOS2/lib/cmake/adios2/adios2-config.cmake:29 (include)
2023-11-07T03:44:31.3382839Z     CMakeLists.txt:443 (find_package)
```

**To Reproduce**
See scripts in https://github.com/openPMD/openPMD-api/pull/1554

**Expected behavior**
Did build with 2.9.1 and a slightly older C-Blosc2, per linked PR.

**Desktop (please complete the following information):**
 - OS/Platform: Linux, macOS and Windows are all affected
 - Build type: static

**Additional context**
https://github.com/openPMD/openPMD-api/pull/1554

ax3l:
If I remove the installed `FindBlosc2.cmake` file then it finds it... #3714

ax3l:
[...]

PR #3910, opened 2023-11-10
Remove FindBlosc2

ax3l:
Going forward, it might be easier if we dump the dependency to 2.10.2+
https://github.com/Blosc/c-blosc2/releases/tag/v2.10.2

Related to #3903


vicentebolea:
I rebased on the top of master, there was some unrelated CI failures in master for a two weeks.

ax3l:
Thanks @vicentebolea 

> I think that the issues that you found are due to the fact that blosc2 could not be installed due to depedencies missing and using as a backup this FindBlosc file.

I cannot see a c-blosc2 install issue here:
https://github.com/openPMD/openPMD-api/pull/1554

pnorbert:
It doesn't hurt to increase the version requirement for blosc2.

pnorbert (approved):

vicentebolea:
Note that we might not have an issue with our current FindBlosc2.cmake file. The error that @ax3l have been seeing it seems to be unrelated to this file. 

The error that was reported at #3903 seems to be related to the blosc2config.cmake not to be found. 

We might not want to arbitrarily drop dependency to older c-blosc2 since dependending on its last release might hinder users that do not have access to older c-blocs2 versions.

vicentebolea:
This was already done at #4093. Closing

PR #3918, opened 2023-11-14
SST with GPU support

### LINKED ISSUE #3706: SST with GPU support
The file engines are capable of receiving GPU buffers in their Put/Get calls. SST/SSC should also be able to do this. 

Even though SST is using the same Serializer, there is something missing that is making it seg fault when receiving CUDA buffers. This needs to be fixed if there is a bug or implemented if there is something missing.

anagainaru:
SST was using MallocV and not ChunkV for copying the memory to internal buffers (for small block not deferred) so it was seg faulting for simple examples.

There are two contributions brought by this PR:
- Bug fix so that SST can use Cuda and Kokkos pointers
- Add an example using directly Kokkos Views with SST

Closes #3706


eisenhauer (approved):
Looks like a straightforward MallocV tweak...

anagainaru:
If someone can merge this bypassing the checks please leave two commits, I would like the example to be separate in case we want to reorganize adios later.

PR #3928, opened 2023-11-18
Dataman with GPU support

anagainaru:
Propagating the logic for GPU support into the DataMan engine + example to test that everything works.

**Note**
Both SST and DataMan work when receiving Kokkos Views because we update the variable memory space based on the type of the View. However, if raw device pointers are sent (regardless of the Cuda or Kokkos backend in ADIOS) this information needs to be detected so each engine needs to trigger the detection of the memory space of each variable. This is done in different parts of each engine and it is a source of possible nasty errors (if the memory space remains `Detect` deep in the engine logic -- which I managed to trigger once). I plan to fix this in the near future by moving the detection before we reach each particular engine. I'll open an issue for this, but this PR can be merged as is.


anagainaru:
If someone else merges this, please leave two commits so I can cherry pick just one in the future

eisenhauer (approved):

Issue #3894, opened 2023-11-06
Move Documentation for adios2-examples.readthedocs.org to adios2.readthedocs.com


spyridon97:
Move Documentation for adios2-examples.readthedocs.org to adios2.readthedocs.com

anagainaru:
I am not sure where to give you this feedback, but the Kokkos gray-scott example is working. For the documentation make a note that, in order to run this example, you need the adios2.xml file from `examples/simulations/gray-scott` and one of the settings json files from `examples/simulations/gray-scott/simulation`

PR #3939, opened 2023-11-23
Move useful docs from ADIOS2-examples to ADIOS2

spyridon97:
Move useful docs from ADIOS2-examples to ADIOS2

Closes https://github.com/ornladios/ADIOS2/issues/3894.

### DIFF
--- a/docs/user_guide/source/components/anatomy.rst
+++ b/docs/user_guide/source/components/anatomy.rst
@@ -1,0 +1,136 @@
+.. _sec:basics_interface_components_anatomy:
+
+***************************
+Anatomy of an ADIOS Program
+***************************
+
+Anatomy of an ADIOS Output
+--------------------------
+
+.. code:: C++
+
+    ADIOS adios("config.xml", MPI_COMM_WORLD);
+    |
+    |   IO io = adios.DeclareIO(...);
+    |   |
+    |   |   Variable<...> var = io.DefineVariable<...>(...)
+    |   |   Attribute<...> attr = io.DefineAttribute<...>(...)
+    |   |   Engine e = io.Open("OutputFileName.bp", adios2::Mode::Write);
+    |   |   |
+    |   |   |   e.BeginStep()
+    |   |   |   |
+    |   |   |   |   e.Put(var, datapointer);
+    |   |   |   |
+    |   |   |   e.EndStep()
+    |   |   |
+    |   |   e.Close();
+    |   |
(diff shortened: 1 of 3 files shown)


docs/user_guide/source/components/anatomy.rst
```diff
@@ -0,0 +1,132 @@
...
+have to retrieve the available variables in an input first as definitions (Variable and Attribute objects).
+
+If we know the particular variable (name and type) in the input stream, we can get the definition using
+InquireVariable(). Generic tools that process any input must use other functions to retrieve the list of variable names
+and their types first and then get the individual Variable objects. The same is true for Attributes.
+
+Anatomy of an ADIOS File-only Input
+-----------------------------------
+
+This only works for files on permanent storage for post-mortem processing. Do not use BeginStep/EndStep, so the
```
anagainaru, line 111: This is not true anymore ever since BP5 became the default file engine, you need to open the file in RandomAccess mode to set the step selection.

docs/user_guide/source/advice/advice.rst
```diff
@@ -87,4 +87,8 @@ The goal is to provide specific advice and good practices about the use of ADIOS
 25. Developers: explore the testing infrastructure ``ADIOS2/testing`` in ADIOS 2 as a starting point for using ADIOS 2 in your own testing environment. 
 
 26. Become a super-user of :ref:`bpls : Inspecting Data` to analyze datasets generated by ADIOS 2.
+
+27. When you need to write intermediate results, such as checkpoints, considering streaming your data instead of writing
```
anagainaru, line 91: Not sure I agree with this for checkpointing. Where do you stream the data and how would you recover a failure if the data is not on the permanent storage?

anagainaru (changes requested):
Please update the section on random access to correspond to the BP5 behavior. The text in the examples was old.

docs/user_guide/source/components/anatomy.rst
eisenhauer: Yeah, I would introduce this section a bit differently.  Start with the two modes of ADIOS input, Read and ReadRandomAccess.  ReadRandomAccess can only be used with file engines and involves loading all the file metadata at once.  So it can be more memory intensive than Read mode, but allows reading data from any timestep using SetStepSelection().  Read mode allows data access only timestep by timestep, but generally it is more memory efficient as ADIOS is only required to load metadata for the current timestep.

PR #3946, opened 2023-11-28
Set the compiler for the Kokkos DataMan example

anagainaru:
Set it to what is used to build Kokkos.

For some reason I was able to compile this code without the fix on Summit, but it was falling on Frontier.


spyridon97 (approved):

anagainaru:
The CI falling has nothing to do with this PR and if I rebase  I will loose the approval so I will merge like this and I take full responsibility for the merge.

Issue #3942, opened 2023-11-27
Operators do not work with HDF5


rupertnash:
**Describe the bug**
Plugin operators do not run when the engine is set to HDF5.

**To Reproduce**
1. Compile ADIOS 2 (I'm using release 2.9.2) with Sodium
2. Modify the Sodium plugin example XML to use HDF5 output
3. Run the write sodium plugin example
4. Observe that HDF5 output is not encrypted
```
rwn-e609@ln01:~/adios2-build/debug-sodium> cp ~/ADIOS2/examples/plugins/operator/example_operator.xml operator_hdf5.xml
rwn-e609@ln01:~/adios2-build/debug-sodium> emacs operator_hdf5.xml 
rwn-e609@ln01:~/adios2-build/debug-sodium> cat operator_hdf5.xml
<?xml version="1.0"?>
<!-- Config XML file for the
        hello_skeleton*
     executables in build/bin -->

<adios-config>

    <!--====================================
           Configuration for the Writer
        ====================================-->

    <io name="writer">
        <variable name="data">
            <operation type="plugin">
                <parameter key="PluginName" value="OperatorPlugin"/>
                <parameter key="PluginLibrary" value="EncryptionOperator" />
                <parameter key="SecretKeyFile" value="test-key" />
            </operation>
        </variable>
        <engine type="HDF5">
        </engine>
    </io>


    <!--====================================
           Configuration for the Reader
        ====================================-->

    <io name="reader">
        <variable name="data">
            <operation type="plugin">
                <parameter key="PluginName" value="OperatorPlugin"/>
                <parameter key="PluginLibrary" value="EncryptionOperator" />
                <parameter key="SecretKeyFile" value="test-key" />
            </operation>
        </variable>
        <engine type="HDF5">
        </engine>
    </io>
</adios-config>
rwn-e609@ln01:~/adios2-build/debug-sodium> bin/exampleOperatorPlugin_write operator_hdf5.xml
rwn-e609@ln01:~/adios2-build/debug-sodium> file testOperator.bp 
testOperator.bp: Hierarchical Data Format (version 5) data
rwn-e609@ln01:~/adios2-build/debug-sodium> h5dump testOperator.bp 
HDF5 "testOperator.bp" {
GROUP "/" {
   ATTRIBUTE "NumSteps" {
      DATATYPE  H5T_STD_U32LE
      DATASPACE  SCALAR
      DATA {
      (0): 1
      }
   }
   GROUP "Step0" {
[...]

PR #3951, opened 2023-12-01
Have HDF5 write raise error if operator(s) requested

rupertnash:
Discussion in #3942 suggests that support for operators in "secondary" (for want of a better term) engines is not planned. My suggestion that ADIOS should let the user know this won't happen is added in this PR.

### DIFF
--- a/source/adios2/toolkit/interop/hdf5/HDF5Common.cpp
+++ b/source/adios2/toolkit/interop/hdf5/HDF5Common.cpp
@@ -1709,6 +1709,16 @@ void HDF5Common::StaticGetAdiosStepString(std::string &stepName, size_t ts)
     stepName = "/Step" + std::to_string(ts);
 }
 
+void HDF5Common::CheckVariableOperations(const core::VariableBase &variable) const
+{
+    if (!variable.m_Operations.empty())
+    {
+        helper::Throw<std::runtime_error>("Toolkit", "interop::hdf5::HDF5Common",
+                                          "CheckVariableOperations",
+                                          "ADIOS2 Operators are not supported for HDF5 engine");
+    }
+}
+
 #define declare_template_instantiation(T)                                                          \
     template void HDF5Common::Write(core::Variable<T> &, const T *);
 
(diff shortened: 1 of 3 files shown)


pnorbert:
Thanks. I have to think about this. I am not sure if every user would want the code abort with an exception rather than quietly ignore the operator. On the other hand, as far as I know, you are the one and only user of the HDF5 engine.

rupertnash:
Hopefully made clang format happy now!

pbartholomew08:
Could there be an option to make warnings into errors (similar to `-Werror`) in that case? Then if users absolutely don't want it silently ignored it will stop?

eisenhauer:
> > Could there be an option to make warnings into errors (similar to `-Werror`) in that case? Then if users absolutely don't want it silently ignored it will stop?
> 
> Yes, I think this would be a good idea. I don't think we have the option now but we should look into this in the future. I am strongly in favor of giving at least some warning if you set a parameter that will be ignored.

Taking this literally, we have long thought it might be useful to warn if you set an engine parameter that is never consumed (because you might have just misspelled it).  But our disaggregated method of looking at parameters makes this hard.  This case of ignoring an operator is at least easier to detect.  But norbert's got a point too...

source/adios2/toolkit/interop/hdf5/HDF5Common.cpp
```diff
@@ -1709,6 +1709,16 @@ void HDF5Common::StaticGetAdiosStepString(std::string &stepName, size_t ts)
     stepName = "/Step" + std::to_string(ts);
 }
 
+void HDF5Common::CheckVariableOperations(const core::VariableBase &variable) const
+{
+    if (!variable.m_Operations.empty())
+    {
+        helper::Throw<std::runtime_error>("Toolkit", "interop::hdf5::HDF5Common",
+                                          "CheckVariableOperations",
+                                          "Operators not supported for HDF5 engine");
```
guj, line 1718: Thanks for adding this warning. 
I think the error message would  be clear if it says something like "ADIOS Operators are not supported for HDF5 engine".  What would you think?

guj (changes requested):

source/adios2/toolkit/interop/hdf5/HDF5Common.cpp
rupertnash: Happy to change the wording; see forced push

PR #3957, opened 2023-12-05
ci: fix false negative asan checks

scottwittenburg:
The asan tests have all been failing like [this](https://open.cdash.org/viewDynamicAnalysisFile.php?id=11065354) for some unknown amount of time, and it seems that kind of failure didn't trigger ctest to exit non-zero, which caused the check to look like it was passing.

To start with, let's see if we can get the test to actually run all those tests and see if there are any real failures.  If there are, we will have to fix them here, before merging this.


scripts/ci/cmake/ci-fedora-asan.cmake
```diff
@@ -4,7 +4,7 @@ set(ENV{CC}  clang)
 set(ENV{CXX} clang++)
 set(ASAN_FLAGS "-fsanitize=address -fno-omit-frame-pointer -pthread -mllvm -asan-use-private-alias=1 -Wno-unused-command-line-argument")
 set(ENV{ASAN_OPTIONS} "use_odr_indicator=1")
-set(ENV{LSAN_OPTIONS} "suppressions=${CMAKE_SOURCE_DIR}/thirdparty/perfstubs/perfstubs.supp")
+set(ENV{LSAN_OPTIONS} "suppressions=${CMAKE_CURRENT_LIST_DIR}/../..//thirdparty/perfstubs/perfstubs.supp")
```
vicentebolea, line 7: try this `CI_SOURCE_DIR` var instead. This is what we use in the CI

scottwittenburg:
Yikes, a lot of errors detected by asan now, but at least they're showing up properly in [cdash](https://open.cdash.org/viewDynamicAnalysisFile.php?id=11078460).

scottwittenburg:
> I can take point on resolving the actual errors for the moment.

@eisenhauer What do you think of using this ctest [feature](https://cmake.org/cmake/help/v3.18/variable/CTEST_CUSTOM_MEMCHECK_IGNORE.html) to ignore the current failures, so we can merge this PR asap?  That way we can prevent new issues from creeping in while we work on fixing the existing failures (in a new PR that adds back the current failing tests).

vicentebolea:
> But a counter-argument might be that if we fix san bugs in separate PRs, that makes it easier to backport those fixes into release_29. I guess I don't have strong feelings either way...

Thats indeed important, we cannot backport API/ABI incompatible changes such as the change of the function signature. I suggest to first add the ignore directive and later we fix this issues one by one, this will make the backporting process much easier.

eisenhauer:
> Thats indeed important, we cannot backport API/ABI incompatible changes such as the change of the function signature. I suggest to first add the ignore directive and later we fix this issues one by one, this will make the backporting process much easier.
Lets follow that approach.  I'm also concerned that the overall MinMax fix may be more complex than my changes above, so I'd propose that you guys roll back my commits, add the CTEST_CUSTOM_MEMCHECK_IGNORE spec and merge this PR.

scottwittenburg:
Hmmm, all checks passed, approval from @vicentebolea, what requirements have yet to be met?

scottwittenburg:
> what requirements have yet to be met?

I guess we require branch to be up to date before merging.

vicentebolea:
> Hmmm, all checks passed, approval from @vicentebolea, what requirements have yet to be met?

The option requires PR to be uptodate was enabled. We had some issues in some earlier pull requests that the CI was passing with this automatic merge commit but once it was merged it broke some tests. Maybe the reason was that a new commit got to master right before that was merge and that change was not tested.

vicentebolea:
I disable it again since our CI is in a good state again. I enabled it because the CI was not in a good state and I was trying to avoid further issues until the current issues were resolved.

PR #3959, opened 2023-12-07
Several fixes for ASAN issues

eisenhauer:
Final Description:  This is a compilation of a variety of ASAN fixes.  They include:

- Fixing a couple of leaks in C bindings blockinfo handling
- Cleaning up Min/Max handling so that we don't access memory in situations where there is no min/max valid for that type.  (This may need further cleanup.  In the past we've quietly returned invalid min/max when you explicitly ask for min/max and none was calculated or none was possible.  We might want to throw an exception in that case.)
- Clean up BP5 attribute marshalling, both in append mode and when handling arrays of strings.  The logic of when to use old-style attribute marshalling and when to do new-style was faulty, resulting in a mix of output.  This wasn't a problem for the reader which transparently handles either, but it caused memory leaks on the writer.
- clean up leaks of minBlocksInfo in bpls.
- fix test testing/adios2/engine/bp/TestBPWriteReadBlockInfo.cpp which didn't properly free the blocksinfo, IO or ADIOS before exit.

This PR significantly reduces the set of ASAN errors that accumulated when that sanitizer was disabled, but it doesn't completely eliminate them.  There remain at least issues with SST buffer handling and with the remote data access facilities.  However if we can get this merged, those can be tackled with the next PR.

ALSO NOTE:  This PR creates a single overall ADIOS suppression file rather than simply using the one in perfstubs (becaause we need another suppression).


eisenhauer:
Hoping that this reduces the sanitizer ignore list...

anagainaru (approved):
We do create the MinMax function for other types (mainly for all types in `ADIOS2_FOREACH_PRIMITIVE_STDTYPE_1ARG`). In order to be consistent to the CPU version, the GPU one does the same: 

```
#define declare_type(T)                                                                            \
    template void GetGPUMinMax(const T *values, const size_t size, T &min, T &max);
ADIOS2_FOREACH_PRIMITIVE_STDTYPE_1ARG(declare_type)
#undef declare_type
```

and then down in the internal implementation I template over Complex numbers specifically and use an empty function. Do we need to update these to correspond to the types that have a chance of reaching the function?

**Later edit:** Also for the CPU backend, I think we have a minmax function for Complex numbers. 

```
template <>
inline void GetMinMax(const std::complex<float> *values, const size_t size, std::complex<float> &min, std::complex<float> &max);
```

eisenhauer:
The "empty function" approach matches what was essentially happening at the CPU level, but that had some unfortunate side effects.  These were patched around to make the samitizer happy in earlier work (init the min/max with all zeroes, then never change it for things that don't have a min/max), but that didn't happen in BP5 because the sanitizers weren't running during development.  I could patch BP5 like before, but instead of handing you back some random thing if you ask for a min or max of something that doesn't have one, I'd rather throw an exception, which probably means also patching things like bpls which may depend upon the current behavior.   BUT, while MIN/MAX handling was the cause of issues when I enabled asan with compiler args on my laptop, there are things happening in CI that I dont seem to be able to duplicate that way.  So I'm going to have to try to get the CI docker image working somewhere before I can push this much further.  Anyway, my hope is that you wouldn't have to change anything on the GPU, but that you'd simply never get a GetGPUMinMax called in a situation where you'd have an empty function.  But first, gotta work on the paper.

scottwittenburg:
@eisenhauer  Here is how I was reproducing the sanitizer checks last week:

<details>
  <summary>Bash script with comments</summary>

```bash
#!/bin/bash

# cd /data/scott/Documents/adios2
# mkdir san_repro
# cd san_repro
# git clone /data/scott/projects/adios2/adios2Src source
# docker run --rm -v /data/scott/Documents/adios2/san_repro:/root/projects -ti ornladios/adios2:ci-fedora-asan

export CIRCLE_BRANCH="scott-local-testing"
export CIRCLE_PR_NUMBER="1"
export CIRCLE_JOB="fedora-asan"
export CIRCLE_SHA1=""
export CIRCLE_WORKING_DIRECTORY="~/projects"

cd ~/projects

bash source/scripts/ci/circle/postCDashStatus.sh
bash source/scripts/ci/circle/run.sh update
bash source/scripts/ci/circle/run.sh configure
bash source/scripts/ci/circle/run.sh build
bash source/scripts/ci/circle/run.sh memcheck
```

</details>

eisenhauer:
> @eisenhauer Here is how I was reproducing the sanitizer checks last week:

Thanks much!  I have discovered that while I've been amazingly successful in running x86 docker images on my arm MacBook (presumably thanks to dynamic instruction translation), running asan is apparently a bridge too far...  Falling back to my older macbook.

eisenhauer:
@scottwittenburg Some of these leaks seem unrelated to ADIOS (maybe due to MPI?).  For example ibv_get_device_list in the reports below.  There is no call to ibv_get_device_list in ADIOS (it's a call to an RDMA library, so some kind of MPI init is possible?).

=================================================================
==55473==ERROR: LeakSanitizer: detected memory leaks

Direct leak Direct leak of 8 byte(s) in 1 object(s) allocated from:
    #0 0x533967 in calloc (/root/project/fedora-asan/bin/TestCommonWriteAttrs+0x533967)
    #1 0x7f386d18a11f in ibv_get_device_list (/lib64/libibverbs.so.1+0x1011f)

-----------------------------------------------------
Suppressions used:
  count      bytes template
     20       4480 ps_make_timer_name_
-----------------------------------------------------

SUMMARY: AddressSanitizer: 8 byte(s) leaked in 1 allocation(s).

=================================================================
==55474==ERROR: LeakSanitizer: detected memory leaks

Direct leak Direct leak of 8 byte(s) in 1 object(s) allocated from:
    #0 0x539ca7 in calloc (/root/project/fedora-asan/bin/TestCommonReadAttrs+0x539ca7)
    #1 0x7f321166311f in ibv_get_device_list (/lib64/libibverbs.so.1+0x1011f)

-----------------------------------------------------
Suppressions used:
  count      bytes template
     18       3815 ps_make_timer_name_
-----------------------------------------------------

SUMMARY: AddressSanitizer: 8 byte(s) leaked in 1 allocation(s).

caitlinross:
@eisenhauer You can create a suppression file with a list of leaks that should be ignored, so we won't continue to get errors for it in our CI, since it's not something we can fix.

caitlinross:
Looks like we already have a suppression file for something in perfstubs that was causing leaks, see `thirdparty/perfstubs/perfstubs.supp`.  Then we'd need to update the `scripts/ci/cmake/ci-fedora-asan.cmake` to point to that suppression file. I'm unsure if we can pass in paths to multiple files or if we'd need to just have one suppression file that has everything in it.

eisenhauer:
Unfortunately it looks like we can't have multiple files, so we'll have to have our own rather than just using the one for Perfstubs...

eisenhauer:
@vicentebolea @anagainaru or @pnorbert.  If any of you are working, this PR should be ready to go.  (˙˙˙ɐılɐɹʇsnɐ uı uǝʌǝ 'ƃuıʞɹoʍ llıʇs ɯ,ı 'sǝʎ)

vicentebolea (approved):

eisenhauer:
Thanks @vicentebolea!

PR #3960, opened 2023-12-08
Add pip packaging

scottwittenburg:
Add python packaging so users can pip install adios2. This PR follows general guidance [here](https://learn.scientific-python.org/development/guides/) regarding choice of tooling and practices.

Supersedes #3949


vicentebolea:
This is awesome! ill do a review this Monday

scottwittenburg:
Finally [success](https://github.com/ornladios/ADIOS2/actions/runs/7146041582/job/19463471646) uploading a package to test.pypi.org!

After moving the PR to a branch on the main repo, I still had repeated failures to push.  Some discussion around [here](https://github.com/scikit-build/moderncmakedomain/issues/29#issuecomment-1564478955) suggested removing and re-adding the trusted publisher on the testpypi side, and when I did that, it worked! :tada:

.github/workflows/pypackaging.yml
```diff
@@ -0,0 +1,90 @@
+name: Python Packaging
+
+on:
+  workflow_dispatch:
+  pull_request:
+  push:
+    branches:
+      - master
```
vicentebolea, line 8: We want release_XXYY branches here too

CMakeLists.txt
```diff
@@ -18,6 +18,39 @@ setup_version(2.9.0)
...
+    list(APPEND CMAKE_INSTALL_RPATH
+      "@loader_path")
+  elseif (UNIX)
+    # ELF loader settings.
+    set(CMAKE_BUILD_WITH_INSTALL_RPATH ON)
+    list(APPEND CMAKE_INSTALL_RPATH
+      "$ORIGIN")
+  endif ()
+endif ()
+
```
vicentebolea, line 53: Here we can keep only the logic affecting ubuntu/Linux, for this info we can add a link in a commit

bindings/Python/CMakeLists.txt
```diff
@@ -47,18 +47,34 @@ set_target_properties(adios2_py PROPERTIES
...
-string(REGEX REPLACE "[^/]+" ".." relative_base "${CMAKE_INSTALL_PYTHONDIR}/adios2")
 if(CMAKE_SYSTEM_NAME MATCHES "Linux")
-  set_target_properties(adios2_py PROPERTIES
-    INSTALL_RPATH "$ORIGIN/${relative_base}/${CMAKE_INSTALL_LIBDIR}"
-  )
+  if (ADIOS2_PIP_INSTALL)
+    set_target_properties(adios2_py PROPERTIES
+      INSTALL_RPATH "${CMAKE_INSTALL_RPATH}"
+    )
+  else()
```
vicentebolea, line 55: We might not need this

bindings/Python/CMakeLists.txt
```diff
@@ -47,18 +47,34 @@ set_target_properties(adios2_py PROPERTIES
...
-)
-install(FILES ${CMAKE_PYTHON_OUTPUT_DIRECTORY}/adios2/__init__.py
-  DESTINATION ${CMAKE_INSTALL_PYTHONDIR}/adios2
-  COMPONENT adios2_python-python
+if (ADIOS2_PIP_INSTALL)
+  install(TARGETS adios2_py
+    DESTINATION ${CMAKE_INSTALL_LIBDIR}
+    COMPONENT adios2_python-python)
+  install(FILES ${CMAKE_PYTHON_OUTPUT_DIRECTORY}/adios2/__init__.py
+    DESTINATION ${CMAKE_INSTALL_LIBDIR}
```
vicentebolea, line 68: Would it make sense to set CMAKE_INSTALL_PYTHONDIR <- CMAKE_INSTALL_LIBDIR

cmake/DetectOptions.cmake
```diff
@@ -395,7 +395,9 @@ if(NOT SHARED_LIBS_SUPPORTED)
   endif()
 endif()
 
-if(ADIOS2_USE_Python STREQUAL AUTO)
+if(ADIOS2_PIP_INSTALL)
```
vicentebolea, line 398: this option could be called ADIOS2_USE_PIP to follow the same standard as other adios2 options

vicentebolea: I Would also added it as an adios2 standard option

pyproject.toml
```diff
@@ -0,0 +1,60 @@
...
+[project]
+name = "adios2"
+dynamic = ["version"]
+authors = [
+  { name="Chuck Atkins", email="chuck.atkins@kitware.com" },
+  { name="Greg S. Eisenhauer", email="eisen@cc.gatech.edu" },
+  { name="William F. Godoy", email="godoywf@ornl.gov" },
+  { name="Junmin Gu", email="jgu@lbl.gov" },
+  { name="Norbert Podhorszki", email="pnorbert@ornl.gov" },
+  { name="Ruonan (Jason) Wang", email="wangr1@ornl.gov" },
```
vicentebolea, line 14: This has to be updated

pyproject.toml
```diff
@@ -0,0 +1,60 @@
...
+ADIOS2_USE_HDF5_VOL = "OFF"
+ADIOS2_USE_BZip2 = "OFF"
+ADIOS2_USE_Blosc2 = "OFF"
+ADIOS2_USE_DataMan = "OFF"
+ADIOS2_USE_SZ = "OFF"
+ADIOS2_USE_ZeroMQ = "OFF"
+ADIOS2_USE_ZFP = "OFF"
+ADIOS2_PIP_INSTALL = "ON"
+BUILD_TESTING = "OFF"
+ADIOS2_INSTALL_GENERATE_CONFIG = "OFF"
```
vicentebolea, line 60: Looks good!

source/adios2/toolkit/sst/util/CMakeLists.txt
```diff
@@ -3,25 +3,27 @@
 # accompanying file Copyright.txt for details.
 #------------------------------------------------------------------------------#
 
-if(ADIOS2_HAVE_MPI)
-  set(maybe_adios2_core_mpi adios2_core_mpi)
-else()
-  set(maybe_adios2_core_mpi)
-endif()
+if (NOT ADIOS2_PIP_INSTALL)
```
vicentebolea, line 6: Should we do this conditionals in the parents CMake?

thirdparty/EVPath/CMakeLists.txt
```diff
@@ -17,6 +17,9 @@ set(EVPATH_LIBRARY_COMPONENT adios2_evpath-libraries)
 set(EVPATH_ARCHIVE_COMPONENT adios2_evpath-development)
 set(EVPATH_HEADER_COMPONENT adios2_evpath-development)
 
+string(REGEX REPLACE "[^/]+" ".." relative_base "${EVPATH_INSTALL_MODULE_DIR}")
+list(APPEND CMAKE_INSTALL_RPATH "$ORIGIN/${relative_base}/${CMAKE_INSTALL_LIBDIR}")
+
```
vicentebolea, line 22: We could possibly address this in the wheel audit

vicentebolea (commented):
Great work there We are almost there, the idea looks great! As for the comments I know that this is still WIP, I just added some comments.

cmake/DetectOptions.cmake
scottwittenburg: I'll change the name, but what do you mean add it as a "standard option"?

pyproject.toml
scottwittenburg: I took these names right from the copyright statement.  If they need to be updated, just need to let me know what names should be there.

thirdparty/EVPath/CMakeLists.txt
scottwittenburg: I'll look into it.

cmake/DetectOptions.cmake
vicentebolea: Something like this https://github.com/ornladios/ADIOS2/blob/master/CMakeLists.txt#L174

Also here https://github.com/ornladios/ADIOS2/blob/master/CMakeLists.txt#L244

source/adios2/toolkit/sst/util/CMakeLists.txt
scottwittenburg: What do you mean?  Just don't include the `util` subdirectory at all?

pyproject.toml
vicentebolea: hmm you are right we have to update the copyright.txt as well, lets have that conversation there with the rest of the team.

source/adios2/toolkit/sst/util/CMakeLists.txt
vicentebolea: yes

cmake/DetectOptions.cmake
scottwittenburg: I see.  My opinion was it should be marked as advanced, and not exposed as an option to users, as it shouldn't be needed or used outside of building with pip.

vicentebolea:
Another thing to add is the badge at the readme.

thirdparty/EVPath/CMakeLists.txt
scottwittenburg: It seems like this is going to be difficult or not possible.  I read that there is some way to tell `auditwheel` to exclude certain files, but the documentation is sparse.  By installing it in a virtual env, I can see help on the `auditwheel repair` command has:

```console
--exclude EXCLUDE     Exclude SONAME from grafting into the resulting wheel (can be specified multiple times)
```

But I don't think we want to omit it from the wheel, we just need to ignore the broken rpaths adios2 puts in those evpath libraries.  I feel like just adding a comment and/or guarding with the `ADIOS2_USE_PIP` option is all we have time for atm.

bindings/Python/CMakeLists.txt
scottwittenburg: That might make sense here, and thus remove the need for the if/else conditional.  However, `CMAKE_INSTALL_PYTHONDIR` is used [elsewhere](https://github.com/ornladios/ADIOS2/blob/master/cmake/ADIOSFunctions.cmake#L80-L91), and I think your suggestion would result in the conditional needing to be moved over there instead.  Or else we would need similar conditional in `DetectOptions.cmake`.

But let me know if you think I missed the point of your suggestion.

.github/workflows/pypackaging.yml
scottwittenburg: Releases were supposed to be covered by the case below, when you manually create a release in the UI.  If that's not how releases get made, then yes, we can add some push rules here.  I would think it would be on release tags rather than branches though.

vicentebolea: I Was thinking more on always running these workflow and when a env var is passed to the workflow it uploads it to pypi

source/adios2/toolkit/sst/util/CMakeLists.txt
scottwittenburg: Done

bindings/Python/CMakeLists.txt
scottwittenburg: You're right, I removed it.

bindings/Python/CMakeLists.txt
vicentebolea: hmm maybe that is ok, maybe for pip builds we would also want to install python source code in the libdir (in the build directory for tests). That would be more consistent.

.github/workflows/pypackaging.yml
scottwittenburg: Ah, ok.  I think we can do that.

cmake/DetectOptions.cmake
vicentebolea: we can declare the options as advanced. The reason of this is to have it printout when we do configure, this might come handy in future troubleshooting sessions

thirdparty/EVPath/CMakeLists.txt
vicentebolea: Thanks for looking into it, it seems that is not worth it to pursue this patch

> I feel like just adding a comment and/or guarding with the ADIOS2_USE_PIP option is all we have time for atm.

Yep, that seems to be the way to go

cmake/DetectOptions.cmake
scottwittenburg: It seems that when we do the printout in configure, it's checking for `ADIOS2_HAVE_OPTION`, but the `adios_option` function only defines `ADIOS2_USE_OPTION`.  Are we supposed to define manually the "`_HAVE_`" variable in `DetectOptions.cmake`?

vicentebolea: > Are we supposed to define manually the "_HAVE_" variable in DetectOptions.cmake?

Yes, thats were we add the logic for it.

bindings/Python/CMakeLists.txt
scottwittenburg: I don't think I understand your whole vision here.  But if it's still not how you want it, maybe we can discuss in person?

vicentebolea: sure lets sync maybe in 30mins?

pyproject.toml
vicentebolea: #3965

scottwittenburg:
If you download the the wheel or sdist from the [artifacts ](https://github.com/ornladios/ADIOS2/suites/19031621136/artifacts/1113669943) of the latest "Python Packaging" build, you can see that now the version we build (the version in the `.tar.gz` or `.whl` file name) matches the version baked into the module.

pyproject.toml
scottwittenburg: Thanks, I've done this.

bindings/Python/CMakeLists.txt
scottwittenburg: We agreed in our discussion that this is fine for now.

scottwittenburg:
Now each wheel undergoes an extra step where it is installed, and a simple test is run against it, e.g. [here](https://github.com/ornladios/ADIOS2/actions/runs/7216573109/job/19662996141?pr=3960#step:4:305).

vicentebolea (approved):

PR #3963, opened 2023-12-12
fix for ASAN issue related to JoinedDimArray handling in BP5 deserializer

pnorbert:
This is a quick fix to free a `vector<...>*` pointer in the destructor. However, this may be revised to use a vector instead of a pointer at all.


eisenhauer (approved):
If this improves san results, it looks good.  I'm still trying to sort out if the MinMax PR improves things, breaks other things, etc.  It seems that you've got to remove the CTEST_CUSTOM_MEMCHECK_IGNORE list to run the sanitizer list that we were ignoring to get that to pass, then you really need to look to see if you've introduced anything new (bad) or just removed issues (good).  My non-CI asan setup isn't flagging everything CI does, so I'm still trying to jump through the hoops that it appears that I need to (duplicating CI locally, etc.).

pnorbert:
It fixes only one of the two problems related to joined array, apparently.

eisenhauer (approved):

PR #3969, opened 2023-12-24
Bug fix building ADIOS2 without the campaign example 

anagainaru:
Currently the campaign example is built even if the `ADIOS2_USE_Campaign` flag is set to OFF


eisenhauer (approved):
Looks reasonable...

PR #3972, opened 2023-12-28
SST ASAN buffer fix

eisenhauer:
This deletes BP5 marshalling buffers for SST after use to fix a memory leak.  We may want to look into reusing those buffers, particularly for SST where we use the MallocV buffer subclass that grows with realloc() to keep things contiguous.  Reuse would potentially avoid the large-copy-on-realloc problem that happens apps that generate a lot of data.


vicentebolea (approved):
Yay! The list is getting smaller

PR #3975, opened 2023-12-29
Bump version to v2.10.0-rc1

vicentebolea:


vicentebolea:
@eisenhauer we are seeing interesting errors here, for once somewhere in the bpserializer it consider each component of the version string to be a single character, thus 2.10.0 is converted to 2.1.0 leading to the erros that we see. Here is a thought, how about we discard the patch release bit here? We already promise ABI compatible patch releases.

vicentebolea:
However by doing this we might just break compatibility with BP files prior 2.10. Maybe a best course of action is to set this as 2.9 for this RC release and let you guys figure out how to resolve this for the next RC.

eisenhauer:
Try this patch.  It lets the BP3/4 minor version go beyond 0x39 ascii, which lets the read side still read it properly because the reader takes the character read and subtracts 0x30 from it:
eisen@192-168-1-6 build % git diff
diff --git a/source/adios2/toolkit/format/bp/BPSerializer.cpp b/source/adios2/toolkit/format/bp/BPSerializer.cpp
index 57c9100fc..51b9a224f 100644
--- a/source/adios2/toolkit/format/bp/BPSerializer.cpp
+++ b/source/adios2/toolkit/format/bp/BPSerializer.cpp
@@ -232,7 +232,8 @@ void BPSerializer::PutMinifooter(const uint64_t pgIndexStart, const uint64_t var
     };
 
     const std::string majorVersion(std::to_string(ADIOS2_VERSION_MAJOR));
-    const std::string minorVersion(std::to_string(ADIOS2_VERSION_MINOR));
+    const char minorVersionChar = '0' + ADIOS2_VERSION_MINOR;
+    const std::string minorVersion(1, minorVersionChar);
     const std::string patchVersion(std::to_string(ADIOS2_VERSION_PATCH));
 
     const std::string versionLongTag("ADIOS-BP v" + majorVersion + "." + minorVersion + "." +
diff --git a/source/adios2/toolkit/format/bp/bp4/BP4Serializer.cpp b/source/adios2/toolkit/format/bp/bp4/BP4Serializer.cpp
index fc28640ba..46e8c59b4 100644
--- a/source/adios2/toolkit/format/bp/bp4/BP4Serializer.cpp
+++ b/source/adios2/toolkit/format/bp/bp4/BP4Serializer.cpp
@@ -61,7 +61,8 @@ void BP4Serializer::MakeHeader(BufferSTL &b, const std::string fileType, const b
     }
 
     const std::string majorVersion(std::to_string(ADIOS2_VERSION_MAJOR));
-    const std::string minorVersion(std::to_string(ADIOS2_VERSION_MINOR));
+    const char minorVersionChar = '0' + ADIOS2_VERSION_MINOR;
+    const std::string minorVersion(1, minorVersionChar);
     const std::string patchVersion(std::to_string(ADIOS2_VERSION_PATCH));
 
     // byte 0-31: Readable tag
eisen@192-168-1-6 build % 

There is still a problem with BP3 and BP4 because they assume a fixed size for the release string, and while this ascii patch will help with the minor version (as long as we don't move too close to three digits), the patch version handling is harder because the patch number might get much bigger than the minor version.  But this might do for now...

vicentebolea:
@eisenhauer your patch seems to have resolved this issue.

eisenhauer (approved):

PR #3976, opened 2024-01-02
A simpler way to deal with the mismatching layouts between the user code and the memory space

anagainaru:
A simpler way of switching the dimensions for GPU by using two Layout members in VariableBase to keep track of the base layout (RowMajor for C++ and ColumnMajor for Fortran) and the buffer layout (RowMajor for CPU pointers and ColumnMajor for GPU pointers). If there is a mismatch between the two dimensions are switched.
This PR closes #3961 

Tested in C++.
```
Write on LayoutRight, read on LayoutRight
1 1 1 1
2 2 2 2
3 3 3 3
1 1 1 1
2 2 2 2
3 3 3 3
Write on LayoutRight, read on LayoutLeft
1 2 3 1 2 3
1 2 3 1 2 3
1 2 3 1 2 3
1 2 3 1 2 3
Write on LayoutLeft, read on LayoutLeft
1 1 1 1
2 2 2 2
3 3 3 3
1 1 1 1
2 2 2 2
3 3 3 3
Write on LayoutLeft, read on LayoutRight
1 2 3 1 2 3
1 2 3 1 2 3
1 2 3 1 2 3
1 2 3 1 2 3
```

Tested in Fortran.
```
$ mpirun -np 2 ./adios2_writeread_f
Reading on Cuda data written on Cuda
0 0 0 0
1 1 1 1
2 2 2 2
1 1 1 1
2 2 2 2
3 3 3 3
Reading on Cuda data written on Host
1 1 1 1
2 2 2 2
3 3 3 3
2 2 2 2
3 3 3 3
4 4 4 4
$ ~/adios/ADIOS2/install-kokkos/bin/bpls BPFortranKokkos.bp/ -d -n 6
  int32_t  bpFloats     {4, 6}
Detect
is host: 1
    (0,0)    1 2 3 2 3 4
    (1,0)    1 2 3 2 3 4
    (2,0)    1 2 3 2 3 4
    (3,0)    1 2 3 2 3 4

  int32_t  bpFloatsGPU  {4, 6}
Detect
is host: 1
    (0,0)    0 1 2 1 2 3
    (1,0)    0 1 2 1 2 3
    (2,0)    0 1 2 1 2 3
    (3,0)    0 1 2 1 2 3
```

Currently the fortran example needs another library that takes care of the interoperability between Kokkos and Fortran: https://github.com/kokkos/kokkos-fortran-interop

Kokkos Views have a concept of Layout. ADIOS2 will capture this when users give Views to their Get/Put call and update the variable correspondingly. There is a new example to test this.

**Future documentation**
I will update the documentation in a future PR on how to compile and run the fortran and the new layout examples as well as the new changes to the `Shape` function that can now take a memory space and a layout to return the correct dimensions.

**Future testing**
The only things tested right now are: the logic when creating a variable,  when calling `SetSelection` and when calling `Shape` (for using Put/Get with different layouts). We might need testing for other ways of users setting the dimensions of the variable (on write and read -- to be designed later)


anagainaru:
Cleaning history after which this PR is ready for review

eisenhauer (approved):
What I see looks good.  Did you have tests to add?

anagainaru:
> What I see looks good. Did you have tests to add?

For now I only have examples, I agree I should add tests to the BP GPU suite. Maybe I'll find time tonight.

anagainaru:
The testing needs to be a separate PR (I will bring this up tomorrow) but basically we only have testing for Cuda and single mpi rank right now, so I want to redesign the whole testing for the GPU-aware functionality to include the layout and other memory spaces. For now, the examples are testing what I would put in the tests anyway so they would pass.

eisenhauer (approved):

Issue #3987, opened 2024-01-08
BP5 + operator for local array


pnorbert:
Apparently, BP5 does not support operators for local arrays, neither for writing nor reading.

eisenhauer:
Hmmm.  No tests for that anywhere?  From a quick glance at BP5 serializer, they should be supported for everything that's not a single value, but maybe there's a non-obvious problem.

eisenhauer:
Well, the logic for operator block sizing is missing on the read side for the Local case in GenerateReadRequests...  Might be more stuff, but it's where I'm starting.

eisenhauer:
Quick PR submitted.  It needs better test cases, but if you have one, give this a go.

pnorbert:
Fixed by #3989

PR #3989, opened 2024-01-08
Fix read request length on BP5 Local arrays with operators

eisenhauer:
Hopefully this addresses Issue #3987


pnorbert (approved):
It does fix my example, thank you.

PR #3997, opened 2024-01-09
Posix transport should fail if you try to read past EOF

eisenhauer:


pnorbert (approved):

source/adios2/toolkit/transport/file/FilePOSIX.cpp
```diff
@@ -417,6 +418,25 @@ void FilePOSIX::Read(char *buffer, size_t size, size_t start)
...
+                    helper::Throw<std::ios_base::failure>(
+                        "Toolkit", "transport::file::FilePOSIX", "Read",
+                        "Read past end of file on " + m_Name + " " + SysErrMsg());
+                }
+                else
+                {
+                    // read past EOF, but we're to wait for data.  Exponential backoff with a limit
+                    // of .5 sec (500,000,000 nanosec)
+                    std::this_thread::sleep_for(std::chrono::nanoseconds(backoff_ns));
+                    constexpr size_t backoff_limit = 500 * 1000 * 1000;
```
vicentebolea, line 434: Does it makes sense to set a inotify watcher on the file that we are reading to get notified when the file is modified?

source/adios2/toolkit/transport/file/FilePOSIX.cpp
```diff
@@ -2,13 +2,12 @@
  * Distributed under the OSI-approved Apache License, Version 2.0.  See
  * accompanying file Copyright.txt for details.
  *
- * FileDescriptor.cpp file I/O using POSIX I/O library
+ * FilePOSIX.cpp file I/O using POSIX I/O library
  *
- *  Created on: Oct 6, 2016
- *      Author: William F Godoy godoywf@ornl.gov
```
williamfgc, line 8: Would be nice to  keep.

source/adios2/toolkit/transport/file/FilePOSIX.cpp
eisenhauer: My gut says no.  Mostly this stuff comes into play because of oddities with filesystem behavior.  When our engines are in streaming mode, pending readers wait until new metadata has been added to the files by readers.  By that time, other files (metadata index, data files, etc.) should have been written, but those updates sometimes don't immediately appear to the reader (maybe because of filesystem caching, or just because the filesystem is only eventually consistent).  In any case, the quick solution is to try to read again.  We were just spinning on read(), which seemed like painful overkill, so we added exponential backoff with a limit, which I think is still fine.  We really don't expect to be waiting a long time here, *unless* there's a bug in ADIOS and we're trying to read more than was ever (or ever will be) written.   That's what the other side of this else was meant to deal with.  The problem is that we're hitting the other side of the else (and throwing an exception) when the writer has done endstep, written data and even closed the file, but the next reader is still not seeing all of the data that it expects (again, presumably for the same reasons as above).   So the result of the discussion of in the call was that we won't immediately throw an exception in the other side of the if.  Instead, we'll do the same sort of backoff, but at the end of a minute throw an exception.  Hopefully it'll do the trick by waiting for the filesystem to sort itself out without adding overhead in normal circumstances.

source/adios2/toolkit/transport/file/FilePOSIX.cpp
anagainaru: If you look at the history of the file it went through multiple revisions and it has many authors. I think we should remove the header from all the files since there is a copyright file that is covering the entire repo and includes all the authors (you are included). I'll bring it up next Tuesday and if everyone agrees I can can create a PR.

PR #4000, opened 2024-01-12
Implementing sqlite database instead of json file. Writing part.

dmitry-ganyushin:


pnorbert (changes requested):
- This is not saving the records as they are being generated, only at Close(), like what the json writer does.
- Not delta_step and delta_time what you need to save for this phase, but step and time itself.
- We talked about two tables, one for the output names only, then using that table's rowid() to put in as an index into the second table that has file-id, step, time. You could put the deltas as extra columns once you are at it.

dmitry-ganyushin:
> * We talked about two tables, one for the output names only, then using that table's rowid() to put in as an index into the second table that has file-id, step, time. You could put the deltas as extra columns once you are at it.

Could you please specify the sql schema for that?

pnorbert (approved):
Approved for the task, but we will need to do the recording step by step in the future, and maintain a second table of step/time information.

PR #4013, opened 2024-01-25
Replace the only PERFSTUBS_SCOPED_TIMER_FUNC() call in BP5 to get rid…

pnorbert:
… of an ASAN error blocking local testing.


pnorbert:
To me, it's more annoying than useful to have this function call in BP5, once I start running with asan

eisenhauer (approved):

PR #4014, opened 2024-01-26
Python updates

pnorbert:
Allow Stream/FileReader constructor overloading and Stream.read()/write() overloading using @singledispatchmethod.
Return status from engine when using Stream.begin_step(). Update examples with the changes to the API
update python examples (hello_world, bpWriter, gray-scott, dataMan) Change stream iterator to use engine BeginStep return status Create python DataMan/SST writers and readers for both high-level and bindings APIs, and make them work with the C++ example versions.


eisenhauer:
There was some magic to getting the docker images rebuilt, but I don't think I was privy to it...

eisenhauer:
Odd.  That failure in macos is spurious, but is happening often enough that we likely have to sort it.  (Test retries means that this had to have happened several times before it gets reported.). The "Errno was 61" come from inside the EVPath communication system and implies that we've tried to do a connection to the remote_server, but the connection failed.  The CI infrastructure should have started the server before all the remote tests and shut it down afterwards, so I'm wondering if it might be related to using a fixed port number for the server.  That would cause problems in CI if we have multiple runs on shared resources.

eisenhauer (approved):

PR #4016, opened 2024-01-29
FFS upstream

eisenhauer:


anagainaru (approved):
Is this the fix for #3926 ?
I did not build this but looks reasonable.

PR #4021, opened 2024-01-31
fix IO.available_attributes(), remove a debug print from Stream, fix …

pnorbert:
…examples printing, fix el8 python3 build

### DIFF
--- a/python/adios2/io.py
+++ b/python/adios2/io.py
@@ -103,10 +103,7 @@ class IO:
                 value
                     attribute information dictionary
         """
-        attributes = {}
-        for name, attr in self.impl.AvailableAttributes():
-            attributes[name] = Attribute(attr, name)
-        return attributes
+        return self.impl.AvailableAttributes()
 
     def remove_attribute(self, name):
         """
--- a/python/adios2/io.py
+++ b/python/adios2/io.py
@@ -66,7 +66,15 @@ class IO:
                 e.g. variable_name + separator + name ("var/attr")
                 Not used if variable_name is empty
         """
-        return Attribute(self.impl, name, content, variable_name, separator)
+
+        # string or list of strings passed on as is
+        if isinstance(content, list) and isinstance(content[0], str):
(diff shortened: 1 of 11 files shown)


eisenhauer (dismissed):

python/adios2/io.py
```diff
@@ -103,10 +103,7 @@ def available_attributes(self):
                 value
                     attribute information dictionary
         """
-        attributes = {}
-        for name, attr in self.impl.AvailableAttributes():
-            attributes[name] = Attribute(attr, name)
-        return attributes
+        return self.impl.AvailableAttributes()
```
vicentebolea, line 106: This will return binding attributes rather than High Level API attributes. Note that we want the user not to use the bindings unless he wants by using the impl member variable. Also note that other hight level python classes expect the high level attributes rather than the bindings for it to work

vicentebolea (changes requested):
See my comment

python/adios2/io.py
pnorbert: Actually, this function returns a dictionary of dictionaries with the attribute names as key and information about each attribute as value. It's the inquire_variable() function that returns a high-level attribute, which I did not change. 

The existing loop simply fails as the dictionary cannot be converted to an attribute but is also shouldn't be.

python/adios2/io.py
```diff
@@ -66,7 +66,15 @@ def define_attribute(
                 e.g. variable_name + separator + name ("var/attr")
                 Not used if variable_name is empty
         """
-        return Attribute(self.impl, name, content, variable_name, separator)
+
+        # string or list of strings passed on as is
+        if isinstance(content, list) and isinstance(content[0], str):
```
vicentebolea, line 71:
```suggestion
        if isinstance(content, list) and len(content) > 0 and isinstance(content[0], str):
```

vicentebolea (changes requested):
Small comment for code safety otherwise approved :)

vicentebolea (approved):

Issue #4025, opened 2024-02-05
CI: All-Static builds (Linux/macOS/Win)


ax3l:
Downstream, it makes sense for some deployments (e.g., PyPA/pip of openPMD-api) to build an all-static ADIOS2 builds with all-static dependencies.

This often breaks in ADIOS2 #4019 #3898 #3716 #2747 #2454 #2316 #2576 #1810 ...

Could we add a CI entry for each Linux/macOS/Win that:
- builds all ADIOS2 dependencies as static
- builds ADIOS2 as static
- does smoke tests?

Currently, we only have three static Linux builds in CI. I am not sure they also test dependencies like C-Blosc2 (static!).

For our use case, covering `serial` is sufficient.

Note: we definitely want to keep shared builds for the same combinations too (esp. Windows DLL exports are tricky and need coverage).

vicentebolea:
I believe that his has been fixed in #4037 and #4041

PR #4027, opened 2024-02-06
Fix Windows Testing build

eisenhauer:


eisenhauer:
This was broken when Python is present for the build, which is apparently not the case for our Windows CI.

vicentebolea:
@eisenhauer that makes sense, however, I cannot see Python enabled in any of our Windows builds.

eisenhauer:
> @eisenhauer that makes sense, however, I cannot see Python enabled in any of our Windows builds.

It's not, but that's just the explanation for why we didn't know this was broken...

vicentebolea:
> It's not, but that's just the explanation for why we didn't know this was broken...

I see, how about enabling python to see if this fixes it(or at least move us to an step closer to have python in Win)?

eisenhauer:
So, I can say that it fixes it in my Windows Visual Studio build (which has python present).  We can also change a CI to do this.  (But if we're adding CIs, I'd probably prioritize Issue #4025 .  We definitely have someone that needs that.)

vicentebolea (approved):

PR #4033, opened 2024-02-12
WIP. Fortran 2008 interface

dmitry-ganyushin:
We need to add Fortran 2008 interface for the Nimrod program. In this branch the new interface is used everywhere.  Should be done with the old one with a conditional variable set up for cmake.


dmitry-ganyushin:
Not really required for Nimrod. Can be postponed.

PR #4035, opened 2024-02-14
Tweak POSIX fail on EOF to kill heisenbug

eisenhauer:


eisenhauer:
@pnorbert FailOnEOF test time has increased here to 43 seconds, representing the amount of time consumed by the exponential backoff before deciding there really isn't going to be any more data.  This seems like a reasonable wait time (for data files where we're really just waiting for filesystem stabilization of data that should already have been written if it was going to be).

vicentebolea (approved):

PR #4036, opened 2024-02-15
Python write scalars

pnorbert:
Allow for these:
```
fh.write("Nx", Nx)
fh.write_attribute("size", Nx, "bpArray")
fh.write_attribute("dimensions", ["Nx"], "bpArray")
...
  int64_t  Nx                  scalar = 10
  double   bpArray             {10} = 0 / 9
  string   bpArray/dimensions  attr   = {"Nx"}
  int64_t  bpArray/size        attr   = 10
```
```
fh.write("N", [Nx, Ny])  
....
  int64_t  N                         {2} = 10 / 10
```
```
c = complex(1.0, 2.0)
s.write("a_c64", c)
s.write("an_int_list", [1, 2, 3, 4])
s.write("a_float_list", [1.0, 2.0, 3.0, 4.0])
...
  double complex  a_c64         scalar = (1,i2)
  double          a_float_list  {4} = 1 / 4
    (0)    1 2 3 4
  int64_t         an_int_list   {4} = 1 / 4
    (0)    1 2 3 4
```
```
int_list = [1, 2, 3, 4]
float_list = [1.0, 3.0, 5.0, 7.0]
s.write_attribute("attr_int_list", int_list)
s.write_attribute("attr_float_list", float_list)
...
  double complex  attr_a_c64              attr   = (1,i2)
  double          attr_float_list         attr   = {1, 3, 5, 7}
  int64_t         attr_int_list           attr   = {1, 2, 3, 4}
```


eisenhauer (commented):
Huh, I'd have sworn that we had DataTypeToString already...

eisenhauer (approved):

eisenhauer:
Hmm.  The failure is an ASAN failure relating to memory leaks in the connection to a remote data server.  This isn't a reliable failure, so I'm not sure what's going on.  For the moment I'd recommend retriggering CI and hoping for a passing result on the next try.  I'll try to sort out if there might be some kind of race condition in tracking these releasable allocations.

eisenhauer:
> > Huh, I'd have sworn that we had DataTypeToString already...
> 
> Isn't it this one in `common/ADIOSTypes.cpp`:
> 
> ```c++
> std::string ToString(DataType type);
> ```

Yep, that's the one!  But it returns a c-style type, with spaces, etc.  Maybe the form added here is needed for Python?  If so, we should name them appropriately.

pnorbert:
Nope, sorry, just did not find it. There is no need for the new one.

pnorbert:
Removed duplicate function and squashed the commits

eisenhauer (dismissed):

pnorbert:
Please don't merge until Vicente reviews it.

python/adios2/io.py
```diff
@@ -67,13 +67,21 @@ def define_attribute(
...
             return Attribute(self.impl, name, content, variable_name, separator)
-        if isinstance(content, str):
+
+        if isinstance(content, list):
+            if len(content) > 0 and isinstance(content[0], str):
+                # string list passed on, becomes std::vector<std::string>
+                return Attribute(self.impl, name, content, variable_name, separator)
+            # non string lists needs to be converted to numpy array
+            content = np.array(content)
+            return Attribute(self.impl, name, content, variable_name, separator)
```
vicentebolea, line 79: Why do we need to convert attributes list (no string list) to numpy arrays?

python/adios2/engine.py
```diff
@@ -102,8 +102,16 @@ def put(self, variable, content, mode=bindings.Mode.Deferred):
         """
         if isinstance(content, np.ndarray):
             self.impl.Put(variable.impl, content, mode)
-        else:
+        elif isinstance(content, str):
+            self.impl.Put(variable.impl, content)
+        elif isinstance(content, list):
+            content = np.array(content)
```
vicentebolea, line 108: Why converting any non string list to numpy? It seems like a further copy to another data type.

python/adios2/stream.py
```diff
@@ -270,15 +269,22 @@ def _(self, name, content, shape=[], start=[], count=[], operations=None):
...
-                variable = self._io.define_variable(name, content_np, shape, start, count)
-            # Scalars variables
-            elif isinstance(content, str) or not hasattr(content, "__len__"):
-                variable = self.define_variable(name)
+                content = np.array(content)
+                if shape == []:
+                    shape = [content.size]
+                    count = shape
+                    start = [0]
+                variable = self._io.define_variable(name, content, shape, start, count)
```
vicentebolea, line 281: This will break local variable/arrays since they always use `shape=[]`.

vicentebolea (changes requested):
Looks good, a few questions/comments

bindings/Python/py11IO.h
```diff
@@ -11,6 +11,7 @@
 #ifndef ADIOS2_BINDINGS_PYTHON_IO_H_
 #define ADIOS2_BINDINGS_PYTHON_IO_H_
 
+#include <pybind11/complex.h>
```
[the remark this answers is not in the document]
pnorbert: indeed, it was not needed at all

python/adios2/io.py
pnorbert: Because we had no glue functions. Now I added them for lists (vectors) and values (pybind11::object).

python/adios2/engine.py
pnorbert: no conversion now

python/adios2/stream.py
pnorbert: Indeed, this broke it but had to add proper parallel testing for local variables to see this. Fixed.

python/adios2/engine.py
```diff
@@ -105,7 +105,7 @@ def put(self, variable, content, mode=bindings.Mode.Deferred):
         elif isinstance(content, str):
             self.impl.Put(variable.impl, content)
         elif isinstance(content, list):
-            content = np.array(content)
+            # content = np.array(content)
```
vicentebolea, line 108:
```suggestion
```

vicentebolea (changes requested):
One last small detail

python/adios2/engine.py
pnorbert: done

pnorbert:
rebased with master and squashed commits

vicentebolea (approved):

PR #4038, opened 2024-02-19
campaign management: update sqlite database after another run

dmitry-ganyushin:
Update  sqlite database if a table with bp files exists. Before there was an error.


pnorbert (approved):
Please remove the rowid variable.

PR #4041, opened 2024-02-20
ci: add macos static build

vicentebolea:
cc @eisenhauer

X-Ref: #4025


vicentebolea:
@pnorbert it turns out that we support static builds in macos.

eisenhauer (approved):
Looks good to me...

PR #4045, opened 2024-02-26
Add timeout parameter to stream.steps(). Necessary for controlling ti…

pnorbert:
…meout in in situ processing. Also very useful for not being stuck forever in processing files that were not closed properly. Default is still infinity, which may not be the ideal setup for python scripts in practice. Usage:

    for _ in f.steps(timeout=1.0):
        ...


anagainaru (approved):
Looks good

PR #4048, opened 2024-02-27
Reading of campaign configuration parameters from ~/.config/adios2

dmitry-ganyushin:
Campaign confrig parameters can be located in the file ~/.config/adios2 in the form 

cachepath /tmp/campaign
verbose 0


pnorbert:
There should be two separate configs in ~/.config/adios2/
One for the user settings for local: cache path, campaign path, cache size etc
Another one (at least) for the project settings:
   summit   dtn.olcf.gov
   frontier   dtn.olcf.gov
   perlmutter ssh://dtn.nersc.gov
   acacia xroot://phy122@galaxy.pawsey.au

pnorbert (approved):

PR #4055, opened 2024-02-27
blosc2: do not export blosc2 lib

vicentebolea:
Blosc2 is a private dependency thus we do not have to export its targets.

### DIFF
--- a/cmake/DetectOptions.cmake
+++ b/cmake/DetectOptions.cmake
@@ -92,15 +92,14 @@ if(Blosc2_FOUND)
 
   set(adios2_blosc2_tgt Blosc2::Blosc2)
   if (Blosc2_VERSION VERSION_GREATER_EQUAL 2.10.1)
-    if (Blosc2_shlib_available)
-      set(adios2_blosc2_tgt Blosc2::blosc2_$<IF:$<BOOL:${ADIOS2_Blosc2_PREFER_SHARED}>,shared,static>)
+    if (Blosc2_shlib_available AND ADIOS2_Blosc2_PREFER_SHARED)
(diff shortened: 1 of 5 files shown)


vicentebolea:
We needed to get rid of the `cmake_dependent_option` since it requires to use a generator expression in the adios2_blocs2 library, this in turn is incompatible with ALIAS types of libraries which we need to add a level of indirection for the blosc2 library without messing the private/public'ness of the library (which happened when using INTERFACE type of lib).

cmake/DetectOptions.cmake
```diff
@@ -92,15 +92,14 @@ if(Blosc2_FOUND)
 
   set(adios2_blosc2_tgt Blosc2::Blosc2)
   if (Blosc2_VERSION VERSION_GREATER_EQUAL 2.10.1)
-    if (Blosc2_shlib_available)
-      set(adios2_blosc2_tgt Blosc2::blosc2_$<IF:$<BOOL:${ADIOS2_Blosc2_PREFER_SHARED}>,shared,static>)
+    if (Blosc2_shlib_available AND ADIOS2_Blosc2_PREFER_SHARED)
+      set(adios2_blosc2_tgt Blosc2::blosc2_shared)
     else()
```
anagainaru, line 97: We will never use the `Blosc2::blosc2_static` target?

anagainaru (approved):
This looks good to me

cmake/DetectOptions.cmake
vicentebolea: only if it blosc2_shared does not exists or the user specified -DADIOS2_Blosc2_PREFER_SHARED=OFF

PR #4059, opened 2024-02-28
backports from master

vicentebolea:
- Python API: support writing Python numeric variables and lists (of one type) as ADIOS variables and attributes. Add extra tests for them and some in the bpWrite and bpReaderHeatMap2D examples. - Add glue to accept lists and single python value for variables and attributes. - Add test for lists and values (int, float and complex) to API tests. - Turn on HDF5 Python test (serial version)
- Add a static windows build to CI (#4037)
- Add arguments to IO.AvailableAttributes() and io.available_attributes() to get attributes of a specific variable.
- Return a 0-dim numpy array of size 1  when reading a scalar variable instead of a 1-dim array of size 1. The difference can be seen with gray-scott gsplot, as step is printed now 10 instead of [10]. Same when running bpReaderHeatMap2D.py.
- Add timeout parameter to stream.steps(). Necessary for controlling timeout in in situ processing. Also very useful for not being stuck forever in processing files that were not closed properly. Default is still infinity, which may not be the ideal setup for python scripts in practice. Usage:     for _ in f.steps(timeout=1.0):         ...
- blosc: do not export blosc lib
- Fix segfault when asking for BlocksInfo for a variable name that does not exist. Return value is now empty dictionary. Also fix handling 1D arrays made from LocalValues. Now this code works for examples/basics/values/valuesWrite.cpp ProcessID:     with adios2.Stream("values.bp", 'rra') as f:         e = f.engine         bi = e.blocks_info("ProcessID", 0)         print(f"Blocks info = {bi}")


vicentebolea:
@ax3l you can use this branch to test the fix for your openPMD builds.

vicentebolea:
@eisenhauer new round of backpors

eisenhauer (approved):
Looks good...

PR #4062, opened 2024-02-29
Bison 3.8 Parser

lizdulac:
Rewrote DerivedVariable Expression Parser to use Bison 3.8 - allows us to do things like group all parser objects under adios2::detail namespace. Also included in rewrite:
- Generated parser files moved to pregen-source directory. CMake checks to see whether to regenerate parser files or use pregen-source
- ASTDriver class created wrapping around parser and storing information related to an individual expression (such as variable lookup table, and stack of parsed ASTNodes that have yet to be attached to ASTree)
- Expression function names (ex: "curl") are no longer checked in parser, limiting how much parser will need to be changed when adding new functions (ADIOS error added if function name cannot be found in ExprHelper mapping. ASTNode objects only contain string operation name, and no ExpressionOperator).
- Indices parsed into vector of tuples (start, end, stride) with default (-1, -1, 1) if left blank (attaches to ASTNode, but not yet translated to Expression class)


source/adios2/toolkit/derived/parser/CMakeLists.txt
eisenhauer, line 1: This is no longer used, right?  Can be deleted from the PR?

source/adios2/toolkit/derived/parser/pregen-source/Makefile
eisenhauer, line 1: Also unnecessary?

source/adios2/toolkit/derived/parser/pregen-source/parser.output
eisenhauer, line 1: Probably we don't need this, this output is informational, not code.

source/adios2/toolkit/derived/parser/pregen-source/simple-c
eisenhauer, line 1: Binary file?  Also needs to go presumably...

source/adios2/toolkit/derived/parser/test-driver.cpp
eisenhauer, line 1: Do we want to convert this to a unit test?  Or just delete?  We probably don't need this in this form in the ADIOS repo...

testing/adios2/derived/TestBPDerivedCorrectness.cpp
eisenhauer, line 1: I don't recall, are we running this test all the time?

lizdulac: This test is only built if ADIOS2_HAVE_Derived_Variable is true (per testing/adios2/CMakeLists.txt). If you're asking about the CI, I'm not sure.

source/adios2/toolkit/derived/parser/test-driver.cpp
lizdulac: I probably should have deleted this; it's a parser-specific demo for development debugging (adios independent). If the parser is particularly broken, it should be picked up in the DerivedVariable Correctness test.

testing/adios2/derived/TestBPDerivedCorrectness.cpp
eisenhauer: We should consider removing that conditional so that this stuff is always built and tested...

source/adios2/toolkit/derived/parser/pregen-source/parser.output
lizdulac: Good point - deleted

testing/adios2/derived/TestBPDerivedCorrectness.cpp
anagainaru: Hmm I am not sure that is possible, the test requires the derived variables to be enabled. I don't think we are ready to turn them on by default (there are still many tests we need to do to be sure the code is stable). I would much rather enable the derived variables in some CIs

eisenhauer: That's reasonable, but turning it on for the moment lets us sort through compilation and other failures.  We can still default the option to off before we merge...

anagainaru: Sounds good, I'll have time to review this tomorrow so I'll help push this soon

eisenhauer:
wait, how did I close this?

eisenhauer:
One sec and I think I can kill the last leak...

eisenhauer:
OK, so msan failed because it timed out.  Only completed 755 of 756 tests.  But I don't get the other failure...

eisenhauer:
OK, this version passes all checks with and without Derived_Variable enabled.  Left it disabled.

anagainaru:
I can find some time to give this a final look tomorrow night (Japan time). If you need this merged before, go ahead and I'll review just for my own sanity :)

eisenhauer (approved):

PR #4064, opened 2024-02-29
Fix stream flush

pnorbert:
Do not flush io and adios in read mode because BP5 reader does not like it.


pnorbert:
I messed up something, this supposed to be a single file

PR #4066, opened 2024-03-01
Using the correct flag to detect the CUDA backend in ZFP

anagainaru:
Trying to create an example using the CUDA backend in ZFP I found this bug. 

@vicentebolea added the `ZFP_WITH_CUDA` env variable in https://github.com/LLNL/zfp/pull/147
This variable has been used in the past 2 years, I think it's safe to rely on it to detect when ZFP can use CUDA.


pnorbert (approved):
No one remembers where the ZFP_CUDA thing came from. Thanks for fixing this.

vicentebolea (approved):
oh I missed, this. looks great!

PR #4069, opened 2024-03-03
Raise an exception if remote open fails

eisenhauer:


eisenhauer:
This removes the "Errno = " printf  (EVPath upstream) and adds an exception if we try to open a file on the remote server and fail for any reason.  At the moment, there is no difference between failing to contact a remote server, the file not existing, not being readable, etc.  All result in the same exception.  We can fine-tune.

pnorbert:
I am not sure this is what I wanted. `bpls -l` will fail to list the content without a valid remote connection

pnorbert:
Maybe throw the error in PerformGetRemote?

eisenhauer:
Possible of course, but requires a few more hoops to jump through.  Currently we don't get to PerformRemoteGets() unless we have a remote connection.  m_remote.Init() is the only place where we know if we should *try* to get a connection.  So, rather than using the m_remote to keep state, we'd have to track more info, like if we were supposed to have a remote connection, but don't.  Or I suppose we could not do the remote open when we do it, but instead do it on-demand or something...  Thoughts?

pnorbert:
I favor on-demand connection. Just opening and using metadata does not require remote connection and probably would not want it with lots of files/campaign files. Imagine a single campaign file with files from multiple locations. A simple bpls would initiate multiple connections even if the user has no intention to get data from all files.

eisenhauer:
OK, let me give that a shot tomorrow...

eisenhauer:
OK, this PR pushes the Open() back to PerformGets() and it checks to see that PerformGets() has something to do before it'll bother with the open.  It works fine with the existing tests.  BUT, those tests were really placeholders for a more detailed test suite.  We should probably beef that up at some point...

anagainaru (commented):
Just out of curiosity, why is the open taking a layout?

We don't care right now, but we might want GPU-aware for remote reading in the future so I just want to understand. I'm going to let Norbert review.

eisenhauer:
> Just out of curiosity, why is the open taking a layout?
> 
> We don't care right now, but we might want GPU-aware for remote reading in the future so I just want to understand. I'm going to let Norbert review.

I assume you're talking about the row-vs-column major stuff.  This comes in to ADIOS from the bindings and represents the programmatic context of the client.  I believe that the server needs to know that too, because the system's behavior depends upon this, so on the client we don't always open as if we are in C++ (despite the server being a c++ client).  Instead we open a file as if we were the remote client...

anagainaru:
That makes sense, the row/column major should stay at the program level and the buffer layout is dealt with at every individual Get, not in the open.

pnorbert (approved):

PR #4070, opened 2024-03-04
WIP. Adios2 connection tool for SSH port forwarding.

dmitry-ganyushin:
Uses aliases for hostnames. Read data from .config/adios2/hosts.ini file


pnorbert:
Does this connection launch the remote server? If not, how do you start it up on demand?

dmitry-ganyushin:
The Paramiko library which is used  can start a remote executable.

Issue #4051, opened 2024-02-27
set memory space functions in the bindings


pnorbert:
fortran, c, and python?

anagainaru:
C bindings in #4074 
Fortran bindings in #4077 

For the python bindings I need to understand how python functions might use the GPU backend so I would leave that for a later time (not in this release). I think this issue can be closed when #4077 is merged.

PR #4074, opened 2024-03-05
C bindings for the memory space related functions

anagainaru:
Adding three functions to the C bindings:
```C
adios2_error adios2_variable_shape_with_memory_space(size_t *shape, const adios2_variable *variable,
                                                     const adios2_memory_space mem);

adios2_error adios2_set_memory_space(adios2_variable *variable, const adios2_memory_space mem);

adios2_error adios2_get_memory_space(adios2_memory_space *mem, adios2_variable *variable);
```

In addition, there is a tests in `bindings/C` to make sure these functions return the correct values. BTW I don't actually need to use GPU pointers to test the functions so we can test this regardless of the Kokkos/CUDA backends being turned on.

This PR is the first part of #4051


bindings/C/adios2/c/adios2_c_types.h
```diff
@@ -159,6 +159,15 @@ union adios2_PrimitiveStdtypeUnion
     char *str;
 };
 
+typedef enum
+{
+    adios2_memory_space_host = 1,
+#ifdef ADIOS2_HAVE_GPU_SUPPORT
+    adios2_memory_space_detect = 0,
```
pnorbert, line 166: If this is user facing, shouldn't they always be available and work properly / throw error in functions where it's in conflict?
If a code use `adios2_memory_space_detect`, it will not compile with a CPU-only adios build

anagainaru: This is the same for the C++ part. If you don't enable the GPU backend there is no point for the Detect or GPU memory spaces. We don't want to go through the logic of trying to detect the memory space if the only option is Host. 

The reason I don't only allow the function if GPU is enabled is for Kokkos codes that can work on any backend without modification. I don't want the users to have to add/remove the memory space function depending on how they build Kokkos

eisenhauer: But this is our external API, right?  If part of it changes depending upon how ADIOS is built (like an enum entry vanishing), that requires user codes to change..  Or am I missing something?

anagainaru: Norbert convinced me :) so I am allowing the Detect memory space to be used regardless of the backend. With the GPU memory space, people should change their codes anyway if they use GPU pointers and I am in favor of a compiler time error than a seg fault if they forget to update this function.

pnorbert (approved):

eisenhauer (approved):

bindings/C/adios2/c/adios2_c_types.h
williamfgc: Good! One minor comment, ADIOS_HAVE_GPU should be sufficient as with other compile-time options. _SUPPORT can be removed.

anagainaru: Sorry William, I didn't see your comment. Right now we have the following:
```
      c        : OFF
      CUDA               : OFF
      Kokkos             : ON
      Kokkos_CUDA        : OFF
      Kokkos_HIP         : ON
      Kokkos_SYCL        : OFF
```
The GPU support is not something the user can turn ON by default so I kinda wanted to separate the name from the typical name the user would expect. I use ` ADIOS_HAVE_GPU_SUPPORT` internally to know if CUDA or any of the GPU backends in Kokkos are ON. We can def discuss this if you think it's important. It doesn't necessary fit in this PR tho so I will leave this merged.

anagainaru: Just a small follow up, if you use ccmake you will only see two options for the user
````
ADIOS2_USE_CUDA
ADIOS2_USE_Kokkos
```
everything else from my previous comment are just internal variables and the user cannot set them.

williamfgc: Nice! Thanks @anagainaru

PR #4076, opened 2024-03-06
Python: add the same treatment to attributes as to variables before: …

pnorbert:
…return scalars (0-dim ndarray) for single value attributes.


eisenhauer (dismissed):
This looks good.

eisenhauer (approved):

PR #4078, opened 2024-03-06
Actually look at return value for read in remote_server

eisenhauer:
Even the youngest padawan knows that read() sometimes returns partial results, how did this live so long?   Maybe because we're not really handling errors in the remote protocol so far...


anagainaru (approved):
Looks good, excellent title and description

eisenhauer:
@anagainaru Fortunately we have CI to test for stupid errors...  Will need another review.

PR #4083, opened 2024-03-10
Update documentation for 2.10 changes to the GPU-backend

anagainaru:
- Updated the paths to the example codes
- Updated the default behavior when building ADIOS2 with Kokkos/CUDA available
- Added info about the new Shape function for the Read side 

This is my contribution to issue #4073


eisenhauer (approved):

anagainaru:
@eisenhauer could you squash merge this? I still can't pass the requirements that all checks must pass before merging

Screenshot 2024-03-10 at 10 31 00 PM

PR #4088, opened 2024-03-12
adios2_define_derived_variable C/Fortran API. C is compiled condition…

pnorbert:
…ally, like the C++ API. The Fortran function is always available, it will print a WARNING if derived variable is not supported. Added Fortran test for magnitude().


anagainaru (approved):
Looks good

anagainaru (approved):

PR #4092, opened 2024-03-12
Fix static blosc2 build

vicentebolea:


vicentebolea:
@ax3l can you try this?

PR #4097, opened 2024-03-18
WIP: Make Fortran tests fail with a non-zero exit code

eisenhauer:


anagainaru:
@eisenhauer When we change the default memory space to be Default regardless of GPU being enabled or not, I forgot to change the Fortran bindings to update this. I think it should pass now.

anagainaru:
Hmm, maybe it's more complicated, even though I don't understand how the changes we made could trigger this new error. I'll try to reproduce it locally (I don't see it on my laptop).

anagainaru:
Turns out the default space needed to be updated in two location (core and bindings). It should pass the tests now, let me know if you need me to clean the history in case you don't want to squash merge.

anagainaru (approved):

eisenhauer:
> Turns out the default space needed to be updated in two location (core and bindings). It should pass the tests now, let me know if you need me to clean the history in case you don't want to squash merge.

I'll squash...  Thank you!

eisenhauer:
So, the deal is this:  Prior to the change in VariableBase, m_MemSpace had value MemorySpace::Host if GPU support was not enabled.  After the change, it was left at MemorySpace::Detect.  The DataMan serializer was looking for Host or GPU values, but couldn't handle the Detect value.   I changed DataMan to handle Detect as if it was Host.  One could also have presumably replaced the Detect with Host at some point in the output process, but I'm not entirely clear where would be best. So I've left it at this for the moment.

anagainaru:
@eisenhauer by the time it reaches dataman there shouldn't be the option of Detect, let me check this before we merge. I did not catch this when I change the default since I am apparently checking this only in the fortran bindings. I might add a test for the default and set memory space when we test the BP engine.

anagainaru (approved):
The problem(s) should be fixed now @eisenhauer. Dataman had a bunch of issues due to changing the default memory space. Should we make a separate PR with the fixes and leave this one clean just with the Fortran CI fixes? Otherwise if we keep just one, take a look at the changes I made (I change the GPU part of your print function) and I think this can be merged (I'll approve so you have this choice).

anagainaru:
I'll add tests for it when I am back, I think it should work ok now. I tested a bunch of things on Frontier, but I did not have time to put them all in a test.

PR #4099, opened 2024-03-19
[Never Merge] PR just to check the Fortran tests

anagainaru:
cherry picked my commit from #4097 to make sure the tests pass without Greg's contribution

I cannot reproduce the error anywhere so this is an attempt to debug this issue


anagainaru:
This is not needed anymore, the issues was with setting the default memory space when GPU is not enabled

PR #4102, opened 2024-03-21
Fix typo in Fortran API docs

JamieJQuinn:
Noticed a small type in Fortran API docs and fixed it. Haven't tested docs build but I presume such a small change should be fine.


eisenhauer:
Thank you!

eisenhauer (approved):

PR #4113, opened 2024-03-28
Fix warnings that happen in Release build

eisenhauer:


eisenhauer:
Release mode inlines some routines and we get a different set of errors.  Mostly here we've abused strncpy in ways that some compilers consider unsafe.  This PR removes those warnings.  Using strncpy_s would have been better in some cases, but that's not guaranteed with c++11.

anagainaru (approved):

PR #4115, opened 2024-03-28
User options

pnorbert:
Add support for user options in `~/.config/adios2/adios2.ini`
Uses `https://github.com/benhoyt/inih` as thirdparty
Currently supported options:

```
[General]
verbose = 0

[Campaign]
active = true
hostname = LAP131864
campaignstorepath = ~/dropbox/adios-campaign-store
cachepath = /tmp/campaign
verbose = 0

[SST]
verbose = 0
```

~ is processed to `$HOME` (`%HOMEPATH%` on Windows) (but only in select path options)
Usage: every engine has `const INIOptions &m_UserOptions` member available and can go from there.
Other parts of the code can call `static const INIOptions ADIOS::GetUserOptions()` and go from there.


pnorbert:
Note: Need to add the support to the campaign python script before this can be merged.

@vicentebolea  Please review the thirdparty business if I did it right. The inih code compiled separately but objects are added into `libadios2_core`, like `pugixml`
@eisenhauer Please review the usage of the user options. I added a verbose flag for SST, but this is just for the engine. I don't know how to propagate this to the CP to use instead of the `SstVerbose` environment variable.

eisenhauer:
You set Params.verbose in SST, which if it is used to set CPVerbosityLevel if it is higher than what is in an environment variable.  So, you're good from that point of view.  What is a little fuzzy is that we have both CP and data plane verbosity, and the latter isn't well developed.

pnorbert:
Note to self: user options are set to default only when the adios2.ini exists and is being parsed but a recognized option is not present. Need to properly initialize the INIOptions struct.

vicentebolea:
Great work @pnorbert, this is will ease the usage of the campaign files features.

In a more higher level conversation, I can see the appeal of using ini files for config files. I also see that we already include a json, a xml and a yaml library as a thirdparty library. I wonder: would it make sense to reuse any of those libraries instead of adding a new lib?

I would vote yaml of course :)

pnorbert:
@vicentebolea Convinced.

PR #4116, opened 2024-03-29
EVpath upstream to make NO_RDMA more robust

eisenhauer:


anagainaru (approved):
My only concern with the changes in the test file is that we don't use this patterns of naming files "_MPI.bp" (when MPI is used) anywhere else.

eisenhauer:
So, yes, Norbert and were having a conversation in the background that the naming is probably the cause of the random failures and that we need to do a new PR making sure that the output filenames change based on Serial or MPI.  Will do that next, but wanted to get this in first...   Thanks!

PR #4119, opened 2024-03-30
Add support for user options in ~/.config/adios2/adios2.yaml

pnorbert:
Currently supported options:
```
General:
  verbose: 0

Campaign:
  active: true
  hostname: LAP131864
  campaignstorepath: ~/dropbox/adios-campaign-store
  cachepath: /tmp/campaign
  verbose: 0

SST:
  verbose: 0
```
`~` is processed to `$HOME` (`%HOMEPATH%` on Windows) (but only in select path options) 
Usage: every engine has `const UserOptions &m_UserOption` member available and can go from there. 
Other parts of the code can call `static const UserOptions ADIOS::GetUserOptions()` and go from there.

Make Campaign AUTO (on when sqlite3 and zlib are found) - still the campaign manager can be deactivated in adios2.yaml


pnorbert:
Note: don't merge until the campaign manager python script uses the same config

eisenhauer (approved):

pnorbert:
I will change the version from 1.0 to the more honest 0.1 version, and add some doc, then merge the commits into one, before merging this PR.

pnorbert:
Well, I need this in the master to write doc with working example. Fixes and doc come in separate PR

PR #4121, opened 2024-04-01
DataPlane Configuration changes

eisenhauer:
This PR does a couple of things:  

1. Change the RDMA transport verbosity in top-level CMake to print a list of possibly available SST DataPlanes rather than simply say "Available" or "Unconfigured"
2. Allows specification of the CMake variable ADIOS2_SST_EXPECT_MPI_DP, which will cause CMake to terminate with an error if the tests for building the MPI dataplane show it unavailable.
3. Allows specification of the CMake variable ADIOS2_SST_BUILD_MPI_DP, which will cause CMake to build the MPI DataPlane without checking to see if Client/Server functionality works.

Number 3 above is necessary on Frontier because MPI client/server works on compute nodes, but not on the login nodes where the user is likely to run cmake.  Currently only doing this for the MPI dataplane because its use is the most conditional upon build conditions.


guj:
All three worked as expected! Thanks Greg

guj (dismissed):

vicentebolea:
@eisenhauer the user can specify `cmake -DADIOS2_HAVE_MPI_CLIENT_SERVER=ON` to bypass/override the configure check. We do not need to add a new cmake variable for this. I should have documented this better.

guj:
For me having  MPI_DP in the name instead of MPI_CLIENT_SERVER is easier to guess and detect.

vicentebolea:
I agree with @guj, while ADIOS2_HAVE_MPI_CLIENT_SERVER speaks of the mpi capabilities that we test, it should be named as what it enables since this is a user flag. I vote for renaming `ADIOS2_HAVE_MPI_CLIENT_SERVER` to `ADIOS2_HAVE_MPI_DP` we should also document this somewhere

pnorbert (dismissed):

CMakeLists.txt
```diff
@@ -443,9 +443,21 @@ foreach(opt IN LISTS ADIOS2_CONFIG_OPTS)
...
+     set (HPCDataPlaneList "${HPCDataPlaneList} fabric")
+  endif()
+  if (ADIOS2_SST_HAVE_UCX)
+     set (HPCDataPlaneList "${HPCDataPlaneList} UCX")
+  endif()
+  if (ADIOS2_SST_BUILD_MPI_DP)
+     set (HPCDataPlaneList "${HPCDataPlaneList} MPI")
+  endif()
+endif()
+if ({HPCDataPlaneList STREQUAL  "")
```
vicentebolea, line 458: typo here.

eisenhauer: Thanks, fixing...

pnorbert (approved):

PR #4125, opened 2024-04-02
Try 32bits in windows

vicentebolea:
cc @eisenhauer


eisenhauer:
Hmm...   Didn't end up showing the errors that Axel was getting...  Wonder why...

vicentebolea:
@eisenhauer our 32bits got new errors :tada:

eisenhauer:
> @eisenhauer our 32bits got new errors 🎉

Oddly, still not the warnings that Axel reported, but different ones.  But it's a start, I'll poke at these.

ax3l:
In case we have the chance to get this merged, still appreciate the coverage :)

eisenhauer:
The challenge unfortunately is far beyond getting this to compile on 32-bits.  I suspect that the assumption that size_t is corresponds to 64-bits may be deeply embedded and may be quite difficult to sort out.  Merging this before sorting that would be problematic.  Getting ADIOS to the point where it runs cleanly on both 32 and 64 bit systems might be even more so.

vicentebolea:
Closing this since this work would require a much greater amount of engineering (compatibility between bp file produced in different arch for instance)

PR #4132, opened 2024-04-04
Compile FileHTTP on Windows

eisenhauer:


dmitry-ganyushin (commented):
For FreeBSD UNIX=TRUE. Network init should be probably rewritten

dmitry-ganyushin:
That code is not ready for windows. One solution would be to shift all
network initialization to the curllib library which works on most
platforms.

eisenhauer:
It's OK, I got this...

pnorbert (approved):

PR #4137, opened 2024-04-09
Quick fix for the issue 4131. Included only for linux.

dmitry-ganyushin:
FileHTTP.cpp is include in build only for linux until further fix.

### DIFF
--- a/source/adios2/CMakeLists.txt
+++ b/source/adios2/CMakeLists.txt
@@ -229,7 +229,7 @@ target_compile_features(adios2_core PUBLIC "$<BUILD_INTERFACE:${ADIOS2_CXX11_FEA
 
 
 target_sources(adios2_core PRIVATE toolkit/transport/file/FilePOSIX.cpp)
-if(UNIX) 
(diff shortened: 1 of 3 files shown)


eisenhauer:
Generally I think we’d rather move in the direction of making the code portable rather than excluding certain functionality from certain platforms.  That’s a slippery slope with an unpleasant end.

dmitry-ganyushin:
That is only a quick fix to make build for FreeBSD (untested on FreeBSD)

eisenhauer:
Dmitry, I'm going to close this because PR #4138 takes care of this problem by making the file in question compile on FreeBSD.

dmitry-ganyushin:
Auto tests are missing. How to test functionality?

eisenhauer:
> Auto tests are missing. How to test functionality?
On step at a time.  PR #4138 just adds a #include so the build isn't broken on FreeBSD.  I'll do windows too because that's pretty quick and easy.  Adding tests and other mods will probably wait until we have a performant use case.  For now, just trying to keep this code around in the anticipation that we might have such a case in the future.

PR #4135, opened 2024-04-04
pugixml: correct target name

vicentebolea:
fixes: #4081

PR #4142, opened 2024-04-17
pugixml,cmake: fix linkage when pugixml is external

vicentebolea:
followup of https://github.com/ornladios/ADIOS2/pull/4135


chenrui333:
patch works for me, adopting it in Homebrew/homebrew-core#169547

pnorbert (approved):

PR #4154, opened 2024-04-27
Upstream EVPath and SST updates to make the Visual Studio compiler happy.

eisenhauer:


eisenhauer:
@anagainaru @pnorbert SST tests are not yet running on Windows, but SST does compile cleanly in both Linux and Windows now.  I'd like to get this merged as an intermediate step and I'll work on the rest later.

anagainaru (approved):
Looks reasonable, I have no way of testing this more than the CI so we might as well merge it.

PR #4158, opened 2024-05-08
Compress the expression string into one line for `bpls`

lizdulac:
Related to PR #4145

Create a function that returns a single line string representation of a derived expression. This is mainly for `bpls` since one line has a more user friendly format. 

ExpressionTree to String Expression:
`ADD({sim1/Ux},{sim1/Uy},{sim1/Uz})`

```
$ ./bin/bpls expAdd.bp --show-derived -l
  float    derived/addU  2*{10, 3, 6} = 2.98842 / 28.085
    Derived variable with expression: ADD({sim1/Ux},{sim1/Uy},{sim1/Uz})
  float    sim1/Ux       2*{10, 3, 6} = 0.000224775 / 9.94031
  float    sim1/Uy       2*{10, 3, 6} = 0.0201616 / 9.83318
  float    sim1/Uz       2*{10, 3, 6} = 0.0031 / 9.95085
```

The current solution is not the best. We do not create derived variables on the read side so the BP5Reader provides an internal function to get the expression string from the deserializer, create an expression on it, and call the pretty-print function on the expression (which is overly complicated). I would still merge this PR like this. We will start creating derived variables on the read side and we can replace the logic with something cleaner at that point. This PR is one step in this direction since it moves the string conversion on the Expression class.

### DIFF
--- a/source/adios2/toolkit/derived/parser/lexer.l
+++ b/source/adios2/toolkit/derived/parser/lexer.l
@@ -149,6 +149,7 @@ void
 adios2::detail::ASTDriver::parse (const std::string input)
 {
   adios2::detail::parser parse (*this);
+  yy_flex_debug = trace_scanning;
   yy_scan_string(input.c_str());
   parse.set_debug_level (trace_parsing);
   parse ();
--- a/source/adios2/core/Engine.h
+++ b/source/adios2/core/Engine.h
@@ -482,7 +482,7 @@ public:
         return false;
     }
 
-    virtual const char *VariableExprStr(const VariableBase &) { return NULL; }
+    virtual std::string VariableExprStr(const VariableBase &) { return NULL; }
 
     /** Notify the engine when a new attribute is defined. Called from IO.tcc
      */
--- a/source/adios2/engine/bp5/BP5Reader.cpp
+++ b/source/adios2/engine/bp5/BP5Reader.cpp
@@ -751,9 +751,18 @@ bool BP5Reader::VariableMinMax(const VariableBase &Var, const size_t Step, MinMa
     return m_BP5Deserializer->VariableMinMax(Var, Step, MinMax);
 }
 
-const char *BP5Reader::VariableExprStr(const VariableBase &Var)
+std::string BP5Reader::VariableExprStr(const VariableBase &Var)
 {
-    return static_cast<const char *>(m_BP5Deserializer->VariableExprStr(Var));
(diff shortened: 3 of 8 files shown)


lizdulac:
Translating an ExpressionTree object instead of an ASTNode object to a single line string expression allows for expressions with the same operator to be merged.

Example:
ASTNode to String Expression:
+( +( {sim1/Ux},{sim1/Uy} ),{sim1/Uz} )

ExpressionTree to String Expression:
ADD({sim1/Ux},{sim1/Uy},{sim1/Uz})

source/adios2/toolkit/derived/parser/lexer.l
```diff
@@ -149,6 +149,7 @@ void
 adios2::detail::ASTDriver::parse (const std::string input)
 {
   adios2::detail::parser parse (*this);
+  yy_flex_debug = trace_scanning;
```
anagainaru, line 152: Do we need to keep this flag in the final version?

source/adios2/toolkit/derived/parser/ASTNode.cpp
```diff
@@ -76,6 +76,43 @@ std::string ASTNode::printpretty(std::string indent)
     return result;
 }
 
+std::string ASTNode::node_to_string_expr()
```
anagainaru, line 79: Since we have a way of printing the Expression string, do we need this at the `ASTNode` level as well?

lizdulac: No you're right, we can replace the ASTNode::node_to_string_expr function with the ExpressionTree function. I've removed the ASTNode function

source/adios2/toolkit/derived/parser/lexer.l
lizdulac: This allows for dynamically setting debug flags without rebuilding the parser (set in ASTDriver). If we don't want this option, we should also remove the boolean members of ASTDriver, but I would argue for leaving them for now while we extend the parser with features like magnitude(curl(x,y,z), 2)

lizdulac: I may have misunderstood, this line should have been in this function from the start, it's the equivalent of the later line:
parse.set_debug_level (trace_parsing);

but for the scanner instead of the parser. This is just relaying the trace flag set in ASTDriver

eisenhauer (approved):

PR #4160, opened 2024-05-10
Remove special ReadFlattenSteps Mode, switch to control by writer-sid…

eisenhauer:
…e FlattenSteps parameter


docs/user_guide/source/engines/bp5.rst
```diff
@@ -156,6 +164,9 @@ This engine allows the user to fine tune the buffering operations through the fo
  StatsLevel                     integer, 0 or 1       **1**, 0
  MaxOpenFilesAtOnce             integer >= 0          **UINT_MAX**, 1024, 1
  Threads                        integer >= 0          **0**, 1, 32
+ FlattenSteps                   boolean               **off**, on, true, false
+ IgnoreFlattenSteps             boolean               **off**, on, true, false
+ Threads                        integer >= 0          **0**, 1, 32
```
anagainaru, line 169: The `Threads` line doesn't seem connected to the rest

anagainaru (commented):
Looks reasonable. So now the option of treating steps as flat (or ignoring this) is given by the writer (or reader) through macros and not part of the API?

docs/user_guide/source/engines/bp5.rst
eisenhauer: Whoops, yes, that line somehow got duplicated.  Will push an update.

eisenhauer:
> Looks reasonable. So now the option of treating steps as flat (or ignoring this) is given by the writer (or reader) through macros and not part of the API?

Yes, pretty much.  While the actual functionality of loading all the steps at once and presenting as a single step is purely reader-side, @ax3l wanted it to be controlled by the writer.  So all that happens on the writer side is we put the FlattenSteps flag into a new file header field.  On the reader side we pull the flag and use that instead of a new read mode.  The IgnoreFlattenSteps was as much for bpls as anything, so we could see the file as written.

anagainaru (approved):

PR #4162, opened 2024-05-12
Reorg Remote as base class in preparation for more remote subclasses

eisenhauer:
No functional changes here.  Just reshuffling the code so that we can subclass remote for various underlying protocols.  This doesn't change any interactions with the current EVPath based remote adios server, but it sets the stage to have an Xroot-based remote protocol implementation without overly disrupting the places these calls are made.


anagainaru (approved):
I didn't realize initially you are just reorganizing things and the `EVPathRemote` looked very familiar haha

PR #4163, opened 2024-05-13
Update CURL implementation to allow the compiler to optimize the code

anagainaru:
I compared the performance of the derived functions to codes outside ADIOS doing the same thing and CURL was 8x slower while magnitude and add were only 1.4x slower. I changed the code to a version that allows the compiler to optimize the linear interpolation. All codes are now 1.4x slower than kernels outside ADIOS (I assume this is because memory allocation for the new variable, but I will have to check further).

Breakdown performance for magnitude and curl:
Screenshot 2024-05-13 at 9 53 49 AM

Legend:
- **Pre-process** measures the code that gets the variable data from Greg's structures
- **Compute**  measures the actual derived function, including allocating data for the new variable (this is 1.4x slower than  computing derived function on preallocated buffers)
- **Post-process** measures creating new ADIOS variables and min/max for it
- **Store data** is the extra time between the two types of derived variables `metadata_only` and `store_data`

All tests are for magnitude(x,y,z) and curl(x,y,z) with x,y,z 3 ADIOS 3D variables of size NxNxN. The horizontal ax show the array size in MB for one of these 3D variables. For e.g. 1GB means x, y, z are 650x650x650.

The time difference between curl and magnitude for the post process/store data phases is due to the fact that CURL creates a 3xNxNxN variable while magnitude creates a NxNxN variable.


eisenhauer (approved):
Seems reasonable...

PR #4080, opened 2024-03-07
WIP. Integration of xrootd client and xrootd adios plug-in 

dmitry-ganyushin:
Initial commit.

PR #4164, opened 2024-05-14
Dmitry's Xrootd implementation, with minor updates

eisenhauer:
This is a replacement for #4080, with minor updates for subclassing the new Remote base class, etc.  This needs significant further work, but would like to get his merged so limit the scope of future updates.


anagainaru (approved):
I am not an expert, but I don't see any issues with your changes

PR #4167, opened 2024-05-17
Basic Xrootd remote fixes

eisenhauer:
Change all RemoteHandle types to void * and make them appear only in the base class, change Xroot request/response sync to use future/promises, fix server-side type handling to use the right size for the type.  These changes are sufficient to pass a the simple TestCommonRead test using Xrootd remote server.  (No actual testing in CI yet, so you'll have to take my word for it.)

### DIFF
--- a/examples/hello/hdf5SubFile/hdf5SubFile.cpp
+++ b/examples/hello/hdf5SubFile/hdf5SubFile.cpp
@@ -14,7 +14,9 @@
 #include <iostream> //std::cout
 #include <mpi.h>
 #include <stdexcept> //std::invalid_argument std::exception
+#ifndef _MSC_VER
 #include <unistd.h>
+#endif
 #include <vector>
 
 void writeMe(adios2::IO &hdf5IO, int rank, int size, const char *testFileName)
--- a/examples/hello/hdf5SubFile/hdf5SubFile.cpp
+++ b/examples/hello/hdf5SubFile/hdf5SubFile.cpp
@@ -36,8 +36,8 @@ void writeMe(adios2::IO &hdf5IO, int rank, int size, const char *testFileName)
     const std::size_t Nx = 1024;
     const std::size_t Ny = 1024 * scale;
 
-    std::vector<float> myFloats(Nx * Ny, 0.1 * rank);
-    std::vector<int> myInts(Nx * Ny, 1 + rank);
+    std::vector<float> myFloats(Nx * Ny, 0.1f * rank);
+    std::vector<int> myInts(Nx * Ny, (int)(1 + rank));
 
     hdf5IO.SetParameter("IdleH5Writer",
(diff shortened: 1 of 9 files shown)


eisenhauer:
@guj, some minor hdf5 file tweaks here.  Something maybe changed in the CI images, because we clearly didn't used to compile these files with Visual Studio on Windows before and now we are.  Things like #include "unistd.h" had to change.

guj:
> @guj, some minor hdf5 file tweaks here. Something maybe changed in the CI images, because we clearly didn't used to compile these files with Visual Studio on Windows before and now we are. Things like #include "unistd.h" had to change.

Thanks Greg.

examples/hello/hdf5SubFile/hdf5SubFile.cpp
```diff
@@ -14,7 +14,9 @@
 #include <iostream> //std::cout
 #include <mpi.h>
 #include <stdexcept> //std::invalid_argument std::exception
+#ifndef _MSC_VER
 #include <unistd.h>
+#endif
```
anagainaru, line 19: I was wondering why the windows test was failing for Liz's PR. I wonder how this was not an issue before

source/adios2/engine/bp5/BP5Reader.cpp
```diff
@@ -299,7 +299,7 @@ void BP5Reader::PerformGets()
         if (getenv("DoXRootD"))
         {
             m_Remote = std::unique_ptr<XrootdRemote>(new XrootdRemote());
-            m_Remote->Open("localhost", 1049, m_Name, m_OpenMode, RowMajorOrdering);
+            m_Remote->Open("localhost", 1094, m_Name, m_OpenMode, RowMajorOrdering);
```
anagainaru, line 302: [optionally] I assume this is the port, should we create a variable for it so it's clear what it is?

anagainaru (approved):

source/adios2/engine/bp5/BP5Reader.cpp
eisenhauer: Yes, there are a number of things like that that I'm assuming that we'll get from the campaign-level stuff, dealing with multiple hosts and multiple configurations.  For now just hardcoding those things until I get the actual functionality in shape.

examples/hello/hdf5SubFile/hdf5SubFile.cpp
eisenhauer: Not sure, but fortunately not too difficult a fix.  Probably could have just killed the unistd include, but there were further warnings because visual studio is anal about type conversion warnings.

anagainaru: I rebased Liz's PR over this one and now it passes the windows CIs :)

PR #4169, opened 2024-05-21
Add API call to reset memory selection

franzpoeschel:
I'm currently adding support for memory selections to the openPMD-api on https://github.com/openPMD/openPMD-api/pull/1620. 
While working on it, I noticed that the selection specified in `Variable<T>::SetMemorySelection(...)` sticks to **all** subsequent `Put()` calls, without a  way to remove the memory selection again.

Is it intended that the setting will stick to the Variable or is that a bug? If it is intended, this PR adds a way to remove the setting again. Otherwise, this PR should be discarded and instead, the memory selection should be removed automatically after calling `Put()`.

cc @pnorbert

### DIFF
--- a/bindings/CXX11/adios2/cxx11/Variable.h
+++ b/bindings/CXX11/adios2/cxx11/Variable.h
@@ -214,7 +214,7 @@ public:
      * variable.Count() = {Ny,Nx}, then memoryCount = {Ny+2,Nx+2}
      * </pre>
      */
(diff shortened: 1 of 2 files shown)


pnorbert:
Thank you for the fix. Indeed, the selection is sticky, at least, that's how the gray-scott example is using it. It defines the variable and sets the memory selection once, and then this is used for all steps. Never occured to me enable a reset.

eisenhauer (approved):

PR #4170, opened 2024-05-21
Add server-side ADIOS file pools to Xrootd

eisenhauer:


source/utils/xrootd-plugin/XrdSsiSvService.cpp
```diff
@@ -519,17 +508,17 @@ void XrdSsiSvService::ProcessRequest4Me(XrdSsiRequest *rqstP)
     else if (TypeOfVar == adios2::helper::GetDataType<T>())                                        \
     {                                                                                              \
         adios2::Variable<T> var = io.InquireVariable<T>(VarName);                                  \
-        std::vector<T> resBuffer;                                                                  \
+        std::vector<T> *resBuffer = new std::vector<T>();                                          \
```
pnorbert, line 511: Why not just use a pointer and use malloc/new?

eisenhauer: The memory handling here was kind of a mess with multiple copies before the return of data.  We need a vector to pass to Get(), but then that memory has to live after the return of this subroutine so that another thread can manage the transmission back to the client.  We can't take ownership of the buffer from the vector, so in order to avoid a copy we need a pointer to the vector in the Xrootd "session" object so that it can be deallocated when we finish transmitting data.  Here we keep it as a void* and cast it back to a std::vector<char> regardless of the original type T, which may not be kosher strictly speaking, but seems OK if all we're going to do on it after the cast is Delete.  (All of which I should probably add comments to explain.) We seem to be clean with valgrind on both the server and client side now.

pnorbert: << We need a vector to pass to Get()
Because you don't know the required size ahead?

eisenhauer: Yes, basically.  But that's a good point, I could use SelectionSize() to get the expected output size and then use malloc() to get a buffer to pass to Get() and avoid the extra cost of C++ zero-intializing the buffer before I overwrite it with data...  Should be doing that in the EVPath remote_server too...

PR #4173, opened 2024-05-23
Xrootd Server Cleanup

eisenhauer:


eisenhauer:
Use malloc'd buffers instead of std::vector to avoid zero-init prior to filling with Get().  Clean up some member names, add comments, remove some tracing, etc.

pnorbert (approved):

PR #4178, opened 2024-05-24
Close more FDs because XRootd plays games in library init

eisenhauer:
Also fix an oversight in EVPath remote

### DIFF
--- a/source/adios2/toolkit/remote/EVPathRemote.cpp
+++ b/source/adios2/toolkit/remote/EVPathRemote.cpp
@@ -171,7 +171,6 @@ EVPathRemote::GetHandle EVPathRemote::Get(char *VarName, size_t Step, size_t Blo
     GetMsg.Start = Start.data();
     GetMsg.Dest = dest;
     CMwrite(m_conn, ev_state.GetRequestFormat, &GetMsg);
-    CMCondition_wait(ev_state.cm, GetMsg.GetResponseCondition);
     return (Remote::GetHandle)(intptr_t)GetMsg.GetResponseCondition;
 }
 
--- a/source/adios2/toolkit/remote/remote_server.cpp
+++ b/source/adios2/toolkit/remote/remote_server.cpp
@@ -20,6 +20,9 @@
 #include <sys/types.h> // open
 #ifndef _MSC_VER
 #include <unistd.h> // write, close, ftruncate
+
+static int fd_is_valid(int fd) { return fcntl(fd, F_GETFD) != -1 || errno != EBADF; }
+
 #else
 #include <io.h>
 #define strdup _strdup
@@ -620,12 +623,23 @@ int main(int argc, char **argv)
         /* I'm the child, close IO FDs so that ctest continues.  No verbosity here */
         verbose = 0;
 
-        //  Why close a bunch of FDs here?  Because something in the xrootd library is opening new
-        //  fds (upon library init, so before main()) and perhaps using it to track instances?
-        //  Whatever it is doing is interfering with CTest and putting things in the background
-        for (int fd = 0; fd < 13; fd++)
+        //  Why close a bunch of FDs here?  Well, if we've linked with XRootD, stderr might have
+        //  been dup()'d in library initialization.  We don't need those FDs and we have to close
+        //  them to make sure we disassociate from the CTest parent (or else fixture startup hangs).
+        //  It doesn't seem to work to close them before the fork, so we close them afterwards.
+        for (int fd = 0; fd <= 128; fd++)
         {
-            close(fd);
+            if (fd_is_valid(fd))
+            {
+                // OK, fd is valid, should we close it?
+                if ((lseek(fd, 0, SEEK_CUR) == -1) && (errno == ESPIPE))
+                {
+                    // In the circumstances we care about (running under CTest), we want to close
+                    // FDs that are pipes.  The condition above tests for that and we should get
+                    // here only if it's a pipe.
+                    close(fd);
+                }
+            }
         }
 #endif
     }
--- a/source/adios2/toolkit/remote/remote_server.cpp
+++ b/source/adios2/toolkit/remote/remote_server.cpp
@@ -619,9 +619,14 @@ int main(int argc, char **argv)
         }
         /* I'm the child, close IO FDs so that ctest continues.  No verbosity here */
         verbose = 0;
-        close(0);
-        close(1);
-        close(2);
+
+        //  Why close a bunch of FDs here?  Because something in the xrootd library is opening new
+        //  fds (upon library init, so before main()) and perhaps using it to track instances?
+        //  Whatever it is doing is interfering with CTest and putting things in the background
+        for (int fd = 0; fd < 13; fd++)
+        {
+            close(fd);
+        }
 #endif
     }
 
--- a/source/adios2/toolkit/remote/remote_server.cpp
+++ b/source/adios2/toolkit/remote/remote_server.cpp
@@ -627,7 +627,7 @@ int main(int argc, char **argv)
         //  been dup()'d in library initialization.  We don't need those FDs and we have to close
         //  them to make sure we disassociate from the CTest parent (or else fixture startup hangs).
         //  It doesn't seem to work to close them before the fork, so we close them afterwards.
-        for (int fd = 0; fd <= 128; fd++)
+        for (int fd = 0; fd <= 16; fd++)
         {
             if (fd_is_valid(fd))
             {


source/adios2/toolkit/remote/remote_server.cpp
```diff
@@ -619,9 +619,14 @@ int main(int argc, char **argv)
...
         /* I'm the child, close IO FDs so that ctest continues.  No verbosity here */
         verbose = 0;
-        close(0);
-        close(1);
-        close(2);
+
+        //  Why close a bunch of FDs here?  Because something in the xrootd library is opening new
+        //  fds (upon library init, so before main()) and perhaps using it to track instances?
+        //  Whatever it is doing is interfering with CTest and putting things in the background
+        for (int fd = 0; fd < 13; fd++)
```
anagainaru, line 626: How do you know how many fds to close? Is there an API in xrootd that can tell you this number?

source/adios2/toolkit/remote/EVPathRemote.cpp
```diff
@@ -171,7 +171,6 @@ EVPathRemote::GetHandle EVPathRemote::Get(char *VarName, size_t Step, size_t Blo
     GetMsg.Start = Start.data();
     GetMsg.Dest = dest;
     CMwrite(m_conn, ev_state.GetRequestFormat, &GetMsg);
-    CMCondition_wait(ev_state.cm, GetMsg.GetResponseCondition);
```
anagainaru, line 174: Was this wait here to block until the write is done?

anagainaru (commented):
I am trying to understand what the code is doing, I can approve if you want it merged

source/adios2/toolkit/remote/remote_server.cpp
eisenhauer: Knowing exactly what was going on (and having it documented) would be nice, but I'd be pretty sure that this is undocumented behaviour.  We're not even calling xrootd subroutines, but apparently library initialization gets you a bunch of open FDs.  This is normally inconsequential, but there is some kind of odd interaction with CTest here and our normal routine of having our remote server fork itself to the background in order to run as a "fixture" in CI doesn't work.  I'd previously observed this with FDs 0,1 and 2, but the Xrootd behaviour seems to expand this list.  We're only doing this in a program that will never use xrootd itself, so this seems safe, albeit weird.  (And I got this number by doing an open() and seeing what was the next FD available.  This could change as xrootd changes, but we'll know because tests that don't use xrootd will start failing.  I guess I could make this configurable, or close the first 128 or something, but lets see how this goes.)

pnorbert:
@abh3 Do you have any idea what is goin on here?

abh3:
Not at all and certainly xrootd is not at fault here. The linker obviously is going to open loads of FD's to service library dependencies during library initialization. That is completely normal and I don't see why it should interfere with CTest. Anyway, closing arbitrary FD's without knowing why they were opened is surely going to cause more problems than a CTest failure.

eisenhauer:
> Not at all and certainly xrootd is not at fault here. The linker obviously is going to open loads of FD's to service library dependencies during library initialization. That is completely normal and I don't see why it should interfere with CTest. Anyway, closing arbitrary FD's without knowing why they were opened is surely going to cause more problems than a CTest failure.

I'll get you more info when I can.  This is a program (an ADIOS data server) that uses ADIOS, but not doesn't call any Xrootd client routines, but it behaves differently when ADIOS is linked with those Xrootd routines available than when it is not (even when not called).  Closing some stdout and stderr are necessary when the server backgrounds itself when running as a fixture under CTest.  If the server doesn't do that CTest thinks the subcommand hasn't finished.  Somehow more get opened when linked with xrootd and closing them resolves a problem.  I need to see if I can do some tracing to see exactly when those get opened.  (I'm also trying to sort what libraries we're really pulling in, if those are just client libraries, etc.  Also need to have cmake find the xrootd executable so we can test what we're doing in our own CI, but having linking with xrootd not break the existing setup is the first step.

eisenhauer:
So, I added a couple of lines to our (non-xrootd) remote data server right after main() like this:
```
{
 . . .
    int nextfd = open("/dev/tty", O_WRONLY);
    printf("Next available fd after main was %d\n", nextfd);

```
When this program is run without xrootd linked in to ADIOS, the output is as one might expect:
```
eisen@Endor build % bin/adios2_remote_server
Next available fd after main was 3

```
I.E. stdin, stdout and stderr were open upon entry to main, so the next available FD was 3.  But when ADIOS is linked with xrootd, this is the output:
```
eisen@Endor build % bin/adios2_remote_server 
Next available fd after main was 13
```
I suspected that something might be going on in library initialization and in particular with logging, so I ran this in a debugger with breakpoints set at likely candidates (open(), dup(), dup2(), etc.).  This led me to:
```
XrdSysFD.hh:
   92  	inline int  XrdSysFD_Dup(int oldfd)
   93  	                 {int newfd = dup(oldfd);
   94  	                  if (newfd >= 0) fcntl(newfd, F_SETFD, FD_CLOEXEC);
   95  	                  return newfd;
   96  	                 }

```
This is called 7 times, each time during the constructor of a XrdSysLogger object during xrootd library initialization and each time calling dup on stderr.  (There are a couple of other opens during network initialization that account for some of the other FDs.). 

So we have a lot of instances where stderr is open on an FD other than 2.  Normally this isn't a problem, but it caused an issue because of the particular way we ran our adios2_remote_server as a ctest fixture in our test suite.  Fixtures need a "setup" command, which is run like a normal test, but which in this case must startup a server that will run in the background (and will exit when the "shutdown" command is run).  For adios2_remote_server, we had a -background flag we used here that simply had it fork() itself, the child continued in the background while the parent exited so that the "setup" test would exit.  However because ctest is trying to read the stdout and stderr or all its tests, it does more than just look for the test process exiting, but also looks to get an EOF on the stdout and stderr of that test.  Because those FDs are inherited by the child of the test, CTest doesn't get an EOF on them unless the child closes them too.  That's why we were originally closing FDs 0,1, and 2.   What xrootd libraries dup'ing stderr upon initialization changed for us is now stderr has to be closed a lot more places before CTest gets that EOF.  

One thing to keep in mind is that this process (and ADIOS in this circumstance) has not and will not call any xrootd subroutines.  So while I'm certain that us closing these FDs would likely mess up XrdSysLogger, that's not a particular concern for us because it'll never be active anyway.  It's just that the extra 7 dup's of stderr and their interaction with ctest were unexpected.

That said, I might have to change how we start our servers in CI anyway.  I'd like to be checking adios xrootd functionality in CI, which means starting up an xrootd server and shutting it down on demand. It looks like there's a -b argument for xrootd, and a mechanism for putting the server PID in a file (which will help with shutdown, maybe with -HUP?).  But I don't know yet if we'd face the same stderr issues we ran into with our remote server, and if we do I can't modify xrootd to fix it in quite the same way.

source/adios2/toolkit/remote/EVPathRemote.cpp
eisenhauer: Yes, but we were inconsistent.  The Get in EVPathRemote returned a GetHandle, which we could have Waited on, but instead we waited in the Get and never used the returned GetHandle.  This didn't matter much because there was no concurrency on the server side.  But there is possible concurrency on the Xrootd server side, so I've changed BP5Reader to actually send all the Gets and then wait on all the handles.  But I forgot to change the EVPath remote to match that.  This change is all that is required to correct that.

abh3:
If at all possible, it would be helpful that at each breakpoint you could include a "where", so we know where this call is actually coming from. Clearly, we did not expect that many "dups" to occur, so it is mystifying. Additionally, review of the code indicates that none of those could occur from a static initialization. So, we are doubly mystified.

Andy

________________________________

eisenhauer:
Not a problem, though at the moment I've been working in an OSX development environment, so it's an lldb log.  I saw the same CTest-hanging-on-server-startup behavior on Ubuntu, and will get you a log from there when I get a chance as well.  
[lldb_log.txt](https://github.com/ornladios/ADIOS2/files/15447131/lldb_log.txt)

eisenhauer:
Looking at this a little bit more, I'd be suspicious of the static XrdSysLogger entries in various cpp files.

XrdCryptoAux.cpp: 
```
// For error logging and tracing
static XrdSysLogger Logger;
```

The constructor is going to cause the dup...

abh3:
Good point! Old code from the dim dark past, sigh. I'll see what we can do about it.

Andy
________________________________

pnorbert (approved):

eisenhauer:
> Good point! Old code from the dim dark past, sigh. I'll see what we can do about it. Andy

@abh3 Well, coming up on two years later I finally got around to submitting a fix for this problem to xrootd.  Wasn't quite sure of the PR protocol there, so it may need adjustment, but at least it's contributed back.

PR #4196, opened 2024-06-05
cmake: add ADIOS2_LIBADIOS_MODE flag

vicentebolea:
We need this for package managers where we first need to install c/c++ components for the first package (libadios2) and later for the adios python/adios binaries package we need to install the remaining components.


eisenhauer (approved):
Looks workable...

ax3l:
Since we need this for conda releases (e.g., https://github.com/conda-forge/adios2-feedstock/pull/104), is there a reason this was intentionally not added to 2.10.2 or do I misread the status?

PR #4197, opened 2024-06-07
Dill upstream in master

eisenhauer:
Make Dill fall back to emulation on architectures where we don't have native CG.


vicentebolea:
Is there a reason not to use the `thirdparty/dill/update.sh` script for the update?

eisenhauer:
I believe it was used for this update...

vicentebolea (approved):
My bad, on other projects we update that update.sh script  with the commit but not here

PR #4205, opened 2024-06-18
Fixing read side json profiling

guj:
Applies to BP5 Reader.


eisenhauer:
Hi Junmin.  Unfortunately, if you look at the "files changed" part here, you'll see that this PR changes much more than what you intended.  Probably this is because you merged master in to your branch rather than rebasing.  You might have to start off with a fresh branch because I'm not sure what the operant change here is...

guj:
Ah, interesting.  
I will rebase and do a new pull request. Thanks. 
> Hi Junmin. Unfortunately, if you look at the "files changed" part here, you'll see that this PR changes much more than what you intended. Probably this is because you merged master in to your branch rather than rebasing. You might have to start off with a fresh branch because I'm not sure what the operant change here is...

PR #4209, opened 2024-06-21
include <functional> to be able to compile std::function<> with GCC c…

pnorbert:
…ompilers


pnorbert:
How did PRs get into master so that I cannot compile it with either gcc 9.4 or 10.5?

anagainaru (commented):
This shouldn't be required. Do you use the latest files that I pushed yesterday?

anagainaru (approved):
Yep I confirm this was a mistake, it seems my compiler doesn't need the header when I tested it.

PR #4210, opened 2024-06-24
ADIOS2 User Side KVCache to accelerate the remote query access.

Change72:
### Structure
In KVCache v1.0, we use Redis (community version) as the local cache server, and hiredis as the local cache client for C++.

### Build and Use
Users can follow the build example scripts (scripts/build_scripts/build-adios2-kvcache.sh) to turn on KVCache (by default is off). After the build, the user needs to start the remote server and cache server by 'source scripts/build_scripts/build-adios2-kvcache.sh --start'

### Features
In KVCache v1.0, we support partially overlapped cases. We can utilize the existing cache to reduce network traffic even though the new request only partially overlaps with existing cached results. Users can also set the depth of overlap searching, by changing the environment variable: maxDepth (in source/adios2/engine/bp5/BP5Reader.cpp).

Redis provides auto-persistence, which means Redis will flush all the data to disk when it exits and re-load to memory when it restarts. Users can adjust Redis memory usage limit and other configurations in Redis configuration files.


eisenhauer:
Before this can go through testing and be merged, your code has to be formatted according to our clang-format rules (which requires running the right version of clang-format).  If you are on a machine with docker installed, the easiest way to accomplish this is to cd into your ADIOS2 directory and run "./scripts/developer/run-clang-format.sh".  This will run docker and download an image with the appropriate version of clang-format.  It'll generate a lot of "diff" output, but what's important for you is that it leaves the re-formatted files in place, so you can just do another git commit and all should be well.  Note that the docker image is x86-based, but I've found that for this simple use even works on a Apple silicon if you're on a MacBook...

cmake/DetectOptions.cmake
```diff
@@ -606,6 +606,21 @@ elseif(ADIOS2_USE_Campaign)
   endif()
 endif()  
 
+# KVCache
+if(ADIOS2_USE_Cache STREQUAL AUTO)
+    find_package(hiredis REQUIRED)
```
eisenhauer, line 611: I think this use of find_package() should not have the REQUIRED keyword.

vicentebolea: I second this, when ADIOS2_USE_Cache is AUTO we should not use required. REQUIRED is only with =ON

Change72: ok I have removed it and finished formatting by docker

eisenhauer: Ah, formatting is OK except for shell scripts.  It's picky about shell scripts and as far as I know those have to be fixed manually.  Unless you know an automatic way @vicentebolea ?  @Change72 click on "details" next to the format line below to see what it is complaining about.

source/adios2/CMakeLists.txt
```diff
@@ -247,7 +247,20 @@ if (ADIOS2_HAVE_SST)
...
+if (ADIOS2_HAVE_KVCACHE)
+  target_sources(adios2_core PRIVATE  toolkit/cache/KVCacheCommon.h
+    toolkit/cache/KVCacheCommon.tcc)
+  target_link_libraries(adios2_core PRIVATE hiredis::hiredis)
+
+  # Ensure the rpath is set correctly
+  # message(STATUS "PACKAGE_PREFIX_DIR: ${PACKAGE_PREFIX_DIR}")
+  set_target_properties(adios2_core PROPERTIES
+    BUILD_RPATH "$ORIGIN:${PACKAGE_PREFIX_DIR}/lib"
+    INSTALL_RPATH "$ORIGIN:${PACKAGE_PREFIX_DIR}/lib"
```
vicentebolea, line 260: Why is this needed?

CMakeLists.txt
```diff
@@ -265,7 +265,7 @@ set(ADIOS2_CONFIG_OPTS
     DataMan DataSpaces HDF5 HDF5_VOL MHS SST Fortran MPI Python PIP Blosc2 BZip2
     LIBPRESSIO MGARD MGARD_MDR PNG SZ ZFP DAOS IME O_DIRECT Sodium Catalyst SysVShMem UCX
     ZeroMQ Profiling Endian_Reverse Derived_Variable AWSSDK XRootD GPU_Support CUDA Kokkos
-    Kokkos_CUDA Kokkos_HIP Kokkos_SYCL Campaign
+    Kokkos_CUDA Kokkos_HIP Kokkos_SYCL Campaign KVCACHE
```
vicentebolea, line 268: need to add as at adios2_option in this file.

cmake/DetectOptions.cmake
vicentebolea: @eisenhauer is correct, CI uses shellcheck for ensuring best practices for shell scripts, we do not force any formatting/indentation style for shell scripts.

source/adios2/CMakeLists.txt
Change72: Otherwise, in '~/project/ADIOS2-Source$ ldd build-cache-test/bin/adios2_remote_server', it will show:
libhiredis.so.1.1.0 =>  not found

and user needs to manually set LD_LIBRARY_PATH by 'export LD_LIBRARY_PATH=${SW_DIR}/adios2/lib:${SW_DIR}/hiredis/lib:$LD_LIBRARY_PATH'

.gitignore
```diff
@@ -44,3 +44,13 @@ CMakeSettings.json
 # Python wheels stuff
 
 *.egg-info/
+
+# CMake generated files
+build/
+build-cache*/
+cmake-build-*/
```
vicentebolea, line 51:
```suggestion
```

source/adios2/CMakeLists.txt
Change72: ${PACKAGE_PREFIX_DIR} refers to hiredis installation path and set in hiredis cmake.

scripts/build_scripts/build-adios2-kvcache.sh
```diff
@@ -0,0 +1,110 @@
...
+build_cache() {
+    # redis - in-memory data structure store
+    redis_dir=${SW_DIR}/redis
+    if [ ! -d ${redis_dir} ]
+    then
+        git clone https://github.com/redis/redis.git ${redis_dir}
+        cd ${redis_dir}
+        git checkout tags/7.2.3
+        # cannot accleerate by 'make -j8'. It will cause error.
+        make
```
vicentebolea, line 37: Use CURL/wget rather than git here

source/adios2/CMakeLists.txt
vicentebolea: I understand the need of rpath, what I do not is why setting it manually. In cmake this is handled by cmake global cmake flags, you do not need to set it yourself unless there is a very particular issue.

The thing is that you do not set the rpath to your particular external dependency, you use other means for this (LD_library_path is the slowest one).

The problem with this is that doing that is not portable since in a binary distribution this assumption of where to find the particular library will fail since the user might have its library in a different dir than the directory that was uses when building adios2 (eg: a package manager case).

I am afraid that you will either need to setup correctly your LD_LIBRARY_PATH or install your dependencies in a place that the linker can find it.

scripts/build_scripts/build-adios2-kvcache.sh
```diff
@@ -0,0 +1,110 @@
+#!/bin/bash
```
vicentebolea, line 1: A future idea is that this could be a docker-compose file

source/adios2/CMakeLists.txt
Change72: Ok, I figure it out by passing -DCMAKE_PREFIX_PATH="${SW_DIR}/hiredis" and updated.

eisenhauer:
@Change72 If you click on "details" next to the "cdash" entry in the checks, then click on the yellow box indicating a failure in a build, you can see the warnings that resulted in that build failure...  ADIOS doesn't allow warnings in CI, so these have to be corrected.

scripts/build_scripts/build-adios2-kvcache.sh
```diff
@@ -0,0 +1,112 @@
...
+            -DCMAKE_INSTALL_PREFIX="${SW_DIR}"/adios2
+
+    make -j32
+    make install
+    cd "${BUILD_DIR}"/../ || exit
+    echo "Build completed."
+    unset BUILD_DIR
+}
+
+start_services() {
```
eisenhauer, line 62: The "DoRemote" environment variable was a quick hack to trigger remote access while we figured out how we really wanted to to it, but I'm a little surprised that it works to have DoRemote set for the adios2_remote_server.  Generally you'd want that set for the clients, not for the server (which runs the same Get() code, but which has to do local access).   If the DoRemote and useKVCache variables are for the clients which follow, maybe move those to the end of this script?  Or is does the server need useKVCache?  (That may be confusion on my part WRT where the caching happens, on the server or on the client.)

Change72: The current cache design is for clients, and I have moved those two declarations to the end of build process.

source/adios2/CMakeLists.txt
Change72: I think I met a weird problem when I use -DCMAKE_PREFIX_PATH="${SW_DIR}/hiredis". The link under bin works fine, but under lib is not found. Do you know how to fix it?

4cv@PC0141240:~/project/ADIOS2-Source/examples/simulations/gray-scott/build$ ldd /home/4cv/project/ADIOS2-Source/build/sw/adios2/lib/libadios2_core.so
	linux-vdso.so.1 (0x00007ffdb4ffc000)
	libhiredis.so.1.1.0 => not found
	libsqlite3.so.0 => /lib/x86_64-linux-gnu/libsqlite3.so.0 (0x000076b73ddbc000)
	libz.so.1 => /lib/x86_64-linux-gnu/libz.so.1 (0x000076b73d5e4000)
	libadios2_perfstubs.so.2.10 => /home/4cv/project/ADIOS2-Source/build/sw/adios2/lib/../lib/libadios2_perfstubs.so.2.10 (0x000076b73ddb6000)
	libadios2_evpath.so.2.10 => /home/4cv/project/ADIOS2-Source/build/sw/adios2/lib/../lib/libadios2_evpath.so.2.10 (0x000076b73d569000)
	libadios2_ffs.so.2.10 => /home/4cv/project/ADIOS2-Source/build/sw/adios2/lib/../lib/libadios2_ffs.so.2.10 (0x000076b73d4fe000)
	libadios2_atl.so.2.10 => /home/4cv/project/ADIOS2-Source/build/sw/adios2/lib/../lib/libadios2_atl.so.2.10 (0x000076b73d4ef000)
	libstdc++.so.6 => /lib/x86_64-linux-gnu/libstdc++.so.6 (0x000076b73d200000)
	libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x000076b73d119000)
	libgcc_s.so.1 => /lib/x86_64-linux-gnu/libgcc_s.so.1 (0x000076b73d4cf000)
	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x000076b73ce00000)
	/lib64/ld-linux-x86-64.so.2 (0x000076b73df1d000)
	libadios2_dill.so.2.10 => /home/4cv/project/ADIOS2-Source/build/sw/adios2/lib/../lib/../lib/libadios2_dill.so.2.10 (0x000076b73d47a000)


4cv@PC0141240:~/project/ADIOS2-Source/examples/simulations/gray-scott/build$ ldd /home/4cv/project/ADIOS2-Source/build/bin/adios2_remote_server 
	linux-vdso.so.1 (0x00007fff52fc0000)
	libadios2_evpath.so.2.10 => /home/4cv/project/ADIOS2-Source/build/lib/libadios2_evpath.so.2.10 (0x000077ab8af0f000)
	libadios2_core.so.2.10 => /home/4cv/project/ADIOS2-Source/build/lib/libadios2_core.so.2.10 (0x000077ab8a600000)
	libadios2_atl.so.2.10 => /home/4cv/project/ADIOS2-Source/build/lib/libadios2_atl.so.2.10 (0x000077ab8af00000)
	libstdc++.so.6 => /lib/x86_64-linux-gnu/libstdc++.so.6 (0x000077ab8a200000)
	libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x000077ab8ae07000)
	libgcc_s.so.1 => /lib/x86_64-linux-gnu/libgcc_s.so.1 (0x000077ab8ade5000)
	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x000077ab89e00000)
	libadios2_ffs.so.2.10 => /home/4cv/project/ADIOS2-Source/build/lib/libadios2_ffs.so.2.10 (0x000077ab8a595000)
	libhiredis.so.1.1.0 => /home/4cv/project/ADIOS2-Source/build/sw/hiredis/lib/libhiredis.so.1.1.0 (0x000077ab8adcd000)
	libsqlite3.so.0 => /lib/x86_64-linux-gnu/libsqlite3.so.0 (0x000077ab8a448000)
	libz.so.1 => /lib/x86_64-linux-gnu/libz.so.1 (0x000077ab8adb1000)
	libadios2_perfstubs.so.2.10 => /home/4cv/project/ADIOS2-Source/build/lib/libadios2_perfstubs.so.2.10 (0x000077ab8a442000)
	/lib64/ld-linux-x86-64.so.2 (0x000077ab8afd6000)
	libadios2_dill.so.2.10 => /home/4cv/project/ADIOS2-Source/build/lib/libadios2_dill.so.2.10 (0x000077ab8a1ab000)

eisenhauer:
Excellent work getting rid if the templating, and I'm thrilled that we don't have to base64 encode the data to store it in Redis.  I'll do a more thorough review in a bit, but I think you've hit the big points I'd have.

source/adios2/engine/bp5/BP5Reader.cpp
eisenhauer, line 1: My lurking worry here is that we have expanded a simple loop with a 2 line body to 100+ lines.  Maybe we can break out the KVCache case to a different subroutine to keep the non-cache case simple to understand?  Also you might use CoreDims/DimsArray rather than Dims.  It's probably not as critical here, but Dims uses std::vector, which means lots of tiny mallocs whenever these are created, destroyed, passed by value, etc.  This caused serious performance problems in metadata manipulation with BP4 at scale.  As a result, BP5 tries to avoid Dims.

CMakeLists.txt
Change72: added

source/adios2/engine/bp5/BP5Reader.cpp
Change72: 1. I want to add a subroutine in:

    if (m_Remote)
    {
        PerformRemoteGets();
    }
    else
    {
        PerformLocalGets();
    }

However, what I can do is like:

#ifdef ADIOS2_HAVE_KVCACHE
        if (getenv("useKVCache"))
        {
            PerformLocalGetsWithKVCache();
        }
        else
        {
            PerformRemoteGets();
        }
#else
    PerformRemoteGets();
#endif

do you think this is fine?

2. It's hard to initialize DimsArray when I declare ReqInfo struct, since I don't know the dimension of each Req, and DimsArray has no default construction method.

anagainaru: What about?
```c++
#ifdef ADIOS2_HAVE_KVCACHE
if (getenv("useKVCache"))
{
PerformLocalGetsWithKVCache();
}
else
#endif
PerformRemoteGets();
```

eisenhauer: I prefer this version...

Change72: Maybe I could get rid of `getenv("useKVCache")` ->
```c++
if (m_Remote)
{
  #ifdef ADIOS2_HAVE_KVCACHE
    PerformRemoteGetsWithKVCache();
  #else
    PerformRemoteGets();
  #endif
}
else
{
  PerformLocalGets();
}
```

If the user builds with KVCache and wants to use remote Gets (DoRemote=1), KVCache will be used automatically. However, if the user prefers to use only the remote service without KVCache, they will need to rebuild (possibly in a separate folder).

We could also consider checking whether the cache connection is active, and if it is not, default to using the pure remote service.

eisenhauer: >     2. It's hard to initialize DimsArray when I declare ReqInfo struct, since I don't know the dimension of each Req, and DimsArray has no default construction method.

So, let me tell you my worry.  Your code in your cache stuff is in a lot of ways similar to the code in NdCopy (which looks at whether or not two blocks intersect, the extent of the overlap, etc.).  We used to have some major performance problems in this when metadata was large (many thousand blocks per variable) because every time it did something like try to determine an intersection it ended up creating a destroying a dozen or more Dims vectors (doing things like converting a taking a StartEnd Box and creating a StartCount box from it, having all those temporary values get free'd when the intersection test returned, only to recreate them later, etc.). We traced the issue back to lots and lots of small malloc/free calls when those Dims were created, initialized and destroyed.  It took a ton of effort to rewrite the worst parts of that code to use ArrayDims (stack allocated dimension array, so cheap to create/destroy, own their own data) or CoreDims (the equivalent of a "span", don't own their data but let us pass it in places and operate on it like a vector without a copy).   With a lot of careful work kill all the object creation/destruction overhead in NdCopy so its only overhead was necessary control and data movement.  (And if you go look at its source in helper/adiosMemory.cpp you'll see extensive use of ArrayDims and CoreDims.). 

The storm of vector construction and destruction in ADIOS metadata handling wasn't noticeable at small scales and it was easy to ignore the implicit overhead of casual use of Dims, but as we ran at higher scales it became seriously problematic.  You're probably not seeing an issue at the scales you're testing caching at either, but we hope to use this in large-metadata situations, so it'll become an issue if it's not done well.   There's just too much similarity in your processing and the NdCopy case for me not to be concerned.  We obviously didn't eliminate Dims from all the ADIOS code, but we killed it where it was used most intensively.  You might do the same.

(And of course you do know the dimensionality of the variable in the place where you declare ReqInfo variable, it's in the Req variable that you're iterating over in that loop.  Or it would be easy to add a default constructor and initialize the dimcount later.  My concern is about places where there might be intensive vector usage.)

Change72: The test script we talked about just now is: https://github.com/Change72/adios-test/blob/master/TestCoreDims.cpp

// will make a value copy of dimsPoint2
DimsArray dimsPointer3(reinterpret_cast<CoreDims&>(dimsPointer2)); 

// will only copy the pointer
DimsArray dimsPointer3(dimsPointer2);

Change72: However, the following code will also give wrong results, but vector works fine.

```c++
helper::NdCopy(reinterpret_cast<char *>(ReqInfo.Data), reinterpret_cast<helper::CoreDims&>(ReqInfo.ReqBox.Start), reinterpret_cast<helper::CoreDims&>(ReqInfo.ReqBox.Count), true, false, reinterpret_cast<char *>(Req.Data), Req.Start, Req.Count, true, false, ReqInfo.TypeSize);
```

Vector code:
```c++
        std::vector<size_t> start;
        std::vector<size_t> count;
        ReqInfo.ReqBox.StartToVector(start);
        ReqInfo.ReqBox.CountToVector(count);
        helper::NdCopy(reinterpret_cast<char *>(ReqInfo.Data), start, count, true, false, reinterpret_cast<char *>(Req.Data), Req.Start, Req.Count, true, false, ReqInfo.TypeSize);
```

eisenhauer (commented):
At a quick page through, I don't have any complaints.  I'll let Norbert look for final approval though.

Change72:
@eisenhauer I just got a warning C4267: 'argument': conversion from 'size_t' to 'const int', possible loss of data

It happened in [NdCopy const int typeSize](https://github.com/ornladios/ADIOS2/blob/17275272fb9c046ede31aab246a08709c015df8b/source/adios2/helper/adiosMemory.cpp#L252) 

While in [size_t GetDataTypeSize(DataType type)](https://github.com/ornladios/ADIOS2/blob/17275272fb9c046ede31aab246a08709c015df8b/source/adios2/helper/adiosType.cpp#L94), the return type is size_t

I just get size_t from GetDataTypeSize and then pass it to NdCopy, which causes the warning.

I am curious why typeSize is int rather than size_t in NdCopy? Could it be a negative value?

eisenhauer:
The windows compilers are annoyingly picky about implicit type conversions.  The typeSize argument to NdCopy is the size of the element.  I.E. it'll be between 1 and the biggest type we support, complex double, which would be 16.  The author of NdCopy chose to use an int for that, presumably not to waste a whole 8 bytes on something that doesn't need to be remotely that big.  You can change your typeSize field in your Req to int, or you can add a cast into your NdCopy call.  Or I guess if you were really ambitious you could change NdCopy to take size_t for that.  But once you go down the rabbit hole of changing types, that stuff tends to snowball...

pnorbert (approved):

eisenhauer:
The MacOS failure is expected at this point, but the sanitizer failure is weird.  I've asked it to rerun to see if it was spurious.

PR #4213, opened 2024-06-25
ci: added rocm gha build

vicentebolea:
This is needed to replicate/fix https://github.com/spack/spack/issues/44594

This PR adds a ROCM 6.1 Kokkos 4.3 build using our spack builds. This PR also adds a change that enables ADIOS2 to be built with ROCM >=6.

PR #4214, opened 2024-06-27
kokkos: support ROCM >=6

vicentebolea:
Backports https://github.com/ornladios/ADIOS2/pull/4213

### DIFF
--- a/source/adios2/helper/kokkos/adiosKokkos.cpp
+++ b/source/adios2/helper/kokkos/adiosKokkos.cpp
@@ -85,7 +85,11 @@ bool IsGPUbuffer(const void *ptr)
     hipError_t ret;
     hipPointerAttribute_t attr;
     ret = hipPointerGetAttributes(&attr, ptr);
+#if defined(ROCM_VERSION_MAJOR) && ROCM_VERSION_MAJOR < 6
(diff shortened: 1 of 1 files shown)


vicentebolea:
@eugeneswalker

eisenhauer (approved):

anagainaru:
I would like to better understand, did rocm 6 change the API on detecting the allocation type? Could you send me a link to the rocm documentation about this change?

I want to test this on Frontier to make sure we can't use something that is the same across all rocm versions before we consider this complete.

vicentebolea:
Hi @anagainaru welcome back! Good point.

hipPointerAttribute_t API in:

- ROCM 5.3: https://rocm.docs.amd.com/projects/HIP/en/docs-5.3.3/doxygen/html/structhip_pointer_attribute__t.html
- ROCM 6.1: https://rocm.docs.amd.com/projects/HIP/en/latest/doxygen/html/structhip_pointer_attribute__t.html

Yep, it would have been nice from the AMD rocm team to put some  ``[[deprecated]]`` before phasing out the old attribute.

anagainaru:
Yes you are right, I don't think there is any better way of handling this. I was hoping there would be some other solution to test for the allocation of a buffer but I can't seem to find anything.

PR #4220, opened 2024-07-02
extend DimsArray constructor with DimsArray and DimsArray & Count

Change72:
add two more DimsArray constructors

CoreDims is designed as a Span and does not own the corresponding memory, but DimsArray is designed to copy the dimension data. However, existing method `DimsArray(const CoreDims &d1)` fails to copy the data in `DimsArray dimsPointer3(dimsPointer2);`

My test script is available at: https://github.com/Change72/adios-test/blob/master/TestCoreDims.cpp

```c++
// will make a value copy of dimsPoint2
DimsArray dimsPointer3(reinterpret_cast<CoreDims&>(dimsPointer2));

// will only copy the pointer
DimsArray dimsPointer3(dimsPointer2);

dimsPointer3[1] = 50;
```

So the first added constructor is to fix this issue. As to the second one, when the dimension count is unknown during construction, one way to use `DimsArray` is to declare with `helper::Max_Dims` first, and save the corresponding count in a separate variable.

However, when passing DimsArray to another function, like `NdCopy`, it will copy the entire array with helper::Max_Dims. That's why we need count, which is not equal to d1.size(), to indicate the actual size.

```c++
        helper::DimsArray startArray(ReqInfo.ReqBox.DimCount, ReqInfo.ReqBox.Start);
        helper::DimsArray countArray(ReqInfo.ReqBox.DimCount, ReqInfo.ReqBox.Count);

        helper::NdCopy(reinterpret_cast<char *>(ReqInfo.Data), startArray, countArray, true, false, reinterpret_cast<char *>(Req.Data), Req.Start, Req.Count, true, false, ReqInfo.TypeSize);
```

### DIFF
--- a/source/adios2/helper/adiosType.h
+++ b/source/adios2/helper/adiosType.h
@@ -177,6 +177,16 @@ public:
     {
         std::copy(d1.begin(), d1.end(), &Dimensions[0]);
     }
+
+    DimsArray(const DimsArray &d1) : CoreDims(d1.size(), &Dimensions[0])
+    {
+        std::copy(d1.begin(), d1.end(), &Dimensions[0]);
+    }
+
+    DimsArray(const size_t count, const DimsArray &d1) : CoreDims(count, &Dimensions[0])
+    {
+        std::copy(d1.begin(), d1.end(), &Dimensions[0]);
+    }
 };
 
 /**


anagainaru (commented):
Please explain in the description of the PR why the copy constructor is needed.

Also, the second constructor taking a count and a DimsArray looks weird. If count is not equal to d1.size() you still copy from begin to end of d1?

eisenhauer:
> Please explain in the description of the PR why the copy constructor is needed.

Well, I can speak to this one...  When I created the class I didn't use it in any circumstance where a copy got invoked, but it one got invoked it was a trivial one which didn't reset the pointer in the CoreDims parent class appropriately.  So, needed to be fixed for other circumstances.  I do tend to wonder about the need for the extra param.  You'd think it would be an odd circumstance where you wanted to use a size other than d1.size(), but also copy the data.

anagainaru:
@eisenhauer that makes sense for the copy constructor. I still don't see why we would need the second one.

eisenhauer:
> @eisenhauer that makes sense for the copy constructor. I still don't see why we would need the second one.

@Change72, can you clarify?  (Or mod your code so that you don't use this if not needed?)

Actually, now that I look at it, the first may not be an actual copy assignment operator, which should use an '=' method name.  Maybe try that in your code?

anagainaru:
> Actually, now that I look at it, the first may not be an actual copy assignment operator, which should use an '=' method name. Maybe try that in your code?

Yeah, the one here is for `DimsArray d_new(d_old)`. I am not sure which one is needed.

Change72:
I just test again:

```c++
    DimsArray dimsPointer3(dimsPointer2);

    DimsArray dimsPointer3 = dimsPointer2;

    dimsPointer3[1] = 50;
```

The above two give the same results: if without my changes, dimsPoint3 copies the pointer and change dimsPoint2 value as well.

With my first constructor, dimsPoint3 copies the value and would be different from dimsPoint2.

The second constructor is to solve the dimension that is not known during the initialization (As my updated description in the very beginning), since I am not allowed to use vector in KVCache.

My kvchache demands are quite simple: 
1. Copy the value from another array, not the pointer, since I need to update the value
2. Declare an array without the exact dimension and update it later, since I do not know the dimension size when I declare my class and struct. Also `new` is not allowed.

Vector can fulfill the requirements, but @eisenhauer says there will be too many small memory allocations, while the current DimsArray cannot fulfill my needs.

eisenhauer:
OK, I'm officially changing my diagnosis of the situation.  What was going on with your test code (TestCoreDims.cpp, not included in this PR but exchanged via Email) is that in these two lines:
    std::vector<size_t> vec = {1, 2, 3, 4, 5};
    CoreDims coreDims(vec);

C++ made a copy of vec in order to call the coreDims(std::vector<size_t>) constructor because I forgot to put an & in front of the parameter.  This apparently wasn't a problem for the NdCopy implementation because the constructor tended to be used in parameters, so the copy was valid for the entire time that the CoreDims was in use.  But it failed in the circumstances of your test.  Please try changing the CoreDims constructor for init from Dims to this:

    CoreDims( const std::vector<size_t> &vec) : DimCount(vec.size()), DimensSpan((size_t*)vec.data()) {}

WRT your note about not knowing the size during init, you do know it, it's in the Req variable that you're iterating over in that loop.  You just have to add a constructor for your QueryBox that passes this in when it inits the DimsArray.   Alternatively you could kill the const on DimsCount in CoreDims and DimsArray and add a resize() method that sets the DimsCount value.  But that's not really the right approach.  The best way to do this is with a smarter constructor.

(And yes, I was just ranting yesterday about C++ making copies of things automagically if you forgot a &, and creating an unexpected (and relatively invisible) inefficiency.  There you go.  Despite effort, we still had bunches of tiny mallocs in NdCopy because of this.  Sigh.

Change72:
However, the issue in this PR is still there. After I use `CoreDims( const std::vector<size_t> &vec) : DimCount(vec.size()), DimensSpan((size_t*)vec.data()) {}`

```c++
    DimsArray dims1(5, 42);  // Initialize with a single value
    std::cout << "DimsArray with single value: " << dims1 << std::endl;

    DimsArray dims2(dims1);
    std::cout << "DimsArray copied from dims1: " << dims2 << std::endl;

    dims2[1] = 100;
    std::cout << "dims1: " << dims1 << std::endl;
    std::cout << "dims2: " << dims2 << std::endl;
```

Output:

DimsArray with single value: {42, 42, 42, 42, 42}
DimsArray copied from dims1: {42, 42, 42, 42, 42}
dims1: {42, 100, 42, 42, 42}
dims2: {42, 100, 42, 42, 42}

eisenhauer:
@Change72 Would you please evaluate PR #4221  and let me know if it works for your stuff.  (Modified test script converted to GTest form added as well.)

Change72:
It works and I have already updated PR https://github.com/ornladios/ADIOS2/pull/4210. I think we can close this one.

PR #4222, opened 2024-07-03
[Bug fix] Switch from `ROCM_VERSION_MAJOR` to `HIP_VERSION_MAJOR`

anagainaru:
Bug introduced in #4214 (PR solving the problem in ROCM API change for hipPointerAttribute_t:

ROCM 5.3: https://rocm.docs.amd.com/projects/HIP/en/docs-5.3.3/doxygen/html/structhip_pointer_attribute__t.html
ROCM 6.1: https://rocm.docs.amd.com/projects/HIP/en/latest/doxygen/html/structhip_pointer_attribute__t.html

@pnorbert this needs to be part of the latest release/patch or the GPU support on Frontier will give compile errors

### DIFF
--- a/source/adios2/helper/kokkos/adiosKokkos.cpp
+++ b/source/adios2/helper/kokkos/adiosKokkos.cpp
@@ -85,7 +85,7 @@ bool IsGPUbuffer(const void *ptr)
     hipError_t ret;
     hipPointerAttribute_t attr;
     ret = hipPointerGetAttributes(&attr, ptr);
-#if defined(ROCM_VERSION_MAJOR) && ROCM_VERSION_MAJOR < 6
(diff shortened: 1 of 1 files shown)


vicentebolea:
@anagainaru this seems to be reverse to what the api requires, rocm < 6 -> use memoryType; rocm > 6 -> use type

vicentebolea:
IT seems to be something else, maybe ROCM_VERSION_MAJOR is not defined in frontier, I am bringing Frontier CI again lets check it there.

anagainaru:
Yes, this is why I just created a draft, I was unable to test rocm 6 yet :) but it's def not working on Frontier as is now

eisenhauer (approved):

eisenhauer:
bypass restrictions because macos-11 is gone.

PR #4223, opened 2024-07-03
ci: add frontier pipelines

vicentebolea:
Bring frontier pipeline


vicentebolea:
Merging without review since I need to merge this to test that it works.

PR #4224, opened 2024-07-03
ci: add sync script workflow

vicentebolea:


vicentebolea:
Merging this without review/ci since I need this in master to test in it in ORNL gitlab instances.

PR #4226, opened 2024-07-05
Kill macos-11 CI

eisenhauer:


eisenhauer:
@vicentebolea Well, I thought I'd see if I could trivially just advance those two macos builds by two versions.  Unfortunately there are a couple of funky warnings popping up that I don't see immediate fixes to.  I won't delete this PR in case it's a useful start, but I'm not going to try to diagnose and fix the issues.

vicentebolea:
superseeded by #4245

PR #4227, opened 2024-07-05
Extend Expression Library: Subtract

lizdulac:
Create subtract operator and add to Derived Variable Correctness Test

    bpOut.DefineDerivedVariable(derivedname,
                                "x =sim/Vx \n"
                                "y =sim/Vy \n"
                                "z =sim/Vz \n"
                                "x-y-z",
                                adios2::DerivedVarType::StoreData);

ApplyOneToOne function is altered to allow initialization of output to be populated with the data of the first operand or set to 0 as originally implemented.

This is necessary because 0-x-y-z is not the same as x-y-z. It's imperative that the first operand is treated differently than subsequent operands.

This change will also be necessary for operations where 0 is not the operator's identity (ex multiplication). We may want to extend this to allow the function to specify how the output data should be initialized similarly to base cases in recursive functions.

Considerations for existing expressions:

Addition output can be initialized with the data of the first operand.

The magnitude function depends on the initialization of the output data as 0 and then recursively adding the next variable to the power of 2, and is left as originally implemented. This could be updated to initialize to the square of the first operand, but mathematically there is no reasonable way to build this function recursively if the output data is set to the first operand.

Curl does not call ApplyOneToOne and is unaffected.


source/adios2/toolkit/derived/Function.cpp
```diff
@@ -38,7 +58,8 @@ DerivedData MagnitudeFunc(std::vector<DerivedData> inputData, DataType type)
 #define declare_type_mag(T)                                                                        \
     if (type == helper::GetDataType<T>())                                                          \
     {                                                                                              \
-        T *magValues = ApplyOneToOne<T>(inputData, dataSize, [](T a, T b) { return a + b * b; });  \
+        T *magValues = ApplyOneToOne<T>(                                                           \
+            inputData, dataSize, [](T a, T b) { return a + b * b; }, true);                        \
```
anagainaru, line 62: Shouldn't the fourth parameter be false and true for substrat?

anagainaru (commented):
I want to re-write a bit the `ApplyOneToOne` function, but other than this it looks ok. I'll push my changes in a bit

anagainaru:
I think this is ready to be merged, @lizdulac take a look at the changes. I did not want to split the functionality of the ApplyOneToOne function so instead I made it so that it applies the operation on ranges of a vector (iterator start to end). This way add becomes "+ from begin to end" and subtract becomes "+ from begin+1 to end" that is subtracted from the first element.

anagainaru (approved):

PR #4228, opened 2024-07-05
First cut at reader-side derived variables

eisenhauer:


source/adios2/engine/bp5/BP5Reader.cpp
```diff
@@ -496,6 +493,8 @@ void BP5Reader::PerformLocalGets()
             m_BP5Deserializer->FinalizeGet(Req, false);
         }
     }
+    m_BP5Deserializer->FinalizeDerivedGets(ReadRequests);
```
anagainaru, line 496: Is this supposed to be included between macros `#ifdef ADIOS2_HAVE_DERIVED_VARIABLE`

source/adios2/toolkit/format/bp5/BP5Deserializer.cpp
```diff
@@ -918,6 +921,17 @@ void BP5Deserializer::InstallMetaData(void *MetadataBlock, size_t BlockLen, size
 
                 if (!VarRec->Variable)
                 {
+                    if (VarRec->Derived)
+                    {
+#ifdef ADIOS2_HAVE_DERIVED_VARIABLE
+                        // define the derived copy first, so we don't throw an error on actual var
+                        // definition
+                        auto tmp = m_Engine->m_IO.DefineDerivedVariable(
+                            VarRec->VarName, VarRec->ExprStr, DerivedVarType::MetadataOnly);
```
anagainaru, line 930: We need to discuss the default of `MetadataOnly`, there is also the option of `ExpressionOnly`

anagainaru (commented):
This makes sense, let me read your email again before I approve.

source/adios2/engine/bp5/BP5Reader.cpp
eisenhauer: FinalizeDerivedGets is defined without or without derived vars, it just has a null body without them.  Could go either way here, but having a null body keeps the number of #ifdef's down at least a little bit.

source/adios2/toolkit/format/bp5/BP5Deserializer.cpp
eisenhauer: Ah, so yes, I have thought about ExpressionOnly, but explicitly did not implement it yet because I hadn't decided the best approach.  The question is whether or not there is any (non-statistical) metadata about the derived variable created when it is written.  In the approach I'm leaning towards, we'd modify the writer to produce the same per-block information that we would have in the MetadataOnly approach, only we'd never do the ApplyExpression, never calculate the statistics and not include the statistics in the metadata.  If we do it that way, we just end up adding a couple more 'if's to the reader and the writer, so implementation is easy and we have an opportunity to throw what errors we might want if the writer does something wrong (like mismatched block counts or sizes with the input variables, etc.) but the downside is that we do have extra information in the metadata that we could reasonably have inferred from what's there.   The alternative, if we truely put the ExpressionOnly in the metadata and no blocksinfo is that we are doing to have to write the code that infers the derived variable's blocksinfo, maybe handle it specially if we want to save it between calls, etc.    

I tend to prefer the first approach because the implementation is more straightforward and I think the downside (more metadata) is marginal, but I wanted to run it by everyone before implementing.  (For the test program, we just need to add the ExpressionOnly to the list of values to test.)

anagainaru:
This is not the first time I am seeing the `macos-12` CI failing without any connexion with the changes in the PR.

eisenhauer:
> This is not the first time I am seeing the `macos-12` CI failing without any connexion with the changes in the PR.

Yeah, it's broken.  But also not required to merge...   So if you approve, this can go in.

anagainaru (approved):

PR #4229, opened 2024-07-05
ci: remove macos11 build

vicentebolea:


vicentebolea:
@eisenhauer lets remove this build so our CI is green and later in #4226 we add newer macos builds.

eisenhauer (approved):
Works for me...

PR #4234, opened 2024-07-10
Connect to tunnel server to launch remote server

pnorbert:
- Added function LaunchRemoteServerViaConnectionManager() to Remote class to talk to local tunnel server to establish a connection to remote host (and launch adios2_remote_server on remote host). Local tunnel server is looked for on fixed port 30000 for now. Remote server's port and local port forwarded to remote server's port are dynamic.

- Ugly hack: added to ADIOS a static Class-member (m_StaticHostOptions) and class-function (StaticGetHostOptions()) that contains one of the ADIOS object's HostOptions, in order to allow access to HostOptions from functions that has no access to any ADIOS objects (e.g. FileRemote transport).

### DIFF
--- a/source/adios2/core/ADIOS.cpp
+++ b/source/adios2/core/ADIOS.cpp
@@ -102,7 +102,8 @@ public:
 };
 
 ADIOS::GlobalServices ADIOS::m_GlobalServices;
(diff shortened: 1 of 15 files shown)


pnorbert:
Merging now as only that single test failure on macos

PR #4235, opened 2024-07-10
Remove reference to xarray backend

JamesMcClung:
The deleted entry causes the generation of an entry_points.txt file with the following content:
```
[xarray.backends]
adios = adios2.xarraybackend:AdiosBackendEntrypoint
```
When [xarray](https://docs.xarray.dev/en/stable/index.html) and ADIOS 2 are both installed with e.g. pip, xarray finds this file and tries to load the `adios2.xarraybackend` module. I assume this module exists somewhere locally, but it doesn't seem to be in this repo, and xarray generates a warning when loading this module fails.

As far as I know, the only ways to make this warning go away are a) remove the reference to the missing xarray backend, or b) implement said backend. This PR obviously does the former, but I'm very curious about the state of progress towards the latter option!

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -48,9 +48,6 @@ Documentation = "https://adios2.readthedocs.io/"
 Discussions = "https://github.com/ornladios/ADIOS2/discussions"
 Changelog = "https://github.com/ornladios/ADIOS2/releases"
 
-[project.entry-points."xarray.backends"]
-adios = "adios2.xarraybackend:AdiosBackendEntrypoint"
-
 [tool.cibuildwheel]
 # Trigger an install of the package, and run a basic test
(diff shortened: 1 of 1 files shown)


pnorbert:
Thank you. The backend exists, but it is not working 100% with the xbout project, so far the only user of this backend, so it was taken out the master repo but forgot the pyproject reference. 
https://github.com/pnorbert/ADIOS2/tree/xarraybackend

JamesMcClung:
Thanks for sharing, that implementation looks pretty nice. I've been working on extracting [this implementation](https://github.com/psc-code/psc/tree/main/python/psc) to a standalone repo with proper type checking, testing, etc., which is how I ran into the problem above.

pnorbert (approved):

PR #4238, opened 2024-07-11
ATL and EVPath Upstream

eisenhauer:


eisenhauer:
This adds an additional attribute to the CManager contact information to uniquely identify each communicating entity.

vicentebolea (approved):

PR #4240, opened 2024-07-11
python: return Attribute data singlevalue as it is

vicentebolea:


vicentebolea:
@pnorbert this is a weird error, in mac the following snippet produces this: `ts.data = [2023 2023 2023]`

```
        adios = Adios()
        with adios.declare_io("BPWriter") as io:
            arr = np.array([2023, 11, 22])
            ts = io.define_attribute("timestamp", arr)
```

Do you know what can be the reason?

python/adios2/attribute.py
```diff
@@ -58,8 +58,6 @@ def data(self):
         Returns:
             Content of the Attribute as a non string.
         """
-        if self.single_value():
```
pnorbert, line 61: I intentionally added this code to return the actual value for single attributes, not an array with 1 element. So attribute reading is similar to reading variables, where single values are read as python variables or numpy 0-dim arrays

PR #4247, opened 2024-07-18
[Bug fix] Read side computation of derived variables should not be triggered when data is stored

anagainaru:
Before this PR, adios2 calls `ApplyExpression` every time there is a derived variable (which is identified by the expression string we encode in the name). However, for variables where we store data we do not want to recompute the values. The easiest fix is to not store the expression string for these variables (since in essence they are no longer derived variables).


eisenhauer (approved):
Looks great, thanks for catching!

PR #4249, opened 2024-07-18
Implementing reader-side Get()s for ExpressionString derived variables

eisenhauer:


eisenhauer:
@lizdulac and @anagainaru This is a revival of the older PR.  Main changes on the writer side are:  1) Like StatsOnly, it adds block-level metadata on the writer side and 2) Unlike StatsOnly, it doesn't actually do any of the allocations or calculations, just goes through the motion so that we get the same blocks info that you'd have generated if you did those derived variable calculations.   On the reader side, we do the same stuff that we do for StatsOnly.  That is, if someone asks us for a block of a derived variable that we don't have data for, we calculate the data and return it.

anagainaru:
Will look at this tonight, it needs more attention than other PRs

eisenhauer:
As of the most recent build, this was run with DerivedVariables enabled, so it does compile and run on all platforms.  But master now has that turned off, so once updated or merged it'll be optional, not compiled or tested in CI.   (Not sure what's going on with that one failure, but it isn't related to his PR.)

source/adios2/toolkit/derived/Expression.cpp
```diff
@@ -354,14 +356,14 @@ ExpressionTree::ApplyExpression(DataType type, size_t numBlocks,
     auto op_fct = OpFunctions.at(detail.operation);
     for (size_t blk = 0; blk < numBlocks; blk++)
     {
-        outputData[blk] = op_fct.ComputeFct(exprData[blk], type);
+        outputData[blk] = op_fct.ComputeFct(exprData[blk], type, DoCompute);
```
anagainaru, line 359: Instead of propagating the `DoCompute` to the compute functions we can just add an empty `DerivedData` here if `DoCompute` is false.

anagainaru (commented):
I would not propagate DoCompute all the way to the math functions. We can stop at the expression level, creating empty DerivedData (as I am suggesting above).

However, the more I think about this, the `ExpressionOnly` flag is on the entire Variable so I am not sure why you need to call ApplyExpression at all.

source/adios2/toolkit/derived/Expression.cpp
eisenhauer: Unfortunately DerivedData contains what I need (the dimensions of the output variable) as well as the data pointer, and as far as I can see we can't guess the dimensions without knowing what the ComputFct wll do.

anagainaru: Ok so it cannot be done at the VariableDerived level. 
But the Expression class should be able to return the dimension, there is a dimension function that needs to correspond to what the output of the compute function is (example `SameTypeFunc`). 

Where we call `outputData[blk] = op_fct.ComputeFct(exprData[blk], type);` you could call `Dims opDims = op_fct.DimsFct(exprDims);` and get the output dimensions of the DerivedData.

anagainaru: Actually we can do it in VariableDerived also, you can call `void Expression::SetDims(std::map<std::string, std::tuple<Dims, Dims, Dims>> NameToDims)` and get the output dimension (shape, start, count) by directly accessing `m_Count`, `m_Start`.

eisenhauer: Hmm.  I had missed DimsFct(), so yes, that's maybe helpful.  But I'm still not quite sure how to apply SetDims.  Maybe you can suggest a new body for the dummy function I used in the first go at this:

std::vector<std::tuple<void *, Dims, Dims>> VariableDerived::GenerateDerivedDims(
    std::map<std::string, std::unique_ptr<MinVarInfo>> &NameToVarInfo){}

anagainaru: Yes, I can try to populate your function after lunch.

anagainaru: I did not run this, but your function should be something like:
```c++
    std::vector<std::tuple<void *, Dims, Dims>> blockData;
    for (size_t i = 0; i < numBlocks; i++)
    {
        std::map<std::string, Dims> nameToCount;
        std::map<std::string, Dims> nameToStart;
        for (const auto &variable : NameToVarInfo)
        {
            Dims start;
            Dims count;
            for (int d = 0; d < variable.second->Dims; d++)
            {
                start.push_back(variable.second->BlocksInfo[i].Start[d]);
                count.push_back(variable.second->BlocksInfo[i].Count[d]);
            }
           nameToCount.insert({variable.first, count});
           nameToStart.insert({variable.first, start});           
         }

      outputCount = m_Expr.GetDims(nameToCount)
      outputStart = m_Expr.GetDims(nameToStart)

      blockData.push_back({void *, outputStart, outputCount});
    }
    return blockData;
```

GetDims might not be exposed outside Expression so we might need to expose this. I can apply the changes over your PR is you want.

anagainaru: Actually we don't need to copy the dimensions, we can just use the ones in each block:
```c++
std::vector<std::tuple<void *, Dims, Dims>> blockData;
    for (size_t i = 0; i < numBlocks; i++)
    {
        std::map<std::string, Dims> nameToCount;
        std::map<std::string, Dims> nameToStart;
        for (const auto &variable : NameToVarInfo)
        {
           nameToCount.insert({variable.first, variable.second->BlocksInfo[i].Count});
           nameToStart.insert({variable.first, variable.second->BlocksInfo[i].Start});           
         }

      outputCount = m_Expr.GetDims(nameToCount)
      outputStart = m_Expr.GetDims(nameToStart)

      blockData.push_back({void *, outputStart, outputCount});
    }
    return blockData;
```

anagainaru (dismissed):
The tests pass with DerivedVariables ON. Greg feel free to revert if you don't like my changes. Otherwise this is ready to merge.

eisenhauer:
Sorry, had to add one more thing to make sure that bpls isn't looking for min/max that doesn't exist for these vars.  Approve again?

PR #4252, opened 2024-07-19
Fix errors and warning related to derived variables

anagainaru:
PR to fix errors and warnings related to derived variables (except the lexer related warnings since @lizdulac is opening a PR on fixing lexer issues)

## Error

**Mismatch in reading count when reading variables that are part of a derived variable**

`Req->Count.data()` are the dimensions of the derived variable in the current request

`VarRec->DimCount` are the total dimensions of one of the variables used by the derived variable.

In all ecause Curl adds one dimension at the end).

However, I am adding now an aggregated operator that will have a lower dimension than the composing variables, so this line seg faults.

### DIFF
--- a/source/adios2/toolkit/derived/Function.cpp
+++ b/source/adios2/toolkit/derived/Function.cpp
@@ -44,13 +44,13 @@ T *ApplyCurl(const T *input1, const T *input2, const T *input3, const size_t dim
     size_t dataSize = dims[0] * dims[1] * dims[2];
     T *data = (T *)malloc(dataSize * sizeof(float) * 3);
     size_t index = 0;
-    for (int i = 0; i < dims[0]; ++i)
+    for (int i = 0; i < (int)dims[0]; ++i)
     {
         size_t prev_i = std::max(0, i - 1), next_i = std::min((int)dims[0] - 1, i + 1);
-        for (int j = 0; j < dims[1]; ++j)
+        for (int j = 0; j < (int)dims[1]; ++j)
         {
             size_t prev_j = std::max(0, j - 1), next_j = std::min((int)dims[1] - 1, j + 1);
-            for (int k = 0; k < dims[2]; ++k)
+            for (int k = 0; k < (int)dims[2]; ++k)
             {
                 size_t prev_k = std::max(0, k - 1), next_k = std::min((int)dims[2] - 1, k + 1);
                 // curl[0] = dv3 / dy - dv2 / dz
--- a/source/adios2/toolkit/derived/Function.cpp
+++ b/source/adios2/toolkit/derived/Function.cpp
@@ -81,20 +81,6 @@ T *ApplyCurl(const T *input1, const T *input2, const T *input3, const size_t dim
     return data;
 }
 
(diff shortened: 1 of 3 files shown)


anagainaru:
Also @eisenhauer I am very disturbed that we use VarRec for two reasons in the same file (`BP5Deserializer.cpp`)
```c++
Line 1574:         auto VarRec = (struct BP5VarRec *)Req->VarRec;
Line 1730:                                     BP5VarRec *VarRec = VarByName.at(varBase->m_Name);
```

Both in the same function (`BP5Deserializer::GenerateReadRequests(const bool doAllocTempBuffers, size_t *maxReadSize)`) 

It makes everything hard to read, should we rename the variable on line 1730 to `VarPrimaryRec` or some name to indicate that this is the rec of the composing variable?

eisenhauer:
> Also @eisenhauer I am very disturbed that we use VarRec for two reasons in the same function (in `BP5Deserializer.cpp`)
> 
> ```c++
> Line 1574:         auto VarRec = (struct BP5VarRec *)Req->VarRec;
> Line 1730:                                     BP5VarRec *VarRec = VarByName.at(varBase->m_Name);
> ```

I completely agree.  Actually this is my fault for not trying this in CI with ADIOS2_USE_Derived_Variable set to on everywhere.  At least one of the compilers in our CI would have warned about the nested variable hiding the more global one.

anagainaru:
I compiled your PR before approving but didn't see the warning either :) so both of our faults.

I'll change the name and push over this tonight. I'll wait until Norbert's PR gets merged so I can rebase and not have all these failures.

anagainaru:
This PR can be reviewed

source/adios2/toolkit/derived/Function.cpp
```diff
@@ -112,7 +98,7 @@ DerivedData AddFunc(std::vector<DerivedData> inputData, DataType type)
                                                 [](T a, T b) { return a + b; });                   \
         return DerivedData({(void *)addValues, inputData[0].Start, inputData[0].Count});           \
     }
-    ADIOS2_FOREACH_PRIMITIVE_STDTYPE_1ARG(declare_type_add)
```
eisenhauer, line 115: Not that this is wrong, but curious about why the change from PRIMITIVE_STDTYPE to PRIMITIVE_ATTRIBUTE_STDTYPE.   This just drops the complex types, from being function candidates.  Is there some problem with allowing them?

anagainaru: It is temporary until we decide how to allocate memory for the derived variables. Right now, I didn't want to deal with complex types when using malloc and memset. I added it as a separate commit so I can revert only this one

eisenhauer (approved):
Looks good then.

vicentebolea:
This PR cannot be backported since it depends on another PR that breaks ABI/API comp #4228

anagainaru:
This would mean that we will be able to read derived variables only from 2.11 on. For now we can just write them and/or store stats. I am fine with this. I will mark both this and #4228 for release 2.11

Issue #4082, opened 2024-03-09
MPI variants of ADIOS2 libraries always require MPI::MPI_C in CMake


lahwaacz:
**Describe the bug**
MPI variants of ADIOS2 libraries always require the `MPI::MPI_C` target in CMake, even when `MPI::MPI_CXX` would be more appropriate.

Recall that [FindMPI](https://cmake.org/cmake/help/latest/module/FindMPI.html) says:

> The module exposes the components C, CXX, MPICXX and Fortran. Each of these controls the various MPI languages to search for. The difference between CXX and MPICXX is that CXX refers to the MPI C API being usable from C++, whereas MPICXX refers to the MPI-2 C++ API that was removed again in MPI-3.

Hence, when building a C++ project and linking to e.g. `adios2::cxx11_mpi`, the `MPI::MPI_CXX` target is the best option and `MPI::MPI_C` might even be unavailable (see the reproducer below).

**To Reproduce**
Create the following `CMakeLists.txt` and an empty `foo.cpp` file:
```cmake
cmake_minimum_required(VERSION 3.28.3)
project(foo LANGUAGES CXX)
add_executable(foo foo.cpp)

find_package(MPI COMPONENTS CXX REQUIRED)
find_package(ADIOS2 REQUIRED)
target_link_libraries(foo PUBLIC MPI::MPI_CXX adios2::cxx11_mpi)
```
Try to configure the project:
```
$ cmake -B build -S .
-- Found MPI_CXX: /usr/lib/libmpi.so (found version "3.1")
-- Found MPI: TRUE (found version "3.1") found components: CXX
-- Found MPI: TRUE (found version "3.1")
-- Found ADIOS2: /usr/lib/cmake/adios2/adios2-config.cmake (found version "2.9.2") found components: CXX MPI
-- Configuring done (0.6s)
CMake Error at /usr/lib/cmake/adios2/adios2-cxx11-targets.cmake:78 (set_target_properties):
  The link interface of target "adios2::cxx11_mpi" contains:

    MPI::MPI_C

  but the target was not found.  Possible reasons include:

    * There is a typo in the target name.
    * A find_package call is missing for an IMPORTED target.
    * An ALIAS target is missing.

Call Stack (most recent call first):
  /usr/lib/cmake/adios2/adios2-config-common.cmake:199 (include)
  /usr/lib/cmake/adios2/adios2-config.cmake:29 (include)
  CMakeLists.txt:6 (find_package)


[...]

PR #4255, opened 2024-07-22
cmake: use MPI_CXX for targets of CXX sources

vicentebolea:
Before CMAKE 3.10 MPI::MPI_CXX was the C++ deprecated MPI library. After cmake 3.10 MPI::MPI_CXX is the MPI C library but specifics to be used with CXX targets. In this PR I move to use `MPI::MPI_CXX` since keep using `MPI::MPI_C` unnecessary forces adios2 applications to initialize C language.

I am also adding another installable INTERFACE library named adios2_core_mpi_c which includes adios2_core_c but avoids the MPI::MPI_CXX target dependency and and the MPI::MPI_C target. This is needed so that cmake C projects do not need to enable CXX for building their adios2 application.

Fixes https://github.com/ornladios/ADIOS2/issues/4082


testing/install/CXX11/CMakeLists.txt
```diff
@@ -15,11 +15,8 @@ target_link_libraries(adios_cxx11_test adios2::cxx11)
 add_test(NAME adios_cxx11_test COMMAND adios_cxx11_test)
 
 if(ADIOS2_HAVE_MPI)
-  # Avoid using MPI::MPI_CXX
-  enable_language(C)
```
vicentebolea, line 19: @lahwaacz

eisenhauer (approved):
Looks OK...

PR #4251, opened 2024-07-19
Don't print kvcache connection closed message when it was not opened.

pnorbert:

PR #4256, opened 2024-07-22
cmake,ci: disable kvcache

vicentebolea:
#4250  broke the CI, the reason is that we did not test this kvcache enable in the corresponding CI due to a typo. 

Disabling kvcache until we resolve this


eisenhauer (approved):

anagainaru (commented):
I thought #4251 solves this. We just need to merge it.

Issue #3754, opened 2023-08-14
ADIOS2 incompatible with latest MGARD


vicentebolea:
When using the latest mgard from spack `    2023-03-31    [git] https://github.com/robertu94/MGARD at commit a8a04a86ff30f91d0b430a7c52960a12fa119589`

```
spack install mgard
```

I get the following errors when running `ctest -R MGARD`:

```
2023-08-12T22:24:10.9023580Z 
2023-08-12T22:24:10.9024327Z 1179: Test command: /__w/ADIOS2/ADIOS2/ubuntu20.04-gcc10-ompi/bin/Test.Engine.BP.WriteReadMGARDPlus.Serial "--gtest_filter=*/BPWriteReadMGARDPlus.BPWRMGARD3D/*" "BP5"
2023-08-12T22:24:10.9024990Z 1179: Working Directory: /__w/ADIOS2/ADIOS2/ubuntu20.04-gcc10-ompi/testing/adios2/engine/bp/operations/bp5
2023-08-12T22:24:10.9025307Z 1179: Test timeout computed to be: 120
2023-08-12T22:24:10.9455266Z 1184: Note: Google Test filter = */BPWriteReadBZIP2.ADIOS2BPWriteReadBZIP23D/*
2023-08-12T22:24:10.9455873Z 1184: [==========] Running 9 tests from 1 test suite.
2023-08-12T22:24:10.9456677Z 1184: [----------] Global test environment set-up.
2023-08-12T22:24:10.9457243Z 1184: [----------] 9 tests from BZIP2Accuracy/BPWriteReadBZIP2
2023-08-12T22:24:10.9457805Z 1184: [ RUN      ] BZIP2Accuracy/BPWriteReadBZIP2.ADIOS2BPWriteReadBZIP23D/0
2023-08-12T22:24:10.9537281Z 1179: Note: Google Test filter = */BPWriteReadMGARDPlus.BPWRMGARD3D/*
2023-08-12T22:24:10.9538609Z 1179: [==========] Running 4 tests from 1 test suite.
2023-08-12T22:24:10.9540757Z 1179: [----------] Global test environment set-up.
2023-08-12T22:24:10.9543426Z 1179: [----------] 4 tests from MGARDAccuracy/BPWriteReadMGARDPlus
2023-08-12T22:24:10.9545405Z 1179: [ RUN      ] MGARDAccuracy/BPWriteReadMGARDPlus.BPWRMGARD3D/0
[...]

PR #4262, opened 2024-07-24
ci: update gcc11 images to ubuntu22.04

vicentebolea:
Fixes: [#3754](https://github.com/ornladios/ADIOS2/issues/3754)

We need this since we need to test with newer mgard which its latest release now works with adios2.

I have also done some refactoring with the builds.

The red build is because I set the build name as tmp, after approval I will retag the images and the error should go away


vicentebolea:
@pnorbert it seems that we are still incompatible with  mgard 2023/03 perhaps we do not have the issue with versions older than that?

eisenhauer (approved):

PR #4267, opened 2024-07-25
KVCache ctest and cache key optimization

Change72:
1. Optimize cache key with filename to support cache different datasets.
      cached key: Md5(remote file path) + variable_name + step + start + count

2. kvcache ctest together with remote server


pnorbert:
Just an observation, a variable name can also potentially be long, there is not much restriction how deep hierarchy one can emulate with a long name. I thought you would make a hash for path+varname.

Change72:
If so, for each request, there will be a hash calculation, which might bring (a lot) extra cost.

In comparison, if only hash the path, there will be only one hash calculation in one whole PerformGets.

eisenhauer:
Just a thought here...  In any remote situation we have separated metadata from data and it would be good to make sure that the metadata we have locally (used to generate the request) is the same as the metadata and data in the remote file (seen by the server), because if they get out of sync because the file has been rewritten bad things are likely to happen.  To this end, it might be wise to have a randomly generated "file id" or fingerprint in the metadata that we can check against when we open.  Such a fingerprint might make a better key than the remote file path.  (At least "better" in the sense that you don't have to do an MD5 at all, just use the value in the metadata.  This doesn't address Norbert's concerns about long variable names (which certainly occur, perhaps more often than you might think).)

Change72:
I think such mapping or fingerprints should be managed by the campaign manager, and cache only records `.aca` file info. But this also has a problem with the access without the campaign.

eisenhauer:
Hmm.  I'd have to see the case for getting the .aca file involved.   For metadata consistency checking, I'd just see the fingerprint of local metadata compared to that of remote metadata and throw an exception if they don't match (which they always should, unless you've changed the file (even if it has the same path) and thus invalidated the metadata).   No need to campaign there.  I would think this would be very similar for the KVcache.  It's just a way to make sure that you're matching entries appropriately.

pnorbert:
How about the BP5 engine calculates this hash/fingerprint in local open based on path + metadata (both on remote server, and when using a campaign, the metadata is local). The cache can use this fingerprint as hash. Then we can later add another PR, where the locally generated hash (from campaign's metadata and remote path) is compared to the remotely generated hash (that the remote server will need to send back) and throw a tantrum when they don't match. This will ensure that the cache is valid only if the remote data is not changed, outdated campaign files will be caught (with exception), and updated campaigns on updated/replaced remote data will work properly by not using old cached data. 

So let's this PR in after an BP5Reader class member variable is used to store the hash that is right now inside the KVCache class.

pnorbert (approved):

eisenhauer:
I'm not sure I buy the rationality of basing the fingerprint on a path vs. simply a random value generated when the file is created (or rewritten).  Files can be moved, accessed via multiple paths, get modified or overwritten with different data, etc.  All of these things are problems for a path-based fingerprint, but are things you probably want to allow or at least not be broken by.  The only time (I can think of) that a random value is at a disadvantage over path-based is if the remote file happens to be overwritten by exactly the same data.  Not that that might not be common in development, but in production I have to think it'd be rare.  A cryptographic hash based over the full contents of the file, metadata and data included, would be superior to either, but I don't think we'd care for the cost.

pnorbert:
I meant a hash based on path + metadata, so basically based on the time of close. If it is generated at time of creation, that would work well too. In this case we need to add this somewhere into the BP5 format and retrieve it during Open in the reader engine.

PR #4282, opened 2024-08-01
Extend Derived Variable Expressions: Trig functions

lizdulac:
Extend Derived Variable Expressions to include sin, cos, tan, asin, acos, atan


source/adios2/toolkit/derived/Function.cpp
```diff
@@ -39,6 +39,23 @@ T *ApplyOneToOne(Iterator inputBegin, Iterator inputEnd, size_t dataSize,
...
+                                             "Error allocating memory for the derived variable");
+    }
+    for (size_t i = 0; i < dataSize; i++)
+    {
+        T data = *(reinterpret_cast<T *>(inputData + i));
+        outValues[i] = compFct(data);
+    }
+    return outValues;
+}
+
```
anagainaru, line 58: This function is basically a `std::transform`. Check how sqrt is implemented now and update the trig functions using the same method.

anagainaru (changes requested):

anagainaru (changes requested):
Trig functions have the same issue as pow/sqrt so the type function used should not be `SameType`.
Also, see my comment on `ApplyOneToOneOnce`

anagainaru (commented):
The new version looks good, I turned ON the derived variables to make sure all tests pass.

anagainaru (dismissed):

source/adios2/toolkit/derived/Function.cpp
anagainaru, line 58:
```suggestion
```

anagainaru (approved):

PR #4291, opened 2024-08-06
Relaxing the restriction of the derived variables to have the same type as the operands

anagainaru:
For derived operations that do not have the same input and output type, e.g. `float sqrt(int)`

DerivedData contains now in addition to the data pointer and the start and count Dims, the type of each input operand. The derived functions will have the following signature:
```
struct DerivedData
{
    void *Data;
    Dims Start;
    Dims Count;
    DataType Type;
};
    std::function<DerivedData(std::vector<DerivedData>, DataType)> ComputeFct;
```

The second argument `DataType` is the output type.

### DIFF
--- a/source/adios2/core/IO.cpp
+++ b/source/adios2/core/IO.cpp
@@ -890,8 +890,8 @@ VariableDerived &IO::DefineDerivedVariable(const std::string &name, const std::s
 
     derived::Expression derived_exp(exp_string);
     std::vector<std::string> var_list = derived_exp.VariableNameList();
-    DataType expressionType = DataType::None;
     bool isConstant = true;
+    std::map<std::string, DataType> name_to_type;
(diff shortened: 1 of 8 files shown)


eisenhauer (commented):
Any testing?  Or just taking this piece-wise?

(I won't approve yet because you need to turn derived variables back off and when you push a commit to do that it'll dismiss any reviews anyway.)

anagainaru:
> Any testing? Or just taking this piece-wise?
I tested it separately with the POW/SQRT functions that are coming in #4248 

I wanted this a separate PR because it creates a backbone for all other functions that need this, not just pow/sqrt (in case we squash merge that one). A slightly complicated way of organizing this, but it made sense at the time.

> (I won't approve yet because you need to turn derived variables back off and when you push a commit to do that it'll dismiss any reviews anyway.)
Yes, I'll turn it OFF once I confirm there's no error.

PR #4294, opened 2024-08-07
KVCache use MinBlocksInfo to get start/count in query

Change72:
Use MinBlocksInfo to get start/count in the query (touched blocks). The start/count will be saved in KVCache and will be used in partial overlap cases in the future.

In the current code, `MinBlocksInfoMap` will temporarily hold the `minBlocksInfo` in memory in order to avoid the repeat `MinBlocksInfo(*VB, Req.RelStep);` for the same variable.

However, `MinBlocksInfoMap` only works for `mode::deferred`, and not for `mode::Sync`.


pnorbert (dismissed):
Just move MinBlocksInfoMap to be a private class member, and free it up in `EndStep()`. Then it should work with Sync mode as well.

pnorbert (approved):

pnorbert (approved):

Issue #4297, opened 2024-08-09
CMake scripts for the MPI dataplane of SST don't work


franzpoeschel:
**Describe the bug**
The MPI dataplane is not correctly propagated from `cmake/DetectOptions.cmake` to `source/adios2/toolkit/sst/CMakeLists.txt`. There are three separate issues:

1. The availability of the dataplane is not correctly inferred upon configuring (at least on Frontier). A workaround for this is to replace `if (ADIOS2_HAVE_MPI_CLIENT_SERVER)` in `cmake/DetectOptions.cmake` with `if (TRUE)`.
2. `cmake/DetectOptions.cmake` sets `ADIOS2_HAVE_MPI_DP`, but `source/adios2/toolkit/sst/CMakeLists.txt` tries reading `ADIOS2_SST_HAVE_MPI` (without `_DP`). Adding a line `set(ADIOS2_SST_HAVE_MPI TRUE)` in `cmake/DetectOptions.cmake` works around that.
3. `source/adios2/toolkit/sst/CMakeLists.txt` tries reading a variable `ADIOS2_SST_BUILD_MPI_DP` that is never set. 

For building on Frontier with the MPI DP, I currently use:
```diff
diff --git a/cmake/DetectOptions.cmake b/cmake/DetectOptions.cmake
index 24e8470b4..a08910a54 100644
--- a/cmake/DetectOptions.cmake
+++ b/cmake/DetectOptions.cmake
@@ -491,8 +491,9 @@ if(ADIOS2_USE_SST AND NOT WIN32)
         }]=]
     ADIOS2_HAVE_MPI_CLIENT_SERVER)
     unset(CMAKE_REQUIRED_LIBRARIES)
-    if (ADIOS2_HAVE_MPI_CLIENT_SERVER)
+    if (TRUE)
       set(ADIOS2_SST_HAVE_MPI_DP TRUE)
+      set(ADIOS2_SST_HAVE_MPI TRUE)
     else()
       if ("${ADIOS2_SST_EXPECT_MPI_DP}") 
           message(FATAL_ERROR "Expected MPI to support Client-server connection model, but test failed.")
diff --git a/source/adios2/toolkit/sst/CMakeLists.txt b/source/adios2/toolkit/sst/CMakeLists.txt
index 4a8252eee..0c13ecc5d 100644
--- a/source/adios2/toolkit/sst/CMakeLists.txt
+++ b/source/adios2/toolkit/sst/CMakeLists.txt
[...]

PR #4298, opened 2024-08-10
Rationalize SST_HAVE_MPI_DP

eisenhauer:
This fixes part of issue #4297.   The first part, correctly detecting MPI client-server on Frontier, I'm going to kick to @vicentebolea.  However @franzpoeschel, you can also just tell CMake that it has client-server on the command line as a workaround.  "cmake -DADIOS2_HAVE_MPI_CLIENT_SERVER=1 ..."


pnorbert (approved):

vicentebolea:
Oh we did have the variable `ADIOS2_SST_HAVE_MPI_DP` that was used for the same thing, we could have use that variable in `source/adios2/toolkit/sst/dp/dp.c`

PR #4303, opened 2024-08-14
Update the script to build adios with Kokkos for Frontier

anagainaru:
The gcc and cray modules on Frontier used in the build scripts have been deprecated as of July 2024:

```
gcc/11.2.0:
   As of July 16, 2024, the 'gcc' module name is deprecated and has been replaced by 'gcc-native'. Running 'module load gcc' from a
   default login shell has undefined behavior due to this name change. Please use 'module load PrgEnv-gnu gcc', or optionally 'module
   load PrgEnv-gnu cpe/23.09' to load older versions of the GNU toolchain. See
   https://docs.olcf.ornl.gov/software/software-news.html#frontier-system-software-update-july-16-2024 for more information.
```

### DIFF
--- a/scripts/build_scripts/build-adios2-kokkos-frontier.sh
+++ b/scripts/build_scripts/build-adios2-kokkos-frontier.sh
@@ -1,10 +1,8 @@
 #!/bin/bash
 # shellcheck disable=SC2191
 
-module load rocm/5.4.0
-module load gcc/11.2.0
+module load PrgEnv-gnu-amd/8.5.0
 module load cmake/3.23.2
-module load craype-accel-amd-gfx90a
 
 ######## User Configurations ########
 Kokkos_HOME=$HOME/kokkos/kokkos
--- a/scripts/build_scripts/build-adios2-kokkos-frontier.sh
+++ b/scripts/build_scripts/build-adios2-kokkos-frontier.sh
@@ -2,6 +2,7 @@
 # shellcheck disable=SC2191
 
 module load PrgEnv-gnu-amd/8.5.0
+module load craype-accel-amd-gfx90a
 module load cmake/3.23.2
 
 ######## User Configurations ########


eisenhauer (approved):

anagainaru:
@eisenhauer Any ideas why the mac-os CIs fail?

eisenhauer:
> @eisenhauer Any ideas why the mac-os CIs fail?

It looks like maybe the Mac images have changed and gfortran-11 isn't there anymore:

CMake Error at /usr/local/Cellar/cmake/3.30.2/share/cmake/Modules/CMakeDetermineFortranCompiler.cmake:33 (message):
    Could not find compiler set in environment variable FC:
    gfortran-11.

@vicentebolea ?

anagainaru:
Yeah, it doesn't seem related to the changes here, so maybe you could bypass the requirements and merge this. We can open an issue with this for @vicentebolea

scripts/build_scripts/build-adios2-kokkos-frontier.sh
```diff
@@ -1,10 +1,8 @@
 #!/bin/bash
 # shellcheck disable=SC2191
 
-module load rocm/5.4.0
```
vicentebolea, line 4: craype-accel-amd-gfx90a is needed for the rocm devices to work

anagainaru: I'll add it back and see if this is the issue.
It's strange though because I was able to run with HIP on Frontier with just `PrgEnv-gnu-amd/8.5.0`

anagainaru: This was not the issue and it doesn't seem we need `craype-accel-amd-gfx90a`
I will look again tonight but I will remove it if all our tests work without it.

eisenhauer (approved):
If this is settled and working, then it looks good.

anagainaru:
> If this is settled and working, then it looks good.

Yes, Vicente is right , we need `craype-accel-amd-gfx90a` if we want to use the cray-compiler as your HIP compiler so I left it. I am very confused by the CI, but I think you can bypass it, or I can try to retrigger it.

PR #4304, opened 2024-08-14
Fix argument ordering in FLEX_TARGET

gsjaardema:
The output file is the third item in the FLEX_TARGET macro.  Fixed the ordering

### DIFF
--- a/source/adios2/CMakeLists.txt
+++ b/source/adios2/CMakeLists.txt
@@ -150,8 +150,8 @@ if (ADIOS2_HAVE_Derived_Variable)
       DEFINES_FILE ${CMAKE_CURRENT_BINARY_DIR}/parser.h)
     FLEX_TARGET(MyScanner
       toolkit/derived/parser/lexer.l
-      COMPILE_FLAGS "-o lexer.cpp --header-file=lexer.h" 
       ${CMAKE_CURRENT_BINARY_DIR}/lexer.cpp
+      COMPILE_FLAGS "--header-file=lexer.h" 
       DEFINES_FILE ${CMAKE_CURRENT_BINARY_DIR}/lexer.h)
     ADD_FLEX_BISON_DEPENDENCY(MyScanner MyParser)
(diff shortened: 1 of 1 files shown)


eisenhauer:
Can you tell us where this was a problem?

eisenhauer:
Haven't seen it here, but at least in the CI you point to the CMake version is newer than what we're using.  That might be the difference.   In your PR, you change two things, the location of the COMPILE_FLAGS line and it's contents.   I think only that only changing the location is necessary (so that it doesn't split the two required FLEX_TARGET arguments).   Can you restore the original COMPILE_FLAGS ("-o lexer.cpp --header-file=lexer.h") and see if that is sufficient?

gsjaardema:
I didn't know whether specifying the output file twice (once with the `-o lexer.cpp` option and once in the required argument  `${CMAKE_CURRENT_BINARY_DIR}/lexer.cpp` would cause issues or confusion.

But, I can add back the `-o lexer.cpp` arguent...

eisenhauer:
Ah, you must be pulling from the master branch and not from any release, right?  It seems that this portion of ADIOS, with ADIOS2_USE_Derived_Variable, is turned on by accident.   A recent PR enabled it by default when it should be disabled by default.   I've submitted a PR to fix that.  The Derived_Variable support in ADIOS is still experimental and had it been properly turned off you'd not have seen this.  If you don't mind, I'm going to close this PR, merge the one that turns this experimental feature back off, and make that feature's developers aware of this problem so that it can go into their branches.   The ADIOS master branch should be updated with this feature turned off as soon as I can push that PR through CI.

eisenhauer:
And, BTW, good point about specifying the output in both places.  Not great practice even if it doesn't cause a problem.

gsjaardema:
That sounds fine.  Yes, we are using the master branch for some reason that I don't remember why...

eisenhauer:
Thanks

PR #4305, opened 2024-08-14
Try MacOS gfortran fixup

eisenhauer:


eisenhauer:
@vicentebolea It appears that gfortran-11 has vanished on macos CI.  It's old anyway.  This tweaks it up to v12 only on macos to get CI running again.  Maybe we want to change so we have the same version everywhere, and maybe we want to go to something more recent than 12.  Thoughts?

vicentebolea (approved):
Looks great!

PR #4312, opened 2024-08-19
Implementing reader side derivation version 2

eisenhauer:
This PR takes a different approach to getting writer-side metadata for uncalculated variables than does #4249.  However, as currently written is is broken because Expression SetDims() doesn't handle Start correctly.


anagainaru:
This PR can be closed. #4249 solves what this PR was doing and it was merged.

PR #4317, opened 2024-08-21
Updating very old tests

eisenhauer:
There were tests in testing/examples/heatTransfer that hadn't been updated forever.  I.E. they were testing BP3/4, but not 5, were testing SST with BP and FFS marshalling methods (neither maintained), etc.  This PR simplifies a bit, reduces SST testing to the default marshalling method, switches out bp3 and adds bp5.


anagainaru (dismissed):
Looks good to me

eisenhauer:
@anagainaru finally cleaned up mistaken substitutions here.  Should be ready to go.

anagainaru (approved):
There are some references to B4 in the documentation of heatmap as well if you want to update everything:

```
The ADIOS2 executable needs an XML config file to select the Engine used for the output. The engines are: File, BP4 and HDF5, the corresponding XML config files are in the examples/simulations/heatTransfer/ directory. The "File" engine will be BP4 or HDF5 depending on the extension of the file name.
```

PR #4328, opened 2024-08-28
New API to allow storing only stats for primary variables

anagainaru:
Add a bool in `VariableBase` to control what we write and corresponding APIs for CXX, C, Python, Fortran.

CXX API
```c++
varI32.StoreStatsOnly(true);
```

C bindings:
```C
adios2_store_stats_only(varI32, adios2_true);
```
Fortran bindings:
```Fortran
call adios2_store_stats_only(varI32, .true., ierr)
```

Python bindings
```python
varI32.StoreStatsOnly(True)
```

The reader can inquire variables and get stats about each block but will not be able to read data. A `std::runtime_error` exception will be thrown if data is read for a `StatsOnly` variable.

```
$ ./bin/bpls -ld ADIOS2BPWriteReadStatsOnly.bp
  float    r32   {8} = 1.1 / 8.1

bpls caught an exception
No data exists for this variable
```


anagainaru:
Ready for review

pnorbert (commented):
I feel like the function name does not convey what it is about. How about an inverse approach, like SetStoreMetadataOnly(bool)? 

One thing is missing though, the new python API in python/adios2/variable.py. New tests should just use that, which will exercise the adios2.bindings indirectly as well.

anagainaru:
@pnorbert this is ready for review.

vicentebolea:
Just an idea: sometime feels better to have `storeMetadataOnlyOn() storeMetadataOnlyOff()`

anagainaru:
> Just an idea: sometime feels better to have `storeMetadataOnlyOn() storeMetadataOnlyOff()`

Normally I would agree, but in this case I don't want users to be able to turn on and off. We either store only stats or we store everything. We talked about the API and decided we do not want to change the `DefineVariable` call to add this option (the cleanest choice), but rather include a new function.

pnorbert (approved):

pnorbert:
> > Just an idea: sometime feels better to have `storeMetadataOnlyOn() storeMetadataOnlyOff()`
> 
> Normally I would agree, but in this case I don't want users to be able to turn on and off. We either store only stats or we store everything. We talked about the API and decided we do not want to change the `DefineVariable` call to add this option (the cleanest choice), but rather include a new function.

I believe he meant the same functionality just changing syntax from having a bool argument to separate functions. Graphics packages are full of these. I don't feel better, however, seeing that in adios as a first.

PR #4331, opened 2024-08-29
Fix in python API: 

pnorbert:
a. BlocksInfo(name, step)  
b. stream.write() sets selection if either 'start' or 'count' is provided (fix for local arrays).

This code below shows the original problem when enumerating the blocks of the `v3` variable in the localArray example, where the number of blocks changes across the steps. 
```
#!/usr/bin/env python3
import numpy as np
import adios2

def process_var(varname):
    print("--------------------------")
    print(f"Process variable {varname}")
    v = f.inquire_variable(varname)
    # b = f.all_blocks_info(varname)
    for step in range(v.steps()):
        v.set_step_selection([step, 1])
        bs = f.engine.blocks_info(varname, step)
        for bid in range(len(bs)):   # could use here len(b[step])
            v.set_step_selection([step, 1])
            v.set_block_selection(bid)
            data = f.read(v)
            print(f"step {step} block {bid} len = {len(data)}")


with adios2.FileReader("localArray.bp") as f:
    process_var('v0')
    process_var('v1')
    process_var('v2')
    process_var('v3')
```

### DIFF
--- a/python/adios2/stream.py
+++ b/python/adios2/stream.py
@@ -315,7 +315,7 @@ class Stream:
         if shape != [] and not variable.single_value():
             variable.set_shape(shape)
 
-        if start != [] and count != []:
+        if start != [] or count != []:
             variable.set_selection([start, count])
 
         if operations:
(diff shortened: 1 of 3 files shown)


python/adios2/stream.py
```diff
@@ -315,7 +315,7 @@ def _(self, name, content, shape=[], start=[], count=[], operations=None):
         if shape != [] and not variable.single_value():
             variable.set_shape(shape)
 
-        if start != [] and count != []:
+        if start != [] or count != []:
```
anagainaru, line 318: I thought start and count need to match in length

pnorbert: For local arrays, start is empty, just like shape.

PR #4336, opened 2024-09-05
Korvo upstream - enable building without perl installed

eisenhauer:


vicentebolea (approved):
looks great @eisenhauer thanks for the quick turnaround time

eisenhauer:
FYI @vicentebolea, I did kill perl on an ubuntu VM to try this and ADIOS built properly, so I'm hopeful that this will work for you.

PR #4338, opened 2024-09-09
If there's an exception in GenerateReadRequests, clear state and rethrow

eisenhauer:
This looks like a lot, but it's just adding "try {" at the beginning and this at the end:
```
    catch (...)
    {
        std::exception_ptr ex = std::current_exception();
        // if something in GenerateReadRequests threw an exception, clear the PendingGetRequests so
        // it doesn't happen again (I.E. if this happened in PerformGets(), then EndStep() should
        // then happen cleanly.)
        PendingGetRequests.clear();
        std::rethrow_exception(ex);
    }
```


anagainaru (approved):
Perfect

PR #4341, opened 2024-09-10
Backport #3759 to Release 2.10 branch. 

pnorbert:
(cherry-picked from master branch #4340)

Organize the processes into groups so that the two steps of metadata aggregation has more or less the same number or participants. This replaces in-node aggregation in first step. The new strategy balances the size of metadata gathered in the two steps.


eisenhauer (approved):

eisenhauer:
If I were guessing, I'd say that the runner images have changed, master CI has been updated to fix it and release_210 hasn't.  Would need to look for the PR that fixed it, probably merged to master but not backported.

eisenhauer:
I haven't sorted out the PR that changed it yet, but we no longer build on macos-11 or -12 in master.  I'm going to bypass branch protections to go ahead and merge this PR, but if we're going to keep updating release_210, we need to sort the CI.

PR #4343, opened 2024-09-11
Setting the memory space for Spans to Host

anagainaru:
Memory selections when GPU pointers are used for BP5 copies the data to internal buffers before min/max is applied. I am updating the memory space for spans to Host when data is moved to temp buffers.


eisenhauer (approved):
This looks like the right fix...

PR #4344, opened 2024-09-11
Support for 4D arrays in ZFP compressor

gregorweiss:
This PR introduces a small yet potentially valuable enhancement for users who work with 4D data arrays and require support for ZFP data compression.

While working on data compression for Neko, we encountered the need to compress 4D data arrays using the ZFP compressor, specifically the `zfp_field_4d()` function. This feature is available starting with ZFP version 0.5.4 <https://zfp.readthedocs.io/en/release0.5.4/versions.html>

The current ADIOS2 release supports ZFP version 0.5.3. Therefore, a CMake source check has been introduced to conditionally enable this feature if the system's ZFP version is 0.5.4 or higher.

The branch has been successfully tested with our Neko use case.


cmake/DetectOptions.cmake
```diff
@@ -137,13 +138,32 @@ if(ZFP_FOUND)
       return !zfp_stream_set_execution(stream, zfp_exec_cuda);
     }"
     ADIOS2_HAVE_ZFP_CUDA)
+
```
vicentebolea, line 141: IS this needed, we can check if zfp version is greater than 0.5.4 to check that we support 4D functions. 

https://github.com/LLNL/zfp/commit/83402a16fb71106d44fe6253195fb9ec03728e4e

vicentebolea (changes requested):

cmake/DetectOptions.cmake
gregorweiss: Commit 9b75283 solves it through checking the version which involves a cumbersome preprocessor if clause. So checking for feature availability seemed to be the option that would not clutter the source file with version checking. Certainly, a rollback to 9b75283 is possible.

vicentebolea: @gregorweiss it appears to me that the https://github.com/ornladios/ADIOS2/commit/9b75283a15d72ac353f908591ba91ce36d139e5a approach is takes less LOC than this and it is a stronger alternative. 

> that would not clutter the source file with version checking

Cmake code is also source code

The rationale for it to be a better alternative is that:
 - It is faster, no need to build/run test in the configure step.
 - The build/run step could fail for other reasons unrelated to whether it supports 4d arrays.

Finally, the reason of using a build/run test for checking if a feature of a lib is available/running is the last resort that we have when we have no other way of checking this directly in the source code.

gregorweiss: @vicentebolea Thank you for sharing your train of thought. Rollback is ready. There was also a flaw that I fixed in the if-condition of commit 9b75283. The new commit should check that version number is greater than 0.5.3.

vicentebolea: looks good to me, please fix format and Ill approve this. Many thanks!

vicentebolea (approved):

vicentebolea:
feel free to merge

gregorweiss:
Test `1130: Bindings.C.ADIOS2_C_API.ADIOS2BPWriteStatsOnly.MPI` in Ubunut20.04, gcc8, ompi is failing after merging. I don't see that this test would cover the code in this PR. What are your thoughts?

eisenhauer:
Probably not a real failure.  I've forced a rerun of that job.

vicentebolea:
WriteStateOnly is a new test, it maybe flaky. It should not hold this PR from being merged.

anagainaru:
> WriteStateOnly is a new test, it maybe flaky. It should not hold this PR from being merged.

I agree it should now. But the test should not be flaky, what was the error, I didn't see it.

eisenhauer:
The reader reported steps=0, then a series of failures after that.  Perhaps the writer died prematurely.  But a rerun fixed it.

PR #4345, opened 2024-09-12
Updating the testing for the GPU backend

anagainaru:
Adding Kokkos and updating CUDA for the latest changes.


anagainaru:
I am in favor of removing the CUDA testing. It's more code to take care of and we are only testing that the CUDA function does the same as Kokkos (I am splitting between them down the line so we don't need to check high level functionality).

And anyway no CI is triggering CUDA testing anyway so there is no point in keeping it.

vicentebolea:
Note that we have build coverage in CI for CUDA. This has some value by itself. Also we might be able to get a CUDA runner later on through the Spack CI.

anagainaru:
And I'm all for testing, but right now we are not trigger it anywhere and it's just code that needs to be maintained. The CUDA tests would not pass right now because we changed the way the backend is used since we did the testing but nobody noticed because we are not triggering it. I'll add it back once we decide to trigger the testing. At least the Kokkos one I can run from time to time on Frontier.

> Note that we have build coverage in CI for CUDA. This has some value by itself. Also we might be able to get a CUDA runner later on through the Spack CI.

vicentebolea (approved):

vicentebolea:
This change added regression at master https://open.cdash.org/builds/9940686

vicentebolea:
We missed this because the frontier CI was unavailable for a few weeks

PR #4352, opened 2024-09-26
Checks and changes in Get(variable, datum)

eisenhauer:
For Get(variable, datum), check that variable shape is Global or Local Value, pass launch mode.  Modify test to cover the situation.

### DIFF
--- a/source/adios2/core/Engine.tcc
+++ b/source/adios2/core/Engine.tcc
@@ -111,9 +111,14 @@ void Engine::Get(const std::string &variableName, T *data, const Mode launch)
 }
 
 template <class T>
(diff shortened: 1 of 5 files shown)


pnorbert (approved):

anagainaru (commented):
SST has a different logic?

eisenhauer:
> SST has a different logic?

SST and all the BP engines do.  That the others do not is a little weird, but I'm actually not entirely sure how firmly we documented the "Value Gets are populated immediately" semantic.  That there is an option for deferred (and it's even the default) implies that the semantics were at least fuzzy at some point.

PR #4357, opened 2024-10-04
Revert "Updating the testing for the GPU backend"

vicentebolea:
Reverts ornladios/ADIOS2#4345


anagainaru (changes requested):
I would not revert it, the testing was broken before. I ran it on Frontier and it was fine, let me reproduce this and fix it. If I can't I will completely remove it but not revert to what was before

anagainaru:
Fixed in #4358

PR #4358, opened 2024-10-07
Fix the Kokkos testing compilation error on Frontier

anagainaru:
Instead of #4357

If `hipcc` is used as the CXX compiler there is no error, but the CI is using `g++` so we need to manually set the CXX compiler to whatever was used to build Kokkos.

The rocm version had to be updated to `rocm/5.7.1` in the frontier CI due to an incompatibility between the g++ version in `PrgEnv-gnu/8.5.0` and the hipcc version in `rocm/5.4.3`

The Kokkos version had to be updated to > 4.0.0 due to a warning with 3.7.01 and newer rocm compilers.

PR #4359, opened 2024-10-08
Fix the code that copies wrongly from host to device GPU buffers instead of device to host

anagainaru:
NdCopy is used to copy data for Spans during write. However, NdCopy is also used by the reader to bring data to the user buffer. For buffers allocated on the CPU this is not a problem since the direction is not important. For the GPU backend however the copy needs to happen from the Host to the Device or from the Device to the Host depending on what operation is being done.

This PR add an extra argument to NdCopy that reads true if the function is called by the writer and false if called by the reader. 

The Kokkos backend did not have this problem because Kokkos::deep_copy internally detects the correct memory spaces and so it copies the data correctly. CudaMemcpy cannot do this. 

Part of this PR I also updated the Kokkos code to use explicit the correct memory space so we don't rely on the internal implementation (in case this will be changed in their future releases).

### DIFF
--- a/source/adios2/engine/bp5/BP5Writer.cpp
+++ b/source/adios2/engine/bp5/BP5Writer.cpp
@@ -1815,7 +1815,8 @@ void BP5Writer::PutCommon(VariableBase &variable, const void *values, bool sync)
         helper::NdCopy((const char *)values, helper::CoreDims(ZeroDims), MemoryCount,
                        sourceRowMajor, false, (char *)ptr, MemoryStart, varCount, sourceRowMajor,
                        false, (int)ObjSize, helper::CoreDims(), helper::CoreDims(),
-                       helper::CoreDims(), helper::CoreDims(), false /* safemode */, memSpace);
+                       helper::CoreDims(), helper::CoreDims(), false /* safemode */, memSpace,
+                       /* duringWrite */ true);
     }
     else
     {
--- a/source/adios2/helper/adiosMemory.cpp
+++ b/source/adios2/helper/adiosMemory.cpp
@@ -251,7 +251,7 @@ int NdCopy(const char *in, const CoreDims &inStart, const CoreDims &inCount,
            const CoreDims &outStart, const CoreDims &outCount, const bool outIsRowMajor,
            const bool outIsLittleEndian, const int typeSize, const CoreDims &inMemStart,
            const CoreDims &inMemCount, const CoreDims &outMemStart, const CoreDims &outMemCount,
-           const bool safeMode, MemorySpace MemSpace)
+           const bool safeMode, const MemorySpace MemSpace, const bool duringWrite)
 
 {
 
@@ -439,7 +439,7 @@ int NdCopy(const char *in, const CoreDims &inStart, const CoreDims &inCount,
             if (MemSpace == MemorySpace::GPU)
             {
                 helper::NdCopyGPU(inOvlpBase, outOvlpBase, inOvlpGapSize, outOvlpGapSize, ovlpCount,
-                                  minContDim, blockSize, MemSpace);
+                                  minContDim, blockSize, MemSpace, duringWrite);
                 return 0;
             }
 #endif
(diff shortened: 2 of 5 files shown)


anagainaru:
@franzpoeschel rebase on this PR and let me know if you still see the problem in https://github.com/ornladios/ADIOS2/discussions/4337

eisenhauer (approved):
Hmm.  I didn't know we supported reading directly into GPU memory.  Changes make sense though...

anagainaru:
The Frontier CI is failing because of the Kokkos testing (which is fixed in #4358) so I'll merge without Frontier passing.

franzpoeschel:
This fixes the issue that I had, memory selections for Cuda pointers work fine now. Thank you!

anagainaru:
> This fixes the issue that I had, memory selections for Cuda pointers work fine now. Thank you!

Looking forward to the next bug :)

vicentebolea:
this breaks ABI, if needed for the patch release we need to retrofit this in ABI compliant manner.

eisenhauer:
Ugh.  NdCopy is really complex and used only very deep in serialization.  We've got to get out of this situation where changes in internal, not documented or thoughtfully exported routines can't happen in minor releases.

vicentebolea:
I agree that we overexpose headers, we could probably achieve this with opaque pointers and things like that. Nothing that can be done at this moment since package managers expect us to be abi compatible between patch releases (Spack most importantly). Note that for minor releases is fine to break api/abi. Patch releases are the tricky ones

vicentebolea:
from Dav-sdk we are exploring to add a CI builds in the projects that we support (Adios2 one of it) that alerts when a change breaks API/ABI (in release branch)

anagainaru:
This was a bug fix so it would be nice to include it in the release (especially since Frantz might be using it).
We either need to modify the signature of NdCopy or include another function NdCopyWrite but both would break the API/ABI 

I would spend some time before the 2.11 release to reduce how much we expose.

The only other very ugly fix I can think of is to detect automatically where each buffer is allocated before I call CudaMemcpy and set the flag correctly. This is wrong and inefficient and would hide potential bugs (so I would def remove it for 5.11), but it would fix this problem without changing the ABI. @vicentebolea @eisenhauer @pnorbert what's your preference?

eisenhauer:
"Overexpose headers" is a considerable understatement.   We install a huge number of undocumented internal headers, I guess in order to support plug-in engines.  But unless I'm mistaken, the plug-in engine we have relies upon not having to do much of anything because it just relies on the reader and writer being actually the same process so everything is in memory.  It doesn't actually use anything that does data manipulation like NdCopy.  We need to find a way to support its narrow needs without impacting our ability to fix internal bugs.

vicentebolea:
>  But unless I'm mistaken, the plug-in engine we have relies upon not having to do much of anything because it just relies on the reader and writer being actually the same process so everything is in memory. It doesn't actually use anything that does data manipulation like NdCopy

If that is the case we could further restrict the headers needed to the client plugin compilation and rely in opaque pointers (Pimpl) to not expose the whole class hierarchy. 

@anagainaru you can always add a new method overload without breaking api

eisenhauer:
Agreed, but the challenge is that it includes InlineWriter.h and the classes InlineReader and InlineWriter inherit from core Engine.h, which includes the world.  Really, I suspect the inline engine approach needs to be re-conceived.  While I love the idea of an external engine that you can load on the fly, I think non-trivial examples are wildly difficult if not impossible.  The interaction surface between engines and the rest of ADIOS is broad and pretty fluid as we add features and capabilities.  I know I've had to change core/Engine.h to support BP5 and SST and to clean up inefficiencies in prior implementations.  Has any of that broken things other ADIOS features in such a way that the wouldn't work in the plug-in engine situation?  Probably, but I don't think we have any idea...

anagainaru:
I am on Greg's side, but clearly we won't deal with the plugin engines for this minor release. For now, I will create on overload of NdCopy that has this extra argument and call this function with the default value for the existing signature. I'll open the PR in 5 min once this is done.

vicentebolea:
I see, this is definitely challenging, I also agree with the points exposed. An alternative approach would be to heavily rely in Pimpl as we do in VTK, in this manner you still get to publish most of the headers but have the possibility of changing classes internals without changing abi/api.

@anagainaru thanks!

PR #4361, opened 2024-10-08
Stop C/TestBPWriteStatsOnly from running with MPI, it's not suitable

eisenhauer:


eisenhauer:
Running with mpi produces probabilistic failures because each rank separately writes the same file.

anagainaru (approved):

PR #4366, opened 2024-10-15
Fix the Kokkos testing compilation error on Frontier

anagainaru:
Reopening #4358 (closing it was a mistake)

If hipcc is used as the CXX compiler there is no error, but the CI is using g++ so we need to manually set the CXX compiler to whatever was used to build Kokkos.

The rocm version had to be updated to rocm/5.7.1 in the frontier CI due to an incompatibility between the g++ version in PrgEnv-gnu/8.5.0 and the hipcc version in rocm/5.4.3

The Kokkos version had to be updated to > 4.0.0 due to a warning with 3.7.01 and newer rocm compilers.

@vicentebolea thank you for catching the mistake.

### DIFF
--- a/.gitlab/gitlab-ci-frontier.yml
+++ b/.gitlab/gitlab-ci-frontier.yml
@@ -101,7 +101,7 @@ stages:
       PrgEnv-gnu
       gcc-native/12
       craype-accel-amd-gfx90a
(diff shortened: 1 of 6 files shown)


vicentebolea (approved):
waiting for frontier CI

PR #4367, opened 2024-10-16
Replacing DefApps with Core/24.00 for the frontier-cray CI

anagainaru:
The frontier cray CI was failing:
```
Lmod has detected the following error: The following module(s) are unknown:
"zstd" "libffi"
```
They both require the Core/24.00 module that is not compatible with DefApps

@vicentebolea please let me know if you think another fix will be better.


vicentebolea (approved):
Good catch @anagainaru

PR #4368, opened 2024-10-17
Extend Expressions: Stats funcs

lizdulac:
Includes basic statistics functions such as min, max, sum, mean, variance, standard deviation.

These functions return a single value

### DIFF
--- a/testing/adios2/derived/TestBPDerivedCorrectness.cpp
+++ b/testing/adios2/derived/TestBPDerivedCorrectness.cpp
@@ -789,6 +789,10 @@ int main(int argc, char **argv)
     ::testing::InitGoogleTest(&argc, argv);
 
     result = RUN_ALL_TESTS();
-
+    
+#if ADIOS2_USE_MPI
+    MPI_Finalize();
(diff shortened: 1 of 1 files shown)


eisenhauer:
msan run timed out.  I'm refreshing this to the latest master so it can be merged.  But am I misremembering, weren't derived variables turned off by default in the 2.10 release because they're not an advertised feature yet?  If so, then does this need to be backported to the 2.10.2 release branch?  Or did I miss that someone is asking for this?

anagainaru:
This is still work in progress, I'll look at it this week but I think it still needs to add MPI support. I don't think we need to back port, we can just add it in the 11 release.

eisenhauer:
OK, I've taken the backport label off of it so that it doesn't stand in the way of the 2.10.2 release.

anagainaru:
I would keep this open until we have the new derived mechanism. I'll remove the label for now.

anagainaru:
There are too many conflicts to fix this PR and for now applications do not need this. I will revisit stats functions some other time.

PR #4372, opened 2024-10-17
Some BP5 Serialization documentation, mostly writer-side perspective.

eisenhauer:


anagainaru (commented):
I think this is great, we should have more of this type of documentation for developers. I'm just wondering if the code is the best place to have this. Could we have a separate location for developer level documentation?

eisenhauer:
Not opposed to a separate location certainly, though in some sense it might be easier to read when looking at code.  Probably also want it versioned.  So, if not in the code, maybe a developer_docs directory with entries in Markdown format?  I worry a bit because I think stuff in the code is more likely to be updated as things change, but there's no perfect system there.

anagainaru (approved):

anagainaru:
Something to discuss, I'll approve for now so this gets in. And we can pull it out if we decide to create such a place.

Issue #4371, opened 2024-10-17
OnDemand race condition.


eisenhauer:
I suspect that there is a race condition associated with this the OnDemand request queue.  I'm guessing that it only happens when on the last step the metadata going to reader crosses paths with the request coming from the reader to the writer.  If we had another step on the writer, we'd notice in EndStep that we had this request and it was no longer useful, but since it's the last step we fall through into Close() and don't properly deallocate the request.   This is quite rare, so probably not a high priority, but it does seem to happen, as per sanitizer output below.

Script directory is /root/project/fedora-asan/bin
current working directory is /root/project/fedora-asan/testing/adios2/engine/staging-common
TestDriver: Writer command line : /root/project/fedora-asan/bin/TestOnDemandWrite SST Staging.OnDemandSingle.1x1.CommMin.BP5.SST CPCommPattern=Min,MarshalMethod=BP5
TestDriver: Reader command line : /root/project/fedora-asan/bin/TestOnDemandRead SST Staging.OnDemandSingle.1x1.CommMin.BP5.SST
TestDriver:MPMD value is False

TestDriver:  Doing simple with file_test = False
[==========] Running 1 test from 1 test suite.
[----------] Global test environment set-up.
[----------] 1 test from SstOnDemandWriteTest
[ RUN      ] SstOnDemandWriteTest.ADIOS2SstOnDemandWrite
[==========] Running 1 test from 1 test suite.
[----------] Global test environment set-up.
[----------] 1 test from SstOnDemandReadTest
[ RUN      ] SstOnDemandReadTest.ADIOS2SstOnDemandRead
SST,Write,1,10,100,10,20317,74537.9
SST,Read,1,10,100,10,11875,975867
[       OK ] SstOnDemandWriteTest.ADIOS2SstOnDemandWrite (1327 ms)
[----------] 1 test from SstOnDemandWriteTest (1327 ms total)

[----------] Global test environment tear-down
[==========] 1 test from 1 test suite ran. (1327 ms total)
[  PASSED  ] 1 test.
[       OK ] SstOnDemandReadTest.ADIOS2SstOnDemandRead (1363 ms)
[----------] 1 test from SstOnDemandReadTest (1363 ms total)

[----------] Global test environment tear-down
[==========] 1 test from 1 test suite ran. (1363 ms total)
[  PASSED  ] 1 test.
TestDriver: Reader exit status was 0
TestDriver: Writer exit status was 1
TestDriver: Writer failed, causing test failure
[...]

PR #4373, opened 2024-10-17
Possibly fix leak when an ondemand request comes in just before close

eisenhauer:
Hopefully address issue #4371 .


anagainaru (approved):
It looks acceptable, too bad this is not an error that occurs all the time to make sure it's fixed.

eisenhauer:
> It looks acceptable, too bad this is not an error that occurs all the time to make sure it's fixed.

Yeah, it would take a more complex distributed-system-style test environment to ensure correctness here.

PR #4376, opened 2024-10-23
Handling compilation error with Fujitsu compilers on Fugaku

mnagaso:
Hello,

This change is required when I compile the ADIOS2 on Fugaku@RIKEN with Fujitsu compilers.

- version fj@4.10.0 on linux-rhel8-a64fx

Thanks for considering it.

### DIFF
--- a/thirdparty/KWSys/adios2sys/Configure.hxx.in
+++ b/thirdparty/KWSys/adios2sys/Configure.hxx.in
@@ -17,7 +17,7 @@
 
 #if defined(__SUNPRO_CC) && __SUNPRO_CC > 0x5130 && defined(__has_attribute)
 #  define @KWSYS_NAMESPACE@_has_cpp_attribute(x) __has_attribute(x)
-#elif defined(__has_cpp_attribute)
+#elif defined(__clang__) && defined(__has_cpp_attribute)
 #  define @KWSYS_NAMESPACE@_has_cpp_attribute(x) __has_cpp_attribute(x)
 #else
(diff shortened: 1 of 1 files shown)


eisenhauer:
@vicentebolea or @caitlinross ?   KWSys request, so presumably this would have to change upstream...

mnagaso:
@eisenhauer Thanks for checking it! I got it.

mnagaso:
FYI, I am also having the same error at

```
"(...)/ADIOS2/thirdparty/nlohmann_json/nlohmann_json_wrapper/single_include/nlohmann/json.hpp", line 1887: error: the argument to a feature-test macro must be a simple identifier
  #elif JSON_HEDLEY_HAS_CPP_ATTRIBUTE_NS(clang,fallthrough)
```

eisenhauer:
Also something we pull in from elsewhere, so ADIOS wouldn't change it directly, but if we can get it changed in the upstream repos we can pull in those changes.  But I'm curious what's going on in the Fujitsu compiler.  Does the __have_cpp_attribute feature somehow show up, but not work?

mnagaso:
Thanks for your comment @eisenhauer 

I am assuming that the Fujitsu compiler is regarded as clang by cmake, while this compiler has 2 mode (trad. and clang mode) which can be controled with `-Nclang` flag. 

I am checking what happens with clang mode. So let me come back later with that result.

mnagaso:
PS: ADIOS2 can be compiled with Fujitsu compiler in clang mode without any error.

eisenhauer:
Excellent...

PR #4387, opened 2024-10-30
Add New SelectiveMetadatAggregation method

eisenhauer:
Still a little tentative, want to make sure it doesn't break at high scales (on machines where gatherv breaks at high scales), but would like to merge at this point to get it into master.

### DIFF
--- a/source/adios2/engine/bp5/BP5Engine.h
+++ b/source/adios2/engine/bp5/BP5Engine.h
@@ -172,6 +172,8 @@ public:
     MACRO(StatsBlockSize, SizeBytes, size_t, DefaultStatsBlockSize)                                \
     MACRO(Threads, UInt, unsigned int, 0)                                                          \
     MACRO(UseOneTimeAttributes, Bool, bool, true)                                                  \
+    MACRO(UseSelectiveMetadataAggregation, Bool, bool, true)                                       \
+    MACRO(OneLevelGatherSizeLimit, Int, int, 6000)                                                 \
     MACRO(FlattenSteps, Bool, bool, false)                                                         \
     MACRO(IgnoreFlattenSteps, Bool, bool, false)                                                   \
     MACRO(RemoteDataPath, String, std::string, "")                                                 \
--- a/source/adios2/engine/bp5/BP5Engine.h
+++ b/source/adios2/engine/bp5/BP5Engine.h
@@ -23,13 +23,6 @@ namespace core
 namespace engine
 {
 
-/**
- * sub-block size for min/max calculation of large arrays in number of
- * elements (not bytes). The default big number per Put() default will
- * result in the original single min/max value-pair per block
- */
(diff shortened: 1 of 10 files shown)


eisenhauer:
This does not yet change the default aggregation in BP5 file engine, it just adds an option.

anagainaru (commented):
I don't see any evident problem with this. Is there any way to test the new aggregation schemes?

eisenhauer:
This is still a bit of a work-in-progress.  I've got numbers on Frontier that shows the new scheme takes half the time as the old for kind-of-normal metadata (~100 vars, 100 arrays), and comparable time for E3SM-sized metadata.  But new involves full-scale gatherv, and Frontier seems to have unpredictable problems with this (sometimes slow, sometimes fast).  So not ready to finalize until that is sorted or at least understood.  Point taken about NewWriteMetadata.  Not sure about overload here though.  Will ponder.

eisenhauer:
This needs to be rebased on top of the FFS upstream of PR #4391, then should pass MSAN.

eisenhauer:
> This needs to be rebased on top of the FFS upstream of PR #4391, then should pass MSAN.

this has been done

eisenhauer:
> Still a little tentative, want to make sure it doesn't break at high scales (on machines where gatherv breaks at high scales), but would like to merge at this point to get it into master.

IMHO, ready as default now.  Changed as appropriate.

source/adios2/engine/bp5/BP5Engine.h
```diff
@@ -172,6 +172,8 @@ class BP5Engine
     MACRO(StatsBlockSize, SizeBytes, size_t, DefaultStatsBlockSize)                                \
     MACRO(Threads, UInt, unsigned int, 0)                                                          \
     MACRO(UseOneTimeAttributes, Bool, bool, true)                                                  \
+    MACRO(UseSelectiveMetadataAggregation, Bool, bool, true)                                       \
+    MACRO(OneLevelGatherSizeLimit, Int, int, 6000)                                                 \
```
pnorbert, line 176: `OneLevelGatherSizeLimit`: Size of what? To me, it first suggests a metadata size limit. Propose new names: `OneLevelGatherMPISizeLimit`, `OneLevelGatherNprocLimit`
Change could go into another PR with lines in the documentation ;-)

anagainaru: I would actually call it `OneLevelGatherRanksLimit` since the threshold is given by the number of ranks that are part of the gather right?

eisenhauer: I like OneLevelGatherRanksLimit.  I've changed to that, eliminated a couple of unused (and undocumented) parameters and added some docs on the parameters introduced here.  Turns out I miscounted and UseSelectiveMetadataAggregation was longer than anything and I had to reformat the table a bit to get it to fit.

anagainaru (approved):

pnorbert (approved):

PR #4394, opened 2024-11-06
[Bug fix] SegFault when a variable that is part of a derived expression doesn't have data

anagainaru:
We were checking if a variable was defined before allowing the derived variables to be defined but we were not checking correctly if there is data written in a given step for all the variables needed by the derived expression. Before this fix the code was seg faulting.

I added a test for what happens when we try to define a derived variable before we define the component variables and what happens when there's no data to compute the derived variable.


eisenhauer (changes requested):
Straightforward enough, I'll approve after it gets through formatting and CI.

eisenhauer (approved):

PR #4397, opened 2024-11-08
Fix name clashes in concurrent tests

eisenhauer:


eisenhauer:
I suspect that test output name clashes are the cause of some random failures in CI.  This removes more conflicts.

anagainaru (approved):

PR #4400, opened 2024-11-15
Update cmake version used for the Frontier build script

anagainaru:
Version `3.23.2` is no longer available on Frontier

I prefer spelling out the version (since we can guarantee that this combination will build succesfully), but I am also fine leaving the default versions if this is the preference (in which case we don't need a PR every time a version is no longer supported but users might get some unexpected errors).


eisenhauer:
Hmm.   I kind of like the idea that we use the latest version and see/fix failures quickly when the version changes.  So I'd probably vote for using the default.

anagainaru:
> Hmm. I kind of like the idea that we use the latest version and see/fix failures quickly when the version changes. So I'd probably vote for using the default.

Ok, I set the default versions for all the scripts. Before we merge this, I want to check that we can build on all systems. I'll start with Frontier/Perlmutter but I am not sure if I still have access to Polaris/Summit. If I can't find anyone to test on those systems (assuming I don't have access) I'll remove the scripts.

eisenhauer (approved):

anagainaru:
Frontier/Perlmutter are building adios2 with Kokkos/CUDA correctly with the new modules. Summit is decommissioned so I removed the scripts.

I could not check Polaris, my account expired. I asked for a renewal, I'll reopen a PR if needed when I get the opportunity to check. This is ready for review.

eisenhauer (approved):

Issue #3941, opened 2023-11-24
adios2-config installation does not respect DESTDIR


drew-parsons:
**Describe the bug**

Preparing linux distribution (Debian) packaging for adios 2.9.2 fails at the point of installing adios2-config because of PR #3850 .  This PR was intended to fix DESTDIR installation but has instead broken it.

The problem is https://github.com/ornladios/ADIOS2/blob/515db03345e200bfcbb9cee1a6dab519b191edb3/cmake/install/post/generate-adios2-config.sh.in#L69

In the case where `${1:0:1}` is `"/"` (absolute, not relative path) l.69 tests if `PREFIX` already exists and only prepends DESTDIR if it does not already exist.

But  PREFIX is initially set to the end-installation value `/usr`, which obviously exists.  So DESTDIR is never applied.

The error message  emitted by the generated generate-adios2-config.sh and cmake (make install) is
```
Configuring adios2-config dummy CMake project
...
Writing /usr/bin/adios2-config
/projects/adios2/build/cmake/install/post/generate-adios2-config.sh: line 202: /usr/bin/adios2-config Permission denied
chmod: cannot access '/usr/bin/adios2-config': No such file or directory
```
The "Permission denied" error is due to the error that I am reporting here, from https://github.com/ornladios/ADIOS2/blob/515db03345e200bfcbb9cee1a6dab519b191edb3/cmake/install/post/generate-adios2-config.sh.in#L201
The script is trying to write to `${BINDIR}/adios2-config`, but BINDIR is defined as `${PREFIX}/bin` (https://github.com/ornladios/ADIOS2/blob/515db03345e200bfcbb9cee1a6dab519b191edb3/cmake/install/post/generate-adios2-config.sh.in#L194)  Since PREFIX does not have DESTDIR applied to it, the script is trying to install into the end-installation directory rather than the DESTDIR location, for which it does not have permission.

**To Reproduce**
I worked around #3669 (adios2 cmake config not found) by adding the installed cmake path, running cmake with `-DCMAKE_PREFIX_PATH=${DESTDIR}/usr/lib/cmake/adios2/"`

DESTDIR is intended to be `DESTDIR=/projects/adios2/install`

**Expected behavior**
DESTDIR installation of /usr/bin/adios2-config should not fail.

**Desktop (please complete the following information):**
 - OS/Platform: Debian unstable
 - Build gcc 13.2.0, cmake 3.27.8, build type: shared

**Following up**
Was the issue fixed? Please report back.

PR #4402, opened 2024-11-17
Fix DESTDIR installation

lahwaacz:
Fixes https://github.com/ornladios/ADIOS2/issues/3941


vicentebolea (approved):
Looks great! Thanks for the contribution

PR #4403, opened 2024-11-17
Fixed typo in GPU-aware docs

lahwaacz:


anagainaru (approved):
Thanks for catching it

PR #4407, opened 2024-11-19
Make adios2_remote_server multithreaded and compress responses using …

pnorbert:
…MGARD or ZFP (preference in this order) when an accuracy is set by user.
- added "-t nThreads" option, default is 8
- Use threads to compress responsesi and to return data to client. Adios Get() is still done in main thread because that is not thread-safe.

updates to Campaign engine:
- Call PerformGets() in Campaign engine
- Pass accuracy between CampaignReader and BP5 variables

update to Python API
- added python `bindings.Accuracy` struct, and `SetAccuracy()`/`GetAccuracy()` to bindings, and `variable.set_accuracy()` and `variable.get_accuracy()` to python API.

### DIFF
--- a/bindings/Python/py11Variable.cpp
+++ b/bindings/Python/py11Variable.cpp
@@ -75,6 +75,24 @@ size_t Variable::SelectionSize() const
     return size;
 }
 
+void Variable::SetAccuracy(const adios2::Accuracy &a)
+{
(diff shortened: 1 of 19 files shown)


pnorbert:
Note that I can run the `Remote.BPWriteMemorySelectionRead.GetRemote` test fine on Frontier, so I will ignore the reported error on the OLCF CI and merge this now.

PR #4412, opened 2024-11-27
thirdparty: Update kwsys

vicentebolea:
Fixes: #4419 

We need to update clang since no current fedora support clang9. Our sanitizer builds have many failures at the test phase but it is not reported if we cannot find any sanitizer issue. This was the case before this update, and now. 

After finishing this I will move these builds to the Spack CI

### DIFF
--- a/thirdparty/KWSys/adios2sys/CMakeLists.txt
+++ b/thirdparty/KWSys/adios2sys/CMakeLists.txt
@@ -298,14 +298,6 @@ endif()
 set(KWSYS_HEADER_INSTALL_DIR)
 set(KWSYS_LIBRARY_INSTALL_DIR)
 
-# Generated source files will need this header.
-string(COMPARE EQUAL "${PROJECT_SOURCE_DIR}" "${PROJECT_BINARY_DIR}"
-  KWSYS_IN_SOURCE_BUILD)
-if(NOT KWSYS_IN_SOURCE_BUILD)
-  configure_file(${PROJECT_SOURCE_DIR}/kwsysPrivate.h
-    ${PROJECT_BINARY_DIR}/kwsysPrivate.h COPYONLY IMMEDIATE)
-endif()
-
 # Select plugin module file name convention.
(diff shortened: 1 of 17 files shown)


vicentebolea:
@pnorbert updating kwsys seems to have fixed this msan build. Still it passed in the minute 59 which seems very risky.

vicentebolea:
New ubsan is picking a memory align issue https://open.cdash.org/viewDynamicAnalysisFile.php?id=11240523. It is located at source/adios2/toolkit/format/bp/bp3/BP3Serializer.tcc:133. 

@eisenhauer

eisenhauer:
> New ubsan is picking a memory align issue https://open.cdash.org/viewDynamicAnalysisFile.php?id=11240523. It is located at source/adios2/toolkit/format/bp/bp3/BP3Serializer.tcc:133.

That there is an issue is not surprising.  The initial Span code paid zero attention to alignment concerns.  @pnorbert did a fix for BP4, so maybe he can reprise it?  (I'm in Australia and not returning for another week.)

eisenhauer:
> That there is an issue is not surprising. The initial Span code paid zero attention to alignment concerns. @pnorbert did a fix for BP4, so maybe he can reprise it? (I'm in Australia and not returning for another week.)

Nevermind, this is a different issue.  I assumed that alignment associated with spans was related to the data pointer returned to applications.  That is not the issue here.  The issue seems to be that the min/max aren't aligned in the metadata buffer.  But they don't really have to be.  What we should be done differently is this:
```
        std::copy(&min, &min + 1, reinterpret_cast<T *>(buffer.data() + minPosition));
        std::copy(&max, &max + 1, reinterpret_cast<T *>(buffer.data() + maxPosition));
```
This is moving data to a potentially unaligned address and we shouldn't be using std::copy<T>.  I'd change this to something like:
std::memcpy(buffer.data()+minPosition, &min, sizeof(T));

eisenhauer (approved):

Issue #4446, opened 2025-01-24
stringop-overflow warnings with gcc 14.2


ZedThree:
Building current master (925234) with gcc 14.2.1 gives me a bunch of warnings about "reading/writing 128 bytes from a region of size 16":

```
In file included from /usr/include/c++/14/string:51,
                 from /home/peter/Tools/ADIOS2/source/adios2/core/AttributeBase.h:15,
                 from /home/peter/Tools/ADIOS2/source/adios2/core/Attribute.h:14,
                 from /home/peter/Tools/ADIOS2/source/adios2/toolkit/format/bp5/BP5Base.h:11,
                 from /home/peter/Tools/ADIOS2/source/adios2/toolkit/format/bp5/BP5Helper.h:11,
                 from /home/peter/Tools/ADIOS2/source/adios2/toolkit/format/bp5/BP5Helper.cpp:8:
In static member function ‘static _Up* std::__copy_move<_IsMove, true, std::random_access_iterator_tag>::__copy_m(_Tp*, _Tp*, _Up*) [with _Tp = const char; _Up = char; bool _IsMove = false]’,
    inlined from ‘_OI std::__copy_move_a2(_II, _II, _OI) [with bool _IsMove = false; _II = const char*; _OI = char*]’ at /usr/include/c++/14/bits/stl_algobase.h:521:30,
    inlined from ‘_OI std::__copy_move_a1(_II, _II, _OI) [with bool _IsMove = false; _II = const char*; _OI = char*]’ at /usr/include/c++/14/bits/stl_algobase.h:548:42,
    inlined from ‘_OI std::__copy_move_a(_II, _II, _OI) [with bool _IsMove = false; _II = const char*; _OI = __gnu_cxx::__normal_iterator<char*, vector<char> >]’ at /usr/include/c++/14/bits/stl_algobase.h:555:31,
    inlined from ‘_OI std::copy(_II, _II, _OI) [with _II = const char*; _OI = __gnu_cxx::__normal_iterator<char*, vector<char> >]’ at /usr/include/c++/14/bits/stl_algobase.h:651:7,
    inlined from ‘void adios2::helper::CopyToBuffer(std::vector<char>&, std::size_t&, const T*, std::size_t) [with T = long unsigned int]’ at /home/peter/Tools/ADIOS2/source/adios2/helper/adiosMemory.inl:144:14,
[...]

PR #4447, opened 2025-01-27
Fix buffer overflow

eisenhauer:


eisenhauer:
Addresses issue #4446 .  Turns out that code isn't exercised in normal circumstances, and I'm not 100% sure our tests cover it.  Please let us know if this solves things or we need followup.

anagainaru (approved):

PR #4451, opened 2025-01-29
Add support for key-encrypted metadata in campaign files. Ask the con…

pnorbert:
…nection manager for the keys when needed. Ignore datasets for which adios fails to retrieve the valid key (missing or invalid password was given).


eisenhauer (approved):

anagainaru (approved):
Except that I would just remove the commented out printfs, this looks good

pnorbert:
I will remove the printf statements when we go out of demo mode to production and settle on useful messages. For now, they are on/off depending on the need of the current demo/tutorial.

Issue #4453, opened 2025-02-01
Bug in the logic for reading a derived variable


anagainaru:
We are not setting the dimensions correctly for derived variables. In the following example, there is a file with 3 variables (3D 200x200x200) and a derived variable that has only stats in the bp file:

```
$ bpls -l out_test.bp/ --show-derived
  double   derived  100*{200, 200, 200} = 0.688576 / 172.425
    Derived variable with expression: MAGNITUDE({var1},{var2},{var3})
  double   var1     100*{200, 200, 200} = -100 / 100
  double   var2     100*{200, 200, 200} = -100 / 100
  double   var3     100*{200, 200, 200} = -99.9999 / 100
```

Now, let's say we have a python code that sets a selection and reads a variable (given as an input parameter):

```python
        rStream.BeginStep()
        adiosVar = ioReader.InquireVariable(givenName)
        adiosVar.SetSelection([(0,0,0), (10,10,10)])
        buffer = np.zeros(selCount)
        rStream.Get(adiosVar, buffer)
        rStream.EndStep()
```

If I run this code with one of the normal variables (var1 for example) the code in `source/adios2/toolkit/format/bp5/BP5Deserializer.cpp` creates an entry in the RR structure for a chunk of {10,10,10} and size 2894480 bytes.

For the derived variable there are 3 RR created (for var1, var2, var3) each set as {200,200,200} and 64000000 bytes, and when I look in the ApplyExpression, this function is called on a derived data of {200,200,200} and magnitude is applied on three arrays of {200,200,200}.

anagainaru:
I think this is still an issue, we haven't addressed it because we were hoping to redesign the whole derived logic, but there was no bandwidth. Something to discuss Tuesday

eisenhauer:
OK, I finally have a PR addressing this, #4986.  This finally wires sub-selection into the reader-side derived variable generation process.  Previously we *always* read the whole of any relevant (intersecting) input block, ran the expression over the whole thing to generate output, then did NdCopy into the destination.  Correct, but painfully inefficient.  Now we work backwards from the output selection to the selection required from each input variable.  Read *that* from however many blocks that contain pieces of it into a contiguous region.  Then do the expression over those then contiguous input selections.  In the specific example described above, this results in a 96x speedup:
```
  Full read/compute: 86 ms. 
  Sub-selection read/compute: 0.9 ms.
```
[...]

PR #4454, opened 2025-02-01
Attempt to set the correct dimensions for derived var on the read side 

anagainaru:
when selection is being used.

Bug described in #4453 

I am starting this PR so I can converge with @eisenhauer on this.

The derived variable is created during setup when the metadata is being read so this variable and all the ones necessary to read the data will have counts equal to the global shape. At Get we know the dimensions we want to read, so I am setting the selection for all the variables.

However the mvi points to blocks that still have the global shape so this PR is not a fix.


anagainaru:
This is still a problem

PR #4455, opened 2025-02-05
Modify remote requests to include step count in a single request ...

pnorbert:
instead of splitting them into many single step requests. Modify cache using steps as part of dimensions.

 - keyPrefix should have ending '|' to distinguish variables with inclusive prefix like 'psi' and 'psi_surf'.

### DIFF
--- a/source/adios2/toolkit/remote/remote_server.cpp
+++ b/source/adios2/toolkit/remote/remote_server.cpp
@@ -267,9 +267,11 @@ static void OpenSimpleHandler(CManager cm, CMConnection conn, void *vevent, void
 template <class T>
 void ReturnResponseThread(CMConnection conn, CMFormat ReadResponseFormat, AnonADIOSFile *f,
                           size_t readSize, T *RawData, void *Dest, int GetResponseCondition,
-                          Accuracy acc, std::string name, adios2::DataType vartype,
-                          adios2::Dims count, adios2::Dims start, size_t blockid, bool boxselection)
+                          Accuracy acc, Variable<T> *var)
+
 {
+    adios2::Dims count = var->Count();
+    bool boxselection = (var->m_SelectionType == SelectionType::BoundingBox);
     _ReadResponseMsg Response;
     memset(&Response, 0, sizeof(Response));
     Response.Size = readSize;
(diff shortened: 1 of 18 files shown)


source/adios2/toolkit/remote/remote_server.cpp
```diff
@@ -390,12 +403,7 @@ void PrepareResponseForGet(CMConnection conn, struct Remote_evpath_state *ev_sta
                 GetMsg->Dest,
                 GetMsg->GetResponseCondition,
                 acc,
-                var->m_Name,
```
eisenhauer, line 393: Changing this to pass in var, rather than the individual items makes the parameter list shorter, but I think it violates thread safety.  var is a data structure shared between threads.  Someone else may change the count, start, blockID, etc to support their own get, meaning that the info you feed into the operator in ReturnResponseThread might get trashed.  (It'd be lovely to see this show up in tsan, but the server isn't run for those tests.)

pnorbert: Fantastic catch. This was the reason originally I had the long list of individual parameters.

eisenhauer: Experience has made me sensitive to our threading-related pain points, and info in vars is high on that list, so that sort of thing pops out...

eisenhauer (approved):
I was trying to think about whether or not relative vs. absolute steps might be an issue here.  I'm thinking it just falls out because the count and start that we receive on the client side and transmit are all relative, then we do the appropriate translations during the Get()s on the server side.  Might be useful to confirm that with a test that had a variable missing steps though.

pnorbert:
> I was trying to think about whether or not relative vs. absolute steps might be an issue here. I'm thinking it just falls out because the count and start that we receive on the client side and transmit are all relative, then we do the appropriate translations during the Get()s on the server side. Might be useful to confirm that with a test that had a variable missing steps though.

I was avoiding touching absolute steps at all cost. The remote server works with the StepSelection that the user sets at the client side. It is the remote side's BP5 engine's private business to translate that to absolute steps on the .

eisenhauer:
Yeah, at the lower level BP5 does have to work with absolute steps and it's a pain.  But I don't know any non-painful way of handling that without copying all the metadata out of a per-absolute-step structure and into a per-variable structure.

PR #4456, opened 2025-02-05
Remote server test fix

eisenhauer:
The test (!m_Remote) was intended as a test to see if the objected pointed to by the m_Remote pointer was valid, but that syntax should have been (!(*m_Remote)).  Fixing this, and maybe this will get a better error for the Remote test failures we sometimes see in CI.

### DIFF
--- a/source/adios2/engine/bp5/BP5Reader.cpp
+++ b/source/adios2/engine/bp5/BP5Reader.cpp
@@ -438,7 +438,7 @@ void BP5Reader::PerformGets()
                 "Remote file " + m_Name +
                     " cannot be opened. Possible server or file specification error.");
         }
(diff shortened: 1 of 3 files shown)


anagainaru (approved):

anagainaru (approved):
I am assuming this part of the code is not multi-threaded?

eisenhauer:
> I am assuming this part of the code is not multi-threaded?

Well, yes and no.  All our file read operations employ multiple threads, but each of those threads does it's own Open() on the file.  Otherwise we'd be sharing FDs between threads (possible, if the file pointer isn't an issue), or sharing socket connections to the server (possible, but done at the CManager level, not here).

PR #4459, opened 2025-02-08
Bindings for cuPy pointers

anagainaru:
**Allow python to feed GPU pointers to ADIOS.**

Writer side will have to provide the pointer to the cupy array in the call to Put.

Reader side can call Shape(adios2.MemorySpace.GPU) to receive the dimensions correctly for the GPU then will have to SetSelection to this shape and allocate memory on the GPU using these dimensions.

Example in  `examples/hello/bpStepsWriteReadCuda/bpStepsWriteReadCuda.py`

Running the examples:
```
 $ python examples/hello/bpStepsWriteReadCuda/bpStepsWriteReadCuda.py
Array allocation:  <CUDA Device 0>
Bytes required to store the gpu array 24
Bytes allocated on the device memory pool 512
Bytes used on the device memory pool 512
Blocks allocated on the pinned memory pool (The allocated pinned memory is released just after the transfer is complete) 1
Write to file StepsWriteReadCuPy.bp: (2, 3) data from GPU and (2, 3) data from CPU
Step 0: read GPU data
 [[0. 1. 2.]
 [3. 4. 5.]]
Step 0: read CPU data
 [[0. 1. 2.]
 [3. 4. 5.]]
Step 1: read GPU data
 [[ 0.  2.  4.]
 [ 6.  8. 10.]]
Step 1: read CPU data
 [[1. 2. 3.]
 [4. 5. 6.]]
```

If we bpls the ADIOS file created, the read is done on the CPU so the dimensions will be flipped for the GPU array:
```
$ bpls StepsWriteReadCuPy.bp/ -d gpuArray -n 2
  float    gpuArray  2*{3, 2}
    (0,0,0)    0 1
    (0,1,0)    2 3
    (0,2,0)    4 5
    (1,0,0)    0 2
    (1,1,0)    4 6
    (1,2,0)    8 10
```

### DIFF
--- a/bindings/Python/py11Engine.cpp
+++ b/bindings/Python/py11Engine.cpp
@@ -78,6 +78,39 @@ void Engine::Put(Variable variable, const pybind11::array &array, const Mode lau
     }
 }
 
+#ifdef ADIOS2_HAVE_CUDA
+void Engine::Put(Variable variable, long array, const Mode launch)
+{
+    helper::CheckForNullptr(m_Engine, "in call to Engine::Put list of cupy pointers");
+    helper::CheckForNullptr(variable.m_VariableBase,
+                            "for variable, in call to Engine::Put list of cupy pointers");
+
+    const adios2::DataType type = helper::GetDataTypeFromString(variable.Type());
+
+    if (type == adios2::DataType::Struct)
+    {
+        // not supported
+    }
+#define declare_type(T)                                                                            \
+    else if (type == helper::GetDataType<T>())                                                     \
+    {                                                                                              \
+        m_Engine->Put(*dynamic_cast<core::Variable<T> *>(variable.m_VariableBase),                 \
+                      reinterpret_cast<const T *>(array), launch);                                 \
+    }
+    ADIOS2_FOREACH_NUMPY_TYPE_1ARG(declare_type)
+#undef declare_type
+    else
+    {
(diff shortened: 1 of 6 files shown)


anagainaru:
I am leaving this message here so it's not lost when I re-trigger it (@eisenhauer).

The asan sanitizer fails with the following error:
```
99% tests passed, 2 tests failed out of 819
Total Test time (real) = 135.23 sec
The following tests FAILED:
	350 - Remote.BPWriteReadADIOS2stdio.GetRemote (Failed)
	351 - Remote.BPWriteMemorySelectionRead.GetRemote (Failed)
```

pnorbert (approved):

eisenhauer:
> I am leaving this message here so it's not lost when I re-trigger the CI (@eisenhauer).
> 
> The asan sanitizer fails with the following error:
> 
> ```
> 99% tests passed, 2 tests failed out of 819
> Total Test time (real) = 135.23 sec
> The following tests FAILED:
> 	350 - Remote.BPWriteReadADIOS2stdio.GetRemote (Failed)
> 	351 - Remote.BPWriteMemorySelectionRead.GetRemote (Failed)
> ```

At this point, I think it's possible that this happens when the server doesn't start up in time during testing.  I.E. it likely doesn't represent a problem outside the test scenario.  I'll submit a PR shortly that makes clients notice the connection failure, wait for a second, then retry the connection.  With luck, it'll take care of this heisen-bug.

Issue #4466, opened 2025-02-13
AvailableAttributes() invariably shows all attributes as modifiable


franzpoeschel:
**Describe the bug**
`IO::AvailableAttributes()` returns a list of attributes and for each attribute a handful of meta information on that attribute, including if it is modifiable or not. This is always set as true.

**To Reproduce**
Seen with this test file:

```cpp
#include <adios2.h>
#include <numeric>

static void create_file()
{
    adios2::ADIOS adios;
    auto IO = adios.DeclareIO("IO");
    IO.SetEngine("bp5");
    auto engine = IO.Open("sample.bp5", adios2::Mode::Write);

    auto variable =
        IO.DefineVariable<int>("/data/meshes/theta", {10}, {0}, {10});

    for (size_t step = 0; step < 5; ++step)
    {
        engine.BeginStep();

        // write default openPMD attributes
        IO.DefineAttribute("/basePath", std::string("/data/%T/"));
        IO.DefineAttribute("/date", std::string("2021-02-22 11:14:00 +0000"));
        IO.DefineAttribute("/iterationEncoding", std::string("variableBased"));
        IO.DefineAttribute("/iterationFormat", std::string("/data/%T/"));
        IO.DefineAttribute("/meshesPath", std::string("meshes/"));
        IO.DefineAttribute("/openPMD", std::string("1.1.0"));
        IO.DefineAttribute("/openPMDextension", uint32_t(0));
        IO.DefineAttribute("/software", std::string("openPMD-api"));
        IO.DefineAttribute("/softwareVersion", std::string("0.17.0-dev"));

        std::vector<size_t> current_snapshots(10);
        std::iota(
            current_snapshots.begin(),
            current_snapshots.end(),
            current_snapshots.size() * step);
        IO.DefineAttribute(
            "/data/snapshot",
            current_snapshots.data(),
            current_snapshots.size(),
            "",
            "/",
            /* allowModification = */ true);

        IO.DefineAttribute("/data/dt", double(1));
        IO.DefineAttribute(
            "/data/meshes/theta/axisLabels",
            std::vector<std::string>{"x"}.data(),
            1);
        IO.DefineAttribute("/data/meshes/theta/dataOrder", std::string("C"));
        IO.DefineAttribute(
            "/data/meshes/theta/geometry", std::string("cartesian"));
        IO.DefineAttribute("/data/meshes/theta/gridGlobalOffset", double(0));
[...]

PR #4468, opened 2025-02-17
Kill Modifiable info item for Attributes

eisenhauer:
closes issue #4466


eisenhauer:
The AttributeBase::GetInfo() call is undocumented, so removing the "Modifiable" element in the "Params" returned by this would seem to be fair game.

anagainaru (approved):
Seems reasonable

Issue #3948, opened 2023-11-29
Add a ZSTD compression operator


vicentebolea:
ZSTD is an industry standard lossless parallel compressor brought by Facebook that promises a strong compression ratio with an high throughput. It also provides the possibility of generating and using a dictionary to increase small files (de)compression, this can be a key feature. https://facebook.github.io/zstd/.

I believe it will be beneficial to replace the BZIP2 compressor to the zstd for better performance overall.

Here is a benchmark performed by CentmintMod

![Screenshot from 2023-11-28 19-03-01](https://github.com/ornladios/ADIOS2/assets/939798/f69745d9-15fb-4c99-8936-37ee7390bdb1)

https://community.centminmod.com/threads/round-3-compression-comparison-benchmarks-zstd-vs-brotli-vs-pigz-vs-bzip2-vs-xz-etc.17259/

williamfgc:
Look neat, could be an optional dependency that coexists with bzip2. Operators in general weren't designed to compete, but rather driven by user requests.

PR #4472, opened 2025-03-04
Draft: Add ZSTD operator with threading support

vicentebolea:
This commit adds a new operator for ADIOS2 that uses the ZSTD compression library.

Features:
- Support for different compression levels (1-22)
- Multithreaded compression via nThreads parameter
- Added FindZSTD.cmake for CMake integration
- Added necessary CMake infrastructure

The operator can be configured with parameters:
- compressionLevel: 1 (fast) to 22 (high compression)
- nThreads: Number of threads for compression (default: 1)


vicentebolea:
implements #3948

PR #4473, opened 2025-03-07
[BugFix] Set the correct dimensions and type for a derived expression

anagainaru:
Setting the type and dimensions of a variable is currently done in two places using two different logics. One in the math expressions (the return `DerivedData` object has the dimensions set by the math function) and the second in the derived variable class (when defining the variable). This PR removes the logic inside the math function so we have consistent dimensions and types.

Before this PR if you defined a derived variable:
```
    bpIO.DefineDerivedVariable("derived/myderived",
                                "x = U \n"
                                "sqrt(pow(x))",
                                adios2::DerivedVarType::StatsOnly);
```
The result your error with the following message:
```
"x = U\n sqrt(pow(x))" -> <Derived> <Function> <SqrtFunc> : Invalid variable types
```

**I tested this with the derived variables on and the CI passes**

### DIFF
--- a/source/adios2/core/VariableDerived.cpp
+++ b/source/adios2/core/VariableDerived.cpp
@@ -71,7 +71,7 @@ VariableDerived::ApplyExpression(std::map<std::string, std::unique_ptr<MinVarInf
         inputData.insert({variable.first, varData});
     }
     std::vector<adios2::derived::DerivedData> outputData =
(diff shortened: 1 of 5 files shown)


anagainaru:
The Frontier CI passed when I turned on the derived variables so this is ready for review. Should I trigger it again or can we merge it around it?

PR #4479, opened 2025-03-12
Cleanup CMake a bit

eisenhauer:
Only try to find Caliper if DAOS is found, because it's not used by non-DAOS code.  Pull in new version of enet with increased minimum CMake version to avoid warnings.


vicentebolea (approved):
Looks good!

PR #4481, opened 2025-03-12
ci,frontier: do not load hdf5

vicentebolea:
HDF5 is not needed since HDF5 adios2 support is already disabled at both: `scripts/ci/cmake/ci-frontier-cray.cmake` and `scripts/ci/cmake/ci-frontier-kokkos-hip.cmake`

### DIFF
--- a/.gitlab/gitlab-ci-frontier.yml
+++ b/.gitlab/gitlab-ci-frontier.yml
@@ -106,7 +106,6 @@ stages:
       git
       ninja
       libffi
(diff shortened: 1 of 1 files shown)


vicentebolea:
@eisenhauer this should fix the existing error in frontier

eisenhauer (approved):
I'll go ahead and pre-approve with the assumption that if it doesn't fix things, you won't merge...

PR #4491, opened 2025-03-20
python: stream.write accept np array w/o extra args

vicentebolea:
This simplify the usage of Streams with numpy arrays.

### DIFF
--- a/python/adios2/stream.py
+++ b/python/adios2/stream.py
@@ -1,6 +1,6 @@
 """License:
-  Distributed under the OSI-approved Apache License, Version 2.0.  See
-  accompanying file Copyright.txt for details.
+Distributed under the OSI-approved Apache License, Version 2.0.  See
+accompanying file Copyright.txt for details.
 """
 
 from functools import singledispatchmethod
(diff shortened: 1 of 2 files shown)


python/adios2/stream.py
```diff
@@ -297,6 +297,11 @@ def _(self, name, content, shape=[], start=[], count=[], operations=None):
         if not variable:
             # Sequence variables
             if isinstance(content, np.ndarray):
+                # If shape and count are not specified, use the numpy array's shape
+                if shape == [] and count == []:
+                    shape = list(content.shape)
+                    count = shape
```
anagainaru, line 303: I think here `count` and `shape` will point to the same memory so if in the future someone changes count, shape will also be updated. Is this something you wanted or you wanted to do `count = list(shape)`?

vicentebolea: Good catch, thanks @anagainaru . Find the last push fixing this for ndarrays and also for python list by wrapping lists as ndarrays and passing it down to adios2 C/C++ layer.

anagainaru (approved):
Looks good to me, I assume Frank's CI is failing for some other reason.

vicentebolea:
@anagainaru yep nothing to do with this PR. I am also working to improve the flakiness of these tests.

Issue #4493, opened 2025-03-26
Latest bpls shows wrong size of an array


zhihuizou:
**Describe the bug**
The bpls in the latest master branch shows that the **faces** array has 18446744073709551614 entries as shown in the following screenshot, which is wrong.
![Image](https://github.com/user-attachments/assets/118b7102-f263-46f0-9afd-34a55ffece22)

The latest released version of bpls (Version 2.10.2) gives the right information as follows,

![Image](https://github.com/user-attachments/assets/bc772ed5-0eb0-4fdb-a514-735153f884cd)

**To Reproduce**
1. uzip the attached meshdb.zip file
2. bpls -lDa meshdb > output.txt
3. open output.txt file and you will see the screenshot I posted above.
[meshdb.zip](https://github.com/user-attachments/files/19475447/meshdb.zip)

pnorbert:
Hi, I won't be able to look at this problem this week but can you please share the xxx.bp/m* files so that we can debug it?

zhihuizou:
> Hi, I won't be able to look at this problem this week but can you please share the xxx.bp/m* files so that we can debug it?

Hi, Thanks for your quick response. I just attached the database for you to debug. Thanks.

eisenhauer:
Interesting, these are coming in as one dimensional with a dimension value of **-2.**  We should probably be throwing an error on the writer side on being given a negative size value but instead we are calculating the ridiculously large size.  Can you check the values that you're providing for the global array size?
`Dims = 1 ,BlockCount = 1 ,DBCount = 1 ,Shape = 0x141817938 -2 ,Count = 0x141817940 24973 ,Offset = NULL,DataBlockLocation = 0x141817948 389856 ,MinMax = 0x141817950 `

pnorbert:
it's probably a joined array then

eisenhauer:
Ah, that makes sense.  Then maybe just an output problem in bpls...

eisenhauer:
So, yes, looks like just a bpls display problem.  Trying to sort how best to fix, but will get an update into Github master soon.

eisenhauer:
This was a BP5 issue, not present in BP4, so earlier versions of ADIOS didn't demonstrate.  It is now fixed in ADIOS master and will be included in the next major release.

zhihuizou:
@eisenhauer @pnorbert  Thanks very much for your quick fix. Do you have an estimate when the next major release will be delivered? Thanks.

[...]

PR #4494, opened 2025-03-27
Fix joined array output in bpls

eisenhauer:


eisenhauer:
Closes  #4493

anagainaru (approved):

eisenhauer:
Not sure why readthedocs isn't completing, but this doesn't have anything to do with documentation, so I'm ignoring that.  WRT the sanitizers, we've some kind of a race condition that seems to result in occasional leaks on shutdown.  Also unrelated, but I'll try to sort in another PR.  In the meantime, merging this one.

PR #4495, opened 2025-03-27
issues: stale issues that have been open been inactive

vicentebolea:
Add daily check that goes through opened issues and set them as stale if open and inactive for over a year.

For issues that are stale, they will be closed a month later if still inactive. Re-opening the issue resets the timers.

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -1,0 +1,20 @@
+name: 'Close stale issues and PRs'
+on:
+  schedule:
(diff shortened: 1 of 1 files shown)


vicentebolea:
@pnorbert this is an idea to organize the existing issues, if we want to whitelist some issues so that they are never closed we can use the 'evergreen' tag.

eisenhauer:
I'm on board.

pnorbert (approved):

PR #4497, opened 2025-03-31
Once more bpls fix

eisenhauer:
Roll back previous incorrect fix, try to fix locally.


eisenhauer:
@pnorbert Please take a look as I'm not entirely sure what some of the logic in bpls is meant to do...

pnorbert (changes requested):
MinBlocksInfo should return information the same way as for GlobalArrays, since a joined array is presented as GlobalArray to a reader, without any differentiation to its origin.

eisenhauer:
The challenge here is that to do this properly the MinVarInfo semantics need to change a bit.  Currently the Shape pointer in the MinVarInfo struct is a simple pointer to memory that is not owned by the struct.  I.E. it can only point to an existing Shape value somewhere in the BP5 metadata (memory for which is managed separately).  For GlobalArrays that were originally LocalValues or JoinedArray, we're constructing a new Shape value that doesn't exist elsewhere in the per-rank metadata, so to make sure that gets free'd properly we need to pass ownership of that memory back with the MinVarInfo struct.  So it needs to be a Dims vector or a smart pointer.  This isn't a problem generally, because MinVarInfo isn't user-exposed, but it pretty certainly violates the ABI stability requirements for minor releases, so it does mean that this fix waits for a major release...

eisenhauer:
@pnorbert please review.  The failures here aren't related to this PR.  (Still another race condition in shutdown I need to investigate.)

PR #4508, opened 2025-04-07
Simpler Joined Array Size mod, backout prior fix

eisenhauer:


eisenhauer:
Looking at this more carefully allowed a simpler fix, and simplification of VarShape() code.

eisenhauer:
The msan errors don't seem to be related...

pnorbert (approved):

PR #4513, opened 2025-04-22
    OpenPMD option for Engine=HDF5

tybulut:
- H5OpenPMD* options and H5 attributes
    - Adios2 Step = OpenPMD Iteration
    - Supports OpenPMD: fileBased and groupBased
    - basePath: Adds step prefix to be multiple sub-keys ("/" separated path)

### DIFF
--- a/source/adios2/engine/hdf5/HDF5ReaderP.cpp
+++ b/source/adios2/engine/hdf5/HDF5ReaderP.cpp
@@ -75,6 +75,7 @@ void HDF5ReaderP::Init()
                                              ", in call to Open");
     }
 
+    m_H5File.PreInitParseParameters(m_IO);
     m_H5File.Init(m_Name, m_Comm, false);
     m_H5File.ParseParameters(m_IO);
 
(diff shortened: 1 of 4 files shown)


guj (commented):
For file based option in openPMD, it expects  one file per step, and I don' see this been taken care of.  This commit uses openPMD style path name /data/step/... in one single one for all steps,   the result is  a group based  openPMD file.

tybulut:
> For file based option in openPMD, it expects one file per step, and I don' see this been taken care of. This commit uses openPMD style path name /data/step/... in one single one for all steps, the result is a group based openPMD file.

It actually generates one file for step. If you run the unit tests you can see the outputs.

tybulut:
> For file based option in openPMD, it expects one file per step, and I don' see this been taken care of. This commit uses openPMD style path name /data/step/... in one single one for all steps, the result is a group based openPMD file.

This patch assumes the layer above will decide whether to write additional steps in the same file or create a new ADIOS2 file for each step.

guj:
> > For file based option in openPMD, it expects one file per step, and I don' see this been taken care of. This commit uses openPMD style path name /data/step/... in one single one for all steps, the result is a group based openPMD file.
> 
> This patch assumes the layer above will decide whether to write additional steps in the same file or create a new ADIOS2 file for each step.

Then it seems only the option  " std::string HDF5Common::PARAMETER_OPENPMD_COMPLIANT "  is needed?   As the expected behavior is  to write to a h5 file with variable names in the openPMD style (e.g.  "/data/step/[var]" ).

PR #4517, opened 2025-05-05
Use Pread() instead of read() in Posix file transport to eliminate kernel state dependency

eisenhauer:


eisenhauer:
Also ready for review, just unrelated failures.

pnorbert (approved):

PR #4523, opened 2025-05-13
Add old output files and ctests to see if we can still read them

eisenhauer:


eisenhauer:
Note I have not disturbed the "backward_compatibility" subdirectory which isn't actively involved in testing...

vicentebolea:
For extracting these zipped bp files you could leverage cmake -E tar command: https://cmake.org/cmake/help/latest/manual/cmake.1.html#cmdoption-cmake-E-arg-tar

eisenhauer:
> For extracting these zipped bp files you could leverage cmake -E tar command: 
Is there some reason to avoid the ARCHIVE_CREATE/EXTRACT functionality in the CMAKE file command?  (With the caveat that I couldn't use ARCHIVE_CREATE because I needed more control over the zip command line to try to keep some info out of the archive.  This is all due to trying to create archives that are identical except for actual changes in ADIOS file data...)

testing/adios2/output_archive/create_archive.cmake
```diff
@@ -0,0 +1,9 @@
+
+message(STATUS "Creating archive from ${SOURCE_DIR} into ${ARCHIVE_NAME}")
+file(GLOB files_to_archive "${SOURCE_DIR}/*")
+list(TRANSFORM files_to_archive REPLACE "${SOURCE_DIR}/" "" )
+list(REMOVE_ITEM files_to_archive "profiling.json")
+list(JOIN files_to_archive " " files_no_semi)
+execute_process (COMMAND bash -c "touch -d 1970-01-01T00:00:00Z ${files_no_semi}" WORKING_DIRECTORY ${SOURCE_DIR})
+execute_process (COMMAND bash -c "rm -f ${SOURCE_DIR}.zip; zip -qX ${SOURCE_DIR}.zip ${files_no_semi}" WORKING_DIRECTORY ${SOURCE_DIR})
```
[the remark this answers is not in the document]
vicentebolea: In the current approach we set the atime and mtime and but we do not save time info (-X) in the zip archive. The remaining extended info in unix systems is the UID and GUID but zip ignores it when extracting. We might just delegate into the `cmake tar`, but maybe it will not be enough, there is little info on how cmake version of it deal with it.

vicentebolea:
> > For extracting these zipped bp files you could leverage cmake -E tar command:
> > Is there some reason to avoid the ARCHIVE_CREATE/EXTRACT functionality in the CMAKE file command?  (With the caveat that I couldn't use ARCHIVE_CREATE because I needed more control over the zip command line to try to keep some info out of the archive.  This is all due to trying to create archives that are identical except for actual changes in ADIOS file data...)

makes sense, file attributes architiving is tricky, I left some comment,

testing/adios2/output_archive/CMakeLists.txt
```diff
@@ -0,0 +1,73 @@
...
+# zipped_output. We expect the filename to start with "<S>.", where
+# <S> is one if the TestSet names in the TestSets variable below.
+# Current output files look like <S>.<version info>.<engine>.zip.  But the
+# <engine> isn't provided to the reader, and the <version info> isn't
+# interpreted.  <Version info> could/should also include relevant
+# architecture information, like byte order.
+#
+SET (TestSets Attribute Common)
+
+ADD_EXECUTABLE(test_file_unique test_file_unique.c)
```
vicentebolea, line 23: upper case.

testing/adios2/output_archive/ParseArgs.h
```diff
@@ -0,0 +1,325 @@
...
+    }
+    return Ret;
+}
+
+void ParseArgs(int argc, char **argv)
+{
+    int bare_arg = 0;
+    while (argc > 1)
+    {
+        if (std::string(argv[1]) == "--num_steps")
```
vicentebolea, line 92: How about instead of this many conditionals a giant `unordered_map<string, function()>` I am not sure if this will hurt readability, but it will definitely make it more concise.

eisenhauer: The whole ParseArgs.h is a byproduct of the tests here being derived from more complex tests elsewhere.  We don't specify any of these args here, so it should just be cut down to the few things required.

testing/adios2/output_archive/create_archive.cmake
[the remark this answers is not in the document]
vicentebolea: cmake tar can also manage zip, but then again there is no option to avoid including these file attributes since it is a very primitive tar. I can see how zip -X just works for this. Thanks for explaining it.

testing/adios2/output_archive/create_archive.cmake
```diff
@@ -0,0 +1,9 @@
+
+message(STATUS "Creating archive from ${SOURCE_DIR} into ${ARCHIVE_NAME}")
+file(GLOB files_to_archive "${SOURCE_DIR}/*")
+list(TRANSFORM files_to_archive REPLACE "${SOURCE_DIR}/" "" )
```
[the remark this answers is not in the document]
eisenhauer: This isn't a valid substitution.  The first two lines create list of filesnames with directories and then remove the directory spec, leaving only the filenames.

testing/adios2/output_archive/ParseArgs.h
eisenhauer: ParseArgs.h removed and code cut down.

testing/adios2/output_archive/CMakeLists.txt
eisenhauer: Lowered.

eisenhauer:
OK, output and uniqueness testing again disabled until I sort discrepancies, but the remainder is significantly cleaned up.  Request re-review.

testing/adios2/output_archive/CMakeLists.txt
```diff
@@ -0,0 +1,90 @@
...
+		-P ${CMAKE_CURRENT_SOURCE_DIR}/create_archive.cmake
+	)
+
+	add_test(NAME Archive.${TestSet}.${Engine}.Unique
+	    COMMAND test_file_unique ${OUTPUT_DIRECTORY}/Current${TestSet}.${Engine}.zip ${${TestSet}S}
+	)
+
+
+	SetupTestPipeline(Archive.${TestSet}.${Engine} "Write;Zip;Unique" TRUE)
+    endforeach()
```
vicentebolea, line 87: indentation is wrong here.

testing/adios2/output_archive/CMakeLists.txt
```diff
@@ -0,0 +1,90 @@
...
+
+    add_executable(Read${TestSet} Read${TestSet}.cpp)
+    target_link_libraries(Read${TestSet} adios2::cxx11 adios2::c adios2_core adios2::thirdparty::gtest)
+    if(ADIOS2_HAVE_MPI)
+        add_executable(Write${TestSet} Write${TestSet}.cpp)
+        target_link_libraries(Write${TestSet} adios2::cxx11_mpi adios2::c_mpi adios2_core_mpi MPI::MPI_CXX adios2::thirdparty::gtest)
+    endif()
+
+    file(GLOB ${TestSet}S ${CMAKE_CURRENT_SOURCE_DIR}/zipped_output/${TestSet}*.zip)
+    set (OUTPUT_DIRECTORY "${CMAKE_BINARY_DIR}/TempOutput")
```
vicentebolea, line 37: We might want to output to the corresponding directory in binary_dir "current"

vicentebolea (changes requested):
couple of little comments. everything else looks good

vicentebolea (dismissed):

vicentebolea (approved):

pnorbert (approved):

PR #4531, opened 2025-05-21
Enable checking of new archive output

eisenhauer:


eisenhauer:
Tried to rerun for sanitizers, but they seem to be broken independent of this.  So, I think this is ready for merge despite failures. @pnorbert, will wait for a review before doing a bypass merge.

anagainaru (commented):
The switch to std::map looks good to me, I am not really the best person to judge the cmake parts. Did Vicente already review this in your previous PR?

pnorbert (approved):

PR #4537, opened 2025-05-30
init_fabric: use Trace verbosity instead of Critical for success message

Lgt2x:
A success/non-error message should not end up in stderr, and in my case made some test scripts fail.


eisenhauer:
CI failure seems to be spurious, maybe a heisen-failure with an odd tell.  Merging on bypass.

PR #4538, opened 2025-05-30
WaitForOpen in EVPath remote was a no-op, should throw if an open fails

eisenhauer:


eisenhauer:
Nothing remarkable here.  There was an apparently one-off msan failure that could only have happened if an remote open failed but we didn't notice.  This makes sure we notice.

PR #4539, opened 2025-05-30
Tweaks to remove DP name dependency from SST

eisenhauer:


eisenhauer:
Restructure a bit so that we're not looking at DP names in the control plane.  This is done by reworking the usage of one of the CP/DP interface routines.

PR #4544, opened 2025-06-09
GoogleTest version update, move to require C++14.

eisenhauer:


vicentebolea (approved):

CMakeLists.txt
```diff
@@ -106,8 +106,8 @@ if(NOT CMAKE_BUILD_TYPE AND NOT CMAKE_CONFIGURATION_TYPES)
   set(CMAKE_BUILD_TYPE "Release" CACHE STRING "Choose the type of build." FORCE)
 endif()
 
-# Force C++11 (No extensions) and C99
-set(CMAKE_CXX_STANDARD 11)
+# Force C++14 (No extensions) and C99
+set(CMAKE_CXX_STANDARD 14)
```
anagainaru: Since we are changing this, should we take this opportunity to check if the users are trying to set the version themselves and issue a warning that we are overwriting what they are trying to set?

eisenhauer: Good point.  Changing a bit...

PR #4546, opened 2025-06-11
Fix mishandling m_AvailableShapes map in TimeSeries engine. 

pnorbert:
Add a variable with fixed dimensions to the test.


pnorbert:
The funny situation when the last fix for the complicated test broke the simple but untested case.

eisenhauer (dismissed):

PR #4547, opened 2025-06-11
Kill orphan files

eisenhauer:


anagainaru (approved):
Before my time but it seems to build without it so I'll approve. I don't know exactly what the files were doing but at a first glance I don't see this helping us with the whole operator discussion so we might as well remove them

PR #4550, opened 2025-06-18
Dill tweaks

eisenhauer:
@anagainaru Please give this a shot to see if it helps building on Apple Silicon without libffi.


eisenhauer:
> I was able to build adios on my laptop ! I'm not approving yet until it passes the CI

Good news!  (The infrastructure to build without Dill was there, it just didn't quite work in the context of the ADIOS meta-build...)

The CI failures here are unrelated.  Some of it is the random ASAN stuff that we can't recreate outside of CI.  The rest are new warnings from a windows compiler about thirdparty code in enet.  We probably need to see if we can suppress those, as we can't fix them.  So, my temptation is to override the rules and force a merge here.  Do you concur?

anagainaru:
Sounds good to me

PR #4551, opened 2025-06-19
Fixes for 32-bit builds on Visual Studio 17 (Windows Server 2022)

franzpoeschel:
This fixes two small build issues affecting builds of ADIOS2 release 2.10.2:

* In the BP5 serializer, there are implicit conversions from `uint64_t` to `size_t` which are narrowing conversions on 32-bit systems. This PR makes these conversions explicit.
* Two locations in EVPath declare `extern int h_errno`, but apparently `h_errno` is declared as a macro in `winsock2.h`.

I am not sure if this is the right place to fix the EVPath issue since it is a third-party library?

### DIFF
--- a/source/adios2/toolkit/format/bp5/BP5Serializer.cpp
+++ b/source/adios2/toolkit/format/bp5/BP5Serializer.cpp
@@ -1629,16 +1629,16 @@ std::vector<core::iovec> BP5Serializer::BreakoutContiguousMetadata(
         {
             uint64_t MEBSize;
             helper::CopyFromBuffer(Aggregate.data(), Position, &MEBSize);
-            MetadataBlocks.push_back({Aggregate.data() + Position, MEBSize});
-            Position += MEBSize;
+            MetadataBlocks.push_back({Aggregate.data() + Position, static_cast<size_t>(MEBSize)});
+            Position += static_cast<size_t>(MEBSize);
         }
         helper::CopyFromBuffer(Aggregate.data(), Position, &ABCount);
         for (uint64_t i = 0; i < ABCount; ++i)
         {
             uint64_t AEBSize;
             helper::CopyFromBuffer(Aggregate.data(), Position, &AEBSize);
-            AttributeBlocks.push_back({Aggregate.data() + Position, AEBSize});
-            Position += AEBSize;
+            AttributeBlocks.push_back({Aggregate.data() + Position, static_cast<size_t>(AEBSize)});
+            Position += static_cast<size_t>(AEBSize);
         }
         uint64_t element;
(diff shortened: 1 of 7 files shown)


eisenhauer:
The EVPath stuff has to be fixed upstream in the GTkorvo package, but it's an odd thing to be popping up as we do compile on windows.  Maybe an oddity of the 32-bit netdb.h?  The right answer to this is probably to rewrite it to not use the obsolete functions present here.

franzpoeschel:
The error only popped up for us after moving our CD runs from Windows Server 2019 to Windows Server 2022, so it might depend on the specific Windows environment.

eisenhauer:
@franzpoeschel If you kill the thirdparty changes, the cast is fine.  However, beyond the compile-without-warnings phase, can I ask the general plan for this?  For anything that comes from or goes to disk, using a fixed-size type is probably the right thing to do so that the basic file format doesn't change.  However much of the adios code (like this bit) uses size_t for internal position calculations.  Even if those don't land on disk, letting them drop to 32-bits would restrict the ability of 32-bit library to access large files.  Now, maybe that's fine, depending upon the goals for the 32-bit port, but it might be worthwhile to be explicit about the strategy.  Can you share?

eisenhauer (approved):

eisenhauer:
I'll see about getting changes into EVPath that fix kill the h_errno reference.  However, I can't easily test on 32-bit VS 2022, so we may need to iterate to get things clean there.

franzpoeschel:
>  For anything that comes from or goes to disk, using a fixed-size type is probably the right thing to do so that the basic file format doesn't change.

Yes, this is why I did not change the declarations for those variables that interact directly with the buffer (MEBSize, AEBSize)

> However much of the adios code (like this bit) uses size_t for internal position calculations. Even if those don't land on disk, letting them drop to 32-bits would restrict the ability of 32-bit library to access large files. Now, maybe that's fine, depending upon the goals for the 32-bit port, but it might be worthwhile to be explicit about the strategy. Can you share?

Yes, I am aware that this is the implication of performing a narrowing cast in this place. Our main motivation for keeping 32-bit builds around as long as it's somehow possible is for ensuring cross-platform portability of our IO routines. So this is mostly about having as many different hardware platforms in the CI/CD as possible.
I don't think that anyone still uses those 32-bit builds, but restricting any such user to files smaller than 2GB is reasonable enough.

> I'll see about getting changes into EVPath that fix kill the h_errno reference. However, I can't easily test on 32-bit VS 2022, so we may need to iterate to get things clean there.

My guess is that the EVPath changes probably don't depend on 32-vs-64 bit, but only on the Windows-Server + Visual Studio versions? But feel free to ping me once you have a solution there and I'll try to test it.

PR #4555, opened 2025-06-24
Suppress windows warning about enums

eisenhauer:


anagainaru (approved):
I am not the best person to review this, but based on Vicente's email this seems to be ok. @vicentebolea can revert it if he doesn't approve

PR #4557, opened 2025-06-25
Fix leaks if server has already exited

eisenhauer:


eisenhauer:
I think the failures here are another heisen-bug, but this PR fixes some of odd ASAN failures that I was previously unable to reproduce.  Apparently sometimes the mechanism to kill the adios2_remote_server fixture is either run twice or in some other way fails to connect to the server.  When that happened, we called exit() right away because there was nothing else for us to do.  However, this left allocated memory unfreed.  This PR makes sure that we exit cleanly in that circumstance.

anagainaru (approved):

PR #4575, opened 2025-07-10
Fix column-major reading with memory selection

ohlmann:
Use the correct variables in BP5Deserializer that have been reversed if the reader is column-major. Also add a test.

Before, Adios2 would crash with a segfault if a memory selection was used when reading in data from Fortran even if the selection was done in exactly the same way as for writing the data out. Executing the test without the fix shows this behavior as well.


ohlmann:
Thanks for reviewing and merging so quickly!

PR #4576, opened 2025-07-10
Operator::AddExtraParameters() to allow engine pass parameters before…

pnorbert:
… Operate()/InverseOperate()

This modification allows now to pass engine name and variable name to external plugin operators as well.


pnorbert:
I did not want to rework the plugin operator and plugin operator interface. The previous solution worked for operators but not for plugin operators, they did not get the parameters from the engine

PR #4581, opened 2025-07-12
Fix in TimeSeries engine for variables without min/max (like char arr…

pnorbert:
…ay) which cause exception at open. Added a char array output to test to make sure opening works fine. Also extended test to test the correct minmax of the double array is reported at every step.

### DIFF
--- a/source/adios2/engine/timeseries/TimeSeriesReader.cpp
+++ b/source/adios2/engine/timeseries/TimeSeriesReader.cpp
@@ -17,6 +17,7 @@
 
 #include <ios>
 #include <iostream>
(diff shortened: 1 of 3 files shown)


pnorbert:
Duh, you cannot compare local min/max to global min/max. Hence the failures with the MPI versions.

pnorbert:
There is probably a new bug discovered though:
- BP3/BP4
- ReadRandomAccess
- MinMax(step) returns the over-all-steps global min/max not the per-step min/max

I thought this was already tested somewhere but I only found per-step Max() testing in streaming mode, where this works properly.

Issue #4586, opened 2025-07-16
BP5: Spans invalidated by defining other Spans on the same variable


franzpoeschel:
**Describe the bug**

When defining multiple `Span`s on a variable in BP5, their `.data()` function will return the same pointer for all spans. That pointer will point to the last span defined.

Note that Spans can get invalidated even in BP4, but in that case, calling `.data()` a second time will return an updated pointer. So not the Span itself, only the returned pointer should be invalidated. This behavior is documented in the C++ interface:

```cpp
    /**
     * Pointer to span data, can be modified if new spans are added
     * Follows rules of std::vector iterator invalidation.
     * Call again to get an updated pointer.
     * @return pointer to data
     */
    T *data() const noexcept;
```

**To Reproduce**
```cpp
#include <adios2.h>
#include <iostream>
#include <numeric>
#include <vector>

int main()
{
    using namespace adios2;
    using type = uint32_t;
    ADIOS adios;
    IO IO = adios.DeclareIO("IO");
    IO.SetEngine("bp5");
        Engine engine = IO.Open("span.bp", Mode::Write);

        Variable<type> spanvar =
            IO.DefineVariable<type>("span", {10, 10});

        engine.BeginStep();

        std::vector<Variable<type>::Span> spans;
        spans.reserve(10);

        for(size_t i = 0; i < 10; ++i)
        {
            spanvar.SetSelection({{i, 0}, {1, 10}});
            auto span = engine.Put<type>(spanvar);
            spans.push_back(std::move(span));
        }

        for(size_t i = 0; i < 10; ++i)
        {
            std::cout << "Span pointer: " << spans[i].data() << '\n';
            std::iota(spans[i].begin(), spans[i].end(), i*10);
        }

        engine.EndStep();
        engine.Close();
}
```

Running this:

```
> ./build/adios_span_bp5 && bpls span.bp/ -d
Span pointer: 0x7fffcfe00178
Span pointer: 0x7fffcfe00178
Span pointer: 0x7fffcfe00178
Span pointer: 0x7fffcfe00178
Span pointer: 0x7fffcfe00178
Span pointer: 0x7fffcfe00178
Span pointer: 0x7fffcfe00178
Span pointer: 0x7fffcfe00178
Span pointer: 0x7fffcfe00178
Span pointer: 0x7fffcfe00178
  uint32_t  span  {10, 10}
    (0,0)    0 0 0 0 0 0
    (0,6)    0 0 0 0 0 0
[...]

PR #4587, opened 2025-07-16
Fix a problem with multiple spans on a single variable in BP5

eisenhauer:
addresses issue #4586


pnorbert (approved):

franzpoeschel:
Thank you for the fix! I just ran a test and this also fixes the original issue that I had observed.

ax3l:
@pnorbert @vicentebolea can you please mark this for a patch release, e.g., in case you plan a 2.10.3?

PR #4590, opened 2025-07-17
Restore remote test and fix bug

eisenhauer:
Or avoid bug.  EVPath doesn't like 8-byte messages.  Add junk so that it's 16.


source/adios2/toolkit/remote/remote_common.h
```diff
@@ -105,6 +105,7 @@ typedef struct _CloseFileResponseMsg
 {
     int CloseResponseCondition;
     int Status;
+    size_t unused; // small messages call stack addressing issues?
```
anagainaru: I would add in the comment the reason why you are adding this

anagainaru (dismissed):
Looks good otherwise

source/adios2/toolkit/remote/remote_common.h
eisenhauer: done

anagainaru (approved):

pnorbert (dismissed):

eisenhauer:
Summary of changes:

- For the moment, disable memory sanitization in two routines in remote_server.  Not sure yet if msan is confused by something there or if there's a real problem, but lets get the other things in.
- Actually test the success of various server operations and raise appropriate client-side exceptions when things fail.
- freopen server stderr to file /tmp/server_stderr.  This is invaluable for getting sanitizer output for the server when it is run in the background as a ctest fixture.
- there's an adiosNetwork fix to handle odd hostname setups in some VMs and dockers images
- Change simpleFile memory management to rely upon shared_ptr.
- Add a "-d filename" option to adios2_remote_server.  This is useful in fixture shutdown and causes a file to be dumped to stdout.  This can at least get a server log or other file into the raw log from CI.

eisenhauer:
Changed logging method to avoid std::cout and introduce a log_output() subroutine that uses a lock for mutual exclusion.  Previous msan errors were really tsan issues with multiple threads having conflicts on the internal IO buffers.

pnorbert (approved):

PR #4593, opened 2025-07-26
Shared remote connection

pnorbert:
Use a single socket connection to a remote server from all engines accessing (different) files from the same host. This eliminates creating too many connections and tunnels to the same host when processing a campaign with lots of files.

### DIFF
--- a/examples/campaign/campaign_many_files.py
+++ b/examples/campaign/campaign_many_files.py
@@ -8,7 +8,7 @@ import adios2
 NFILES = 100
 
 # User data
-myArray = numpy.array([0, 1.0, 2.0, 3.0, 4.0, 5.0, 6.0, 7.0, 8.0, 9.0])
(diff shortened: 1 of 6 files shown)


eisenhauer (dismissed):

examples/campaign/campaign_many_files.py
```diff
@@ -0,0 +1,17 @@
...
+import adios2
+
+#
+# Create many files for a campaign, each with a single small array
+#
+
+NFILES = 100
+
+# User data
+myArray = numpy.array([0, 1.0, 2.0, 3.0, 4.0, 5.0, 6.0, 7.0, 8.0, 9.0])
```
vicentebolea:
```suggestion
myArray = numpy.arange(0.0, 9.0, 1)
```

eisenhauer (approved):

PR #4605, opened 2025-08-08
largeblocks: Skip test if not enough space

vicentebolea:
This PR restrict the executing of the test LargeBlock only in systems that have enoughs space


testing/adios2/engine/bp/TestBPLargeBlocks.cpp
```diff
@@ -108,6 +130,9 @@ TEST_F(LargeBlocks, MultiBlock)
         }
         reader.Close();
     }
+
+    auto rmCmdStr = std::string("rm -rf ./") + fname;
```
vicentebolea: I am normally against removing things from the builddir, but given that this test will always generate large files, it appears to be the place to make an exception.

pnorbert: shouldn't we use kwsys functions to make this command portable?

vicentebolea: Yes, we should as long as there are no stdlib alternative since they are platform independent. I was taking liberties here (avoid writing a recursive remove dir function) since we only enable these tests in UNIX-like systems

vicentebolea: Oh, `SystemTools::RemoveADirectory` seems to do the recursive rm for dirs.

PR #4607, opened 2025-08-08
Remove system('cat file')

eisenhauer:


eisenhauer:
Merging because san failure is unrelated and this does have an approval.

PR #4625, opened 2025-09-09
Throw exceptions to gauge stdio usage in CI

scottwittenburg:
DO NOT MERGE

Test how much the `FileStdio` transport is used/tested in CI.


scottwittenburg:
In [917a1f0](https://github.com/ornladios/ADIOS2/pull/4625/commits/917a1f03478856a8c5ee62d484d9597f8f88408e) I just verified that nothing in CI tests append mode with FileStdio, as can be seen in the test results [here](https://open.cdash.org/index.php?compare1=61&filtercount=1&field1=revision&project=ADIOS&showfilters=0&limit=100&value1=917a1f03478856a8c5ee62d484d9597f8f88408e&showfeed=0).

PR #4548, opened 2025-06-12
Data size based aggregation

scottwittenburg:
Goal: Add a new aggregation approach on the writer side, where ranks are grouped into MPIChains of roughly equal data size, for writing.

* Add new helper module to handle partitioning of ranks based on data size.  It should be easy to add new partitioning strategies, and eventually we should allow selecting one at runtime using config options
* Add an MPIChain constructor that will put ranks into substreams based on a computed data-size based partitioning of the ranks
* Update `BP5Writer` to collect some member fields into a new structure containing the things that should be cached. The purpose of this is to avoid opening the same files repeatedly for subsequent time steps, when it can be avoided.
* Update `BP5Writer` initialization logic to support data-size based aggregation (which is off by default and enabled with a configuration setting).

PR #4627, opened 2025-09-10
transport: Fix FileStdio::OpenChain() behavior

scottwittenburg:
Fix some apparent issues with `FileStdio` behavior in `Open()` and `OpenChain()`:

- Do not open async unless there is only one process
- First process in chain opens write mode, others open append mode
- Fix the append mode open flags ("rwb" looks like UB)
    - https://en.cppreference.com/w/cpp/io/c/fopen


eisenhauer (dismissed):
Looks good...

scottwittenburg:
Testing this with `EveryoneWritesSerial` results in the same failures I saw when I cherry-picked it for use it in #4548, so let's not merge this.

PR #4629, opened 2025-09-11
libfabric/RDMA dataplane of SST: Fix detection of CXI key

franzpoeschel:
This fixes a bug introduced in #3964: After this PR, CXI keys were no longer detected unless explicitly setting FABRIC_IFACE.

TODO:

- [ ] Testing
- [ ] Ensure that `init_fabric()` picks up the same interface for which the CXI key was retrieved. I think that it currently just assumes the right ordering of interfaces, but we cannot be sure of that.


eisenhauer:
This looks good, so I'm going to merge.  If testing reveals issues, we can fix.

PR #4634, opened 2025-09-15
generate-adios2-config: do not hide `CC` settings

mathstuf:
Forcing the compiler to be `mpi*` compiler wrappers via the environment does not work if the default compiler is selected via this mechanism. Instead, tell CMake to use the compiler paths using its `_COMPILER` variables so that the environment can continue to select a different compiler.

---
Cc: @vicentebolea


mathstuf:
I was able to avoid needing this with `MPICH_CC` and `OMPI_CC` (similar for `_CXX` variables) here, but it still seems useful in case other setups also want to use non-default underlying toolchains.

vicentebolea (approved):

PR #4637, opened 2025-09-16
Changes to the build system for the bpcmp utility

### LINKED ISSUE #4333: Update the util build system based on suggestions in #4308
Based on the discussion in #4308 the build system for the utility scripts in `source/util` needs a few updates.
1. Use `CMAKE_CURRENT_BINARY_DIR` instead of `PROJECT_BINARY_DIR`
2. Add quotes for all paths
3. Use the `maybe_xyz` convention in all cases when something might be generated or not
4. Use using syntax=yaml to highlight the syntact in the generated cmake files for bpcmp and bpls

@vicentebolea, did I forget anything?

anagainaru:
Closes #4333 

The changes requested in the issue were:
1. Use `CMAKE_CURRENT_BINARY_DIR` instead of `PROJECT_BINARY_DIR`
2. Add quotes for all paths
3. Use the `maybe_xyz` convention in all cases when something might be generated or not

I made the changes everywhere relevant in the utils folder.

In a hierarchical project with subdirectories, when CMake processes a CMakeLists.txt file within a subdirectory, `CMAKE_CURRENT_BINARY_DIR` will reflect the specific build output directory for that subdirectory. `PROJECT_BINARY_DIR` refers to the build directory associated with the most recently encountered project() command. 

This means, in some places it's better to use `PROJECT_BINARY_DIR`.
Example, if I change it from `PROJECT_BINARY_DIR` to `CMAKE_CURRENT_BINARY_DIR` in `source/utils/CMakeLists.txt`:
```cmake
OUTPUT "${PROJECT_BINARY_DIR}/$<CONFIG>/bpcmp.cmake" 
```
the bpcmp.cmake file will be in `/path/to/adios/build/source/utils/Release` instead of `/path/to/adios/build/Release`

But this file is used for Testing for which `PROJECT_BINARY_DIR` is `/path/to/adios/build` and `CMAKE_CURRENT_BINARY_DIR` is `/path/to/adios/build/Testing/..` so I would manually have to correct the path. 

Same is for where I am storing `bpcmp.py` file, it should be in the `bin` folder of the build directory and not in the `source/utils/bpcmp/bin` folder.

### DIFF
--- a/source/utils/bpcmp/CMakeLists.txt
+++ b/source/utils/bpcmp/CMakeLists.txt
@@ -1,7 +1,6 @@
 configure_file(bpcmp.py "${PROJECT_BINARY_DIR}/bin/bpcmp.py" COPYONLY)
 
 install(PROGRAMS bpcmp.py
-  RENAME bpcmp
   DESTINATION ${CMAKE_INSTALL_BINDIR} COMPONENT adios2_scripts-runtime
   ${ADIOS2_MAYBE_EXCLUDE_FROM_ALL}
 )
--- a/source/utils/bpcmp/CMakeLists.txt
+++ b/source/utils/bpcmp/CMakeLists.txt
@@ -1,4 +1,4 @@
-configure_file(bpcmp.py ${PROJECT_BINARY_DIR}/bin/bpcmp.py COPYONLY)
+configure_file(bpcmp.py "${PROJECT_BINARY_DIR}/bin/bpcmp.py" COPYONLY)
 
 install(PROGRAMS bpcmp.py
   RENAME bpcmp
--- a/source/utils/CMakeLists.txt
+++ b/source/utils/CMakeLists.txt
@@ -27,12 +27,12 @@ install(TARGETS bpls EXPORT adios2
 )
 configure_file(
   "${CMAKE_CURRENT_SOURCE_DIR}/bpls/bpls.cmake.gen.in"
-  "${PROJECT_BINARY_DIR}/bpls.cmake.gen"
+  "${CMAKE_CURRENT_BINARY_DIR}/bpls.cmake.gen"
   @ONLY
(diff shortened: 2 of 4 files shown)


source/utils/bpcmp/CMakeLists.txt
```diff
@@ -1,7 +1,6 @@
 configure_file(bpcmp.py ${PROJECT_BINARY_DIR}/bin/bpcmp.py COPYONLY)
 
 install(PROGRAMS bpcmp.py
-  RENAME bpcmp
```
vicentebolea, line 4: Is this intended?

vicentebolea:
Looks good otherwise

source/utils/bpcmp/CMakeLists.txt
anagainaru: Yes, I found it weird for bpcmp to loose the py extension in the install bin folder. If there is any reason for this to happen I can re-add the line but if I was a user I would prefer using the python extension.

anagainaru:
> Looks good otherwise

Great, I'll make the same changes for bpls and the other utilities that need it and update the PR

source/utils/bpcmp/CMakeLists.txt
pnorbert: It's a trend I started with the python scripts installed into bin/. Feel free to override this trend and do as people usually would prefer.

vicentebolea: Sounds good, it is customary to omit the extension in installed scripts since uses do not care which lang it was the script written for them it is just another command. However, if the rest of the scripts are `.py` I say make this also `.py` since what we def do not want are some with `.py` and some without it.

vicentebolea (approved):

PR #4647, opened 2025-10-03
Rerouting aggregator

scottwittenburg:
Implement aggregation method inspired by the paper "Can I/O Variability Be Reduced on QoS-Less HPC Storage Systems?" ([link](https://doi.org/10.1109/TC.2018.2881709)).

Currently only works with `DataSizeBased` aggregation.

Does not yet do any actual re-routing, only re-implements writing using a new multithreaded messaging scheme based on the paper above. 

Adds some missing functionality to `adiosComm` (probe api and support for `MPI_ANY_SOURCE`)

### DIFF
--- a/source/adios2/engine/bp5/BP5Engine.h
+++ b/source/adios2/engine/bp5/BP5Engine.h
@@ -151,6 +151,7 @@ public:
     MACRO(DirectIOAlignOffset, UInt, unsigned int, 512)                                            \
     MACRO(DirectIOAlignBuffer, UInt, unsigned int, 0)                                              \
     MACRO(AggregationType, AggregationType, int, (int)AggregationType::TwoLevelShm)                \
+    MACRO(EnableWriterRerouting, Bool, bool, false)                                                \
     MACRO(AsyncOpen, Bool, bool, true)                                                             \
     MACRO(AsyncWrite, AsyncWrite, int, (int)AsyncWrite::Sync)                                      \
     MACRO(GrowthFactor, Float, float, DefaultBufferGrowthFactor)                                   \
--- a/source/adios2/engine/bp5/BP5Engine.h
+++ b/source/adios2/engine/bp5/BP5Engine.h
@@ -150,8 +150,8 @@ public:
     MACRO(DirectIO, Bool, bool, false)                                                             \
     MACRO(DirectIOAlignOffset, UInt, unsigned int, 512)                                            \
     MACRO(DirectIOAlignBuffer, UInt, unsigned int, 0)                                              \
-    MACRO(AggregationType, AggregationType, int, (int)AggregationType::TwoLevelShm)                \
-    MACRO(EnableWriterRerouting, Bool, bool, false)                                                \
+    MACRO(AggregationType, AggregationType, int, (int)AggregationType::DataSizeBased)              \
+    MACRO(EnableWriterRerouting, Bool, bool, true)                                                 \
     MACRO(AsyncOpen, Bool, bool, true)                                                             \
     MACRO(AsyncWrite, AsyncWrite, int, (int)AsyncWrite::Sync)                                      \
     MACRO(GrowthFactor, Float, float, DefaultBufferGrowthFactor)                                   \
--- a/source/adios2/engine/bp5/BP5Writer.cpp
+++ b/source/adios2/engine/bp5/BP5Writer.cpp
@@ -369,7 +369,20 @@ void BP5Writer::WriteData(format::BufferV *Data)
             WriteData_EveryoneWrites(Data, false);
             break;
         case (int)AggregationType::EveryoneWritesSerial:
+            WriteData_EveryoneWrites(Data, true);
+            break;
         case (int)AggregationType::DataSizeBased:
+            // First initialize aggregator and transports if we haven't done it yet this step
+            if (!m_AggregatorInitializedThisStep)
+            {
+                // We can't allow ranks to change subfiles between calls to Put(), so we only
+                // do this initialization once per timestep. Consequently, partition decision
+                // could be based on incomplete step data.
+                InitAggregator(Data->Size());
+                InitTransports();
+                m_AggregatorInitializedThisStep = true;
+            }
+
             // For rerouting to be useful, there must be multiple writers sending
             // data to multiple subfiles.
             if (m_Parameters.EnableWriterRerouting && m_Comm.Size() > 1 && m_Aggregator->m_SubStreams > 1)
@@ -398,19 +411,6 @@ void BP5Writer::WriteData(format::BufferV *Data)
 
 void BP5Writer::WriteData_EveryoneWrites(format::BufferV *Data, bool SerializedWriters)
 {
-    if (m_Parameters.AggregationType == (int)AggregationType::DataSizeBased)
-    {
-        if (!m_AggregatorInitializedThisStep)
-        {
-            // We can't allow ranks to change subfiles between calls to Put(), so we only
-            // do this initialization once per timestep. Consequently, partition decision
-            // could be based on incomplete step data.
-            InitAggregator(Data->Size());
-            InitTransports();
-            m_AggregatorInitializedThisStep = true;
-        }
-    }
-
     const aggregator::MPIChain *a = dynamic_cast<aggregator::MPIChain *>(m_Aggregator);
 
     // new step writing starts at offset m_DataPos on aggregator
--- a/source/adios2/engine/bp5/BP5Writer.cpp
+++ b/source/adios2/engine/bp5/BP5Writer.cpp
@@ -90,15 +90,15 @@ helper::RankPartition BP5Writer::GetPartitionInfo(const uint64_t rankDataSize, c
     int numPartitions = subStreams <= 0 ? std::max(parentSize / 2, 1) : subStreams;
     m_Profiler.AddTimerWatch("PartitionRanks");
     m_Profiler.Start("PartitionRanks");
-    helper::Partitioning partitioning = helper::PartitionRanks(allsizes, numPartitions);
+    m_Partitioning = helper::PartitionRanks(allsizes, numPartitions);
     m_Profiler.Stop("PartitionRanks");
 
     if (parentRank == 0 && m_Parameters.verbose > 0)
     {
-        partitioning.PrintSummary();
+        m_Partitioning.PrintSummary();
     }
 
-    return partitioning.FindPartition(parentRank);
+    return m_Partitioning.FindPartition(parentRank);
 }
 
 StepStatus BP5Writer::BeginStep(StepMode mode, const float timeoutSeconds)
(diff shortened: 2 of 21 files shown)


scottwittenburg:
I think this is ready for your review @pnorbert and @eisenhauer. Once we finish iterating on the changes, and you're nearly ready to approve, let me know so I can remove the WIP commit at the tip of the branch. That change should not be merged.

testing/adios2/helper/TestRerouteMessage.cpp
eisenhauer, line 1: Should this be here?  Not really a BP test per se.

scottwittenburg: Ok, let me look into moving it somewhere more appropriate.

testing/adios2/engine/bp/TestBPDataSizeAggregate.cpp
eisenhauer, line 1: Lots of repeated edits in all the affected BP tests.  This isn't unexpected, but I'm wondering if we should take this opportunity to do it differently and maybe make a more general change instead of one that supports only this specific purpose.  In particular, making all the BP tests take a list of engine params as the extra argument (rather than just the string WithRerouting) would be more general and perhaps useful in the future.  Rather than changing the output filename, we could also just use a special BP5_RR_DIR rather than BP5_DIR for the output to avoid conflicts.  That keeps the BP test changes generic.  (Might need a test support file to hold the parameter parsing code so that it isn't repeated in every test.)

source/adios2/engine/bp5/BP5Writer.cpp
```diff
@@ -1069,6 +1094,34 @@ void BP5Writer::EndStep()
...
+        std::cout << "  m_AppendAggregatorCount: " << m_AppendAggregatorCount
+                  << ", m_Aggregator->m_NumAggregators: " << m_Aggregator->m_NumAggregators
+                  << std::endl;
+        std::cout << "  m_AppendSubfileCount: " << m_AppendSubfileCount
+                  << ", m_Aggregator->m_SubStreams: " << m_Aggregator->m_SubStreams << std::endl;
+    }
+
+    if (!m_WriterStep || m_AppendWriterCount != static_cast<unsigned int>(m_Comm.Size()) ||
+        m_AppendAggregatorCount != static_cast<unsigned int>(m_Aggregator->m_NumAggregators) ||
+        m_AppendSubfileCount != static_cast<unsigned int>(m_Aggregator->m_SubStreams ||
```
eisenhauer: This is probably a typo.  it's doing an || of m_SubStreams and EnableWriterRerouting (producing a bool), and then comparing that to m_AppendSubfileCount (an int).  Probably EnableWriterRerouting snuck into the static_cast unintentionally?

eisenhauer: And I'm curious about actual intent.  Is the implication that a new writer map is needed (per the comment) on every step (because it's only new condition is EnableWriterRerouting)?  Or is this just gathering values to see if a new map should be done?

scottwittenburg: The entire if is supposed to be an || or 5 things: a boolean, 3 inequality expressions, another boolean.

eisenhauer: I think that last || crept one parenthesis too far in...

scottwittenburg: This existed before (see down below), but it was just a check of 4 conditions, any of which could trigger a new writer map to be added. I just moved it here and added a condition (if we're doing rerouting).

eisenhauer: I positioned my comment poorly, so it broke up the relevant line.  This was the prior line, an or of 4 conditions (sorely in need of clarifying parens IMHO):
```
    if (!m_WriterStep || m_AppendWriterCount != static_cast<unsigned int>(m_Comm.Size()) ||
        m_AppendAggregatorCount != static_cast<unsigned int>(m_Aggregator->m_NumAggregators) ||
        m_AppendSubfileCount != static_cast<unsigned int>(m_Aggregator->m_SubStreams))
```

This is the new line, still an or of 4 conditions:

```
    if (!m_WriterStep || m_AppendWriterCount != static_cast<unsigned int>(m_Comm.Size()) ||
        m_AppendAggregatorCount != static_cast<unsigned int>(m_Aggregator->m_NumAggregators) ||
        m_AppendSubfileCount != static_cast<unsigned int>(m_Aggregator->m_SubStreams ||
                                                          m_Parameters.EnableWriterRerouting))

```
Your added "|| m_Parameters.EnableWriterRerouting" is inside the static_cast<unsigned int>() when it should be outside of it so that it's the 5th boolean.  The indentation is the clue.

scottwittenburg:
I don't know what's going on with github right now. It's duplicating my comments, and for some reason they're all marked "pending" for me, even though I just added comments (not part of a review).

source/adios2/engine/bp5/BP5Writer.cpp
scottwittenburg: Yes, I see that now. The result of this typo was likely that a new writer map was created on every time step under all conditions. I'll fix it, great catch!

eisenhauer: Not necessarily a bad thing that this happened, the new map on every step probably helped catch that StepSelection bug.  I'm also curious if this will resolve the Archive test (will that routing be stable and the same as before?) or if we'll just have to force a stable aggregator for those if rerouting becomes the new default.

testing/adios2/engine/bp/TestBPDataSizeAggregate.cpp
scottwittenburg: I'm looking into how to do this, and I've noticed some confounding issues in the codebase. On one hand, there's already a function to take a string of parameters from the command line and turn it into a parameter map, so yay!. It's called `ParseEngineParameters`. Only problem is, it is redefined, identically, in six different TUs under the `testing` directory. 

Well, that is until today, where in #4831, identical changes were made to two copies of the function, while the other four copies were left untouched :stuck_out_tongue_winking_eye:

eisenhauer: That is almost certainly my bad.  I had originally had this in staging-common, it got reused, rather than refactored, in many places because it was a simple little thing.  But then it needed to be more complex, so it got expanded but again not refactored.  I'd say it should go in helper, but I'm not sure the tests have direct access to that.   I'll look tomorrow.

eisenhauer: A recent refactor created  testing/adios2/engine/TestHelpers.h to try to keep the file cleanup stuff in one place.  I probably should have moved ParseEngineParams there too, but didn't catch that at the time.  That would be well positioned to at least get the uses that are under engine testing.  Shall I try that tomorrow so that you don't have that mixed in with this PR?

scottwittenburg: I just tried moving the version from `ParseArgs.h` to `TestHelpers.h`, but the problem is that is included from the c bindings tests, which don't know about `adios2`, only `adios2_c`. So currently I'm putting it into it's own helper file, a sibling with `TestHelpers.h`.  I should have a reorganizing commit for you to look at by tomorrow sometime.

scottwittenburg: Ok, see what you think of #4901 as a cleanup. If it seems acceptable, I'll rebase this PR on it and use it to make the BP5 tests take lists of engine params, instead of the special `WithRerouting` argument I added initially.

scottwittenburg:
@eisenhauer Let me when you think you don't have any more questions or suggestions. At that point I will rebase this on the SZ3 fix from yesterday (to clean up unrelated CI failures), remove the WIP commit making rerouting the default, and do one last CI run.

source/adios2/engine/bp5/BP5Engine.h
eisenhauer, line 1: Are all the changes here in BP5Engine.h just for testing this as the default aggregator?  Assume so, but checking.

scottwittenburg: Yes, that's correct.

eisenhauer: Then I think my concerns are resolved.  I'm happy with it.

scottwittenburg: Awesome, thanks @eisenhauer!

scottwittenburg:
You can't see it on CDash yet for some reason (looking into that), but the issue is that `Archive.Common.bp5.Unique` is failing on those two jobs. This is surprising because this happened without DSB+Rerouting as the default. When I reproduce it locally it says:

```
The file /.../build/testing/adios2/output_archive/CurrentCommon.bp5.zip has a unique size and checksum value and should be placed in the zipped_output directory.
```

@eisenhauer Is there any way to see what is different? Or is there only the size/checksum stored from the original run?

scottwittenburg:
I *suspect* it's because I added a new attribute into the `transport_N` dictionaries of the `profiling.json`, namely, the filename. So now those look like this:

<details>
  <summary>profiling.json excerpt</summary>

```
    "transport_0": {
      "type": "File_POSIX",
      "name": "data.0",
      "wbytes": 33920,
      "close": {
        "mus": 67,
        "nCalls": 1
      },
      "write": {
        "mus": 173,
        "nCalls": 40
      },
      "open": {
        "mus": 163,
        "nCalls": 1
      }
    },
    "transport_1": {
      "type": "File_POSIX",
      "name": "md.0",
      "wbytes": 53200,
      "close": {
        "mus": 25,
        "nCalls": 1
      },
      "write": {
        "mus": 46,
        "nCalls": 40
      },
      "open": {
        "mus": 150,
        "nCalls": 1
      }
    }
```

</details>

However, I'd like to be sure that's the only difference.

eisenhauer:
> I _suspect_ it's because I added a new attribute into the `transport_N` dictionaries of the `profiling.json`, namely, the filename. So now those look like this:
> profiling.json excerpt
> 
> However, I'd like to be sure that's the only difference.

It won't be the profiling.json.  That has timing information in it, so we simply don't add it to the zip files.  (We also reset timestamps to 0 and other things so that we can actually compare 2 adios files in a sensible way.). So it's something else.  Basically you can just unzip your CurrentCommon.bp5.zip and Common.2.10.bp5.zip into different directories and do diff -r. Maybe something you've changed is writing extra writer records or something?  At least diff -r will tell you what file is different.

scottwittenburg:
> Maybe something you've changed is writing extra writer records or something? At least diff -r will tell you what file is different.

Using `bp5dbg`, I can see that my change has caused there to be more writer map entries in the file, and to me it appears that shouldn't have been needed since the additional writer maps all look identical.

The two things I changed in that department are:

1. I moved the decision point for needing a new writer map into `EndStep()` (though maybe that was unnecessary, since I don't really check whether rerouting actually *happened* or not, just whether it is enabled)
2. I added two conditions to the list of conditions that trigger a new writer map (rerouting is enabled or agg type is DSB)

At least I can see what changed now, and I can try to figure out why. Thanks @eisenhauer.

scottwittenburg:
I've updated the conditions that are used to decide whether a new writer map is needed, and now we no longer see the extra writer maps in the `Archive.Common` zip file. Let me know what you think @eisenhauer.

PR #4649, opened 2025-10-06
Making build with HDF5 more informative

guj:
1) display whether HDF5 is parallel or serial if present 
2) the behaviour of ADIOS2_USE_HDF5=AUTO/TRUE is consoliated when HDF5
   is present

 FYI: SERIAL HDF5 compiles fine


guj:
win2022 had many failed tests, not sure why.   Any clue on that platform?

source/h5vol/H5VolReadWrite.c
```diff
@@ -396,17 +417,30 @@ bool gRemoveUnderGrp(H5VL_ObjDef_t *owner, const char *obj_name)
...
             return true;
+        }
     if (adios2_error_none == adios2_remove_variable(&result, owner->m_FileIO, fullPath))
         if (adios2_true == result)
+        {
+            free(fullPath);
             return true;
+        }
 
+    free(fullPath);
```
scottwittenburg, line 443: There appears to be a problem here (caught by rebasing #4666 on `master` today): `fullPath` is still used below, after this call to `free` it.

I'm not sure how this isn't causing any problems in CI. Maybe the function never gets called in CI? Or some early-return codepath above is always taken? Or `NEVER` is, in fact, defined?

eisenhauer: It's in a failure path, likely never executed in CI...

scottwittenburg: Ok, fix attempted in #4698

PR #4650, opened 2025-10-07
CURL derived variable applied on an aggregated array

anagainaru:
This is needed by the WarpX derived variable that computes the `curl(cross(E,B))`

Curl currently requires three 3D arrays `curl(Ex, Ey, Ez)` while cross returns an array with all three vectors merged into a single N+1 dimension array. After this PR curl can be applied as `curl(E)` with Ex, Ey, Ex concatenated one after another into a 4D array (E) similar to how cross creates it.


eisenhauer (dismissed):
Looks good...

anagainaru:
I tuned off the derived variables for merging so I need another review @eisenhauer

PR #4651, opened 2025-10-07
Fixing a bunch of small bugs/typos for the derived variable engine

anagainaru:


anagainaru:
I merged your fix @eisenhauer with the rest of the fixes in the firs commit. I turned off the derived variables since it was passing all tests (except Frontier that is constantly down). This is ready for review

PR #4653, opened 2025-10-08
Python API fixes and examples:

pnorbert:
- import from bindings the names LocalValueDim, Mode, ShapeID, StepMode, StepStatus, DerivedVarType, Accuracy
- define JoinedDim (since it is not present in bindings)
- example for LocalValueDim, JoinedDim, and deferred read

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -83,4 +83,4 @@ include = 'python/adios2/.*.py|testing/adios2/python/.*.py'
 
 [tool.pylint]
 disable = ['E0401', 'W0102', 'R0904', 'R0913', 'R0401']
-good-names = ['io']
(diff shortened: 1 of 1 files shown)


pnorbert:
`python/adios2/__init__.py:32:0: C0103: Constant name "JoinedDim" doesn't conform to UPPER_CASE naming style (invalid-name)`

I want it `JoinedDim` so it looks similar to the other enums/constants in bindings. `JOINEDDIM` is understandably preferred I just don't like it. 

@vicentebolea @anagainaru What should it be?

anagainaru:
I prefer `JoinedDim` so it's the same as the CPP version. There are ways to suppress the warning, I'd go with this version

Issue #4657, opened 2025-10-10
Missing #include <iomanip> when using std::setw


Micket:
With GCC 14.3.0 I'm seeing a few compilation errors. 

testing/adios2/interface/TestADIOSSelection.cpp
testing/adios2/engine/bp/TestBPBufferSize.cpp
testing/adios2/engine/skeleton/TestSkeletonReader.cpp

all use `std::setw`, e.g.

https://github.com/ornladios/ADIOS2/blob/859ac9770ff51ea57d351ca4ed31c2ea615f7c1a/testing/adios2/interface/TestADIOSSelection.cpp#L100

They should `#include <iomanip>`
https://en.cppreference.com/w/cpp/io/manip/setw.html

Edit: also, std::setprecision in 
testing/adios2/engine/skeleton/TestSkeletonReader.cpp
testing/adios2/engine/staging-common/TestCommonRead.cpp
testing/adios2/engine/staging-common/TestCommonReadAttrs.cpp
also requires iomanip.

eisenhauer:
Interesting.  I buy this, but I'm actually not able to replicate it after trying gcc 14.3.0 in several different environments.  Can I ask where you are running?

Micket:
I'm trying to package ADIOS2 for EasyBuild  (ref. https://github.com/easybuilders/easybuild-easyconfigs/pull/24194)
I happened to build it on Rocky Linux 9, pretty average HPC node.
I don't think I'm doing anything particularly special here, checking the build output

```
[...]

PR #4660, opened 2025-10-13
Fix Warnings

eisenhauer:


eisenhauer:
As noted in issue #4657, the weirdness with this in the testing directory seems to be that this file is included by gtest.h, so we're not getting warnings about it not being present there.  Elsewhere (core source and probably examples) we may already include it (unless we've been ignoring warnings).  Just exactly why OP is seeing warnings in the testing code while we are not remains a mystery.

PR #4662, opened 2025-10-14
Bug fix for derived variables when used on Joined Arrays

anagainaru:
Bug fix needed (once again) by the Warpx derived variables. The particles are joined arrays and when I set the mapping between variables and their dimensions I was getting a seg fault trying to get start values (that are empty for joined arrays and I was expecting start and count to have the same size). I fixed the bug and added testing for this.

While testing the `curl(cross)` expression on doubles I was sometimes getting seg fault (UB at its finest) so the PR includes one extra fix for this.

### DIFF
--- a/testing/adios2/derived/TestBPDerivedCorrectnessMPI.cpp
+++ b/testing/adios2/derived/TestBPDerivedCorrectnessMPI.cpp
@@ -25,6 +25,64 @@ protected:
     adios2::DerivedVarType GetThreads() { return GetParam(); };
 };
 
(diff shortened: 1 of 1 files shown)


anagainaru:
> So now we are turning Derived Variables ON for everyone? Or is this just an oversight here?

I will turn it back off when the CI passes. I was overly optimistic that everything will pass when I asked for your review (unsurprisingly it did not pass first go)

anagainaru:
Rebased and cleaned history. The CI (except Frontier but unrelated) passes with derived variables on. This can be reviewed again and merged finally.

PR #4666, opened 2025-10-15
ci: Update os and spack/mirror versions for ubuntu docker images

scottwittenburg:
* Update to use Ubuntu 22.04 instead of 20.04
* Replace gcc8 with gcc12
* Replace clang 6 and 10 with clang 11 and 14
* Update to spack `v1.0.2`, and use mirror `https://binaries.spack.io/v2025.07.0`
* Remove custom blosc2 build in favor of spack-built version
* Remove unneeded adios2 patch, update the xrootd patch
*  Replace macos 13 jobs with macos 15 (now all macos jobs are arm64)

For now, leave the rocm build as-is. Spack v1.0.2 builds many of the libraries included in the rocm dev base image, and sorting it out will take a little more effort. Another option would be to avoid spack and just install dependencies from the package manager.

### DIFF
--- a/testing/adios2/performance/manyvars/PerfManyVars.c
+++ b/testing/adios2/performance/manyvars/PerfManyVars.c
@@ -91,7 +91,7 @@ void alloc_vars()
     }
 
     char fmt[16];
-    snprintf(fmt, sizeof(fmt), "v%%%d.%dd", digit, digit);
+    snprintf(fmt, sizeof(fmt), "v%%%d.%dd", digit % 333, digit % 333);
     printf("fmt=[%s]\n", fmt);
     for (i = 0; i < NVARS; i++)
     {


scottwittenburg:
The new rocm image is more than 2x larger than before, and it seems there are two culprits: a new dependency on `llvm-amdgpu`, and `concretizer:unify:true` config doesn't seem to be working as well as before.

<details>
  <summary>New spack packages/sizes:</summary>

```
root@5b0b6614647c:/spack/opt/spack/linux-ubuntu22.04-x86_64_v3/gcc-11.4.0# du -sh * -- * | sort -rh
4.3G	llvm-amdgpu-6.3.2-rfb2ij3x4eqbtc3tpa3cfmh6tdk4mx6t
319M	python-3.13.2-idqwntoqahowcuqq3tapf2gpvwffnons
290M	python-3.11.11-zecmeqishqsprpv3hpoh5ml3libjg7bf
103M	openblas-0.3.29-elsyplctasbovgg6kib7l6du6if3sxv4
98M	    cmake-3.31.6-sy6oisonvapix6tiaiqojdehsy5ib2gj
98M	    cmake-3.31.6-5kskfvdrlk72ervfp646rgpf2mjvqg3j
98M	    cmake-3.31.5-44vqqmcelkowo36kdu67pdbtcu7sglxp
84M	    re2c-3.1-oekbeicvvtdmsgwamkvokbgoqr3wtxk4
84M	    re2c-3.1-lu3u7xvhgddtk33wrkfduqiqtwt43yfd
76M	    cmake-3.25.3-nb4mmztfjb2zsbt3neqydyb2ny45kgdg
67M	    z3-4.12.4-dnjwiunlajd72wzwaudyb75udml34zkv
65M	    perl-5.40.0-nytabtxapnpwodt2vw4qn5uawp2mq62b
65M	    perl-5.40.0-7hpzkhzodhb53kqbtzmtymbbnvk2mbob
56M	    libzmq-4.3.5-hv3ruwz4npau3nvmmdmllptvr4jyhodk
46M	    py-numpy-2.2.4-y2cunu6ebdg5uyl7lxjhnbkzspv2doeg
45M	    py-cython-3.0.11-qw6seqzfmzpijziy3wlfjnrrctak6u6u
44M	    gettext-0.23.1-ft3tslwgv4ink5xbdfy4bvsc74uhpgnc
34M	    elfutils-0.192-b3rpq4vyt7suap4huikb5kl37ktpifq7
30M	    ncurses-6.5-3gr2eej675rjr5k3bgqnprcbbfziizg2
29M	    mgard-2023-12-09-iaclfionw2yo33dn3mbskvrpd2agcmro
28M	    sqlite-3.46.0-cnftqr77mkl747a46oqz6eoynvchm4t3
26M	    hdf5-1.14.5-hdcculsb4i2xp6ehbsp64h4v5irawbc3
26M	    gcc-runtime-11.4.0-f63c77kavzjtpmnhucd2oyfaxagwjzla
25M	    openssl-3.4.1-isb3fwjeatupjzn3umwu6bgmttsi7n4k
25M	    openssl-3.4.1-4f4zgr3fnnzpvwkfww2tvee4nksgqavw
25M	    openssl-3.4.0-ckalxk3vs33br2xu6u3jwy22r7by4p7d
15M	    py-pip-24.3.1-ey2q3xmnaq2rqrmbxfhw7j7hmpxqpgdw
15M	    py-pip-24.3.1-dhb5ujvvzva53i4owikb5klri5hs4qkh
15M	    libarchive-3.7.6-j2hjpgc4223ltq7jmpsdfezso7druy6h
13M	    py-setuptools-76.0.0-uvdgy5273uwexg7fuf66gbld5g36olkf
13M	    py-setuptools-76.0.0-lfbtsbh3wukhgr7usskw2t73abddznpu
13M	    protobuf-3.21.12-lfb4qzyldi2pi5klpnc3vb437pmvqo7q
12M	    util-linux-uuid-2.40.4-rc3nt2qvkcq7adyuyrp23jyuovyhj36z
12M     meson-1.7.0-ydozucrkigs5kd33yfbxnod3qhu4m2yo
11M	    meson-1.7.0-xw2ocyvesjebjldaxcm52rrq2563viuq
8.8M	libxml2-2.13.5-r3bbnyomqkf42o22auiikspoibyzuimg
8.3M	ninja-1.12.1-jfskofq3sfbipg55o3agnzydvc55qstq
8.3M	berkeley-db-18.1.40-ts7yab4czwndtli4koyst6q4nqh7g4fs
8.2M	ninja-1.12.1-fmpxiyn6ezzghwbn2eksslr6te4la63l
7.0M	hsa-rocr-dev-6.3.2-5r4ukox6oheyeba2bnj72dhwl26b2m5p
6.1M	xz-5.6.3-fxpjaxtjfum5jgj5msax4rwk4b6ttrfm
5.0M	libiconv-1.17-pls4r2ckybf2cjwkqckvfuba2ebr5tiy
4.6M	findutils-4.10.0-6ijijtjswxu5lub2wqgldxrz4i5p2pn4
4.5M	libfabric-1.22.0-shhhrzrzqndevpcujivyuw2rexdyolzt
4.4M	readline-8.2-q4odcrx7xvvvsf6su3dnyso4gf53y3pi
4.1M	curl-8.11.1-rm7cxxek7lcd73lhev7wgua47vwfuymv
4.1M	curl-8.11.1-ei6ugfoaafflyv2pcs7uuhxwpepiyngy
4.1M	curl-8.11.1-cmdq2hrpqzdcpf2g6hm6l23ez3hchdp5
4.1M	autoconf-2.72-su2oye753ruglddg3u3qmyxu5hzy4ojj
3.8M	tar-1.35-rhgmeaff3edcdv6m22okanlf2y6sipz7
3.7M	zstd-1.5.6-c5vvwejfl3mrmwofcdwhml2wjtnfsups
3.6M	libtool-2.4.7-dhyw5lpgndo63hvc2v6avs5ej4jye2i2
3.4M	diffutils-3.10-rrxy6iwevvx65eil3jlfa46ki4ie6ih4
3.3M	sz-2.1.12.5-h6vf4lmm6t3zaupygbevc5u3we3nfjcj
3.1M	c-blosc2-2.15.1-3zwggadbap22a2uf7443l5na5dtnjmfj
3.0M	nghttp2-1.64.0-f3ay3yae4btihhc7cwadgdmdg3mgtbig
3.0M	gdbm-1.23-67ajshqetnukqfmigdotqyhfaxrkub7a
3.0M	automake-1.16.5-fro26mmvmrwq6n7mzxrkqz73vkfjuszr
2.9M	nghttp2-1.65.0-k2rvlhlqagfi2my73gefc67kjwomgxcz
2.7M	libbsd-0.12.2-5ificcq6myoramfzfvysn544iah3og4p
2.5M	libxcrypt-4.4.38-k5f2qc76qjsfdaf2pnfhrqaaamccfbjl
2.4M	libdrm-2.4.124-6jyiseecjfyxt6tkjriuvsmpasq63yl2
2.3M	m4-1.4.19-rsbk2hekprs6psvgn2irv6oosfm6fr6e
2.2M	lzo-2.10-vc5quinilggm5evpbfypbgxg66p3vapf
2.1M	libpng-1.6.39-copukfwonbk62udmigfv5ysjcpmufl4g
2.1M	expat-2.7.0-f5tfumor6dsqfl4wxkahayblhxdcb2qz
1.9M	sed-4.9-ub66wo2jidzw6oulgliqcq627d6m3yyl
1.9M	numactl-2.0.18-w5ylmswvjjaxmvmrem5s7qfmy4uyr72q
1.9M	libsodium-1.0.20-4iu4kr64g7d3xoxkj33xxb727xotcwj5
1.6M	py-wheel-0.45.1-gp63kews6gdzsq3zau6tufyzp533h2nw
1.6M	py-wheel-0.45.1-bxzgf2wecglso6jck3e5dbgh46mrzoyp
1.5M	py-packaging-24.2-vpr7re36xndzsfep4um4medaidl3ym2m
1.4M	yaml-cpp-0.8.0-gdtyofju63cti77eh3jiowjphjb4ekxk
1.3M	zfp-0.5.5-npoblsvjwulg6rle5nxgq3postg2io5a
1.3M	pkgconf-2.3.0-tsnrfeyllxu2sk6icdi4d6bppaoi2xh3
1.2M	py-meson-python-0.16.0-y5bmu224yygmmp3skstpxwhijll5tgfo
1.2M	py-flit-core-3.10.1-3e47dx6xw6ro2gzu6mr4ltb7yhpkzppk
1.2M	lz4-1.10.0-46cmkx7rt3xayhgyb3rbtk6chhnsvees
1.1M	lizard-2.0-ms7q7iz755f6yovzyiceogwpowjgsaeu
1016K	gmake-4.4.1-qp5blvcyuzgzhqsrp2ew6gq2nlos34b2
972K	py-pyproject-metadata-0.7.1-uu5vgwbu6e4nl6gpvmkzysyvhfuge5uj
940K	bzip2-1.0.8-y3fhvdex6tamsxixzhxv5yptprhbnn6p
940K	bzip2-1.0.8-wst3dom2tj76lqrd7jas7loyiuekmiac
864K	python-venv-1.0-pwij7nazzq3obiajei5bufx6msj5ayap
856K	python-venv-1.0-jk5ktsnpucffwwbjzngmi5jnba6rffd5
824K	libmd-1.1.0-i3v67zolj4i4fy76bujtadl553fe52qd
708K	libpciaccess-0.17-3togtxaksrza37udedldjk4iomkahx2f
680K	snappy-1.2.1-7hmiug2clkatcnw3fjuw6xiisk3hgbjv
672K	zlib-ng-2.2.3-t3ltsgdmknshkorrbjfjywtlbpu47adu
656K	rocm-core-6.3.2-42alddtzpvpnaxrppxhhdwnom5ywieop
592K	libsigsegv-2.14-byltaktkr5bwp6jbvxp3gnh7rgjybzqp
420K	libffi-3.4.6-xvd7m5gldf7ogzw67sn4nx7mriqeixyd
316K	pigz-2.8-hbkqxrr6uz6df2vmsxepfzt3ayrdtrus
316K	ca-certificates-mozilla-2025-02-25-qpb66jhnwximky77junbrvduqrll4lsb
316K	ca-certificates-mozilla-2024-12-31-d74kt4tmvcdtgqorvk53thiwq3gilncf
232K	util-macros-1.20.1-rkbngyczxuhmgtyapno6ed7l3xvey7kv
144K	libpthread-stubs-0.5-pftxgkbb6lejl3bk75aek74w6j6jpld5
140K	xxd-standalone-8.2.1201-3mphokhxiyr37mx4fic2hu466ol7xjyu
```

</details>


<details>
  <summary>Previous spack packages/sizes:</summary>

```
root@f73b9db30d3a:/spack/opt/spack/linux-ubuntu22.04-x86_64/gcc-11.4.0# du -sh * -- * | sort -rh
287M	python-3.11.7-beq7w7gyojt45jvcdcneocen6ikp2nkf
116M	openblas-0.3.26-lwfqkzcgsflx6kht5tl2zit5bjptq4qd
90M	    cmake-3.27.9-3gq4owhidx3exqfpga3a465lrn25wvbd
64M	    perl-5.38.0-fqkbmyzke2usautdxbtyrjm6tnx6vmjo
56M	    libzmq-4.3.5-lylawfryrncagxvxkja35hzkjrwijyuw
45M	    py-numpy-1.26.4-lawgvacsxeg3m6esocebp5adctook7cc
42M	    gettext-0.22.4-7pjbhkg343r5m6ohgs7cetv2b3vplnaj
30M	    ncurses-6.4-7a4mbz3l2vv46p5ferdpngzd5y7n5ad4
29M     mgard-2023-12-09-fyotdh6snlghjh4iebefbkp7n5yqrqoq
28M	    sqlite-3.43.2-cggmg334a2ux7ajsx3py26yq4gewkrs5
27M	    re2c-2.2-ubsvgtpahemjnd6f5seodv4ejhm2doqn
26M	    gcc-runtime-11.4.0-6cqtdncxvqlh2eh766nhbnb5dvkctcpj
24M	    openssl-3.2.1-ms4sbshdfzfigj2knaltqfwbcv5bysou
23M	    py-cython-0.29.36-cr3n6rf6re74cbrose7d5gsmq4entdzq
18M	    py-pip-23.1.2-7iflr6erktssqda77qumcbxtbrw5w67z
15M	    protobuf-3.25.3-wedunkokab2yi4cu2q6hzqqzd3potach
14M	    libarchive-3.7.1-foi6iacgxixyqbjjprupsnomrhjqqulw
11M	    meson-1.3.2-7uxm2d6re5fxvkxzjdnq2agkptys37a6
9.5M	abseil-cpp-20240116.1-ee74qz54lamv4tmt3xtob2m24seaz3j2
9.3M	libxml2-2.10.3-tomx3ziqawzwz74emtapsuowe3emcmuw
8.9M	util-linux-uuid-2.38.1-wjsrjs3f4rvgzjuge4nu35gbstkvtibm
8.5M	py-setuptools-69.2.0-kdxc323hlzgddirjh4eumcmucjwlebxp
8.4M	ninja-1.11.1-ciukme5adt66a2vxkzw6vq6hrj65o7uq
8.3M	berkeley-db-18.1.40-y2dkgvnczd5sa6y3omulo7bhutbtqhae
6.9M	xz-5.4.6-7w2vrb7ssxkoffwkeevj2wwz6b33hknk
6.7M	curl-8.6.0-bqjvf6ejuburx4f4uwuv3xvh2xj2f6kp
6.5M	tar-1.34-4kecmsc5tfj5nyq3z6meg6fql6gmdfet
5.0M	libiconv-1.17-vn2i3otirztvtyudi3ycdafgh7xy6cyu
4.5M	libfabric-1.21.0-tvt2a5gvqqfsh6jerj4ezrkhlom66ll6
4.4M	readline-8.2-fl4uzzks5vq6ux5bmondccalcys6rhpo
3.8M	zstd-1.5.6-pmda3i7gfaml65yu4ijoio3pke6vpc23
3.4M	sz-2.1.12.5-wb2ylj7dfqfpvtaxi7vwwpy2sj24btha
3.4M	diffutils-3.10-ppgdy35us4ltslujvxs5ooewwq2wy467
3.1M	c-blosc2-2.11.1-oqersnhyoc2jkmrtit737bo3aesk2jqo
3.0M	gdbm-1.23-xc5bypnuisvis242g3kl7kjph2wnacie
2.9M	nghttp2-1.57.0-pz4ulynaomnry6bdgsx2ypnrixhxvknc
2.7M	libbsd-0.12.1-ipezmagi5y7ssbijgv565asq3qthylxx
2.4M	libxcrypt-4.4.35-ifj2qvdausq5mrvz5jounggwcf6fckux
2.2M	lzo-2.10-3d6ay7fe2pyu72lffpwfpclitrzcl63v
2.1M	libpng-1.6.39-t4bvllv3pnaavr66pu2zvvp7uwnmve5t
2.0M	py-flit-core-3.9.0-texmlekzvos4fanx5o3tcuui77ppdo4i
2.0M	expat-2.6.2-47hbnkz2jaijesyabre6hcihaevwqufx
1.9M	sed-4.9-kzdcd4w5a7sb4wl7ru67puit6vwpibn2
1.9M	libsodium-1.0.19-43jx5r4vcgizqs7bxopyn4h4tywgmtni
1.8M	lz4-1.9.4-p4zke22rqjyc7huysws4tyoktwfgcpd6
1.6M	py-wheel-0.41.2-6a7c3gr5xo3km7lsba3mwkxus7jqmm3s
1.5M	yaml-cpp-0.7.0-5gdypvkpb273w7jbrlrgm5eghwzc2ugn
1.4M	py-packaging-23.1-inqb4jvox6o5kcgypwdold3372tzrdkx
1.3M	py-meson-python-0.15.0-zozobatmmnrtyb26yxddhvqqut42ewru
1.3M	pkgconf-1.9.5-xbzeeool3el44d7uznxiusg2iitr7dod
1.2M	zfp-0.5.5-n2etbebn7avzshsliof4ftded5oq4ksx
1.1M	py-pyproject-metadata-0.7.1-v3cu2nrx6rfhfbydfz2lwtlievmnidji
1.1M	lizard-1.0-h2jq3akmspfp67vkhpjj7awjxuzqadvy
1008K	gmake-4.4.1-kunl7zm5panr4w5y4g3nboeovufduaom
932K	bzip2-1.0.8-mvv5tmfnuga2tt3fx5tmq6gegli4efhd
728K	libmd-1.0.4-4o7v5qtqfmgsnir7hyshnjzox5zlkf7r
720K	snappy-1.1.10-35xzcjqtstkfq7wcljefimhrqd2o6y2f
712K	zlib-ng-2.1.6-nsjxqfdrox7d4gis42kb4hrmxapkserz
416K	libffi-3.4.6-gr7x2kqxutnk33lmmdd6ktdilrwgdkn7
352K	pigz-2.8-kox732v27ctbgrztkogyjljwq6gdubz5
296K	ca-certificates-mozilla-2023-05-30-5bui7u7xy26ktxj7lc62ypy5w5n6xahn
```

</details>

scottwittenburg:
@vicentebolea Any idea what might cause (or fix!) [this](https://github.com/ornladios/ADIOS2/actions/runs/18568562123/job/52936307735?pr=4666#step:10:1024) link error? 

```
  /usr/bin/ld: /opt/rocm-7.0.2/lib/libamdhip64.so.7: undefined reference to `hsa_ext_sampler_create_v2@ROCR_1'
  /usr/bin/ld: /opt/rocm-7.0.2/lib/libamdhip64.so.7: undefined reference to `hsa_amd_memory_get_preferred_copy_engine@ROCR_1'
```

To me it appears that everything is in order within the container, but I must be missing something:

```
root@650afa16704f:/opt/rocm-7.0.2/lib# readelf -d libamdhip64.so.7.0.70002  

Dynamic section at offset 0x4a1980 contains 36 entries:
  Tag        Type                         Name/Value
 0x0000000000000001 (NEEDED)             Shared library: [librocprofiler-register.so.0]
 0x0000000000000001 (NEEDED)             Shared library: [libhsa-runtime64.so.1]
 0x0000000000000001 (NEEDED)             Shared library: [libnuma.so.1]
 0x0000000000000001 (NEEDED)             Shared library: [libstdc++.so.6]
 0x0000000000000001 (NEEDED)             Shared library: [libm.so.6]
 0x0000000000000001 (NEEDED)             Shared library: [libgcc_s.so.1]
 0x0000000000000001 (NEEDED)             Shared library: [libc.so.6]
 0x0000000000000001 (NEEDED)             Shared library: [ld-linux-x86-64.so.2]
 0x000000000000000e (SONAME)             Library soname: [libamdhip64.so.7]
 0x000000000000001d (RUNPATH)            Library runpath: [$ORIGIN:$ORIGIN]

root@650afa16704f:/opt/rocm-7.0.2/lib# nm -D libamdhip64.so.7.0.70002 | grep -i hsa_ext_sampler_create_v2
                 U hsa_ext_sampler_create_v2@ROCR_1

root@650afa16704f:/opt/rocm-7.0.2/lib# nm -D libhsa-runtime64.so.1 | grep -i hsa_ext_sampler_create_v2
0000000000076540 T hsa_ext_sampler_create_v2@@ROCR_1
```

What does the double `@` mean in the symbol name in the `libhsa-runtime64.so.1` library, where the symbol is listed as `T`?  In the `libamdhip64.so.7.0.70002` library, where the symbol is `U`, there is only a single `@`.

scottwittenburg:
@vicentebolea Regarding the rocm changes we discussed: 

It seems that the `hip` package requires the same `<major>.<minor>.<patch>` version of many of its dependencies. The latest version available in the `e4s-25.06` branch of spack is `6.3.2`, so I could not use `rocm/dev-ubuntu-22.04:6.4.4` as the base image, or else spack wants to build it's own `hip`, which depends on `llvm` and blows the image up to 20GB.

I also tried with `rocm/dev-ubuntu-22.04:6.3.2` as the base image and `^hip@6.3.2`, but oddly, that had the same problem.  The only thing that worked was picking the same base as was used in the previous image, `rocm/dev-ubuntu-22.04:6.1.2`, and pinning `hip` to the same. That still results in a new dependency on `llvm-amdgpu` (as well as many duplicated packages, as I posted above), so the resulting image is still 10GB instead of the previous 4.6GB.

The build completed without errors or warnings though.

vicentebolea:
@scottwittenburg it then appears that latest e4s spack has an issue with the unify feature.

scottwittenburg:
Ugh, I've just noticed [these](https://open.cdash.org/index.php?compare1=61&filtercount=1&field1=revision&project=ADIOS&showfilters=0&limit=100&value1=5b2a10a95a2965f04b76e06faa0863c18dc51aee&showfeed=0) hundreds of configure warnings on the gcc builds as well.  For seemingly every target, it warns:

```
CMake Warning at source/adios2/toolkit/sst/util/CMakeLists.txt:13 (add_executable):
  Cannot generate a safe dependent library path for target sst_conn_tool
  because files in some directories may conflict with libraries in implicit
  directories:

    runtime library [libgomp.so.1] in /usr/lib/gcc/x86_64-linux-gnu/10 may be hidden by files in:
      /spack/var/spack/environments/adios2-ci-mpich/.spack-env/view/lib

  Some of these libraries may not be found correctly.
```

I noticed because I saw the same in the local image I built to specify `openmpi@4.1.8`.

vicentebolea:
> Ugh, I've just noticed [these](https://open.cdash.org/index.php?compare1=61&filtercount=1&field1=revision&project=ADIOS&showfilters=0&limit=100&value1=5b2a10a95a2965f04b76e06faa0863c18dc51aee&showfeed=0) hundreds of configure warnings on the gcc builds as well. For seemingly every target, it warns:
> 
> ```
> CMake Warning at source/adios2/toolkit/sst/util/CMakeLists.txt:13 (add_executable):
>   Cannot generate a safe dependent library path for target sst_conn_tool
>   because files in some directories may conflict with libraries in implicit
>   directories:
> 
>     runtime library [libgomp.so.1] in /usr/lib/gcc/x86_64-linux-gnu/10 may be hidden by files in:
>       /spack/var/spack/environments/adios2-ci-mpich/.spack-env/view/lib
> 
>   Some of these libraries may not be found correctly.
> ```
> 
> I noticed because I saw the same in the local image I built to specify `openmpi@4.1.8`.

Is this an issue of using newer spack or `openmpi`?

scottwittenburg:
> Is this an issue of using newer spack or `openmpi`?

I'm not sure, I guess it could be either.  It happens on this branch with either version of OpenMPI I tried (`4.1.8` and `5.0.6`). I verified that the previous base E4S image also had `libgomp.so.1`, and in the same location, but we didn't see the warnings there for some reason.

vicentebolea:
@scottwittenburg I see, is it possible to move away from the E4S tags? Perhaps move to 1.0.2? As I understood latest spack has much better buildcache hit ratio since it only considers `specs+glibc` we can still base our image of the e4s-runner images which it is what is used in spack ci jobs.

scottwittenburg:
> is it possible to move away from the E4S tags? Perhaps move to 1.0.2?

@kwryankrattiger Is there a binary mirror we could try with spack version 1.0.2?

kwryankrattiger:
Because Spack is now using the split repos for packages and Spack core, you should be able to get a good hit rate using the `https://binaries.spack.io/v2025.07.0` mirror with any version of Spack >= v1.0.0. In general, the packages drive the most change in re-usability of specs.

The top level mirror only recently got published, but it should be accessible now. If you run into issues or poor hit rate performance I would be interested to know, we are trying to make the public mirrors better.

scottwittenburg:
> I verified that the previous base E4S image also had `libgomp.so.1`, and in the same location, but we didn't see the warnings there for some reason.

So looking closer, the previous image was built with spack before it modeled the compiler runtime.  So while there was a libgomp.so in the same system directory, there was no `gcc-runtime` package, which is what now brings in the conflicting libgomp.so.

I've noticed that we only get the conflict warnings with the images that installed an extra gcc/g++ compiler (9, 10, and 12), but not with the gcc11 image. I'm not yet sure why that is the case.

scottwittenburg:
> Perhaps move to 1.0.2?

DId anyone try to upstream the `0002-adios2-add-xrootd-variant.patch` to spack? I have to update it every time I try a different version of spack, which is quite a pain.  Especially so, now that packages are in a separate repo after 1.0.2.

scottwittenburg:
@kwryankrattiger The cache hit rate, when using `v1.0.2` along with `https://binaries.spack.io/v2025.07.0`, was as good or better than when we were using `e4s-25.06` and the associated mirror.

One thing I noticed though, is that the rocm image size went up another 1.5GB due to a new dependency on gcc@12.5. This is unfortunate since on top of the spack environments image, we build images with more versions of gcc by apt-get installing them. So in the gcc12 image we now have a system gcc12 along with a spack one, which I'm guessing will cause some more of those configure warnings like I mentioned [here](https://github.com/ornladios/ADIOS2/pull/4666#issuecomment-3413256239).

vicentebolea:
@scottwittenburg somehow we are pulling "newer" rocm symbols. For instance `hsa_amd_vmem_handle_create` did no exist yet in 6.1.X:
https://rocm.docs.amd.com/projects/ROCR-Runtime/en/docs-6.1.5/api-reference/api.html

scottwittenburg:
> somehow we are pulling "newer" rocm symbols

I think the only thing that changed for the rocm image, since the last ci run, when it worked, is that I'm now using newer spack.  But I'm investigating.

scottwittenburg:
Poking around inside the rocm container, everything seems in order to me.  The missing symbols are indeed marked `U` in `libamdhip64.so.6`, but it shows that `libhsa-runtime64.so.1` is `NEEDED`, and the symbols are marked as `T` in there.  So I don't understand what the problem is.

<details>
  <summary>Container library details</summary>

```
root@a5c78f45ef17:/opt/rocm-6.1.2/lib# nm -D libamdhip64.so.6 | grep hsa_amd_vmem
                 U hsa_amd_vmem_address_free@ROCR_1
                 U hsa_amd_vmem_address_reserve@ROCR_1
                 U hsa_amd_vmem_get_access@ROCR_1
                 U hsa_amd_vmem_handle_create@ROCR_1
                 U hsa_amd_vmem_map@ROCR_1
                 U hsa_amd_vmem_set_access@ROCR_1
                 U hsa_amd_vmem_unmap@ROCR_1
root@a5c78f45ef17:/opt/rocm-6.1.2/lib# readelf -d libamdhip64.so.6

Dynamic section at offset 0x4c2998 contains 36 entries:
  Tag        Type                         Name/Value
 0x0000000000000001 (NEEDED)             Shared library: [libamd_comgr.so.2]
 0x0000000000000001 (NEEDED)             Shared library: [libhsa-runtime64.so.1]
 0x0000000000000001 (NEEDED)             Shared library: [libnuma.so.1]
 0x0000000000000001 (NEEDED)             Shared library: [libstdc++.so.6]
 0x0000000000000001 (NEEDED)             Shared library: [libm.so.6]
 0x0000000000000001 (NEEDED)             Shared library: [libgcc_s.so.1]
 0x0000000000000001 (NEEDED)             Shared library: [libc.so.6]
 0x0000000000000001 (NEEDED)             Shared library: [ld-linux-x86-64.so.2]
 0x000000000000000e (SONAME)             Library soname: [libamdhip64.so.6]
 0x000000000000001d (RUNPATH)            Library runpath: [$ORIGIN]
 0x000000000000000c (INIT)               0x21000
 0x000000000000000d (FINI)               0x399e30
 0x0000000000000019 (INIT_ARRAY)         0x4bd330
 0x000000000000001b (INIT_ARRAYSZ)       512 (bytes)
 0x000000000000001a (FINI_ARRAY)         0x4bd530
 0x000000000000001c (FINI_ARRAYSZ)       16 (bytes)
 0x000000006ffffef5 (GNU_HASH)           0x298
 0x0000000000000005 (STRTAB)             0x6158
 0x0000000000000006 (SYMTAB)             0x1040
 0x000000000000000a (STRSZ)              22374 (bytes)
 0x000000000000000b (SYMENT)             24 (bytes)
 0x0000000000000003 (PLTGOT)             0x4c4000
 0x0000000000000002 (PLTRELSZ)           8280 (bytes)
 0x0000000000000014 (PLTREL)             RELA
 0x0000000000000017 (JMPREL)             0x1e2f0
 0x0000000000000007 (RELA)               0xc3b0
 0x0000000000000008 (RELASZ)             73536 (bytes)
 0x0000000000000009 (RELAENT)            24 (bytes)
 0x000000006ffffffc (VERDEF)             0xbf80
 0x000000006ffffffd (VERDEFNUM)          12
 0x000000000000001e (FLAGS)              STATIC_TLS
 0x000000006ffffffe (VERNEED)            0xc120
 0x000000006fffffff (VERNEEDNUM)         8
 0x000000006ffffff0 (VERSYM)             0xb8be
 0x000000006ffffff9 (RELACOUNT)          2803
 0x0000000000000000 (NULL)               0x0
root@a5c78f45ef17:/opt/rocm-6.1.2/lib# nm -D libhsa-runtime64.so.1 | grep hsa_amd_vmem
00000000000a3400 T hsa_amd_vmem_address_free@@ROCR_1
00000000000a33e0 T hsa_amd_vmem_address_reserve@@ROCR_1
00000000000a34e0 T hsa_amd_vmem_export_shareable_handle@@ROCR_1
00000000000a34c0 T hsa_amd_vmem_get_access@@ROCR_1
00000000000a3540 T hsa_amd_vmem_get_alloc_properties_from_handle@@ROCR_1
00000000000a3420 T hsa_amd_vmem_handle_create@@ROCR_1
00000000000a3440 T hsa_amd_vmem_handle_release@@ROCR_1
00000000000a3500 T hsa_amd_vmem_import_shareable_handle@@ROCR_1
00000000000a3460 T hsa_amd_vmem_map@@ROCR_1
00000000000a3520 T hsa_amd_vmem_retain_alloc_handle@@ROCR_1
00000000000a34a0 T hsa_amd_vmem_set_access@@ROCR_1
00000000000a3480 T hsa_amd_vmem_unmap@@ROCR_1
root@a5c78f45ef17:/opt/rocm-6.1.2/lib# 
root@a5c78f45ef17:/opt/rocm-6.1.2/lib# 
root@a5c78f45ef17:/opt/rocm-6.1.2/lib# ldd libamdhip64.so.6
	linux-vdso.so.1 (0x00007ffde1feb000)
	libamd_comgr.so.2 => /opt/rocm-6.1.2/lib/./libamd_comgr.so.2 (0x00007860a8ac1000)
	libhsa-runtime64.so.1 => /opt/rocm-6.1.2/lib/./libhsa-runtime64.so.1 (0x00007860a87cb000)
	libnuma.so.1 => /lib/x86_64-linux-gnu/libnuma.so.1 (0x00007860a87b9000)
	libstdc++.so.6 => /lib/x86_64-linux-gnu/libstdc++.so.6 (0x00007860a858d000)
	libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007860a84a6000)
	libgcc_s.so.1 => /lib/x86_64-linux-gnu/libgcc_s.so.1 (0x00007860a8484000)
	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007860a825b000)
	/lib64/ld-linux-x86-64.so.2 (0x00007860b2de2000)
	libz.so.1 => /lib/x86_64-linux-gnu/libz.so.1 (0x00007860a823f000)
	libtinfo.so.6 => /lib/x86_64-linux-gnu/libtinfo.so.6 (0x00007860a820d000)
	libelf.so.1 => /lib/x86_64-linux-gnu/libelf.so.1 (0x00007860a81ef000)
	librocprofiler-register.so.0 => /opt/rocm-6.1.2/lib/./librocprofiler-register.so.0 (0x00007860a816d000)
	libdrm.so.2 => /opt/amdgpu/lib/x86_64-linux-gnu/libdrm.so.2 (0x00007860a8151000)
	libdrm_amdgpu.so.1 => /opt/amdgpu/lib/x86_64-linux-gnu/libdrm_amdgpu.so.1 (0x00007860a8143000)
```

</details>

scottwittenburg:
We also now have some new compiler warnings treated as errors [here](https://open.cdash.org/viewBuildError.php?type=1&buildid=10766489).

<details>
  <summary>Warning messages</summary>

```
../adios2/toolkit/sst/dp/rdma_dp.c: In function 'init_fabric':
[CTest: warning matched] ../[toolkit/sst/dp/rdma_dp.c:403](https://github.com/ornladios/ADIOS2/blob/7ebd3efd662d64b9bde4736fdda9fced95de2e81/toolkit/sst/dp/rdma_dp.c#L403):13: warning: 'FI_LOCAL_MR' is deprecated
  403 |         FI_CONTEXT | FI_LOCAL_MR | FI_CONTEXT2 | FI_MSG_PREFIX | FI_ASYNC_IOV | FI_RX_CQ_DATA;
      |             ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~
[CTest: warning matched] ../[toolkit/sst/dp/rdma_dp.c:567](https://github.com/ornladios/ADIOS2/blob/7ebd3efd662d64b9bde4736fdda9fced95de2e81/toolkit/sst/dp/rdma_dp.c#L567):13: warning: 'FI_LOCAL_MR' is deprecated
  567 |     if (info->mode & FI_LOCAL_MR)
      |             ^~~~~~~~~~~~~~~~~~~~~        
[CTest: warning matched] ../[toolkit/sst/dp/rdma_dp.c:608](https://github.com/ornladios/ADIOS2/blob/7ebd3efd662d64b9bde4736fdda9fced95de2e81/toolkit/sst/dp/rdma_dp.c#L608):13: warning: 'FI_MR_BASIC' is deprecated
  608 |     if (info->domain_attr->mr_mode != FI_MR_BASIC)
      |             ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~
../adios2/toolkit/sst/dp/rdma_dp.c: In function 'RdmaGetPriority':
[CTest: warning matched] ../[toolkit/sst/dp/rdma_dp.c:2510](https://github.com/ornladios/ADIOS2/blob/7ebd3efd662d64b9bde4736fdda9fced95de2e81/toolkit/sst/dp/rdma_dp.c#L2510):13: warning: 'FI_LOCAL_MR' is deprecated
 2510 |         FI_CONTEXT | FI_LOCAL_MR | FI_CONTEXT2 | FI_MSG_PREFIX | FI_ASYNC_IOV | FI_RX_CQ_DATA;
      |             ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~
[CTest: warning matched] ../[toolkit/sst/dp/rdma_dp.c:2540](https://github.com/ornladios/ADIOS2/blob/7ebd3efd662d64b9bde4736fdda9fced95de2e81/toolkit/sst/dp/rdma_dp.c#L2540):13: warning: 'FI_MR_BASIC' is deprecated
 2540 |         hints->domain_attr->mr_mode = FI_MR_BASIC;
```

</details>

Updating to spack `v1.0.2` resulted in us picking up `libfabric@2.2.0` when in the old image we had `libfabric@1.21.0`.  I notice that the mpich project saw similar [issue](https://github.com/pmodels/mpich/issues/7453) with `libfabric` >= 2.0.0, maybe we can do something similiar.  What do you think @eisenhauer @pnorbert ?

vicentebolea:
hmm it seems that `hsa_amd_vmem_address_free` and others where existing in rocm 6.1.X but were just undocumented.

I can see that that in `libamdhip64` the symbols have a single $ and in `libhsa-runtime64` they have two $$. Do you think that this intended?

scottwittenburg:
> I can see that that in `libamdhip64` the symbols have a single $ and in `libhsa-runtime64` they have two $$. Do you think that this intended?

I noticed the [same thing](https://github.com/ornladios/ADIOS2/pull/4666#issuecomment-3412076907) last week with rocm 7. As I mentioned in that comment, I don't know what the `@@` character sequence means in the library where the symbols are defined. Maybe it will be important to figure that out, as otherwise, it would appear to me that everything is fine.

scottwittenburg:
The manpage for `readelf` says one `@` indicates the version of the symbol while `@@` is the default version of the symbol (used when resolving unversioned references). Since in this case it's `ROCR_1` that follows both, it seems to me it should be fine.  But obviously it's not :sweat_smile: 

<details>
  <summary>man readelf excerpt</summary>

```
-s
       --symbols
       --syms
           Displays the entries in symbol table section of the file, if it has one.  If a symbol has version information associated with it then this is displayed as well.  The version string is displayed as a suffix to the symbol name, preceded by an @ character.  For example foo@VER_1.  If the version is the default version to be used when resolving unversioned references to the symbol then it is displayed as a suffix preceded by two @ characters.  For example foo@@VER_2.
```

</details>

scottwittenburg:
And readthedocs build failed because it took too long to create the required conda environment.

eisenhauer:
> We also now have some new compiler warnings treated as errors [here](https://open.cdash.org/viewBuildError.php?type=1&buildid=10766489).
> Warning messages
> 
> Updating to spack `v1.0.2` resulted in us picking up `libfabric@2.2.0` when in the old image we had `libfabric@1.21.0`. I notice that the mpich project saw similar [issue](https://github.com/pmodels/mpich/issues/7453) with `libfabric` >= 2.0.0, maybe we can do something similiar. What do you think @eisenhauer @pnorbert ?

I'm taking a look at this.  Unfortunately that MPICH issue doesn't address the core difficulty here.  To avoid, for example, FI_MR_BASIC we've really got to change the FI_VERSION spec to something newer (as is done specially when we notice that UCX is likely the provider), then make sure we've got a working MR mode for that newer spec.  But this probably isn't a "ask chatgpt the answer" sort of situation, it's a "change it and go check on as many of our special-snowflake HPC machines as possible to see if it works there" sort of thing.  Maybe not something to do as part of this PR?

vicentebolea:
RTD build being fixed here: https://github.com/ornladios/ADIOS2/pull/4681

scottwittenburg:
> Maybe not something to do as part of this PR?

Yeah, possibly.  If there is a way to avoid the warnings as errors in those particular cases, it could make sense to do that with a `TODO` indicating that it needs to be fixed at some point.  Is it possible though?  Can we add a compiler flag for just the offending builds that avoid warning about deprecations?  Or is even that too broad of an exception?

vicentebolea:
> > Maybe not something to do as part of this PR?
> 
> Yeah, possibly. If there is a way to avoid the warnings as errors in those particular cases, it could make sense to do that with a `TODO` indicating that it needs to be fixed at some point. Is it possible though? Can we add a compiler flag for just the offending builds that avoid warning about deprecations? Or is even that too broad of an exception?

A new entry in `CTestCustom.cmake` when RDMA is enabled (and perhaps has a newer version) will do the trick

scottwittenburg:
@vicentebolea It appears that with spack `v1.0.2`, spack installs its own copy of some of the same libraries provided by the base image.  In particular, it installs a `libhsa-runtime64.so.1`, which is missing the symbols mentioned in the link errors.

I've noticed that the `build-rocm.sh` script expects that `spack external find hip rocprim` will work, but only `hip` is found.

If I use spack to satisfy all the dependencies of the rocm-enabled adios spec, the install tree alone is 22GB.  On the other hand, if I try to install as much as possible via system libraries, spack still doesn't seem able to find an external `rocprim`.

vicentebolea:
@scottwittenburg if the

> @vicentebolea It appears that with spack `v1.0.2`, spack installs its own copy of some of the same libraries provided by the base image. In particular, it installs a `libhsa-runtime64.so.1`, which is missing the symbols mentioned in the link errors.
> 
> I've noticed that the `build-rocm.sh` script expects that `spack external find hip rocprim` will work, but only `hip` is found.
> 
> If I use spack to satisfy all the dependencies of the rocm-enabled adios spec, the install tree alone is 22GB. On the other hand, if I try to install as much as possible via system libraries, spack still doesn't seem able to find an external `rocprim`.

I wonder if we can spack external find another package. Perhaps only rocm?

Maybe checking how this is done in places such as frontier as shown here can help `https://github.com/E4S-Project/facility-external-spack-configs/tree/master/frontier/rocm-6.2.4`

scripts/ci/images/Dockerfile.ci-spack-ubuntu22.04-base
```diff
@@ -12,26 +13,25 @@ ARG PATCH_VARIANT_XROOTD
...
 # Install base specs
 RUN . /spack/share/spack/setup-env.sh && \
     if test -n "${EXTERNAL_PACKAGES}"; then spack external find ${EXTERNAL_PACKAGES}; fi && \
-    spack mirror add E4S "https://cache.e4s.io/${E4S_VERSION}" && \
+    spack info adios2 && \
+    if test -n "${PATCH_VARIANT_XROOTD}"; then cd ~/.spack/package_repos/fncqgg4 && git apply /spack/0002-adios2-add-xrootd-variant.patch && cd /spack; fi && \
+    spack mirror add binaries "${SPACK_MIRROR}" && \
     export ADIOS_SPEC_NO_MPI="adios2~mpi${EXTRA_VARIANTS}" && \
-    export ADIOS_SPEC_OPENMPI="adios2+mpi${EXTRA_VARIANTS} ^openmpi" && \
+    export ADIOS_SPEC_OPENMPI="adios2+mpi${EXTRA_VARIANTS} ^openmpi@4.1.8" && \
```
scottwittenburg: use the packages.yaml instead

scripts/ci/images/build-ubuntu.sh
```diff
@@ -2,39 +2,33 @@
 
 set -ex
 
+# export TAG_PREFIX="ghcr.io/ornladios/adios2"
+export TAG_PREFIX="ghcr.io/scottwittenburg/adios2"
+
 # Build the base image
-podman build --progress=plain \
-  --build-arg EXTRA_VARIANTS="+blosc+ssc ^mgard@2023-01-10" \
+docker build --progress=plain \
```
scottwittenburg: use an environment variable for the tool to use, default to podman

.github/workflows/everything.yml
```diff
@@ -136,43 +136,53 @@ jobs:
...
       matrix:
-        os: [ubuntu20.04]
-        compiler: [gcc8, gcc9, gcc10, clang6, clang10]
+        os: [ubuntu22.04]
+        compiler: [gcc9, gcc10, gcc12, clang11, clang14]
         parallel: [mpich]
         include:
-          - os: ubuntu20.04
-            compiler: gcc8
+          - imgprefix: ghcr.io/scottwittenburg/adios2:ci-spack
```
scottwittenburg: try to put this in the matrix and see if we can remove the duplicates below

scottwittenburg: @vicentebolea Putting the imgprefix in the matrix as you suggested doesn't work for any extra jobs brought in by the includes, see [here](https://github.com/ornladios/ADIOS2/actions/runs/18890251590/job/53916013934?pr=4666#step:2:25).

Also using an environment variable doesn't work because those aren't available in the job context.

So I don't see a way to avoid the duplication at the moment, but me know if you have any other suggestions.

scottwittenburg: But when I fix it to refer only to ornladios images, I can remove all the duplicates.  We only need them because we need something different for the rocm build job during this pr.

vicentebolea: I see, no need to waste time on this.

scottwittenburg:
@vicentebolea I expect this will pass, so you can probably start reviewing.  Unless you have change requests, I will do the following once you approve:

1. Re-run `build-ubuntu.sh` to build and push `ornladios` images
2. Remove `scottwittenburg` references (only in `.github/workflows/everything.yml` now)
3. Push the final version and re-request your review

Thanks!

scottwittenburg:
It seems the same windows test failures happened on #4681 last week.

Any idea what could be going on there? But either way, it doesn't seem related to this PR at all.

vicentebolea:
@scottwittenburg

vicentebolea (approved):
approved!

PR #4685, opened 2025-10-23
Configure Dependabot for GitHub Actions and Docker

vicentebolea:
Added configuration for GitHub Actions and Docker updates.


vicentebolea:

PR #4690, opened 2025-10-23
Stopgap measure for deprecated libfabric macros

eisenhauer:
libfabric 2.x has deprecated some macros used in the rdma data plane.  These macros represent values that are still used internally in a few extant libfabric providers, so we don't really want to kill that code.  Instead we borrow a trick from libfabric/include/ofi.h and simply declare our own versions of those macros without the deprecated flag (different names in order to avoid conflicts with those ofi.h declarations).  As we move code to new platforms we may need to modify this approach, but for the moment this preserves existing and past functionality without requiring retesting in situations that may be hard to recreate.

This addresses issues raised in PR #4666.

### DIFF
--- a/.github/workflows/everything.yml
+++ b/.github/workflows/everything.yml
@@ -36,6 +36,9 @@ on:
       - master
       - release*
 
+permissions:
+  contents: read
+
 concurrency:
   group: ${{ github.head_ref || github.run_id }}
   cancel-in-progress: true
@@ -113,6 +116,9 @@ jobs:
     if: needs.git_checks.outputs.num_code_changes > 0
 
     runs-on: ubuntu-22.04
+    permissions:
+      contents: read
+      actions: write  # for cache save
     container:
       image: ghcr.io/ornladios/adios2:ci-spack-${{ matrix.os }}-${{ matrix.compiler }}
       options: --shm-size=1g
@@ -211,6 +217,9 @@ jobs:
     if: needs.git_checks.outputs.num_code_changes > 0
 
     runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      actions: write  # for cache save
     container:
       image: ghcr.io/ornladios/adios2:ci-el8-${{ matrix.compiler }}
       options: --shm-size=1g
@@ -275,6 +284,9 @@ jobs:
     if: needs.git_checks.outputs.num_code_changes > 0
 
     runs-on: ${{ matrix.image }}
+    permissions:
+      contents: read
+      actions: write  # for cache save
     env:
       # Only way to source a file in a non interactive/non login shell.
       BASH_ENV: "/Users/runner/.bash_profile"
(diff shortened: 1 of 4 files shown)


source/adios2/toolkit/sst/dp/rdma_dp.c
```diff
@@ -64,6 +64,18 @@ static pthread_mutex_t fabric_mutex = PTHREAD_MUTEX_INITIALIZER;
 pthread_mutex_t wsr_mutex = PTHREAD_MUTEX_INITIALIZER;
 pthread_mutex_t ts_mutex = PTHREAD_MUTEX_INITIALIZER;
 
+/*
+ * Internal version of deprecated APIs.
+ * These are used internally to avoid compiler warnings.
+ */
```
scottwittenburg, line 70: Once these values are actually removed from libfabric, won't this possibly result in undefined or incorrect behavior?

Is there some convention used in ADIOS2 to mark things we don't want to forget to fix eventually? If so, does it make sense to invoke that here?

eisenhauer: One of the challenges with libfabric is that (IMHO) it's less of a unified piece of software than a collection of "providers" that share an external API, but each kind of picks and chooses what to bits implement beneath it, each has different owners, gets updated separately, etc.  So coherent across-the-board changes are difficult/rare.  Providers are specific to networking hardware, so what gets built on any machine depends upon its configuration and we just have to react to what we find at run-time (no compile-time query possible generally).  The macros in question just set bits in a bitmask that is passed in to specify certain characteristics (I.E. specify what behavior we want/need).  I was hoping to look at the code for the providers that we support to see that they were using instead of the "deprecated" bits, or how they managed backwards compatibility, but instead I see that at least one is still directly looking at those bits, so there's not really a substitute.  If at some point those providers get changed to do something else, then presumably we'll find out when our dataplane doesn't work.  (I don't know how else we'd know unless we're monitoring PRs to libfabric and happen to catch it.)  But really, I don't expect those providers to change.  Instead I expect that our code will evolve as it has for the libfabric cxi provider, where we have to write new/different code to support new/different provider behavior and somehow finagle it so that we still behave the same for older providers.  If libfabric was the ideal generic RDMA interface of our dreams, that wouldn't be necessary, but reality falls a bit short.

Anyhow, I can add a comment that those macros are changed to avoid the deprecated version, and that that presumably means that at some future date there will be something that we should be doing instead, but I'm not sure what to say beyond that.  (And really, I expect we're at stage one of a progression.  First those macros are marked deprecated but still used internally.  At some point their use will be removed and the bits they specify in the bitmask will be ignored.  At some later time _maybe_ they'll reuse those bits.  That's when our use might break things, but they have compatibility worries too, so I wouldn't expect that soon, maybe not ever.)

scottwittenburg: Thanks for the background information, that's quite interesting to learn. And I agree on that progression you described, since they have their own compatibility worries, they may never reuse those bits.

scottwittenburg (approved):

PR #4698, opened 2025-10-28
H5VolReadWrite: Do not use variable after memory is freed

scottwittenburg:
Fix issue found by compiler warning newly emitted in #4666

### DIFF
--- a/source/h5vol/H5VolReadWrite.c
+++ b/source/h5vol/H5VolReadWrite.c
@@ -440,13 +440,9 @@ bool gRemoveUnderGrp(H5VL_ObjDef_t *owner, const char *obj_name)
             return true;
         }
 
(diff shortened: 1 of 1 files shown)


guj:
Sorry I just saw the emails. Thanks Scott.  You can remove the  NEVER macros: e.g. 
if ..
if ..

printf("\n......... NOTE: unable to remove GROUP %s \n\n", fullPath);
free(fullPath);
return false; // because nothing was removed.

scottwittenburg:
@guj Thanks for the suggestion, I've updated it as you suggested.  CI is all green now, if you are willing to approve, I'll merge and rebase my ci pr.

guj:
@vicentebolea  can approve. I do not have permission

PR #4699, opened 2025-10-28
Raise an exception when an XRootD get fails.

eisenhauer:
This isn't an ideal implementation.  XRootD is more like httpd than it is our EVPath-based remote server.  We don't actually know on Open() if the file exists on the XRootD server because we don't communicate with the server on Open.  Instead, we find out when the first Get() fails, but actually since we do those in parallel there are multiple Get()s in-flight before we actually get one back that has failed, so there's a bit of noise in stderr because of this.  But this is better than not raising an exception at all, which is what we have been doing.


eisenhauer:
Looks like this CI ran clean except for expected failures (LargeBlock in msan and asan) and on macos14 the test results showed zero failures, but the results themselves failed to upload to cdash.  Should be ready to merge.

anagainaru (approved):

Issue #4673, opened 2025-10-20
Restrict workflow token permissions


vicentebolea:
GitHub Actions workflows are using tokens with excessive permissions. Should explicitly set minimal permissions using the `permissions:` key in each workflow.

Most workflows should be read-only unless they specifically need write access.

Reference: https://github.com/ossf/scorecard/blob/c22063e786c11f9dd714d777a687ff7c4599b600/docs/checks.md#token-permissions

vicentebolea:
Scorecard detailed results for Token-Permissions (0/10):

```
Debug: no jobLevel permission defined: .github/workflows/everything.yml:49
Debug: no jobLevel permission defined: .github/workflows/everything.yml:74
Debug: no jobLevel permission defined: .github/workflows/everything.yml:209
Debug: no jobLevel permission defined: .github/workflows/everything.yml:273
Debug: no jobLevel permission defined: .github/workflows/everything.yml:412
Debug: no jobLevel permission defined: .github/workflows/everything.yml:477
Debug: no jobLevel permission defined: .github/workflows/everything.yml:111
Debug: no jobLevel permission defined: .github/workflows/everything.yml:349
Info: jobLevel 'actions' permission set to 'read': .github/workflows/everything.yml:546
Info: jobLevel 'contents' permission set to 'read': .github/workflows/everything.yml:547
Debug: no jobLevel permission defined: .github/workflows/everything.yml:621
Debug: no jobLevel permission defined: .github/workflows/external.yml:9
Debug: no jobLevel permission defined: .github/workflows/pypackaging.yml:32
Debug: no jobLevel permission defined: .github/workflows/pypackaging.yml:51
Debug: no jobLevel permission defined: .github/workflows/stale.yml:8
Warn: no topLevel permission defined: .github/workflows/everything.yml:1
Warn: topLevel permissions set to 'write-all': .github/workflows/external.yml:6
```

Multiple jobs missing explicit permission definitions. The external.yml workflow has write-all permissions which is overly permissive. Most workflows should explicitly set read-only permissions and only grant write where necessary.

PR #4702, opened 2025-10-29
ci: restrict file scope perms in workflows

vicentebolea:
fixes: #4673

### DIFF
--- a/.github/workflows/external.yml
+++ b/.github/workflows/external.yml
@@ -5,7 +5,6 @@ on:
 
 permissions:
   contents: read
-  statuses: write
 
 jobs:
   generate_statuses:
--- a/.github/workflows/pypackaging.yml
+++ b/.github/workflows/pypackaging.yml
@@ -25,6 +25,9 @@ on:
     types:
       - published
 
+permissions:
+  contents: read
+
 env:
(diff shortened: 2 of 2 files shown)


.github/workflows/external.yml
```diff
@@ -5,7 +5,6 @@ on:
 
 permissions:
   contents: read
-  statuses: write
```
scottwittenburg, line 8: Out of curiosity, if there is only one job here which has explicit permissions it needs, what is the reason for having any permissions at the top level?

.github/workflows/pypackaging.yml
```diff
@@ -25,6 +25,9 @@ on:
     types:
       - published
 
+permissions:
+  contents: read
```
scottwittenburg, line 29: If this workflow was working before, why are new permissions needed at the top level here?

.github/workflows/external.yml
vicentebolea: This is to comply with the OpenSSF best practices that dictates top level read only permissions.

.github/workflows/pypackaging.yml
vicentebolea: Same here

source: https://github.com/ossf/scorecard/blob/main/docs/checks.md#token-permissions

Issue #4565, opened 2025-06-30
libraries names


guj:
cmake says cxx 14 is in use: 
---- 
ADIOS2 build configuration:
  ADIOS Version : 2.10.0.690
  C++ standard  : 14
----

and libraries are still named lib*cxx11,
e.g. libadios2_cxx11_mpi.so  libadios2_cxx11.so  

should they match?

eisenhauer:
That's a very good question...  We're requiring C++14 now simply because we had to upgrade gtest (because it's CMake was too old) and *it* requires C++14.  But we have a lot of stuff with "11" in the name, including our C++ and Python bindings.  The latter maybe still makes sense because pybind11 is still "11".   Generally we don't change the name of our libraries with the compiler (I.E. you can use c++17 to build ADIOS and you don't get a library with 17 in the name).  But why was 11 in the name to begin with?  I can't think of a good reason, but maybe there was one.  If not, I'd think we should drop the 11 rather than changing it to 14.

PR #4707, opened 2025-11-04
bindings: renamed cxx11 bindings to cxx

vicentebolea:
fixes: #4565

- Renamed bindings/CXX11 to bindings/CXX and updated all references including CMake targets (adios2_cxx11 → adios2_cxx).
- Added backward compatibility wrappers in bindings/CXX/adios2/cxx11/ that include their corresponding files from bindings/CXX/adios2/cxx/ with deprecation warnings to ease transition forexisting code.
- Added backward compatibility regression tests for both the headers files and the cmake targets.


scottwittenburg:
Is the goal to merge this before the release? If so, it seems like a sudden breaking change for consuming projects. Would it be possible to somehow provide duplicate library targets (and headers too I guess) under both names, so that the `cxx11` usage could just be deprecated initially?

vicentebolea:
> Is the goal to merge this before the release? If so, it seems like a sudden breaking change for consuming projects. Would it be possible to somehow provide duplicate library targets (and headers too I guess) under both names, so that the `cxx11` usage could just be deprecated initially?

Great point, find the duplicated targets added to the adiosconfig.h anlso the duplicated headers pointing to the new ones.

testing/install/CXX/CMakeLists.txt
```diff
@@ -4,12 +4,17 @@
 #------------------------------------------------------------------------------#
```
vicentebolea, line 5: In this file I added the regression test for the deprecated adios2::cxx11* targets

scottwittenburg: That's a good idea :+1: 

Does it also make sense to add a test including one or more of the deprecated headers? Not sure, just asking :grinning:

testing/install/CXX/main_nompi.cxx
```diff
@@ -5,6 +5,13 @@
 
 #include <adios2.h>
 
+// Test for CXX11 deprecated headers
```
vicentebolea, line 7: @scottwittenburg test for headers is here

bindings/CXX/adios2/cxx11/ADIOS.h
```diff
@@ -0,0 +1,20 @@
...
+ *
+ * DEPRECATION WARNING: The bindings/CXX/adios2/cxx11 directory has been
+ * renamed to bindings/CXX/adios2/cxx. Please update your includes to use
+ * bindings/CXX/adios2/cxx instead of bindings/CXX/adios2/cxx11.
+ */
+
+#ifndef ADIOS2_BINDINGS_CXX_CXX11_ADIOS_H_
+#define ADIOS2_BINDINGS_CXX_CXX11_ADIOS_H_
+
+#warning                                                                                           \
```
scottwittenburg: The `#warning` preprocessor directive seems not to be supported on Windows.

vicentebolea: addressed in last push, I now generate these headers file at conf time

scottwittenburg (approved):
Hard to be 100% sure with a change this size, but it all seems reasonable to me. Thanks @vicentebolea!

vicentebolea:
@eisenhauer this is a big change, can you also have a look before we merge?

eisenhauer (approved):
Huge changes of course, but it looks good.  I did try it with unmodified ADIOS2-Examples and we get appropriate deprecation messages for the old targets but everything builds cleanly.  Somehow on my mac I'm getting a failure on ADIOS2 install saying that mpicc can't create executables.  But I'm reasonably convinced that that has to be something in my setup, so I'm ignoring for the moment.  @pnorbert we should update other things we have, like ADIOS2-Examples, to the new cmake targets as well.

PR #4709, opened 2025-11-05
Fix: add missing cstdint include in adiosNetwork.h

Lgt2x:
Caused the build to fail on recent gcc. Likely introduced by 70b09272d0be099617afb5edb8837eadc5d3f137 that changed include order in adiosNetwork.cpp


Lgt2x:
cc @vicentebolea

Lgt2x:
Thanks for merging! It's not PVSB that failed, but my local ADIOS2 build (clang++ 21) that I update from time to time, to make sure [AdiosCatalyst](https://gitlab.kitware.com/paraview/adioscatalyst) still works with latest ADIOS2.

PR #4710, opened 2025-11-05
Update common tracking

mathstuf:
---
Cc: @vicentebolea


mathstuf:
Note that I found patches in other projects importing the script; I'd like to sync with those first before merging.

PR #4714, opened 2025-11-07
use the same base handle as in extract.sh

guj:
in case keywords "default" "flatten" "joined" are in the path name.


pnorbert (approved):

guj:
@vicentebolea can this be merged? thanks

PR #4717, opened 2025-11-13
Fix logic error in BP5 selective metadata aggregation

eisenhauer:


pnorbert (approved):
I approve this even though I don't understand why the previous summation could become non-zero with a zero array.

eisenhauer:
> I approve this even though I don't understand why the previous summation could become non-zero with a zero array.

It was the opposite problem, we had a non-zero array that added up to zero (with overflow).

PR #4722, opened 2025-11-14
Do not install example binaries in adios2 deployments

vicentebolea:


vicentebolea:
@eisenhauer the linking error without any other hint is really hard to debug.

eisenhauer (approved):
Ugh, yes.  This looks good though.  Except still failing some cases...

vicentebolea:
The issue was due to this: `##[warning]You are running out of disk space. The runner will stop working when the machine runs out of disk space. Free space left: 28 MB`

We need to trim the static builds

vicentebolea:
I have trimmed the static builds so that they remain withing github actions space limits. I have also made the examples only install if they are outside of an adios2 project.

vicentebolea:
@eisenhauer please re-review this

eisenhauer (approved):
Looks good.

PR #4729, opened 2025-11-17
cmake: define cxx11 targets once

vicentebolea:


vicentebolea:
The win vms seems to have an issue, it might be transient.

PR #4734, opened 2025-11-19
Add support to read a BP5/TEXT/IMAGE from TAR files…

pnorbert:
… if provided with a detailed index "name,offset,size;..." for all metadata and data files. Used for supporting TAR files added to Campaigns with a tar index. For now it supports reading from a local TAR file and from a TAR file on a HTTPS server.

### DIFF
--- a/source/adios2/engine/bp5/BP5Reader.cpp
+++ b/source/adios2/engine/bp5/BP5Reader.cpp
@@ -403,6 +403,35 @@ void BP5Reader::EndStep()
     MinBlocksInfoMap.clear();
 }
 
+std::string BP5Reader::UpdateWithTarInfo(const std::string &path, Params &params)
+{
+    if (m_dataIsRemote)
+        return path;
+    auto fileName = adios2sys::SystemTools::GetFilenameName(path);
+    std::string retval = path;
+    if (m_Parameters.verbose > 1)
+    {
+        std::cout << "--- UpdateWithTarInfo path = " << path << ", fileName = " << fileName
+                  << "  map size = " << m_TarInfoMap.size() << std::endl;
+    }
+    auto it = m_TarInfoMap.find(fileName);
+    if (it != m_TarInfoMap.end())
+    {
+        std::string offset = std::to_string(std::get<0>(it->second));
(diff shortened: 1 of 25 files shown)


pnorbert:
Testing is seriously falling behind, as I can test the new functions I have been adding by generating data, putting them on a https server, tarring them up, adding them to a campaign file, etc. Maybe we need to fire up the external testing again and build a complicated environment?

source/adios2/engine/campaign/CampaignData.cpp
```diff
@@ -86,11 +86,11 @@ static int sqlcb_host(void *p, int argc, char **argv, char **azColName)
     }
     std::cout << std::endl;
     */
-    ch.hostname = std::string(argv[0]);
+    ch.hostname = std::string(argv[1]);
```
anagainaru, line 89: These are bug fixes that are unrelated to the TAR reader right?

anagainaru (approved):
This looks good but without a unit test I might have missed something. I guess we don't have any unit test for campaign reading in general

source/adios2/engine/campaign/CampaignData.cpp
pnorbert: Yeah, that comes from turning on ORDER BY in all selects to ensure correct order of rowids.

source/adios2/engine/bp5/BP5Reader.cpp
```diff
@@ -540,7 +570,15 @@ void BP5Reader::PerformGets()
             int localPort = pair.second;
             if (m_Remote && localPort > -1)
             {
-                m_Remote->Open("localhost", localPort, RemoteName, m_OpenMode, RowMajorOrdering);
+                Params p;
```
pnorbert, line 569: @eisenhauer I added Parameters to the Open for EVPathRemote. Will need similar thing for xrootd.

eisenhauer: Seems reasonable.

pnorbert:
Closing this PR since I forgot to implement TEXT/IMAGE read from https.

PR #4737, opened 2025-11-21
just testing what goes wrong with windows CI

pnorbert:


pnorbert:
@vicentebolea I don't think I am the culprit for Windows CI failing. Something is wrong with the Python installation on the Windows images.

vicentebolea:
This was already fixed

PR #4740, opened 2025-11-23
Use Transports directly

eisenhauer:
This PR takes first steps towards eliminating TransportMan in BP5.  This tackles the easy things (metadata files) first to clear the way for future work on the data files.  The one test change was where TestBPBufferSize relied on a weirdness where if you used a bad transport name it just silently failed to use a transport. (Now it actually writes data but still works the same.). The change to every transport was to align the default value for the start parameter to be the same as it is if you go through the TransportManager.  TransportMan changes are just to make OpenFileTransport() public and to kill some commented-out code.

### DIFF
--- a/source/adios2/engine/bp5/BP5Writer.cpp
+++ b/source/adios2/engine/bp5/BP5Writer.cpp
@@ -1015,9 +1015,12 @@ void BP5Writer::EndStep()
             m_AsyncWriteLock.unlock();
         }
     }
-    m_MetadataIndexFile->Flush();
-    m_MetadataFile->Flush();
-    m_MetaMetadataFile->Flush();
+    if (m_Comm.Rank() == 0)
+    {
+        m_MetadataIndexFile->Flush();
+        m_MetadataFile->Flush();
+        m_MetaMetadataFile->Flush();
+    }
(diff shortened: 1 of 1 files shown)


eisenhauer:
> This looks good to me. I approve before looking at the failed tests.

Now I think it's ready to merge.  But I think it can be only to master because if we merged to release_211 we'd break internal ABI compatibility.  @vicentebolea, do you agree?

vicentebolea:
> > This looks good to me. I approve before looking at the failed tests.
> 
> Now I think it's ready to merge. But I think it can be only to master because if we merged to release_211 we'd break internal ABI compatibility. @vicentebolea, do you agree?

As it is right now yes, we would have to retro-fit this to avoid braking ABI comptability

eisenhauer:
> > > This looks good to me. I approve before looking at the failed tests.
> > 
> > 
> > Now I think it's ready to merge. But I think it can be only to master because if we merged to release_211 we'd break internal ABI compatibility. @vicentebolea, do you agree?
> 
> As it is right now yes, we would have to retro-fit this to avoid braking ABI comptability

Well, the point of this is really to rearrange internals so that we can more easily change functionality in the future.  No actual functionality changes here, so nothing to really retro-fit.

eisenhauer:
CI was right, but better now...

PR #4742, opened 2025-11-24
BP5Writer: Subfile config for DSB works like other aggregators

scottwittenburg:
Reduce user surprise level when configuring the number of subfiles, by making DataSizeBased aggregation behave similarly to EveryoneWrites and EveryoneWritesSerial. If NumAggregators is specified, use it. If NumAggregators is zero, use NumSubFiles. If both are zero, compute the number of nodes and use that.


scottwittenburg:
Hmmm, CDash is blank for the current run.  Any idea @vicentebolea?

vicentebolea:
> Hmmm, CDash is blank for the current run. Any idea @vicentebolea?

slow cdash, happens at times

eisenhauer (approved):

PR #4746, opened 2025-11-28
Documentation fixes for beginner users

hahnjo:
* Fix typo in "Performance profiling" header
 * Fix description of Hello World tutorial
 * `IO::Open` with MPI is always collective


docs/user_guide/source/tutorials/helloWorld.rst
```diff
@@ -95,7 +94,7 @@ Start editing the skeleton file `ADIOS2/examples/hello/helloWorld/hello-world_tu
 .. note::
 
   In Mode::Read, the ``BeginStep`` and ``EndStep`` calls are required when **reading** one step and multiple steps. We will see in
```
eisenhauer, line 96: Some rationale behind the "should" here.  ADIOS relatively recently separated Mode::Read from Mode::ReadRandomAccess in order to solve a problem and add more consistency.  Originally, BP3 and BP4 always read _all_ available file metadata on Open() regardless of whether or not you expected to use Begin/EndStep() or select the timestep to which each Get applied (SetStepSelection).  This had the advantage that InquireVariable() was possible before or after BeginStep().  However, as ADIOS applications grew to output with many ranks and timesteps, this meant that Open() was very expensive, and unnecessarily so if you were going to only process one timestep at a time.  Also, streaming engines like SST only supported the Begin/EndStep get interface and didn't know about any variables until BeginStep succeeded.  So ADIOS 2.9(?) introduced the open mode ReadRandomAccess as separate from Mode Read and BP5 started forcing apps to choose up front how they were going to access variables.  BP3/4 were not so updated.  BP3 and 4 are little used these days, so I'd say it's been long enough that we can change this to "must".

eisenhauer:
Bypassing rules as required for docs-only PRs.

PR #4757, opened 2025-12-02
Use specified engine rather than BPFile in tests

eisenhauer:


eisenhauer:
Static build is still failing, apparently for it's own reasons.   This PR was born out of an accidental observation that several tests labeled BP3 or BP4 were actually running the default (BP5) engine because of erroneous code.  In fixing that, I discovered that some tests actually don't work with BP3/4 because of some differing semantics.  For example, if a serial BP4 writer code does BeginStep/EndStep without an intervening Put(), ADIOS doesn't actually record a step, whereas it does in BP5.  Rather than rewrite those tests, this PR simply skips them for BP3/4.  (This decision is arguable.). @pnorbert

pnorbert (approved):

PR #4759, opened 2025-12-04
Updated profiler tags used  in json

guj:
Cleaned up the profiling entries . Hope any future entries follow the X_Y  pattern with  _ indicates Y is a subcomponent of X. 
Added the profiler_simplied/  to use the updated profilers to do quick plots.


guj:
> Lots of changes and diff is a bit confused, so hard to be sure but I think this looks good (except for the format failing of course).

because was resolving conflicts .. 
do you know why  the static serial build on ubuntu failed?  the log displayed is partial and did not show the failed parts.

eisenhauer:
> do you know why the static serial build on ubuntu failed? the log displayed is partial and did not show the failed parts.

Not sure, but vicente was working with the static builds, so it may be related to that.  Not worried about it.

PR #4768, opened 2025-12-19
Order remote replicas in order of most likely accessible: …

pnorbert:
…known hosts in hosts.yaml and not archive locations, s3 hosts with known setup, https locations, then "fs" archives, then finally unknown hosts (just to cause some useful error message later).


pnorbert:
This PR is needed for the UEDGE campaigns, to prefer NERSC access for users who has that set up in their config, and https for everyone else.

eisenhauer (approved):

PR #4770, opened 2025-12-26
Integrate ProDM changes, added new operator RefactorProDM

AkiraFukuda:
This is an integration between ADIOS2 and ProDM (Dr. Xin Liang's software on data decomposition). A new operator called RefactorProDM is added, and will be installed when ProDM is available.

PR #4771, opened 2025-12-26
Integrate ProDM changes, added new operator RefactorProDM

AkiraFukuda:
This is an integration between ADIOS2 and ProDM (Dr. Xin Liang's software on data decomposition). A new operator called RefactorProDM is added, and will be installed when ProDM is available. (Supersedes #4770)

### DIFF
--- a/source/adios2/operator/refactor/RefactorProDM.cpp
+++ b/source/adios2/operator/refactor/RefactorProDM.cpp
@@ -20,8 +20,8 @@
 #include <cstring>
 #include <iostream>
 
-#include "MDR/Refactor/Refactor.hpp"
 #include "MDR/Reconstructor/Reconstructor.hpp"
+#include "MDR/Refactor/Refactor.hpp"
 
 namespace adios2
 {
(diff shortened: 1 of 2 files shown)


eisenhauer:
Replying to @vicentebolea's comment originally on #4770...  I think it's reasonable to question whether or not this should be a built-in operator or should use the plug-in interface (I.E. like the example in ADIOS2/plugins/operators).  Maybe @pnorbert can opine?  ADIOS has a lot of connections to the ProDM team, and having something built-in would be better if this is expected to have some longevity, but it does add an implied onus of maintenance on the ADIOS side  Before taking that up, I'd be nice to see that the software we're building against itself has maintenance, releases, documentation, etc.  I'd also like to see an entry in ADIOS docs describing the operator in a more-than-cursory way.  I'm curious myself how this might relate to MGard, SZ3, etc., and what I see in a quick glance online hasn't enlightened me so far.  That we have the plug-in interface means that the ProDM team could still do interesting things without having to have this build-in, so I don't think that's a bottleneck, whatever is decided.

PR #4778, opened 2026-01-07
Add automatic cleanup of generated test files in BP engine tests

eisenhauer:
Implement cleanup functionality for all engine/bp tests to automatically remove generated .bp files and directories after test execution. This prevents test artifacts from accumulating in the build directory.

### DIFF
--- a/testing/adios2/engine/bp/TestBPWriteReadADIOS2fstream.cpp
+++ b/testing/adios2/engine/bp/TestBPWriteReadADIOS2fstream.cpp
@@ -1632,17 +1632,8 @@ TEST_F(BPWriteReadTestADIOS2fstream, OpenEngineTwice)
         EXPECT_THROW(io.Open(fname, adios2::Mode::ReadRandomAccess), std::invalid_argument);
     }
 
-    // Cleanup generated files - note: no mpiRank in this test
-#if ADIOS2_USE_MPI
-    int mpiRank = 0;
-    MPI_Comm_rank(MPI_COMM_WORLD, &mpiRank);
-#else
-    int mpiRank = 0;
-#endif
-    if (mpiRank == 0)
-    {
-        CleanupTestFiles(fname);
-    }
+    // Cleanup generated files
+    CleanupTestFiles(fname);
 }
 
 int main(int argc, char **argv)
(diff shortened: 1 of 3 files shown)


eisenhauer:
Just cleaning up after tests under testing/adios/engine/bp.  It's a start...

eisenhauer:
OK, I think I'm done with this now...

vicentebolea:
Oh it did work, nice!

testing/adios2/engine/TestHelpers.h
```diff
@@ -0,0 +1,51 @@
...
+#define TEST_ADIOS2_ENGINE_TESTHELPERS_H_
+
+#include <algorithm>
+#include <cstdlib>
+#include <string>
+
+// Helper function to cleanup test files/directories
+// If engine is specified, only cleanup for file engines (not streaming engines)
+// If engine is empty/null, always cleanup (default behavior for non-staging tests)
+inline void CleanupTestFiles(const std::string &path, const std::string &engine = "")
```
vicentebolea, line 18: engine param is not used, we might want to add a safety mechanism not avoid rm anything outside of build_bir/binary_dir

eisenhauer: It did work, but it took a lot of cycles...  

Some kind of safety mechanism isn't a bad idea, but I'm not sure what form it would best take.  The usual path parameter when used in these tests is a simple filename, so interpreted relative to PWD.  Some tests take fname as a parameter, so it could potentially be most anything, which is a bit worrisome.  If you gave a filename as ~/, our test might not succeed, but we might remove the homedir anyway.   Possible approaches:  Refuse any fname which starts with '/' and don't do anything if getcwd() doesn't contain build_dir/binary_dir?   Then again, I think all the tests which accept filename as an argument are in staging-common, where we use the same executable in many different roles.  I'm not that fond of my solution there anyway.  TestCommonWrite creates the file, but then TestCommonRead always removes them.  That actually conflicts with some debugging I commonly do where I write data and then read repeatedly trying to fix problems.  How about if we change those particular tests (the only ones which accept fname as a parameter) so that they only remove the output file if you give them a "-cleanup" flag, and we supply that flag in run_test.py so that cleanup happens in the ctest context, but not kind of willy-nilly.  Does that sound OK without other guard rails?

PR #4780, opened 2026-01-07
Consolidate test utilities and improve cleanup handling

eisenhauer:
@vicentebolea In addition to other random cleanup (move SmallTestData.h contents into TestHelpers.h and eliminate that now-extra file), this changes the tests that don't use relatively-fixed internal fnames for their output so that those tests only cleanup files if you specify a --cleanup flag.  That flag is specified in run_test.py, so files should always be cleaned up in CI (where we are in control of arguments).  This is at least a step closer to safety.  What do you think?


vicentebolea:
@eisenhauer this has added private keys for xrootd, is this intended?

eisenhauer:
It was not!  Thanks.  Stuff I was trying out that leaked in here..  Not sure how it got added.

vicentebolea:
looks good! It would be nice if we move to c++17 and we can use std::filesystem and consolidate all of fs interactions

eisenhauer:
> looks good! It would be nice if we move to c++17 and we can use std::filesystem and consolidate all of fs interactions

Yeah.  I'm not opposed to requiring a newer C++ than 14, and the std::filesystem stuff is a pretty good incentive.  Something we should chat about...

vicentebolea:
> > looks good! It would be nice if we move to c++17 and we can use std::filesystem and consolidate all of fs interactions
> 
> Yeah. I'm not opposed to requiring a newer C++ than 14, and the std::filesystem stuff is a pretty good incentive. Something we should chat about...

It would be nice, also the parallel algorithms, better constexpr (we can refactor our poor man SFINAE for Kokkos/adios views), and many other things. It might also be a good chance to move to C11 as well.

PR #4783, opened 2026-01-08
refactor code to utilize c++14 features

vicentebolea:
- Use std::make_unique whenever it can
- Use the auto keyword whenever possible (specially lambda params)
- Use of [[deprecated]]

### DIFF
--- a/scripts/developer/run-clang-format.sh
+++ b/scripts/developer/run-clang-format.sh
@@ -1,9 +1,14 @@
 #!/bin/bash
 
-sudo docker run -itt --mount type=bind,source="$(pwd)",target=/root/adios2 \
+CONTAINER_DRIVER=${CONTAINER_DRIVER:-docker}
+
+"${CONTAINER_DRIVER[@]}" run -itt --mount type=bind,source="$(pwd)",target=/root/adios2 \
   ghcr.io/ornladios/adios2:ci-formatting sh -c \
   "git config --global --add safe.directory /root/adios2 &&
    cd /root/adios2 &&
    ./scripts/ci/scripts/run-clang-format.sh"
 
-git status --porcelain | awk '{print $2}' | xargs sudo chown "$USER:$(id -g)"
+if [ "${CONTAINER_DRIVER[0]}" != "podman" ]
+then
+  git status --porcelain | awk '{print $2}' | xargs sudo chown "$USER:$(id -g)"
+fi


vicentebolea:
low hanging fruits for claude-code :)

source/adios2/common/ADIOSTypes.h
```diff
@@ -245,7 +245,7 @@ const std::string DefaultTimeUnit("Microseconds");
 constexpr TimeUnit DefaultTimeUnitEnum(TimeUnit::Microseconds);
 
 /** default initial bp buffer size, 16Kb, in bytes */
-constexpr size_t DefaultInitialBufferSize = 16 * 1024;
```
eisenhauer, line 248: I tend to think that this change does not add clarity or other apparent value.  In the original form it was obvious that the value matched the comment.  Now it is less so, but the only gain is to save a few compiler cycles.

source/adios2/common/ADIOSTypes.h
```diff
@@ -257,15 +257,15 @@ constexpr float DefaultBufferGrowthFactor = 1.05f;
 
 /** default Buffer Chunk Size
  *  128Mb */
-constexpr uint64_t DefaultBufferChunkSize = 128 * 1024 * 1024;
+constexpr uint64_t DefaultBufferChunkSize = 134'217'728;
```
eisenhauer: Same for all of these, except it's more problematic for bigger numbers. (And gack on the use of ' as a comma.  (I understand that the comma convention varies around the world and it couldn't be used lexically for this in C++ anyway, but gack...)

source/adios2/core/Group.cpp
```diff
@@ -39,7 +39,7 @@ Group::Group(std::string path, char delimiter, IO &io)
 {
     if (mapPtr == nullptr)
     {
-        mapPtr = std::shared_ptr<TreeMap>(new TreeMap());
+        mapPtr = std::make_shared<TreeMap>();
```
eisenhauer, line 42: All of these make_* changes, I heartily approve of.

source/adios2/helper/adiosMemory.cpp
```diff
@@ -276,37 +276,37 @@ int NdCopy(const char *in, const CoreDims &inStart, const CoreDims &inCount,
     size_t minContDim, blockSize;
     const char *inOvlpBase = nullptr;
     char *outOvlpBase = nullptr;
-    auto GetInEnd = [](CoreDims &inEnd, const CoreDims &inStart, const CoreDims &inCount) {
+    auto GetInEnd = [](auto &inEnd, const auto &inStart, const auto &inCount) {
```
eisenhauer: I'm also less sold on always using auto wherever possible.  Sometimes it helps with clarity.  Now maybe this particular one I'm more sensitive about because the separation between Dims and CoreDims was part of a hard-won fight to improve code efficiency.

eisenhauer (commented):
Just some comments for discussion...

vicentebolea:
@eisenhauer all comment addressed in the last push

PR #4801, opened 2026-01-13
Add adios2_get_string() to C API for safe string variable retrieval

eisenhauer:
This function supports a two-call pattern to retrieve string variables of unknown length: first call with NULL buffer to get length, then allocate and call again to get data. Strings are always retrieved synchronously since they are stored in metadata.


eisenhauer:
> Why not just add `adios2_get_string_len()` and then use the current API?

We have several functions that extract strings in the C api, and this is how they work (call the function once with NULL to get the length if you don't know it somehow statically, call again to get the actual value after allocation), so doing it this way for consistency...  But yes, you could just do the first call using this API and then do the 2nd with the adios2_get() api.  Either way we're adding a function to the C bindings.  I don't have strong feelings, but the draw of consistency pulled me this way.

PR #4812, opened 2026-01-20
GTKorvo Upstream

eisenhauer:


vicentebolea:
oh there were some CI failures

eisenhauer:
Ah, saw it go green and merged.  Let me look at the fails and see if they might be related.

eisenhauer:
One failure in YAML parsing.  Kind of has to be spurious.  The common LargeBlock fails.  And an odd BP4 failure (which can't really be related to Korvo since BP4 doesn't use any of that software stack).  So I think this didn't cause any of those.  However, I do get more and more odd merge conflicts with every invocation of update.sh.  Not at all sure why these occur, but I edit the files manually to remove the conflict text and then commit the result.  Not a CI failure problem, just another weirdness with update.sh.

vicentebolea:
I see, on the jobs sides It is very odd, it seems that the jobs are duplicated. I look into that.

PR #4815, opened 2026-01-22
Add HTTPS transport for XRootD remote access

eisenhauer:
This adds an alternative HTTPS-based transport for accessing ADIOS data through XRootD SSI services.

Components added:
- XrootdHttpsRemote: Client-side HTTPS transport using libcurl
- XrdHttpSsiHandler: Server-side HTTP-to-SSI bridge for XRootD
- CURL detection and configuration in CMake
- Test infrastructure for HTTPS-based remote tests

The implementation uses URL encoding for filenames and variable names to properly handle paths with special characters (spaces, &, =, etc.).

To use, set environment variables:
- DoXRootDHttps=1 to enable HTTPS transport
- XRootDHttpsHost=hostname:port to specify server (default: localhost:8443)
- XRootDHttpsNoVerify=1 to skip SSL verification (for self-signed certs)


eisenhauer:
Not sure why we've developed a readthedocs failure.  The msan and tsan errors are unrelated, so this is ready for review.

pnorbert:
> Not sure why we've developed a readthedocs failure. The msan and tsan errors are unrelated, so this is ready for review.

I did not know we build the code first before building the docs in that image. It fails linking for remote server. Maybe something with new curl dependency?

eisenhauer:
I'm going to bypass requirements because of the readthedocs error.  If that doesn't self-correct will debug on next PR.

PR #4817, opened 2026-01-23
Allow Dill to be build or not depending upon the results of that configuration

eisenhauer:
@ax3l


eisenhauer:
@ax3l Can you confirm that this set of changes is sufficient for you to build release_211?  (These are divorced from a general GTkorvo upgrade in release 2.11.  We've made too many changes to pull in everything.)

PR #4818, opened 2026-01-26
adios2.Stream.minmax()

pnorbert:
Add adios2.Stream.minmax(name, [step], [block_info_list]) to return min/max in current step (r mode or step argument) or global minmax (rra mode no step argument)

### DIFF
--- a/python/adios2/stream.py
+++ b/python/adios2/stream.py
@@ -5,6 +5,7 @@ accompanying file Copyright.txt for details.
 
 from functools import singledispatchmethod
 from sys import maxsize
+from typing import Optional
 import numpy as np
 
 # pylint: disable=duplicate-code
@@ -922,7 +923,7 @@ class Stream:
 
         return self._engine.all_blocks_info(name)
 
-    def minmax(self, name, step: int | None = None, block_info_list: list = []) -> tuple:
+    def minmax(self, name, step: Optional[int] = None, block_info_list: list = []) -> tuple:
         """
         Get min/max value for variable:
         - 'r' mode: min/max in the current step


python/adios2/stream.py
```diff
@@ -921,3 +922,49 @@ def all_blocks_info(self, name):
...
+
+        if self._mode == bindings.Mode.Read and step is not None:
+            raise ValueError("adios2.Stream.minmax() in 'r' mode must not set step")
+
+        variable = self._io.inquire_variable(name)
+        if variable is None:
+            return None, None
+        if variable.type() == "string":
+            return "", ""
+
```
anagainaru, line 952: I would return None for both. I would rather just check for None for variables instead of having first to know their type before knowing how to check if there is min/max for this variable

pnorbert:
> What is returned if the stats are turned off?

Excellent, provocative question. I forgot to check for that

pnorbert:
> > What is returned if the stats are turned off?
> 
> Excellent, provocative question. I forgot to check for that

min = 1.79769e+308  max = -1.79769e+308

Apparently we always have min/max in adios2 now. The type's min/max if not stored

eisenhauer:
> > > What is returned if the stats are turned off?
> > 
> > 
> > Excellent, provocative question. I forgot to check for that
> 
> min = 1.79769e+308 max = -1.79769e+308
> 
> Apparently we always have min/max in adios2 now. The type's min/max if not stored

Note that the Min is type Max and the Max is Type Min.  So if Min > Max you know they aren't proper values...  You could return NaN or something instead if you detect that.

anagainaru:
Or None since that is what we return for variables that do not support min/max

pnorbert:
Well, this is what the C++ api returns, and this is what bpls reports. So I am not sure I want to make the python api different (smarter)

Issue #4807, opened 2026-01-18
ADIOS 2.11.0 DILL CMake issue on macOS GH Action Runners


ax3l:
In https://github.com/openPMD/openPMD-api/pull/1833, I am trying to build an all-static build of openPMD-api dependencies, including ADIOS2.
I am updating that build logic from ADIOS 2.10 to 2.11 right now.

My Linux builds are all good. Using the exact same build logic, but on macOS-14 arm64 and macOS-15 x86_64 GitHub runners alike, I get the following error from just updating from 2.10 to 2.11:
```
2026-01-18T05:33:31.8570680Z -- Found atl: /Users/runner/work/openPMD-api/openPMD-api/src/build-adios2/thirdparty/atl/atl/atl-config.cmake (found version "2.2.2")
2026-01-18T05:33:31.8733430Z CMake Error at /private/var/folders/01/3lf990bj2yxcpypsx6hqj_m80000gn/T/cibw-run-h0bofn0j/cp310-macosx_arm64/build/venv/lib/python3.10/site-packages/cmake/data/share/cmake-4.2/Modules/CMakeFindDependencyMacro.cmake:93 (find_package):
2026-01-18T05:33:31.8734340Z   By not providing "Finddill.cmake" in CMAKE_MODULE_PATH this project has
2026-01-18T05:33:31.8734710Z   asked CMake to find a package configuration file provided by "dill", but
2026-01-18T05:33:31.8735000Z   CMake did not find one.
2026-01-18T05:33:31.8735110Z 
2026-01-18T05:33:31.8735250Z   Could not find a package configuration file provided by "dill" with any of
2026-01-18T05:33:31.8735520Z   the following names:
2026-01-18T05:33:31.8735610Z 
2026-01-18T05:33:31.8735660Z     dillConfig.cmake
2026-01-18T05:33:31.8736050Z     dill-config.cmake
2026-01-18T05:33:31.8736150Z 
2026-01-18T05:33:31.8736290Z   Add the installation prefix of "dill" to CMAKE_PREFIX_PATH or set
2026-01-18T05:33:31.8736620Z   "dill_DIR" to a directory containing one of the above files.  If "dill"
2026-01-18T05:33:31.8736980Z   provides a separate development package or SDK, be sure it has been
2026-01-18T05:33:31.8737220Z   installed.
2026-01-18T05:33:31.8737360Z Call Stack (most recent call first):
[...]

PR #4820, opened 2026-01-26
Fix MacOS Static build without FFI

eisenhauer:


ax3l:
@eisenhauer unfortunately still the same issue with this patch.
Doesn´t the [error look like](https://github.com/ornladios/ADIOS2/issues/4807#issue-3826371484) we need to expand the `CMAKE_PREFIX_PATH` / set the `dill_DIR` to the download location of DILL?

eisenhauer:
> @eisenhauer unfortunately still the same issue with this patch. Doesn´t the [error look like](https://github.com/ornladios/ADIOS2/issues/4807#issue-3826371484) we need to expand the `CMAKE_PREFIX_PATH` / set the `dill_DIR` to the download location of DILL?

The deal is that this version DILL wont/cant build build on macos without libffi.  It can't do dynamic code generation (it's main job) so it falls back to emulation of DCG, but to do that it needs libffi.  It doesn't have that, so it errors out and doesn't build.  FFS is OK with that, it has it's own substitute if it can't get DCG, but something is going wrong with the logic somehow.  But anyhow, there's no appropriate path to set for DILL because it doesn't build here. Looking at this again, it's still not getting past ffs, so the EVPath mods are inoperative.  I've added some diagnostic output to the ADIOS PR.  Can you retrigger?

ax3l:
Thank you, refreshed the patch: https://github.com/openPMD/openPMD-api/pull/1833

eisenhauer:
I'm not seeing the output that I should see from the diagnostics.  After it says: "  -- ADIOS2 ThirdParty: Configuring ffs", the next line should be either:  "ADIOS2 FFS: DID NOT FIND dill::dill target, setting FFS_USE_DILL to OFF" or "ADIOS2 FFS: found dill::dill target, setting FFS_USE_DILL to ON".  Neither appears.  Possible that something is going wrong with the patching?  Or CI caching?

ax3l:
Def had an issue with applying the patch, but now it works :tada: 

Your patch overcomes the issue! Sorry for fumbling the patch initially.

eisenhauer:
Excellent!

ax3l:
Sadly, it looks like it needs the EVPath changes, too to find DILL.

eisenhauer:
Ugh.  Unfortunate.   I’ve put those changes into EVPath master.  The only question is whether we or not we want to pull all the recent changes to the GTkorvo stack into release_211 or just patch it.  In the meantime, I’ll try to get it all into master.


From: Axel Huebl ***@***.***>
Date: Tuesday, January 27, 2026 at 2:25 PM
To: ornladios/ADIOS2 ***@***.***>
Cc: Eisenhauer, Greg S ***@***.***>, Mention ***@***.***>
Subject: Re: [ornladios/ADIOS2] Fix MacOS Static build without FFI (PR #4820)

[https://avatars.githubusercontent.com/u/1353258?s=20&v=4]ax3l left a comment (ornladios/ADIOS2#4820)<https://github.com/ornladios/ADIOS2/pull/4820#issuecomment-3807088412>

Sadly, it looks like it needs the EVPath changes, too to find DILL.

eisenhauer:
Closing as redundant - the changes in this PR are included in #4817

PR #4822, opened 2026-01-29
Xrootd https remote

eisenhauer:


eisenhauer:
The docker stuff here will need to change after this is merged and we've can build ADIOS from an official source, but including it in this PR because it was necessary for testing.

vicentebolea:
@eisenhauer is this intended to run in the CI?

eisenhauer:
> @eisenhauer is this intended to run in the CI?

Clarifying.  The docker scripts are not for CI, but just to help build what we would need to have an ADIOS/XRootD server in a docker image for deployment to environments like NERSC Spin.  The new transports *would* be built and tested in CI if we had a build that had both xrootd and libcurl.  I haven't looked to see if we do.  (And actually I guess this would be like the other situation where we might have to add CURL to the spack after this is integrated?)

vicentebolea:
sounds good! I suggest to make these scripts also podman compatible (so they can build without root)

eisenhauer:
That sounds like a reasonable plan.  Once this is merged, I can rewrite those scripts to run out of master and use podman when available.

eisenhauer:
WTF?  I didn't ask copilot do to anything, yet it is self-initiating?  Or did someone else ask?

vicentebolea:
Oh @eisenhauer sorry for the disruption, I was testing copilot capabilities and it did make the changes (it also opened another PR), I can see that it went beyond what I asked him. Feel free to remove its changes. haha perhaps in the future copilot will self-invoke himself.

eisenhauer:
No worries.  I was just somewhat surprised.

PR #4829, opened 2026-02-02
Stabilize CI tests

eisenhauer:
- Skip LargeBlocks test under MSAN (times out with ~4GB allocation)
- Fix remote server startup race condition

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/source/adios2/toolkit/remote/remote_server.cpp
+++ b/source/adios2/toolkit/remote/remote_server.cpp
@@ -970,6 +970,8 @@ int main(int argc, char **argv)
         {
             printf("Forking server to background\n");
         }
+        // Ready file used to synchronize parent/child - parent waits for this before exiting
+        char ready_filename[256];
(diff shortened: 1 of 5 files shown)


eisenhauer:
Hoping that this PR gets rid of the predictable failures in our sanitizers.  FIxing the timeouts on LargeBlocks by skipping that test with the sanitizers.  The tests correctness checks should mean that msan doesn't add anything and tsan isn't helpful here, so skipping seems best.  Also adding something to ensure that the adios remote server is actually started before we return from CI Fixture setup.

And now added another fix.  In the "OpenEngineTwice" tests, there was a possible race condition because the destructor might have been called immediately (since we didn't save the opened file, destruction could happen at any time).  Saving the opened file avoids this problem.

pnorbert (approved):

PR #4834, opened 2026-02-03
BP5: Fix segfault with DataSizeBased aggregation and async write

scottwittenburg:
Fix an oversight with `DataSizeBased` aggregation that resulted in the aggregator not getting initialized when async write mode was enabled.

Help ensure that `DataSizeBased` aggregation continues to work with async mode by expanding the async write tests to include DSB (async write was previously only tested with EWS and TLS).

### DIFF
--- a/source/adios2/engine/bp5/BP5Writer_EveryoneWrites_Async.cpp
+++ b/source/adios2/engine/bp5/BP5Writer_EveryoneWrites_Async.cpp
@@ -217,6 +217,18 @@ int BP5Writer::AsyncWriteThread_EveryoneWrites(AsyncWriteInfo *info)
 
 void BP5Writer::WriteData_EveryoneWrites_Async(format::BufferV *Data, bool SerializedWriters)
 {
+    if (m_Parameters.AggregationType == (int)AggregationType::DataSizeBased)
+    {
+        if (!m_AggregatorInitializedThisStep)
+        {
+            // We can't allow ranks to change subfiles between calls to Put(), so we only
+            // do this initialization once per timestep. Consequently, partition decision
+            // could be based on incomplete step data.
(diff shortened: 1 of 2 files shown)


scottwittenburg:
@guj I discovered that even when I made DSB the default aggregation type, the tests of async write were overriding that setting, and it was not getting tested.  When I added DSB to the async write tests, I was able to reproduce a segfault that appeared to be the same as the one you emailed me about a couple months ago. The small change here fixes the segfault for me, and I assume it will do so for your use case as well, based on the stack trace you shared back then.

scottwittenburg:
@vicentebolea @pnorbert This seems like a candidate for a patch release, assuming it can be merged quickly. What do you think? What is the process for backporting to release?

scottwittenburg:
Here is a [list](https://open.cdash.org/builds/11017536/tests?filters=%7B%22all%22%3A%5B%7B%22eq%22%3A%7B%22status%22%3A%22PASSED%22%7D%7D%2C%7B%22contains%22%3A%7B%22name%22%3A%22Async%22%7D%7D%2C%7B%22contains%22%3A%7B%22name%22%3A%22DSB%22%7D%7D%5D%7D) of the new tests added by this PR.

eisenhauer (approved):
Looks good.  Lotsa new tests.  Hopefully we've got time to run everything.

scottwittenburg:
> Hopefully we've got time to run everything.

I think it should be ok. According to the test results I linked, the total time for all 48 of those to run on `ubuntu22.04-clang-mpich` was under 2 seconds. Somewhat oddly, running that same subset of tests on my machine locally seems to take about 5 seconds, I'm not sure why they ran so much faster in CI. In any event, I don't think it will have much impact on the overall test time.

PR #4841, opened 2026-02-05
Support host aliases

pnorbert:
in ~/.config/hpc-campaign/hosts.yaml including aliases to the local machine

    E.g.

    NERSC:
      dtn-ssh:
          protocol: ssh
          ...

    NERSC.UEDGE: NERSC
    NERSC.HPSS: NERSC

    TESTMACHINE: local

### DIFF
--- a/source/adios2/common/ADIOSTypes.h
+++ b/source/adios2/common/ADIOSTypes.h
@@ -421,7 +421,8 @@ enum class HostAccessProtocol
     Invalid,
     SSH,
     XRootD,
-    S3
+    S3,
+    LocalHost
 };
 
 /** Host authentication protocols */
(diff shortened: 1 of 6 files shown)


pnorbert:
This is needed because I realized "NERSC" is not a good host name for campaigns from different projects. It may mean different access methods for different projects. Aliases can point a special name to a commonly set up access name for convenience, and also for sharing remote connections to NERSC, NERSC.UEDGE, NERSC.HPSS in the above example. 

A special alias is `local`, so that e.g. on Perlmutter, users can set NERSC.UEDGE to local and thus make adios read data from disk, without forcing them to name the hostname in `config.yaml` to NERSC.UEDGE, which is a project specific name. Same users will want to read NERSC.ELMO data in the near future as well.

eisenhauer (approved):

PR #4842, opened 2026-02-05
BP5Deserializer: capture selection state at queue time for thread safety

eisenhauer:
Add MemoryStart, MemoryCount, and AccuracyRequested fields to BP5ArrayRequest struct. These values are now captured when requests are queued (in QueueGetSingle and QueueGetSingleRemote) rather than being read from VariableBase during FinalizeGet.

This is the first step toward thread-safe Get() operations. Previously, the deserializer would read m_MemoryStart, m_MemoryCount, and accuracy from the VariableBase at finalization time, which created a race condition when multiple threads shared a Variable object.

No behavioral change - this is an internal refactoring to prepare for a future Selection-aware API.


anagainaru (approved):
Seems straight forward

PR #4845, opened 2026-02-06
Cleanup files that might be lingering from earlier output

eisenhauer:
As discussed in ADIOS developers meeting weeks ago.  In order not to create a storm of filesystem metadata operations to remove files that we're just going to replace with fresh data, we don't just remove the whole directory and start over.  Instead we take the 'whitelist' approach and only remove what we're not going to overwrite.

### DIFF
--- a/source/adios2/engine/bp5/BP5Reader.cpp
+++ b/source/adios2/engine/bp5/BP5Reader.cpp
@@ -85,6 +85,11 @@ void BP5Reader::GetMetadata(char **md, size_t *size)
        so we have to read it from file again
     */
     std::vector<char> mdbuf(sizes[0]);
+    if (!m_MDFile)
+    {
+        std::string metadataFile(GetBPMetadataFileName(m_Name));
+        m_MDFile = m_DataFiles->Acquire(metadataFile, m_dataIsRemote);
+    }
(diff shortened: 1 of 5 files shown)


source/adios2/engine/bp4/BP4Writer.cpp
```diff
@@ -249,6 +250,25 @@ void BP4Writer::InitTransports()
...
 
+    // Cleanup stale files from previous writes (only in Write mode)
+    if (m_OpenMode == Mode::Write)
+    {
+        std::vector<std::string> filesToKeep;
+        filesToKeep.push_back(m_BP4Serializer.GetBPMetadataFileName(m_Name));      // md.0
+        filesToKeep.push_back(m_BP4Serializer.GetBPMetadataIndexFileName(m_Name)); // md.idx
+        filesToKeep.push_back(m_BP4Serializer.GetBPVersionFileName(m_Name));       // .bpversion
+        filesToKeep.push_back(m_BP4Serializer.GetBPActiveFlagFileName(m_Name));    // active
+        // data.0 through data.(size-1) - BP4 uses one file per rank
```
anagainaru, line 261: Isn't this different depending on the aggregation strategy or is that just BP5?

anagainaru (approved):

PR #4846, opened 2026-02-07
BP5Reader: close metadata FDs when writer is inactive

eisenhauer:
When the writer's active flag is false, all metadata is in its final state. Close each metadata file descriptor individually once all of its data has been read into memory:

- mmd.0: always read in full, closed immediately
- md.idx: closed when ParseMetadataIndex has consumed all entries (respects the 16 MB streaming-mode limit across multiple calls)
- md.0: closed once all index entries are parsed, since every metadata section they reference has already been read

This reduces file descriptor pressure on remote servers (e.g. XRootD) when many datasets are opened concurrently via campaign files.

Also adds an fstat retry loop for md.idx to handle filesystem metadata lag: after a writer closes a file, fstat() may briefly report a stale size. The loop polls until the reported size stabilizes.

Finally, it adds a test, counting open FDs to make sure that the metadata FDs are actually closed.  This test only functions on non-windows platforms.

### DIFF
--- a/source/adios2/engine/bp5/BP5Reader.cpp
+++ b/source/adios2/engine/bp5/BP5Reader.cpp
@@ -85,6 +85,11 @@ void BP5Reader::GetMetadata(char **md, size_t *size)
        so we have to read it from file again
     */
     std::vector<char> mdbuf(sizes[0]);
+    if (!m_MDFile)
+    {
+        std::string metadataFile(GetBPMetadataFileName(m_Name));
+        m_MDFile = m_DataFiles->Acquire(metadataFile, m_dataIsRemote);
+    }
     m_MDFile->Read(mdbuf.data(), sizes[0], 0);
 
     size_t mdsize = sizes[0] + sizes[1] + sizes[2] + 3 * sizeof(uint64_t);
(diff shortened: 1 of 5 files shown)


source/adios2/engine/bp5/BP5Reader.cpp
```diff
@@ -1440,10 +1445,11 @@ void BP5Reader::UpdateBuffer(const TimePoint &timeoutInstant, const Seconds &pol
 {
     size_t newIdxSize = 0;
     m_MetadataIndex.Reset(true, false);
-    if (m_Comm.Rank() == 0)
+    m_MetadataIndex.m_Buffer.resize(0);
```
anagainaru, line 1448: Can you do a reset of the metadata index without resizing the buffer to 0? In no, I would put the resize inside the reset function

eisenhauer: Most of the time we call Reset and want the buffer to survive.  Here, the next bit of code needs to see the buffer as size zero, so we can't really move it inside (without adding something like a parameter to say we wanted the buffer to go away or something).

PR #4847, opened 2026-02-08
ADIOSFilePool: per-SubPool locking for concurrent file access

eisenhauer:
- Drop `pool_mutex` after map lookup so the slow file-open path only blocks requests for the same file, not the whole server
- Add per-SubPool `subpool_mutex` for fine-grained locking
- Store SubPools as `shared_ptr` to keep them alive after `pool_mutex` is released; return `PoolEntry` struct to callers
- `FlushUnused()` now correctly holds `pool_mutex` (was missing before — race condition)

### DIFF
--- a/source/utils/xrootd-plugin/AdiosFilePool.cpp
+++ b/source/utils/xrootd-plugin/AdiosFilePool.cpp
@@ -5,37 +5,43 @@
 namespace adios2
 {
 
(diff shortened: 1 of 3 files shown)


eisenhauer:
@pnorbert I did this to see if this was a server-side holdup for campaigns.  It was not, but it's a useful improvement, so keeping it.

pnorbert (approved):

PR #4853, opened 2026-02-10
Campaign reader engine must call PerformGets in DoClose to enable mul…

pnorbert:
…tithreaded handling of multiple engines reading


pnorbert:
@eisenhauer Can you look at the tests with xrootd+https. Is this a new heisenbug?

eisenhauer:
> @eisenhauer Can you look at the tests with xrootd+https. Is this a new heisenbug?

It is!  I've seen it once before, but it went away on re-run.  Will look again.

PR #4854, opened 2026-02-10
Decouple plugin operator interface from ADIOS internals

eisenhauer:
Extract DataType, Dims, Params, Accuracy into ADIOSBaseTypes.h as a self-contained header with no internal dependencies. Rewrite PluginOperatorInterface as a standalone abstract class (no core::Operator inheritance) so external plugin operators need only include a single header and no longer link against adios2_core.

This is a first step to not exporting all of our internal .h files for the sake of plugins.  Fixing the plugin engine interface will be much harder.  

I'm not wildly happy about the duplication of DataTypeSize and GetTotalSize here, but moving those functions out of helper while still making sure that plug-in operators don't link against adios_core is a little more involved.  For the moment this seemed like an easy solution that can be cleaned up later if desired.


source/adios2/operator/plugin/PluginOperatorInterface.h
```diff
@@ -5,36 +5,167 @@
...
+        std::memcpy(&ret, buffer + pos, sizeof(T));
+        pos += sizeof(T);
+        return ret;
+    }
+
+    /**
+     * Return the size in bytes of a single element of the given DataType.
+     * Replaces helper::GetDataTypeSize() for plugin use.
+     */
+    static size_t DataTypeSize(const DataType type) noexcept
```
pnorbert: Can't these two functions go into the base type header, and use it both in the plugin and in adios core to have a single implementation of it?

pnorbert (dismissed):
I approve it already but maybe the GetTotalSize and DataTypeSize could go into that single exported header.

eisenhauer:
> I approve it already but maybe the GetTotalSize and DataTypeSize could go into that single exported header.

Yeah, the easy thing is to put them as inline functions in the new BaseTypes header.  Then the choice is to either 1) just change the bodies of those functions in helper so that they use the new version (least code disruption), or 2) kill those functions in helper and then modify everywhere they are called to remove the helper:: prefix (if there wasn't a use statement that made it unnecessary).   Honestly, the latter seemed like the right approach, but I wasn't sure I wanted to touch all those files in this PR.  But we've probably got time to decide, Caitlin is the code owner, so no merging until she has a look.  So what's your preference?  And do it here or another PR?

pnorbert:
Either way is fine with me.

eisenhauer:
> Either way is fine with me.

OK, I've de-dup'd.

pnorbert (approved):

eisenhauer:
@caitlinross

plugins/operators/CMakeLists.txt
```diff
@@ -7,7 +7,11 @@ if(ADIOS2_HAVE_Sodium)
   add_library(EncryptionOperator MODULE
     EncryptionOperator.cpp
   )
-  target_link_libraries(EncryptionOperator adios2_core sodium)
+  target_link_libraries(EncryptionOperator sodium)
+  target_include_directories(EncryptionOperator PRIVATE
+    ${PROJECT_SOURCE_DIR}/source
+    ${PROJECT_BINARY_DIR}/source
+  )
```
vicentebolea, line 14: Why is this needed

source/adios2/common/ADIOSBaseTypes.h
```diff
@@ -0,0 +1,110 @@
...
+        return sizeof(std::complex<float>);
+    case DataType::DoubleComplex:
+        return sizeof(std::complex<double>);
+    default:
+        return 0;
+    }
+}
+
+/** Return the total number of elements (product of dimensions) times elementSize. */
+inline size_t GetTotalSize(const Dims &dimensions, const size_t elementSize = 1) noexcept
```
vicentebolea, line 102: Previous version use to throw an exception this could cause unintended behaviors

source/adios2/common/ADIOSBaseTypes.h
```diff
@@ -0,0 +1,110 @@
...
+{
+    double error;
+    double norm;
+    bool relative;
+};
+
+/** Return the size in bytes of a single element of the given DataType. */
+inline size_t GetDataTypeSize(const DataType type) noexcept
+{
+    switch (type)
```
vicentebolea, line 71: Why not using the macro ADIOS2_FOREACH_STDTYPE_1ARG

source/adios2/operator/plugin/PluginOperatorInterface.h
```diff
@@ -5,36 +5,119 @@
...
+     * @param type data type to check
+     * @return true if this operator supports the given data type
+     */
+    virtual bool IsDataTypeValid(const DataType type) const = 0;
+
+    /** Give an upper bound estimate how big the transformed data could be */
+    virtual size_t GetEstimatedSize(const size_t ElemCount, const size_t ElemSize,
+                                    const size_t ndims, const size_t *dims) const
+    {
+        return ElemCount * ElemSize + 128;
```
vicentebolea, line 75: why the 128?

source/adios2/common/ADIOSBaseTypes.h
eisenhauer: Then I have to get that macro in here.  This is supposed to be pretty standalone.

plugins/operators/CMakeLists.txt
eisenhauer: Because we're not linking this against core, so we don't get the include paths from that.

source/adios2/common/ADIOSBaseTypes.h
eisenhauer: Maybe??  But I think it's a stretch.  This got rid of helper::throw(), but really operators that only work on specific types probably don't count on GetTotalSize() throwing an exception to protect themselves.

source/adios2/operator/plugin/PluginOperatorInterface.h
eisenhauer: Just a (hopefully) generous guess as to what the operator might add for a header.  They can give us a better guess of course.

vicentebolea:
Would it make sense to create a new library in adios2 for `plugins/operators` that could further organize the dependencies?

eisenhauer:
> Would it make sense to create a new library in adios2 for `plugins/operators` that could further organize the dependencies?

At least for the operator, the hope is that it the only thing it gets from ADIOS is relatively static .h files.  There were some simple helper routines, but those are pretty trivial, easily duplicated or declared inline in a .h.  Do you have other dependencies in mind?

vicentebolea:
> > Would it make sense to create a new library in adios2 for `plugins/operators` that could further organize the dependencies?
> 
> At least for the operator, the hope is that it the only thing it gets from ADIOS is relatively static .h files. There were some simple helper routines, but those are pretty trivial, easily duplicated or declared inline in a .h. Do you have other dependencies in mind?

I see, still there is dependency to the IO class, meaning that it will depend on adios_core, in cmake an interface library could fit this profile (only headers but publicly depends on other components). It might help organize these dependencies.

eisenhauer:
> I see, still there is dependency to the IO class, meaning that it will depend on adios_core, in cmake an interface library could fit this profile (only headers but publicly depends on other components). It might help organize these dependencies.

There may be some confusion.  The source in ADIOS2/source/adios2/operator/plugin/PluginOperator is part of core adios and does have dependencies.  However, that is the mechanism through which external plugin operators are loaded and executed.  That thing is not why we've been installing all of our internal .h files.  

The stuff in ADIOS2/plugins/operators/EncryptionOperator *used* to transitively include all of our internal include files.  In this PR it no longer does.   It still includes "adios2/operator/plugin/PluginOperatorInterface.h", but that file is dramatically simplified so that it only includes "adios2/common/ADIOSBaseTypes.h".  The point being now we are installing a couple of simple .h files that lets us create external plugin operators that do not reference ADIOS internals.  They are sandboxed to this couple of .h files with no need to link to or otherwise access ADIOS internals.

source/adios2/operator/plugin/PluginOperatorInterface.h
vicentebolea: I see, I would then name this somehow to avoid magic numbers. Perhaps naming it inside the function to understand its rationale would help future code readers (or llms).

source/adios2/common/ADIOSBaseTypes.h
vicentebolea: Makes sense

plugins/operators/CMakeLists.txt
vicentebolea: Depending in adios2_core only if for a few headers files I think that should be ok and would not defeat the purpose of these changes. I would also in this PR remove the installation of those private headers since they are not needed anymore. I propose to depend on adios2_core in this operator since it is more idiomatic and safer.

vicentebolea:
I see, thanks for explaining I understand it now: it is not needed for us to further split this into different cmake "libs" (interface lib I had in mind). The idea is that these headers will still be part of adios2_core and only those few headers are needed for plugins/operators to be built.

plugins/operators/CMakeLists.txt
eisenhauer: This PR only addressed the plugins operators, not the engines. The engines need the include core include files and that isn't addressed until the next PR (where we stop installing all those .h files).

eisenhauer:
Closing this because the plugin engine PR has all this content.

PR #4855, opened 2026-02-10
Fix XRootD HTTPS test startup race

eisenhauer:
Poll HTTPS port for readiness instead of sleep+kill-0. Fixes transient CI failures where the TLS listener isn't ready before tests run.


eisenhauer:
@pnorbert Hoping to kill that heisen-bug.

PR #4857, opened 2026-02-10
ci: fix readthedocs job

vicentebolea:


eisenhauer:
@vicentebolea I have a different PR that tries to eliminate blockdiag entirely to remove this problem.  Can you review?

eisenhauer:
The mermaid diagrams look a lot like the graphviz diagrams.  The IO diagram in particular I think fails to work nearly as well as the original blockdiag did.

vicentebolea:
> The mermaid diagrams look a lot like the graphviz diagrams. The IO diagram in particular I think fails to work nearly as well as the original blockdiag did.

yep the results were not so great :(

eisenhauer:
> > The mermaid diagrams look a lot like the graphviz diagrams. The IO diagram in particular I think fails to work nearly as well as the original blockdiag did.
> 
> yep the results were not so great :(

I did mess with it for a while in a docker image offline.  Just couldn't make it look right (in graphviz).

PR #4861, opened 2026-02-11
Eliminate blockdiag in favor of uploaded svg and png images.

eisenhauer:
Hoping that this is a more permanent fix to our longstanding readthedocs problems.


vicentebolea:
This while working seems to be cumbersome since we will have to locally generate these figures whenever they need to change.

eisenhauer:
> This while working seems to be cumbersome since we will have to locally generate these figures whenever they need to change.

They haven't changed since they were created in 2018.  In comparison, readthedocs has broken continuously.  Given the choice, I like putting the figures in the repo rather than trying to keep blockdiag working.  (And any changed images could be generated from the commented-out blockdiag description (if someone got it working again using old tools), or if all else fails the svg can be edited with tools like inkscape.  But realistically, the frequency of change is likely to remain roughly where it is, negligible.)

vicentebolea:
@eisenhauer make sense. It appears that blockdiag is long discontinued which explains why we had issues with it keep up with newer pythons. It appears that the best solution would be to move to mermaid, which is the current industry standard, it also plays well with github which can render it out of the box, I still favor it since it will be easier to maintain them (if we ever have too) and LLMs are better at using those diagram languages than actually doing their diagrams.

eisenhauer:
Yeah, we talked years ago about redoing those figures, but it never happened.  I played with graphviz for a while, but it was non-trivial to render those figures in anything that kind of conveyed the same thing as the originals did.  I honestly don't think the SVG is a bad way to go.  It looks just like the blockdiag since it's generated from that, and it's editable should we ever need to change it (unlikely at this point, since they illustrate basic adios concepts).  It's a pity that we need the png's for the backends that don't understand svg, but meh.  I don't know anything about mermaid, so maybe that's an option, but give the trouble I had with graphviz, I might temper my expectations until I see actual renders.

eisenhauer:
Works for me. I think we have to override rules to merge this as docs only.

PR #4863, opened 2026-02-12
Plugin engine interface

eisenhauer:
This PR builds on the previous plugin operator proposal to also encompass plugin engines.  The main change is to make plugin engines use the _public_ ADIOS APIs rather than calling into core directly.  This required some minor tweaks to the bindings classes, not to change their external visibility, but to let us call them from PluginEngine (the internal bridge to actual plug-ins).  This is pretty much the biggest change.  The changes to the actual plugin example pretty much consist of calling adios2:: rather than core::.   There was also a minor change, moving ToString(Dims) to be inline in adiosTypes.h to avoid ToString depending upon core.

Note that this PR **_does_** stop installing all of our internal #include files.  I included that here to make sure everything in the build/install tests works after that is done.  In practice, we maybe want to do this piece-wise, introduce the new plugin interfaces in one release, then blow away those includes in the next major release.

Also Note, this PR is on top of the unmerged operator plugin PR, so it also includes all those changes in the diff.  If we merge the other PR, this will get rebased.


pnorbert (dismissed):
I approve this and vote for blowing away all that includes from the next release. So that we can have minor releases again.

pnorbert:
A caveat though, I did want to install some headers in the past so that I could use useful helper functions and kwsys functions in my apps instead of copy-pasting their implementation all the time. `DimsToString` is the prime example. Counting dimensions, minmax of array.

eisenhauer:
kwsys we probably can't do much about, though it is handy.  I will suggest that moving to C++17 so that we have std::filesystem might go a long way towards replacing that.  WRT DimsToString, there's no reason why simple utility functions operating on the AdiosTypes or AdiosBaseTypes shouldn't have inline definitions in the appropriate .h file.

eisenhauer:
@caitlinross

eisenhauer:
Great, thanks @caitlinross !

PR #4865, opened 2026-02-13
BP5Reader: Fix a possible issue when writer map changes between steps

scottwittenburg:
If the emplace call returned an existing WriterMapStruct, then that struct already had a RankToSubfile vector of the correct size. In that case, the reserve() call did not change anything, but using push_back() to add elements increased the size beyone the number of writers. So those correct values were never reached later when accessing, and instead the previous (incorrect) subfile values were found.


scottwittenburg:
@eisenhauer Here is the reader fix we were discussing the other day, thanks again for your help.

PR #4870, opened 2026-02-16
Support xrootd connections in hosts.yaml, with xrootd/http/https tran…

pnorbert:
…sfer protocols. For BP5 only for now.

### DIFF
--- a/source/adios2/engine/bp5/BP5Reader.cpp
+++ b/source/adios2/engine/bp5/BP5Reader.cpp
@@ -1170,19 +1170,25 @@ void BP5Reader::Init()
         {
             if (getenv("DoXRootDHttps"))
             {
-                m_RemoteHost = getenv("XRootDHttpsHost");
+                char *env = getenv("XRootDHttpsHost");
+                if (env)
+                    m_RemoteHost = std::string(env);
                 m_RemoteProtocol = HostAccessProtocol::XRootD;
                 m_XrootdTransferProtocol = XRootDTransferProtocol::HTTPS;
             }
             else if (getenv("DoXRootDHttp"))
             {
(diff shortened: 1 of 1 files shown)


pnorbert:
@eisenhauer Assuming 8443 and 8080 are not really default ports, now 443 and 80 are used as defaults if not specified in host.yaml properly, or in environment as hostname:port. So this PR will break your local tests if you rely on default port.

pnorbert:
host.yaml format for xrootd access:
```
NERSC.UEDGE:
  spin:
      protocol: xrootd
      transfer_protocol: https  # http, https or xrootd (default is xrootd)
      host: <host>
      port: 443
      verbose: 1
```

pnorbert:
I turned off VerifySSL since that's the only thing working. Should we have an option for this in the host.yaml?

eisenhauer:
> I turned off VerifySSL since that's the only thing working. Should we have an option for this in the host.yaml?

Probably, but I'm not 100% sure.  Verifying SSL certificates is the curl default.  Certainly we want to skip that for local testing because it's a pain, but maybe not in real life?

pnorbert:
> > I turned off VerifySSL since that's the only thing working. Should we have an option for this in the host.yaml?
> 
> Probably, but I'm not 100% sure. Verifying SSL certificates is the curl default. Certainly we want to skip that for local testing because it's a pain, but maybe not in real life?

Yeah, probably not, but the spin server returns an error: 
` <Remote> <XrootdHttpRemote> <WaitForGet> : Get failed: SSL peer certificate or SSH remote key was not OK`

eisenhauer (approved):

pnorbert:
Local testing failed so I had to settle for default ssh when localhost is used without any setup. That's only happens in testing.

PR #4881, opened 2026-02-19
remove thirdparty ci files

vicentebolea:
These files create false security alerts.


vicentebolea:
@eisenhauer this also updates some of the thirdparties, let me know if it is better to stay in their current commit/tag/

PR #4883, opened 2026-02-21
Add Asymmetric encryption support EncryptionOperator

vicentebolea:
Extends EncryptionOperator with a hybrid asymmetric mode using
crypto_box_seal (Curve25519 + XSalsa20-Poly1305 with libsodium). Write requires only
a public key; read requires both. Public key can be derived from secret
automatically.

It adds a simple python script to help generate these key pairs (PyNaCl). Adds pre-generated test keypair and install tests. This script is only installed when the new opt: `ADIOS2_USE_SealKeygen=ON`.

New Params in the Encryption Operator:

  - PublicKeyFile — path to 32-byte Curve25519 public key file (write + read)
  - PublicKey — hex-encoded public key, 64 chars (alternative to file)
  - SecretKeyFile — path to 32-byte secret key file (read only)
  - SecretKey — hex-encoded secret key, 64 chars (alternative to file)

  Secret key lookup order (first found wins):
  1. SecretKeyFile param
  2. SecretKey param
  3. ADIOS2_SECRET_KEY_FILE env var
  
  
  4. ADIOS2_SECRET_KEY env var


eisenhauer:
Shall we sort the mods to plugin-operator and then get this on top of that?

vicentebolea:
> Shall we sort the mods to plugin-operator and then get this on top of that?

It sounds good, I wanted to open this so that I do not forget it later.

vicentebolea:
@eisenhauer @caitlinross this is ready for review

PR #4892, opened 2026-02-27
Catch error thrown by Campaign engine when engine.all_blocks_info() i…

pnorbert:
…s called, and provide an alternative implementationfor this engine.

### DIFF
--- a/python/adios2/engine.py
+++ b/python/adios2/engine.py
@@ -87,7 +87,6 @@ class Engine:
         try:
             for step in range(0, self.steps()):
                 output.append(self.blocks_info(name, step))
-            return output
         except ValueError:  # Campaign engine has no steps()
             vstep = 0
             while True:


pnorbert:
The main problem here is that Campaign Reader has no steps conceptually and CampaignReader::DoSteps() is not implemented.

python/adios2/engine.py
```diff
@@ -84,8 +84,21 @@ def all_blocks_info(self, name):
             list of BlockInfos
         """
         output = []
-        for step in range(0, self.steps()):
-            output.append(self.blocks_info(name, step))
+        try:
+            for step in range(0, self.steps()):
+                output.append(self.blocks_info(name, step))
+            return output
```
anagainaru: You don't need this return, if the try is successful you return after it anyway.

PR #4894, opened 2026-02-27
More complete plugin engine interface, document changes to all plugin interfaces, add whatsnew entry noting breaking change.

eisenhauer:
Since we're already changing the plugin interfaces, this PR goes a bit further to add useful engine calls that were missing from the previous plugin interface.  This does not implement any of these calls in any existing plugin, but just allows us to pass them down to any future plugin that might implement them.

new calls: include selection-based Gets (Selections recently added to BP5 and SST), DoPut with Span, Steps() and MinBlocksInfo().

This also updates the documentation that didn't get updated in the previous PR (which actually introduced a breaking change in the plugin interface).

This did involve moving the new selection class from core to common so that we can more easily use it in plugins.


pnorbert (approved):
approving without actually trying to make a plugin engine

PR #4895, opened 2026-03-01
Better XRootD eviction and tar FD reuse

eisenhauer:
There's really three things here:

1. cost-based eviction strategy for the xrootd server watching both FD counts and metadata size.  We start evicting at 90% of FD limits and when over 2Gb metadata (configurable).
2. introduce a new singleton to try to reuse the FDs of open tar files.  To the extent that we have tarballs in campaigns containing many BP5 files, they can essentially all be opened and share a single FD per tar file.  This essentially removed FD pressure as a constraint in our eviction strategy.  (Posix transport and BP5 only.)
3. the server dumps stats (human readable units) into the log every second (when there have been new requests since the last dump).


eisenhauer:
@pnorbert Unfortunately this isn't quite ready.  turns out that getrlimit in some docker images (like the ones on Spin) is wholly unreliable.  I.E. it says we can open a million FDs even if the kernel limit is much smaller.  That's not a huge problem for the open-file limit in BP5 since that is mostly cautionary, but for setting a resource limit for the server it's a huge pain.  I don't want to set a "reasonable" default like 1024 when we could really have 4096 before evicting.  Reading /proc doesn't help because it also returns that crazy million FD value.  So I'm replanning...  I need to push branches so that my docker build script works, but I should have left the PR for later...  

Well, I was wrong.  Fedora really does build with essentially a million FD limit and it works.  Replanning again.

PR #4896, opened 2026-03-02
ci: test fix sz3 by adding zstd

vicentebolea:


vicentebolea:
@eisenhauer ready for review

PR #4901, opened 2026-03-03
Refactor engine parameter parsing

scottwittenburg:
The goal of this PR is reduce the number of redefinitions of a couple helper methods in various places under the `testing` directory. There is now a single definition of `ParseEngineParams(std::string input)` and `Trim(std::string& str)`, located in a new helper file `ParamsHelpers.h`.


eisenhauer (approved):
Looks good to me.  Curious about how I managed to get both ':' and '=' versions active at the same time... Sigh.

PR #4903, opened 2026-03-03
Campaign in memory metadata

pnorbert:


eisenhauer (approved):
Speed improvements with this?

pnorbert:
CampaignReader hands over the in-memory-metadata read from database to BP4/5 engine, instead of saving to disk and pointing the BP engine to read it from there. I see big improvement when opening the first time. Second time, the metadata is already on disk, so BP engines can just read it from there (and no need to read anything from the database).

PR #4904, opened 2026-03-03
Rename BP5 DataTransport parameter to DataFileTransport

eisenhauer:


eisenhauer:
@franzpoeschel Note this change to AWS support to avoid an engine parameter name conflict with SST.

anagainaru (approved):

PR #4906, opened 2026-03-03
ci: add wheel integration tests

vicentebolea:
This is needed since it adds a simple import test for both our sdist and wheels in macos and linux. Windows seems to have problems that I do not have time to solve know, I will open a PR later to fix the windows sdist installation.

This incidentally fixes an issue with the macos sdist. We will probably to start making wheels for all the platforms, but lets wait for the nanobind transition so that we can reduce the number of wheels using abi3 compat

### DIFF
--- a/.github/workflows/pypackaging.yml
+++ b/.github/workflows/pypackaging.yml
@@ -25,6 +25,10 @@ on:
     types:
       - published
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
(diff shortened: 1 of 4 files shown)


vicentebolea:
@eisenhauer this is ready for review

vicentebolea:
@berkgeveci after merge I will backport this to 2.11 release branch so that we can push new sdist that fixes import issue in macos

eisenhauer (approved):
Looks like an improvement!

PR #4907, opened 2026-03-03
cmake: bump to c++ standard to c++17

vicentebolea:
It sets C++17 as the minimum C++ standard required, it also updates the docs/readme to reflect this change. It also do some small fixes to build in different compilers


vicentebolea:
@eisenhauer ready for review

PR #4923, opened 2026-03-05
Process and save in-memory-metadata object if file is part of a time-…

pnorbert:
…series, since the TimeSeries engine cannot support in-memory objects.

Fix to Engine::Close() so that DoClose() is not called on an engine in invalid state (after an exception, during desctruction)

### DIFF
--- a/source/adios2/engine/inline/InlineWriter.cpp
+++ b/source/adios2/engine/inline/InlineWriter.cpp
@@ -15,6 +15,8 @@
 #include "adios2/helper/adiosFunctions.h"
 #include <adios2-perfstubs-interface.h>
 
+#include <cstdlib>
+#include <cstring>
 #include <iostream>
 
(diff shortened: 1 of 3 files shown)


source/adios2/core/Engine.cpp
```diff
@@ -89,14 +89,17 @@ void Engine::PerformDataWrite() { return; }
 
 void Engine::Close(const int transportIndex)
```
eisenhauer, line 90: So, looking at this prompted to wonder why Engine::Close() has a transportIndex parameter.  The answer of course is that it's completely vestigial.  Never anything passed in, no engine looks at it (except SSC which passes it to an internal engine which then doesn't look at it).  Makes me wonder if at some point we should clean up these things that seemed like a good idea at the time but which never went anywhere.

PR #4926, opened 2026-03-07
Various HTTP/XROOTD changes

eisenhauer:
- Switch to GET in https/xrootd URLs.  Still parse old for now.
- fine tune curl wait loop to avoid busy-wait and reduce cpu overhead
- increase CampaignReaders' thread cap to improve HTTP get throughput
- Add retry logic for transient HTTP failures.


eisenhauer:
@pnorbert The thread count and curl loop changes here produced about 2.5x speedup in remote fetching in the torture test.  The cost is (very) occasional http retries.  That might go away if we go to production.  This change switches the type of URL we are sending the server.  For now the server accepts both new and old.  When all of your work is migrated, then we can kill the old stuff.  Here GET replaces POST and /adios replaces /ssi.

PR #4928, opened 2026-03-08
reuse: add REUSE compliance

vicentebolea:
Adds REUSE (reuse.software) support so every file has an explicit SPDX
copyright and license identifier. Satisfies the OSSF Scorecards license
check and makes it unambiguous for contributors and downstream consumers
what license applies to each file.

- `LICENSES/` with Apache-2.0, BSD-3-Clause, MIT, CC0-1.0 texts
- `.reuse/dep5` for thirdparty libs, binaries, and dotfiles
- SPDX headers added to ~600 source files
- CI workflow using `fsfe/reuse-action` to enforce compliance on PRs

### DIFF
--- a/python/adios2/adios.py
+++ b/python/adios2/adios.py
@@ -2,7 +2,7 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 
-""""""
+"""ADIOS2 ADIOS Python wrapper."""
 
 from adios2 import bindings
 from adios2.io import IO
--- a/python/adios2/derived_variable.py
+++ b/python/adios2/derived_variable.py
@@ -2,7 +2,7 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 
-""""""
+"""ADIOS2 DerivedVariable Python wrapper."""
 
 
 class DerivedVariable:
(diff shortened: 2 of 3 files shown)


anagainaru (commented):
Since we are touching all the files anyway, should we remove the header that has creation date and author since this info is already in GitHub. We could say something like "Copyright contributors of ADIOS" after your SPDX lines

eisenhauer:
> Since we are touching all the files anyway, should we remove the header that has creation date and author since this info is already in GitHub. We could say something like "Copyright contributors of ADIOS" after your SPDX lines

I'd second this.  Always considered it unnecessary...

eisenhauer (approved):

bindings/C/adios2/c/adios2_c_adios.cpp
```diff
@@ -1,11 +1,15 @@
...
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
 /*
  * Distributed under the OSI-approved Apache License, Version 2.0.  See
  * accompanying file Copyright.txt for details.
  *
  * adios2_c_adios.cpp
  *
```
anagainaru: I would remove this part, Apache license is include in your header

eisenhauer: Concur.

vicentebolea: Done!

PR #4937, opened 2026-03-13
Restructure OnDemand test for accuracy and reliability

eisenhauer:
We had an OnDemand test, but it didn't really cover all the possible circumstances (like the request queue going empty), _and_ it sometime misbehaved because the writer had no delays, so sometimes our reader would try to join and the writer would simply exit mid-join because it was heading that way anyway.  This improves things.


eisenhauer:
> I am not a fan of sleeps but I don't have a better solution. Looks good to me.

Yeah, with everybody going full speed, we really didn't test the OnDemand stuff, *and* sometimes we exited before some readers could join.  With strategic sleeps, we actually let the on-demand queue be empty at some points (letting the event queue build up), and then let the situation reverse.  Far better at testing all the code paths.

Issue #4661, opened 2025-10-13
Support for using external PerfStubs instead of built-in version


Sichao25:
Remember to motivate and summarize considering the community. Provide enough information to motivate such as:

**Why is this feature important?**
Our project depends on both PerfStubs and ADIOS2, and we currently build both from source in our workflow. This creates redundancy since ADIOS2 includes its own copy of PerfStubs. If ADIOS2 supports using an externally installed PerfStubs library, it would help avoid potential conflicts and reduce build complexity.

**What is the potential impact of this feature in the community?**
This would benefit other projects that use both PerfStubs and ADIOS2.

**Is your feature request related to a problem? Please describe.**
N/A

**Describe the solution you'd like and potential required effort**
ADIOS2 can have a CMake option (similar to some existing third-party dependencies) to use an externally installed PerfStubs library. For example, `ADIOS2_USE_EXTERNAL_PERFSTUBS`.

**Describe alternatives you've considered and potential required effort**
N/A

**Additional context**
If there are specific reasons why this should not be implemented for PerfStubs, please let me know.

Note: contents adapted from the [VTK contribution guidelines](https://github.com/Kitware/VTK/blob/master/Documentation/dev/git/develop.md#create-a-merge-request)

vicentebolea:
@Sichao25 I have a PR fixing this. Out of curiosity are there any other project using perfstubs?

PR #4949, opened 2026-03-17
CI: add ubuntu/gcc12/mpich job using only external dependencies

vicentebolea:
fixes: #4661


vicentebolea:
@eisenhauer this is ready for review. The introduction of a CI job that build adios2 with the external versions of its vendored libraries is a long time need that will greatly help us iron out issues in package managers downstream.

PR #4951, opened 2026-03-18
Defer hosts.yaml parsing until first access

eisenhauer:
A little rearrangement to use a singleton, and also only trigger parsing when we actually need the host info (when we have remote data in the reader engines that support it).

### DIFF
--- a/source/adios2/engine/bp5/BP5Writer.cpp
+++ b/source/adios2/engine/bp5/BP5Writer.cpp
@@ -326,27 +326,34 @@ uint64_t BP5Writer::WriteMetadata(const std::vector<char> &ContigMetaData,
             MDataTotalSize += sizeof(uint64_t);
         }
     }
+    // Assemble all output into a single WriteV call:
+    //   [header: MDataTotalSize|SizeVector|AttrSizeVector] [ContigMetaData] [AttributeBlocks...]
     MetaDataSize = 0;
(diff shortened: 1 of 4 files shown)


source/adios2/core/ADIOS.cpp
```diff
@@ -101,9 +99,32 @@ static std::atomic_uint adios_count(0);    // total adios objects during runtime
 const adios2::UserOptions &ADIOS::GetUserOptions() { return m_UserOptions; };
 
 /** A constant reference to the host options from ~/.config/hpc-campaign/hosts.yaml */
-const adios2::HostOptions &ADIOS::GetHostOptions() { return m_HostOptions; };
-/** A constant reference to the host options from ~/.config/hpc-campaign/hosts.yaml */
-const adios2::HostOptions &ADIOS::StaticGetHostOptions() { return *StaticHostOptions; };
+const adios2::HostOptions &ADIOS::GetHostOptions()
+{
+    static adios2::HostOptions hostOptions = LoadHostConfig();
```
pnorbert, line 104: On it's own, does this guarantee that we only read it once?

eisenhauer: Yes, this line is the guarantee.  A function-local static is initialized exactly once on the first call, and C++ makes it thread-safe (as of C++11).

PR #4954, opened 2026-03-19
try just specifying the full path for clang in the individual job

pnorbert:


eisenhauer:
> u

Just to clarify.  A cat left that comment by walking on my keyboard at an inopportune time.  I'm sure you find it deeply insightful.

pnorbert:
> > u
> 
> Just to clarify. A cat left that comment by walking on my keyboard at an inopportune time. I'm sure you find it deeply insightful.

Now, yes.

vicentebolea:
Here is the list of software of those ci macos images: https://github.com/actions/runner-images/blob/main/images/macos/macos-14-Readme.md#language-and-runtime

It appears that the `clang` alias used the homewbrew clang that comes without some deps (that could be installed with brew).

This is definitely the right approach to fix this since the alternative of if which is using the clang-`version` is destined to fail after an update of the compilers.

PR #4959, opened 2026-03-22
Use finite timeout in RDMA progress thread to fix shutdown hang

eisenhauer:
The 100 in the timeout is milliseconds, while the original -1 means wait forever.  Waiting forever seems to be fine in normal operation, but sometimes we'd get hangs on shutdown with CXI and pthread_cancel() wouldn't seem to release the thread either.  Waiting 100mSec is arbitrary, redoing the poll 10 times a second won't have a performance impact but also won't delay shutdown too much.

Amended now to add a test that is more likely to trigger this issue (not in CI, as this targets a problem with the CXI fabric provider, but having a real test is handy).

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -52,6 +52,9 @@ Changelog = "https://github.com/ornladios/ADIOS2/releases"
 # Trigger an install of the package, and run a basic test
 test-command = "python -m unittest adios2.test.simple_read_write.TestSimpleReadWrite"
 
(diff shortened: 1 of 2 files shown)


eisenhauer:
@vicentebolea Unrelated Windows failures?

vicentebolea:
#4964 brings the fix for it, it is due to changes in the github actions windows image

PR #4961, opened 2026-03-23
Bump version to v2.12.0-rc1

vicentebolea:


docs/user_guide/source/introduction/whatsnew.rst
```diff
@@ -154,6 +154,122 @@ Key features:
 
 See :ref:`BP5 S3 Object Storage<S3 Object Storage>` for configuration details.
 
+C++17 Minimum Requirement
+--------------------------
+
+The minimum C++ standard required to build ADIOS2 has been raised from C++14
+to C++17.
+
+BP5 DataFileTransport Parameter Rename
```
eisenhauer: I'd drop or reword this because 2.11 BP5 never had a DataFile transport, it only had that name in unreleased master versions.  Could just note the new DataFileTransport parameter...

docs/user_guide/source/introduction/whatsnew.rst
```diff
@@ -154,6 +154,122 @@ Key features:
...
+
+Python Stream: minmax()
+------------------------
+
+``adios2.Stream`` gains a new ``minmax(name, [step], [block_info_list])`` method
+that returns the minimum and maximum values of a variable. In read mode or when
+a step is specified it returns per-step statistics; in random-access mode
+without a step argument it returns the global min/max.
+
+BP5 Concurrent File Access
```
eisenhauer: I'd drop this.  There are internal changes, but nothing in the API (yet).  It'll be there, but current XRootD dev didn't need it, so it's not finished.  This is just internal work that supports it.

vicentebolea:
@eisenhauer I just dropped those two entries, we can revise this anytime before the final release.

PR #4964, opened 2026-03-23
pip: enable curl and openssl

vicentebolea:
Enable OpenSSL and Curl support in pip wheels for Linux.

- Install `openssl-devel` inside manylinux containers via `cibuildwheel` `before-all` hook
- Disable OpenSSL and Curl on Windows to avoid linking against unbundled system DLLs


vicentebolea:
@eisenhauer can you review this again :)

PR #4983, opened 2026-04-02
BP5: Rerouting improvements

scottwittenburg:
This PR includes one correctness fix, and several small, potential performance improvements to the rerouting aggregator (#4647).

This PR should not be merged while it contains the final WIP commit that forces rerouting (whenever possible) in CI.

### Correctness fix

Previously the buffer pool for non-blocking sends was a fixed size (set at 100), and it had no guards against re-using (and thus, obliterating) a buffer while a previous `ISend` using the buffer may still have been incomplete. This commit introduces a role-appropriate pool size, and also adds logic that forces the next `ISend` to wait if all buffers in the pool are actively being used by a prior `ISend`.

### Performance improvements

To reduce the likelihood of the background messaging thread spinning and consuming CPU while nothing is happening, this PR introduces a conditional yield within the background thread loop. If no work is done during an iteration of the loop (message receipt and response, rerouting pair identified and messaged, etc...) then `yield()` is called at the end of the loop.

MPI send-to-self can sometimes be slow, and it was used by each subcoordinator to signal itself to write at the beginning of the workflow. This PR replaces the self-send with direct signalling of the main thread to begin writing.

Previously, once the loop finished, the thread exited immediately, possibly causing buffers associated with incomplete `ISend` to be destroyed. Now that we're collecting the request object from each `ISend` (stored in the buffer pool), this PR iterates the pool one final time to wait for any incomplete `ISend` to complete.

### DIFF
--- a/source/adios2/helper/adiosComm.cpp
+++ b/source/adios2/helper/adiosComm.cpp
@@ -180,6 +180,17 @@ Comm::Status Comm::Req::Wait(const std::string &hint)
     return status;
 }
 
+bool Comm::Req::Test(const std::string &hint)
+{
+    bool result = false;
+    if (m_Impl)
+    {
+        result = m_Impl->Test(hint);
+        m_Impl.reset();
+    }
+    return result;
+}
+
 Comm::Win::Win() = default;
 
 Comm::Win::Win(std::unique_ptr<CommWinImpl> impl) : m_Impl(std::move(impl)) {}
--- a/source/adios2/helper/adiosComm.cpp
+++ b/source/adios2/helper/adiosComm.cpp
@@ -180,6 +180,20 @@ Comm::Status Comm::Req::Wait(const std::string &hint)
     return status;
 }
 
+bool Comm::Req::TestComplete(const std::string &hint)
+{
+    bool result = false;
+    if (m_Impl)
+    {
+        result = m_Impl->TestComplete(hint);
+        if (result)
+        {
+            m_Impl.reset();
+        }
+    }
+    return result;
+}
+
 Comm::Win::Win() = default;
 
 Comm::Win::Win(std::unique_ptr<CommWinImpl> impl) : m_Impl(std::move(impl)) {}
--- a/source/adios2/engine/bp5/BP5Engine.h
+++ b/source/adios2/engine/bp5/BP5Engine.h
@@ -143,13 +143,13 @@ public:
     MACRO(verbose, Int, int, 0)                                                                    \
     MACRO(NumAggregators, UInt, unsigned int, 0)                                                   \
     MACRO(AggregatorRatio, UInt, unsigned int, 0)                                                  \
-    MACRO(NumSubFiles, UInt, unsigned int, 0)                                                      \
+    MACRO(NumSubFiles, UInt, unsigned int, 2)                                                      \
     MACRO(StripeSize, UInt, unsigned int, 4096)                                                    \
     MACRO(DirectIO, Bool, bool, false)                                                             \
     MACRO(DirectIOAlignOffset, UInt, unsigned int, 512)                                            \
     MACRO(DirectIOAlignBuffer, UInt, unsigned int, 0)                                              \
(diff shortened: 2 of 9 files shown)


scottwittenburg:
@eisenhauer I have left a WIP commit on here for testing in CI, but I will remove that commit in preparation to merge, once you are comfortable with the changes.

source/adios2/helper/adiosComm.cpp
```diff
@@ -180,6 +180,17 @@ Comm::Status Comm::Req::Wait(const std::string &hint)
     return status;
 }
 
+bool Comm::Req::Test(const std::string &hint)
+{
+    bool result = false;
+    if (m_Impl)
+    {
+        result = m_Impl->Test(hint);
+        m_Impl.reset();
```
eisenhauer: This bit looks suspicious:
```
       result = m_Impl->Test(hint);
        m_Impl.reset();
```
Shouldn't the .reset() be conditional?  I.E. if the Test returns false, don't you want to be able to test again?  Not reset the unique_ptr and drop the reference?

scottwittenburg: Yes I believe you're right, thanks for taking a careful look!

Using the behavior of `Wait()`, above, as a guide, I guess once the request is complete (i.e. `Test()` returns `true` here), we can reset the unique_ptr since the request has become "dead"?

In that case, inside the `m_Impl->Test()` definition, should I instead return `true` if there are no `MPI_Request` instances registered internally? Currently I'm returning `false` in that case, but inside `m_Impl->Wait()` a default-constructed status is returned immediately in the same situation, which would seem to indicate the request is "dead".

eisenhauer: I need to get back to this to look at what this is doing more broadly, but yeah, not immediately clear if the default case (no item) should be true, false, or something else.  Hope to look more this afternoon.

eisenhauer: I'm hating on the "Test" name, because it's so generic.  Test what?  TestActive?  TestComplete?  ugh.  It's MPI's fault, but I guess we could change the name in our wrapper to be TestComplete if we wanted a bit more clarity.

Anyhow I think the answer to both your questions is Yes.  Yes, you can reset the request once Test returns true.  Also, I think the m_Impl->Test() should return true if the are no requests.

One maybe minor nit.  Rather than a loop in m_Impl->Test(), you could use MPI_Testall() as something that would be atomic over all those requests.  I.E. something like:
```
CheckMPIReturn(MPI_Testall(m_MPIReqs.size(), m_MPIReqs.data(), &flag, MPI_STATUSES_IGNORE),                                                                              
                     hint); 
```

scottwittenburg: > I'm hating on the "Test" name, because it's so generic. Test what? TestActive? TestComplete? ugh. It's MPI's fault, but I guess we could change the name in our wrapper to be TestComplete if we wanted a bit more clarity.

I agree that as a name, `Test` leaves something to be desired. I like `TestComplete` since that's how it's currently behaving, but let's go with whatever you like best.

> Anyhow I think the answer to both your questions is Yes. Yes, you can reset the request once Test returns true. Also, I think the m_Impl->Test() should return true if the are no requests.

Ok thanks, I'll do that.

> One maybe minor nit. Rather than a loop in m_Impl->Test(), you could use MPI_Testall() as something that would be atomic over all those requests.

Will do.

scottwittenburg: Ok, I have made these changes, this is ready for another look.

eisenhauer:
OK, I think this looks good now.  I think you can remove the bit that forces rerouting and lets get this merged!

scottwittenburg:
@eisenhauer I think this is ready for approval/merge whenever you're ready.

eisenhauer (approved):

scottwittenburg:
Thank you for your help getting this merged @eisenhauer!

PR #4984, opened 2026-04-03
Rework derived variable compute functions and type system

eisenhauer:
Rewrote the operator stuff entirely with the idea of having vectorizable loops for both type-uniform operations and some common mixed-type operations (mixing float, double and long for example). This rewrite also directly handles constants in the express with vectorizable loops (not expanding constants to an array).  We also added a unified semantics pass that does C-style type promotion, constant folding strength reduction (x^2 → x*x), and validation.  More unusual type combinations work (uint_8 + double), but there is an additional type conversion pass inserted.

sqrt/trig/pow can now stay float, rather than always promoted float to double. Eliminated extra memory passes in the compute functions (old code zero-initialized then accumulated per-input) and removed the pool-to-caller memcpy in Execute.

This PR also adds sub-selection infrastructure (I.E. the ability to back-propagate read selections to the input selections that we'd need to produce it), but does not yet employ it to optimize read-side derived expressions.

### DIFF
--- a/source/adios2/core/IO.cpp
+++ b/source/adios2/core/IO.cpp
@@ -970,12 +970,12 @@ VariableDerived &IO::DefineDerivedVariable(const std::string &name, const std::s
                               (itVariable->second)->m_Shape}});
     }
 
-    // Build the expression code stream: GenerateCode -> ResolveTypes -> ConstantFold -> PlanBuffers
+    // Resolve types on tree, then generate code, then run linear passes
+    derived::ResolveTreeTypes(exprTree, name_to_type);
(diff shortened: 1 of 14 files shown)


anagainaru (dismissed):
Looks good, I will re-approve once the derived variables are turned off

eisenhauer:
> Looks good, I will re-approve once the derived variables are turned off

OK, I've removed the commit that turned them on, so they are back to defaulting to off.  (We should consider turning them on by default.  Perhaps something to talk about Tuesday.)

anagainaru (approved):

Issue #4502, opened 2025-04-04
Performance considerations for CPU-side Min/Max computation (StatsLevel=1)


franzpoeschel:
**Why is this feature important?**
Computation of Min/Max statistics on the CPU is considerably slower than on the GPU. Data production from the CPU is relevant since we need to transform our data before output and generally do not have GPU memory to spare for this.

Timings from the IO plugin of PIConGPU when creating a 34Gb checkpoint (single GPU run, no MPI parallelization in this test, time includes data preparation and output):

* **BP4, StatsLevel=0:** 1min 6sec 208msec
* **BP4, StatsLevel=1, Threads=1:** 1min 21sec 25msec
* **BP4, StatsLevel=1, Threads=8:** 1min 9sec 881msec
* **BP5, StatsLevel=0:** 1min 0sec 324msec
* **BP5, StatsLevel=1:** 1min 15sec 727msec

Detailed statistics (`profiling.json`) at the end.

**Describe the solution you'd like and potential required effort**

There were two suggestions on how to deal with this:

* By @pnorbert: Implement threaded CPU Min/Max computation in BP5 (feature already exists in BP4), this would enable PIConGPU users (and probably others) to have statistics with a reasonable overhead
* Some weeks ago, I talked to @anagainaru about an option to compute statistics only for GPU variables, i.e. something like `StatsLevel=GpuOnly`. This would allow us to specify statistics more selectively to only include low-overhead computations by default.
    **Possible downsides:** Users might be confused why some variables have statistics and others not.

**Describe alternatives you've considered and potential required effort**
We currently deactivate statistics by default in openPMD and users need to opt-in to using them.

**What is the potential impact of this feature in the community?**
PIConGPU users can activate statistics for ADIOS2 without compromising too much performance.

**Additional context**
PIConGPU uses the Span API for Put() operations.

```json
{
  "bp4_no_stats": [
    {
      "rank": 0,
      "start": "Fri_Apr_04_16:48:52_2025",
      "threads": 8,
      "bytes": 36223373054,
      "mkdir_mus": 74,
      "aggregation_mus": 0,
      "meta_sort_merge_mus": 226,
      "minmax_mus": 0,
      "memcpy_mus": 0,
      "buffering_mus": 14643165,
      "transport_0": {
        "type": "File_POSIX",
        "close_mus": 5,
        "write_mus": 22837467,
        "open_mus": 69
      },
      "transport_1": {
        "type": "File_POSIX",
[...]

PR #4985, opened 2026-04-05
vectorize min/max in BP5

eisenhauer:
See comments in issue #4502 for details.


pnorbert (dismissed):

eisenhauer:
@pnorbert, since you approved this has been changed to use SIMD on all types for which it is possible.  Prior version was just float and double.  Currently impossible types are 64-bit ints, long doubles, and complex.

pnorbert (approved):

PR #4986, opened 2026-04-06
minimum reads for reader-side derived variables

eisenhauer:
Details in issue #4453.


pnorbert (approved):

anagainaru (commented):
The derived variables were not turned off before merge

PR #4989, opened 2026-04-09
Evpath transport modules off

eisenhauer:
Always build-in EVPath transports so that they don't cause confusion, require a module path, etc.  This has EVPath updates to support this and some ATL updates as well.


thirdparty/atl/atl/http_atom_client.c
```diff
@@ -110,13 +109,31 @@ parse_url(const char *url)
     return 1;
 }
 
+/* Connect timeout in milliseconds.  Override with ATL_HTTP_TIMEOUT env var. */
+#define DEFAULT_CONNECT_TIMEOUT_MS 3000
```
pnorbert, line 113: Isn't 3s timeout too optimistic for WAN?

eisenhauer: It is kinda, but this is an odd case.  Too long to explain here, but the short version is that this is for an HTTP version of the server that ADIOS will never hit, and other changes here make sure that it will probably never be hit by anybody, so nobody should ever be making this connection, but if for some crazy reason they try I don't want them waiting long before failing...

pnorbert (approved):

PR #5015, opened 2026-04-22
Make MGARD lossless type configurable

rtobar:
This is enabled by a new `lossless_type` operator parameter that can strictly take the values `huffman` or `huffman_ztsd`. The default remains Huffman_Zstd for backwards compatibility.


eisenhauer:
I was going to call out the lack of update to the mgard page in docs, but it appears that we don't have one!  That would seem to be a problem, but it is not your problem to fix.  I'd say we should also allow the other values of lossless_type to be specified, but I'm happy to leave that until this gets documented...

eisenhauer (approved):

rtobar:
@eisenhauer thanks for the prompt review and merge. I didn't stop to think about the other values in the `lossless_type` enumeration, and you're right that they should have been considered. For the case in point we were only interested in these two. Thanks again!

PR #5021, opened 2026-04-27
Complete MGARD Lossless spec, do docs for MGARD only (not MGARDcomplex)

eisenhauer:


anagainaru (approved):
This looks good. I think in the future we should have a section "Operators" in the GPU section of the documentation as well.

PR #5029, opened 2026-04-29
backport: apply one by one commit

vicentebolea:


eisenhauer (approved):
So, happy to try this, but I asked Claude's opinion and it pooh-poohed the approach, saying the --mainline 1 will error out for non-merge commits.  It's suggestion was a loop that skips merge commits.  I don't know enough about this to know if it's wrong or not, but I thought I'd approve and let you try...

vicentebolea:
> So, happy to try this, but I asked Claude's opinion and it pooh-poohed the approach, saying the --mainline 1 will error out for non-merge commits. It's suggestion was a loop that skips merge commits. I don't know enough about this to know if it's wrong or not, but I thought I'd approve and let you try...

claude is wrong an example was the previous backport that failed, it was able to cherry-pick the first commit of a "Rebase" commit (no merge commit)

PR #5034, opened 2026-05-06
DAOS engine: per-rank array data path with async writes

eisenhauer:
Re-implement the DAOS engine around a per-rank DAOS array for data. Each writer rank issues async daos_array_write submits during AddToVec; EndStep drains and flushes any tail synchronously.  No aggregator.

Highlights:
  - DaosChunkV (new ChunkV subclass) issues async daos_array_write once a chunk-fill byte threshold is crossed, with a multi-iov SGL over a single contiguous array range.
  - ChunkV gains cross-step chunk recycling and an m_NoShrink flag for subclasses with in-flight async I/O against chunk memory. Stable chunk addresses keep the libfabric MR cache hot and eliminate per-step first-touch zeroing.
  - data_oids.txt at the dataset root maps writer rank to per-rank DAOS array OID; the reader parses it at Init.
  - daos_fini intentionally not called at Close: libdaos's reinit path after fini is unreliable.  Per-handle close is.

Per-rank metadata location is selected by the DAOS_METADATA_LAYOUT env var (default 'array-1mb-aligned'; also 'array' and 'kv'). Engine parameters narrowed to those actually used by the DAOS path. Profiling output is the standard <dataset>/profiling.json; BP5Serializer-side and DaosChunkV-side counters mirror into registered timers via SyncExternalCountersToProfiler.

Removes legacy data-path variants (TwoLevelShm, TwoLevelShm_Async, EveryoneWrites_Async) and their associated aggregator / burst- buffer / async-thread infrastructure.

Limitations: Append mode is rejected; Span Puts are unsupported.

Docs: developer_docs/daos_engine.md, user_guide engines/daos.rst.


eisenhauer:
I should add that this engine is far from complete.  Not only does it not implement a lot of ADIOS features yet, it also doesn't have things that you'd want in ADIOS DAOS support, like a way to turn a BP data stored in DAOS objects into a BP5 file-based representation so that you could, for example, ftp it elsewhere for analysis.  This is just a placeholder so that we can explore the performance envelope a bit.

pnorbert (approved):

PR #5035, opened 2026-05-09
XRootD HTTP: more path-encode changes

eisenhauer:
This mod completely gets rid of query arguments in favor of encoding everything in a couple of trailing (dummy) filespecs. That is, the URLs now look like this:
`/adios<filename>/<file-config>/<request>`
where file-config encodes the rowmajor order and any engine params:
```
 - r<digit> — RMOrder.  r0 = column-major, r1 = row-major.                                                                                   
 - p<base64url-blob> 
```
The request segment is:
  Single get: `g~<base64url-varname>~<paramstring>`
         where paramstring is something like "b0" for block 0, or "c4,8,12,o0,0,0" for a count and offset spec.
-or-
  Batch get: `b~N~<v1>~<p1>~<v2>~<p2>~...~<vN>~<pN>` 
         where each vi is base64url-encoded and each pi is a paramstring

### DIFF
--- a/scripts/docker/spin-xrootd/Dockerfile
+++ b/scripts/docker/spin-xrootd/Dockerfile
@@ -2,7 +2,7 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 
-FROM --platform=linux/amd64 fedora:latest@sha256:781b7642e8bf256e9cf75d2aa58d86f5cc695fd2df113517614e181a5eee9138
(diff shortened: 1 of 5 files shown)


eisenhauer:
@pnorbert The server on Spin accepts old and new query styles for now.  This update retires the old client style and switches to the new one unconditionally.  This is to make life easier with Pelican, functionality is identical.

pnorbert (approved):

PR #5047, opened 2026-05-20
Add OpenChain() to NullTransport so that BP5 can use it.

pnorbert:


vicentebolea:
looks good to me but Greg is more knowledgeable in these modules

eisenhauer (approved):
Looks fine, but I'm vaguely curious why the OpenChain isn't Null itself.

pnorbert:
I just wanted it to work the same way as in other transports, i.e. serialize the callers. Open() has some intializations to make sure the calls work as expected but I did not analyze what is really necessary and what is not.

PR #5048, opened 2026-05-26
Refactor to drop TransportMan where we can (not bp3/4)

eisenhauer:


source/adios2/engine/bp5/BP5Writer.cpp
```diff
@@ -530,7 +534,10 @@ void BP5Writer::WriteData_EveryoneWrites(format::BufferV *Data, bool SerializedW
     m_DataPos += Data->Size();
     std::vector<core::iovec> DataVec = Data->DataVec();
     AggTransportData &aggData = m_AggregatorSpecifics.at(GetCacheKey(m_Aggregator));
-    aggData.m_FileDataManager.WriteFileAt(DataVec.data(), DataVec.size(), m_StartDataPos);
+    for (auto &t : aggData.m_DataSubstreams)
+    {
+        t->WriteV(DataVec.data(), static_cast<int>(DataVec.size()), m_StartDataPos);
+    }
```
pnorbert: We can have >2GB blocks of data, can't we?

eisenhauer: Two issues here.  One is that I preserved the array of substreams abstraction that would be a part of the TransportManager even though there's never ever more than one.  So that needs to be fixed.  Honestly I'd love to fix it everywhere that we pass around vectors of things and only ever use [0], but that would be messing with BP3/4 also and I'd rather leave them alone.

But the cast to `int` here shouldn't be problem.  We can have huge blocks in our vector, but the int there is the number of entries in the vector, not their sizes (which is a size_t in the vector).

pnorbert (dismissed):

eisenhauer:
@pnorbert One more fix after eliminating the loop...

pnorbert (approved):

PR #5060, opened 2026-06-04
BP5/XRootD: identity/version/byte-order checks on remote reads

eisenhauer:
This PR is in some sense a roll-up of new information that may be (but likely currently isn't) part of an HTTPS request.  This items include : 
- the UUID of the file requested to ensure it matches what the remote has (omitted for UUID=0, which is true for all files except BP5 files newer than PR #5057).
- the byte order of the requesting client (omitted if little-endian).
- a request-version number (omitted if zero, which it currently is, but here to handle the inevitable changes).

### DIFF
--- a/source/adios2/engine/bp5/BP5Reader.cpp
+++ b/source/adios2/engine/bp5/BP5Reader.cpp
@@ -574,6 +574,9 @@ void BP5Reader::PerformGets()
                 params["SelectSteps"] = m_Parameters.SelectSteps;
             if (m_Parameters.IgnoreFlattenSteps)
                 params["IgnoreFlattenSteps"] = "true";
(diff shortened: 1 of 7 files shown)


eisenhauer:
The SPIN-deployed server is now running with these changes and it properly rejects queries to files with the right path but the wrong metadata UUID.  This, plus the in-CI XRootD tests (which will exercise the matching-UUID path).  Should be sufficient testing for now.  We've added a version number to the request to cover the circumstance where we change the format and want to throw errors on server/request mismatch (no changes anticipated yet).  We've also got the client byte-order in the request, should the world get more heterogeneous again...

pnorbert (approved):

Issue #5061, opened 2026-06-07
Generated `ADIOS2_FEATURE_LIST` causes compile issues


SunBlack:
In case no feature is enabled, the generated `ADIOS2_FEATURE_LIST` is:
```cpp
#define ADIOS2_FEATURE_LIST "BP3" "BP4" "BP5" nullptr
```

This causes:
```
<source>:3:47: error: expected '}' before 'nullptr'
    3 | #define ADIOS2_FEATURE_LIST "BP3" "BP4" "BP5" nullptr
      |                                               ^~~~~~~
<source>:8:42: note: in expansion of macro 'ADIOS2_FEATURE_LIST'
    8 |     const char *adios2_feature_list[] = {ADIOS2_FEATURE_LIST};
      |                                          ^~~~~~~~~~~~~~~~~~~
<source>:8:41: note: to match this '{'
    8 |     const char *adios2_feature_list[] = {ADIOS2_FEATURE_LIST};
      |                                         ^
<source>:3:47: note: probably missing a comma or an operator before
    3 | #define ADIOS2_FEATURE_LIST "BP3" "BP4" "BP5" nullptr
      |                                               ^~~~~~~
<source>:8:42: note: in expansion of macro 'ADIOS2_FEATURE_LIST'
    8 |     const char *adios2_feature_list[] = {ADIOS2_FEATURE_LIST};
      |                                          ^~~~~~~~~~~~~~~~~~~
```

In case a feature is present, e.g., HDF5, the generated code is:
```cpp
#define ADIOS2_FEATURE_LIST "BP3" "BP4" "BP5" "HDF5","HDF5_VOL",nullptr
```
This compiles], but not sure whether it gives the wanted result as following code 
```cpp
#include <iostream>

#define ADIOS2_FEATURE_LIST "BP3" "BP4" "BP5" "HDF5","HDF5_VOL",nullptr

int main()
{

    const char *adios2_feature_list[] = {ADIOS2_FEATURE_LIST};

    for (auto feature : adios2_feature_list) {
        std::cout << feature << " \n";
    }    
}
```
produces
```
BP3BP4BP5HDF5 
HDF5_VOL 
```

I think this line:
https://github.com/ornladios/ADIOS2/blob/f5267290f06980acaecaf54688d0980958eb86bf/source/adios2/common/ADIOSConfig.h.in#L45
 should be changed to:
```cpp
[...]

PR #5062, opened 2026-06-08
Fix ADIOS2_FEATURE_LIST generation and engine reporting

eisenhauer:
Fix the adios_feature_list bug reported in #5061, and while we're messing with that:
- Use the original mixed-case option names in the generated feature list so they match the CMake spelling (DataMan, not DATAMAN).
- The DataMan/DataSpaces engine guards checked ADIOS2_HAVE_DataMan/ DataSpaces, but the generated macros are uppercased, so those engines never appeared in adios2_available_engines even when built.
- Strengthen the Info tests to assert features/engines are distinct entries and that a built engine is actually reported.


cmake/ADIOSFunctions.cmake
```diff
@@ -109,7 +109,7 @@ function(GenerateADIOSHeaderConfig)
 ")
     if(ADIOS2_HAVE_${OPT})
       set(ADIOS2_HAVE_${OPT_UPPER} 1)
-      string(APPEND ADIOS2_CONFIG_FEATURE_LIST "\"${OPT_UPPER}\",")
+      string(APPEND ADIOS2_CONFIG_FEATURE_LIST "\"${OPT}\",")
```
SunBlack: Just in case you want to unify the spacing between default present items and the auto-generated:
```suggestion
      string(APPEND ADIOS2_CONFIG_FEATURE_LIST "\"${OPT}\", ")
```

eisenhauer: Agree, that would be better...

anagainaru (approved):

vicentebolea (approved):

PR #5068, opened 2026-06-10
ci make windows 2025 to use vs2026

vicentebolea:

PR #5070, opened 2026-06-10
ci: keep using cibuildwheel v3 to avoid error

vicentebolea:


vicentebolea:
Combined into #5071 along with #5068.

PR #5078, opened 2026-06-12
Remote: XrdCl HTTP backend

eisenhauer:
The goal here is to restructure the HTTP remote-read path so that the previous libcurl interaction with our XRootD server is just one possibility, and to add an XrdCl-based approach alongside it. Most of the changes are in service of that goal, moving code rather than changing it: the curl-multi machinery comes out of XrootdHttpRemote verbatim into a LibcurlBackend behind a small RemoteHttpBackend interface, while URL building and response parsing stay in XrootdHttpRemote where both backends share them. The libcurl path is largely unchanged.

The XrdCl backend is new, and it's deliberately less capable and complete than the libcurl one. It fetches whole responses through XrdCl::File with the xrdcl-curl plugin in full_download mode (our origin computes a complete body per GET, with no HEAD or range support, so that's the mode that fits).  Honestly I'm not sure this is more than a starting point for Pelican collab, but hopefully it is at least that.

Selecting it: transfer_protocol "xrdcl" on an XRootD entry in ~/.config/hpc-campaign/hosts.yaml, alongside the existing xrootd/http/https values, reached as usual through the reader's RemoteHost parameter. For testing and development there's also a DoXRootDXrdCl environment variable (with XRootDHttpsHost giving host:port), following the existing DoXRootDHttps pattern. libcurl remains the default for the http and https lanes. XrootdHttpRemote now builds when either CURL or XRootD is available, and asking for a backend whose library wasn't built produces an error at Open.

There are server-side changes to accommodate XrdCl-style clients, which append opaque query parameters to request URLs (xrdclcurl.timeout=..., authz=...). The handler now distinguishes the legacy query form from the path-encoded form by the query's first token rather than by the mere presence of a '?'. While in there, the server also now rejects responses over 2 GiB with a clear error instead of silently truncating them (the SSI response API takes an int length; lifting that limit needs the streaming path we don't have yet).

There is a standard set of tests routing through XrdCl.  I've also added a new regression test that issues v2.12.0-style query-string requests directly and verifies the returned bytes.  This can be removed when we think 2.12 clients are sufficiently far in the past

### DIFF
--- a/source/adios2/CMakeLists.txt
+++ b/source/adios2/CMakeLists.txt
@@ -267,15 +267,23 @@ if(ADIOS2_HAVE_OpenSSL)
 endif()
 
 if(ADIOS2_HAVE_XRootD)
-  target_sources(adios2_core PRIVATE toolkit/remote/XrootdRemote.cpp)
+  # Native XRootD (XrdSsi) remote, plus the XrdCl HTTP-lane backend (uses libXrdCl,
+  # the CLIENT component, to reach a Pelican federation or an HTTPS origin)
+  target_sources(adios2_core PRIVATE toolkit/remote/XrootdRemote.cpp toolkit/remote/XrdClBackend.cpp)
   target_include_directories(adios2_core PRIVATE ${XROOTD_INCLUDE_DIRS})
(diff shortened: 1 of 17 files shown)


eisenhauer:
Note that TLS is still a problem and means that we cannot run XrdCl against the Spin server as XrdCl has no way to force it ignore bad or missing TLS.

pnorbert (approved):

PR #5089, opened 2026-06-16
cmake: find ZeroMQ via pkg-config instead of FindZeroMQ.cmake

### LINKED ISSUE #5088: Use config mode to find ZeroMQ
Currently when you have enabled `ADIOS2_USE_ZeroMQ` you will get:
```
C:/dev/vcpkg_dev/buildtrees/adios2/src/v2.12.1-ca42063886/cmake/FindZeroMQ.cmake(21):  file(STRINGS C:/dev/vcpkg_dev/installed/x64-windows/include/zmq.h _ver_strings REGEX ZMQ_VERSION_[^ ]* [0-9]+ )
CMake Warning (dev) at cmake/FindZeroMQ.cmake:21 (file):
  Policy CMP0159 is not set: file(STRINGS) with REGEX updates
  CMAKE_MATCH_<n>.  Run "cmake --help-policy CMP0159" for policy details.
  Use the cmake_policy command to set the policy and suppress this warning.

  For compatibility, CMake is leaving CMAKE_MATCH_<n> unchanged.
  C:/dev/vcpkg_dev/installed/x64-windows/share/zeromq/vcpkg-cmake-wrapper.cmake:1 (_find_package)
  C:/dev/vcpkg_dev/scripts/buildsystems/vcpkg.cmake:864 (include)
  cmake/DetectOptions.cmake:344 (find_package)
  CMakeLists.txt:216 (include)
This warning is for project developers.  Use -Wno-dev to suppress it.
```

Instead of fixing this warning, you could also remove the find script, as ZeroMQ exports it's [CMake targets](https://github.com/zeromq/libzmq/commit/7c709e546637c3058012e6f5f6a46bd68f41ba45) since v4.2.2 (I think it is okay to raise the minimum required version from 4.1 to 4.2.2 as this version is also already 9 years old). This would also fix the issue, that the current script doesn't work with vcpkg, as the lib contains the version number (`libzmq-mt-4_3_5.lib` ) and is therefore not found by fins_library.

So basically:
- Remove `FindZeroMQ.cmake`
- Replace [`find_package(ZeroMQ 4.1)`](https://github.com/search?q=repo%3Aornladios%2FADIOS2%20find_package(zeromq&type=code)) by `find_package(ZeroMQ CONFIG)`
- Replace `ZeroMQ::ZMQ` by `libzmq`

eisenhauer:
Fixes the CMP0159 warning and vcpkg's decorated-library-name failure. A CMake config ships only with CMake-built libzmq (vcpkg), not autotools ones (Homebrew), so pkg-config is the portable choice.   Cleanup some other zmq things while we're editing here.

Fixes #5088


vicentebolea:
It seems that this might cause errors in linux when there is a system packaged zeromq with a pkg-config pc file but we want to use another zeromq that do not bring the pc file such as in Spack.

eisenhauer:
So I see!  Well, trying another with backup...

cmake/DetectOptions.cmake
```diff
@@ -337,11 +337,30 @@ if(MPI_FOUND)
...
+    else()
+        find_path(ZeroMQ_INCLUDE_DIR NAMES zmq.h)
+        find_library(ZeroMQ_LIBRARY NAMES zmq)
+        if(ZeroMQ_INCLUDE_DIR AND ZeroMQ_LIBRARY)
+            set(ZeroMQ_FOUND TRUE)
+            add_library(ZeroMQ::ZeroMQ UNKNOWN IMPORTED)
+            set_target_properties(ZeroMQ::ZeroMQ PROPERTIES
+                IMPORTED_LOCATION "${ZeroMQ_LIBRARY}"
+                INTERFACE_INCLUDE_DIRECTORIES "${ZeroMQ_INCLUDE_DIR}")
+            set(ADIOS2_ZeroMQ_TARGET ZeroMQ::ZeroMQ)
```
```suggestion
        find_package(ZeroMQ 4.1 REQUIRED)
```

vicentebolea:
Lets also depend zeroMQ>=4.2.2 (9 years old) which provides the cmake.config file. @eisenhauer would it be better to patch the homebrew recipe to install the cmake config file?

vicentebolea:
@eisenhauer thinking about it a bit more perhaps the best route is to modify the findzeromq.cmake to solely do the pkg-config package search.

eisenhauer:
Never mind, I was reading too quickly and misinterpreted.  However I'd like to raise another issue with DataMan, our only consumer of ZMQ.  That engine has been orphaned since Jason left (in 2022?) and is unlikely to be used in any new situation.  I guess this doesn't fix the issue, but I don't think DataMan is worth the time we're spending here.

vicentebolea:
> Never mind, I was reading too quickly and misinterpreted. However I'd like to raise another issue with DataMan, our only consumer of ZMQ. That engine has been orphaned since Jason left (in 2022?) and is unlikely to be used in any new situation. I guess this doesn't fix the issue, but I don't think DataMan is worth the time we're spending here.

My is idea is that the findZeroMQ.cmake can be repurpose just to find zeromq (new versions too) with pkg-config. in that manner when we do find_package(zeromq), it will first try to find the pkg-config one and if fails than the cmake.config file. We can reverse this if needed. This seems to me to be a clean approach.

SunBlack:
I haven't tested the current version yet, but it doesn't follow modern CMake best practices. The problem with `find_library` and `find_path` is that they don't detect compiler definitions and also ignore dependencies of the library being searched for. That's why this approach is no longer recommended. Because of the indirect dependencies, you should also run pkg-config in the appropriate mode.

Untested:

```cmake
find_package(ZeroMQ ${ZeroMQ_FIND_VERSION} CONFIG
                     ${ZeroMQ_FIND_VERSION_EXACT}
                     ${ZeroMQ_FIND_QUIET}
                     ${ZeroMQ_FIND_REQUIRED})

# If native package was found, stop here
if(ZeroMQ_FOUND)
  return()
endif()

# pkg-config fallback
find_package(PkgConfig REQUIRED)

# Version handling
if(ZeroMQ_FIND_VERSION)
  if(ZeroMQ_FIND_VERSION_EXACT)
    pkg_check_modules(ZeroMQ REQUIRED IMPORTED_TARGET libzmq=${ZeroMQ_FIND_VERSION})
  else()
    pkg_check_modules(ZeroMQ REQUIRED IMPORTED_TARGET libzmq>=${ZeroMQ_FIND_VERSION})
  endif()
else()
  pkg_check_modules(ZeroMQ REQUIRED IMPORTED_TARGET libzmq)
endif()

# Create alias target to match imported target name of the native package
add_library(ZeroMQ::ZeroMQ ALIAS PkgConfig::ZeroMQ)
```

PR #5091, opened 2026-06-17
Modernize FindZeroMQ.cmake and fix some ZeroMQ related compile bugs

### LINKED ISSUE #5088: Use config mode to find ZeroMQ

SunBlack:
Supersedes #5089 and fixes #5088.

Since the discussion in #5089 showed that `pkg conf` is necessary, and therefore a separate `find` script is still required, I tested this locally with vcpkg and ran into several other issues with ZeroMQ, so it’s probably worth creating a separate PR after all.

Changes:
- Modernized CMake script to use always imported targets (and therefore get also retrieve maybe necessary compiler flags and other indirect dependencies)
- Didn't raised minimum required version to 4.2.2 - didn't checked whether 4.1 exports pkg files
- To prevent recursive calls, added `MODULE` and `CONFIG` to the `find_package` calls
- Removed alternative operator representations, as they caused compile issues with MSVC 2026 (and are deprecated [anyway](https://en.cppreference.com/cpp/header/ciso646))

Notes:
- Didn't verified the pkg path
- Is it okay to add pkgconf to `scripts/ci/images/Dockerfile.ci-spack-ubuntu22.04-base` or is there really a another legacy path necessary?


cmake/FindZeroMQ.cmake
```diff
@@ -2,49 +2,51 @@
...
+find_package(ZeroMQ ${ZeroMQ_FIND_VERSION} ${_zmq_version_exact} ${_zmq_quiet} CONFIG)
+
+# If native package was found, stop here
+if(TARGET libzmq)
+  add_library(ZeroMQ::ZMQ ALIAS libzmq)
+  return()
+endif()
+
+# pkg-config fallback
+find_package(PkgConfig ${_zmq_quiet} ${_zmq_required})
```
vicentebolea: find_dependency will avoid having to explicitly forward these flags.

cmake/FindZeroMQ.cmake
```diff
@@ -2,49 +2,51 @@
...
-        INTERFACE_INCLUDE_DIRECTORIES "${ZeroMQ_INCLUDE_DIR}"
-      )
-    endif()
+find_package(ZeroMQ ${ZeroMQ_FIND_VERSION} ${_zmq_version_exact} ${_zmq_quiet} CONFIG)
+
+# If native package was found, stop here
+if(TARGET libzmq)
+  add_library(ZeroMQ::ZMQ ALIAS libzmq)
+  return()
+endif()
```
vicentebolea: Let's move the CONFIG find this to the detect options, in that manner it remains consistent with the other deps. There if it fails you can invoke this MODULE find.

cmake/FindZeroMQ.cmake
```diff
@@ -2,49 +2,51 @@
...
-      set_target_properties(ZeroMQ::ZMQ PROPERTIES
-        IMPORTED_LOCATION             "${ZeroMQ_LIBRARY}"
-        INTERFACE_INCLUDE_DIRECTORIES "${ZeroMQ_INCLUDE_DIR}"
-      )
-    endif()
+find_package(ZeroMQ ${ZeroMQ_FIND_VERSION} ${_zmq_version_exact} ${_zmq_quiet} CONFIG)
+
+# If native package was found, stop here
+if(TARGET libzmq)
+  add_library(ZeroMQ::ZMQ ALIAS libzmq)
```
vicentebolea: lets just use libzmq thru adios2? Looking at libzmq source code it seems that they do not namespace their targets in the config cmake file.

cmake/FindZeroMQ.cmake
SunBlack: True, always forgetting this command :-/

cmake/FindZeroMQ.cmake
vicentebolea: Thanks for the changes, what I meant is to move the CONFIG find to the `DetectOptions.cmake`

SunBlack: Not sure whether this is a good idea, as there are always more duplication in the `DetectOptions.cmake` due to the `AUTO/ON` paths:
```cmake
if(ADIOS2_USE_ZeroMQ STREQUAL AUTO)
    find_package(ZeroMQ 4.1 CONFIG)
    if (NOT ZeroMQ_FOUND)
      find_package(ZeroMQ 4.1 MODULE)
	endif()
elseif(ADIOS2_USE_ZeroMQ)
    find_package(ZeroMQ 4.1 CONFIG REQUIRED)
    if (NOT ZeroMQ_FOUND)
      find_package(ZeroMQ 4.1 MODULE REQUIRED)
	endif()
endif()
```

Another downside: The (3) examples could not use the Config code or need also the duplicated paths there

vicentebolea: The reason to have this in detectoptions.cmake is keep that file consistent, the dance of find_package is common there.

> Another downside: The (3) examples could not use the Config code or need also the duplicated paths there

Can you develop this point?

Thanks!

SunBlack: I do think it's poor design to move it to DetectOptions, since that means anyone looking for ZeroMQ has to implement both variants instead of having a central location, but if that's what you want...

The examples don't actually need ZeroMQ, since the target is only linked privately.

cmake/DetectOptions.cmake
```diff
@@ -339,9 +339,15 @@ endif()
...
-    find_package(ZeroMQ 4.1)
+  find_package(ZeroMQ 4.1)
+  if(NOT ZeroMQ_FOUND)
+    find_package(ZeroMQPkg 4.1)
+  endif()
 elseif(ADIOS2_USE_ZeroMQ)
-    find_package(ZeroMQ 4.1 REQUIRED)
+  find_package(ZeroMQ 4.1)
+  if(NOT ZeroMQ_FOUND)
+    find_package(ZeroMQPkg 4.1 REQUIRED)
```
vicentebolea: No need for the rename, you can use find_package(... MODULE)

cmake/FindZeroMQ.cmake
vicentebolea: > The examples don't actually need ZeroMQ, since the target is only linked privately.

This is true since we find_dependency(zeroMq) in the adios2-config.cmake.

cmake/DetectOptions.cmake
SunBlack: I know. I put that there on purpose to make it clear that it's only intended for the pkg variant. If the module is ever exported and someone runs `find_package(ZeroMQ)`, they might not realize that it's only meant for that one variant.

vicentebolea: This add extra complexity, please keep it simple using the module is more succinct.

vicentebolea (approved):
Thanks, looks good! sorry for the delay.

vicentebolea (changes requested):
We are hitting some configure warnings that indicate the zeromq is not being picked up:

```

-- Found MPI_C: /spack/var/spack/environments/adios2-ci-mpich/.spack-env/view/lib/libmpi.so (found version "4.1")
-- Found MPI_CXX: /spack/var/spack/environments/adios2-ci-mpich/.spack-env/view/lib/libmpicxx.so (found version "4.1")
-- Found MPI_Fortran: /spack/var/spack/environments/adios2-ci-mpich/.spack-env/view/lib/libmpifort.so (found version "4.1")
-- Found MPI: TRUE (found version "4.1") found components: C Fortran CXX
CMake Warning at cmake/DetectOptions.cmake:347 (find_package):
  By not providing "FindZeroMQ.cmake" in CMAKE_MODULE_PATH this project has
  asked CMake to find a package configuration file provided by "ZeroMQ", but
  CMake did not find one.

  Could not find a package configuration file provided by "ZeroMQ" (requested
  version 4.1) with any of the following names:

    ZeroMQConfig.cmake
    zeromq-config.cmake

  Add the installation prefix of "ZeroMQ" to CMAKE_PREFIX_PATH or set
  "ZeroMQ_DIR" to a directory containing one of the above files.  If "ZeroMQ"
  provides a separate development package or SDK, be sure it has been
  installed.
  CMakeLists.txt:216 (include)


-- Checking for module 'libzmq'
--   Found libzmq, version 4.3.5
CMake Warning (dev) at /usr/local/spack-tools/install/linux-centos7-x86_64/gcc-10.2.1/cmake-3.31.0-catu6wyzxyt5rool7lhqc3wok3kzp2y6/share/cmake-3.31/Modules/FindPackageHandleStandardArgs.cmake:441 (message):
  The package name passed to `find_package_handle_standard_args` (ZeroMQ)
  does not match the name of the calling package (ZeroMQPkg).  This can lead

```

SunBlack:
> We are hitting some configure warnings that indicate the zeromq is not being picked up:

It is - it is just the warning, that the Config-Search didn't found anything (see line hint), but the Module-Call below will (`Found libzmq, version 4.3.5`). This is the result of choosing to encapsulate the config variant and the pkgconf variant outside the FindZeroMQ module rather than within it. This leads to the following effect:
- `find_package(ZeroMQ CONFIG)` Will fail with the warning you quoted in case it fails, otherwise it will print the found version
- `find_package(ZeroMQ CONFIG QUIET)` Will silently fail, but also not print the found version found in case it succeeded

vicentebolea:
> > We are hitting some configure warnings that indicate the zeromq is not being picked up:
> 
> It is - it is just the warning, that the Config-Search didn't found anything (see line hint), but the Module-Call below will (`Found libzmq, version 4.3.5`). This is the result of choosing to encapsulate the config variant and the pkgconf variant outside the FindZeroMQ module rather than within it. This leads to the following effect:
> 
>     * `find_package(ZeroMQ CONFIG)` Will fail with the warning you quoted in case it fails, otherwise it will print the found version
> 
>     * `find_package(ZeroMQ CONFIG QUIET)` Will silently fail, but also not print the found version found in case it succeeded

No, this PR is adding a regression that results in not finding zeromq as opposed to the baseline (current master). You can check this in the gcc9 serial job

vicentebolea:
> find_package(ZeroMQ CONFIG QUIET)`

Disregard my previous comment, further looking at the logs I can see that it found it thru the module. Please add the param QUIET to the CONFIG find_package() call

cmake/DetectOptions.cmake
```diff
@@ -339,9 +339,15 @@ endif()
 
 # ZeroMQ
 if(ADIOS2_USE_ZeroMQ STREQUAL AUTO)
-    find_package(ZeroMQ 4.1)
+  find_package(ZeroMQ 4.1 CONFIG)
```
```suggestion
  find_package(ZeroMQ 4.1 CONFIG QUIET)
```

PR #5096, opened 2026-06-19
cmake: honor the ADIOS2_USE_SysVShMem option

### LINKED ISSUE #5095: Unused option `ADIOS2_USE_SysVShMem`
Here `ADIOS2_USE_SysVShMem` will be defined:
https://github.com/ornladios/ADIOS2/blob/1f6446c1d001048969298ca5189e342c4867560b/CMakeLists.txt#L196

But this will be [never used](https://github.com/search?q=repo%3Aornladios%2FADIOS2+SysVShMem&type=code) - instead it will be automatically detected:
https://github.com/ornladios/ADIOS2/blob/1f6446c1d001048969298ca5189e342c4867560b/cmake/DetectOptions.cmake#L617-L627

eisenhauer:
It was defined but never read, so SysV shared memory was force-enabled wherever shmget exists. Now OFF disables it, per the usual tri-state.

Closes #5095


cmake/DetectOptions.cmake
```diff
@@ -614,16 +614,17 @@ elseif(ADIOS2_USE_DAOS)
 endif()
 
 #SysV IPC
-if(UNIX)
-  include(CheckSymbolExists)
-  CHECK_SYMBOL_EXISTS(shmget "sys/ipc.h;sys/shm.h" HAVE_shmget)
+set(ADIOS2_HAVE_SysVShMem OFF)
+if(ADIOS2_USE_SysVShMem STREQUAL AUTO OR ADIOS2_USE_SysVShMem)
```
```suggestion
if(ADIOS2_USE_SysVShMem)
```

eisenhauer:
Accepted the suggestion.  Always a little leary of relying upon CMake's true/false interpretation when 'auto' is an option.

vicentebolea (approved):

PR #5099, opened 2026-06-22
Make reader engine selection collective and fail cleanly

### LINKED ISSUE #5098: Some exceptions cause non-root nodes to hang
In my application (a distributed library), I accept a string from a user which ultimately informs the argument to `adios2::IO.Open`, in read-mode. The user can pass invalid strings, such as the name of a non-existent file/directory, which will make `Open` throw an exception. I must gracefully handle this, reporting an error message, and perform MPI finalisation. My application's unit tests even include negative tests which deliberately trigger these exceptions.

But alas, ADIOS2 appears to irrevocably hang non-root nodes when encountering an exception! For example, consider C++ MWE
```cpp
#include <iostream>
#include <adios2.h>
#include <mpi.h>


int main() {

    // Prepare MPI-aware ADIOS
    MPI_Init(NULL, NULL);
    auto adios = adios2::ADIOS(MPI_COMM_WORLD);
    auto io = adios.DeclareIO("Foo");

    // Print on every rank
    int rank;
    MPI_Comm_rank(MPI_COMM_WORLD, &rank);
    std::cout << "rank " << rank << " is about to call Open..." << std::endl;
    std::cout.flush();
    MPI_Barrier(MPI_COMM_WORLD);

    // Try to io.Open with every node
    bool localSuccess = false;
    adios2::Engine engine;
    try {
        engine = io.Open("NonExistentFile.bp", adios2::Mode::Read);
        localSuccess = true;
    } catch (...) {
        std::cout << "rank " << rank << " encountered an exception" << std::endl;
        std::cout.flush();
    }

    // Non-root ranks are now permanently hung!
    std::cout << "rank " << rank << " passed the try/except" << std::endl;
    std::cout.flush();

    // It now seems impossible to recover non-root nodes from the ADIOS2 error,
    // which we might have imagined by doing something like this:
    int globalSuccess;
    MPI_Allreduce(&localSuccess, &globalSuccess, 1, MPI_INT, MPI_LAND, MPI_COMM_WORLD);
    std::cout << "rank " << rank << " is aware of global success = " << globalSuccess << std::endl;

    MPI_Finalize();
    return 0;
}
```
When run with four nodes, this will output:
```
rank 1 is about to call Open...
rank 0 is about to call Open...
rank 3 is about to call Open...
rank 2 is about to call Open...
rank 0 encountered an exception
rank 0 passed the try/except

< and we're now hung >
```
This is because only the **root node** of the communicator will ever continue execution from:
```cpp
io.Open("NonExistentFile.bp", adios2::Mode::Read);
```
The non-root nodes forever hang, probably waiting for some MPI collective operation that is never fulfilled due to the root-node exception.

Is this intentional? It seems an incredible foot-gun and suggests it is _impossible_ to gracefully recover from an ADIOS2 exception. My only option appears to be to call `MPI_Abort` from root, killing the entire distributed application, but this means I cannot ever perform a negative test of this exception.

Is there any way to validate that a series of ADIOS2 functions will complete without a hang, before actually performing them? It's insufficient to just check the file/directory `"NonExistentFile.bp` exists, since that obviously does not guarantee ADIOS2 parsing will succeed.

## Build info

Just in case this behaviour is unexpected my CMake build disables almost everything except `USE_MPI`, and I am running on MacOS with MPI compiler `Apple clang version 16.0.0`.
```cmake
    # Use MPI-aware ADIOS
    set(ADIOS2_USE_MPI  ON CACHE BOOL "" FORCE)
   
    # Forego unused facilities
    set(ADIOS2_USE_CUDA       OFF CACHE BOOL "" FORCE)
    set(ADIOS2_BUILD_TESTING  OFF CACHE BOOL "" FORCE)
    set(ADIOS2_BUILD_EXAMPLES OFF CACHE BOOL "" FORCE)
    set(ADIOS2_USE_SODIUM     OFF CACHE BOOL "" FORCE)
    set(ADIOS2_USE_Fortran    OFF CACHE BOOL "" FORCE)
    set(ADIOS2_USE_HDF5       OFF CACHE BOOL "" FORCE)
    set(ADIOS2_USE_ZeroMQ     OFF CACHE BOOL "" FORCE)
    set(ADIOS2_USE_SST        OFF CACHE BOOL "" FORCE)
    set(ADIOS2_USE_DataMan    OFF CACHE BOOL "" FORCE)
    set(ADIOS2_USE_SSC        OFF CACHE BOOL "" FORCE)
    set(ADIOS2_USE_MHS        OFF CACHE BOOL "" FORCE)
    set(ADIOS2_USE_DAOS       OFF CACHE BOOL "" FORCE)
    set(ADIOS2_USE_MGARD      OFF CACHE BOOL "" FORCE)
    set(ADIOS2_USE_BZip2      OFF CACHE BOOL "" FORCE)
    set(ADIOS2_USE_Blosc      OFF CACHE BOOL "" FORCE)
    set(ADIOS2_USE_Blosc2     OFF CACHE BOOL "" FORCE)
    set(ADIOS2_USE_SZ         OFF CACHE BOOL "" FORCE)
    set(ADIOS2_USE_ZFP        OFF CACHE BOOL "" FORCE)
    set(ADIOS2_USE_PNG        OFF CACHE BOOL "" FORCE)
    set(ADIOS2_USE_Profiling  OFF CACHE BOOL "" FORCE)
    set(ADIOS2_USE_Python     OFF CACHE BOOL "" FORCE)
```
with ADIOS2 `v2.12.1`.

eisenhauer:
Clean up reader-side engine selection to 1) eliminate paths where very rank probed the filesystem, and 2) provide a more reasonable failure for non-existant files.  Add a quick test and a whatsnew entry.

Closes #5098

### DIFF
--- a/docs/user_guide/source/introduction/whatsnew.rst
+++ b/docs/user_guide/source/introduction/whatsnew.rst
@@ -26,6 +26,16 @@ only the **BP5** reader supports them, and only when opened in
 ``Mode::ReadRandomAccess`` with a reentrant file transport (POSIX). See the
 Selection chapter for full details.
 
+Collective reader engine selection
+----------------------------------
+
+When a reader does not set an engine explicitly, the engine type is
+auto-detected from the filesystem. That probing now happens on rank 0 and is
+broadcast, instead of every rank issuing its own ``stat()`` calls. If nothing
+exists at the given location, ``IO::Open`` now fails collectively with a clear
+error (and a hint to call ``SetEngine`` for a stream that has not been created
+yet) rather than selecting BP3 and hanging the non-root MPI ranks.
+
 SST Mercury Data Plane
 -----------------------
 
--- a/source/adios2/core/IO.cpp
+++ b/source/adios2/core/IO.cpp
@@ -12,7 +12,9 @@
 #include <memory>
 #include <mutex>
 #include <sstream>
+#include <stdexcept>
 #include <utility> // std::pair
+#include <vector>
 
 #include "adios2/common/ADIOSMacros.h"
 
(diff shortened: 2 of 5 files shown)


pnorbert (approved):

TysonRayJones:
That's an impressively fast solution! 👏

eisenhauer:
> That's an impressively fast solution! 👏

I had noticed that engine selection was a bit of a mess when adding the case for the DAOS engine a month or so ago.  Just decided this was better fixed sooner than later...

PR #5104, opened 2026-06-27
BP5: 32 Bit Crashes Fixed, 32bit-32bit Round-Trips

ax3l:
Ensures BP5 can round-trip writes & reads on 32bit. So far, ADIOS2 BP5 simply segfaulted when used on 32bit systems (tested on Linux, Windows).
Does not yet support cross-bitness portability to 64bit written hosts and back.

Credits to Codex + Claude for the typing.

cc @eisenhauer @pnorbert

### DIFF
--- a/source/adios2/common/ADIOSTypes.h
+++ b/source/adios2/common/ADIOSTypes.h
@@ -263,7 +263,7 @@ constexpr bool end_step = true;
 constexpr bool LocalValue = true;
 constexpr bool GlobalValue = false;
 
-constexpr size_t UnknownStep = MaxU64;
+constexpr size_t UnknownStep = std::numeric_limits<size_t>::max();
 constexpr double UnknownTime = std::numeric_limits<double>::infinity();
 
 // Dims, Params, vParams are defined in ADIOSBaseTypes.h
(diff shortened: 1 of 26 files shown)


eisenhauer:
I've been thinking that coding agents might finally make 32-bit support, and maybe even interoperability, practical...

ax3l:
I was thinking the same. I have a bitop commit locally that looks similar + a bit wild that I can submit separately if you like 😅

eisenhauer:
> I was thinking the same. I have a bitop commit locally that looks similar + a bit wild that I can submit separately if you like 😅

Submit the bitop fix.  (Assume this is the BitField operations, which aren't written in a portable way for 32-bit.)  That's a nice separable thing that we can change without dealing with more global interoperability concerns...

ax3l:
@eisenhauer the 32 bit round-trip works in this PR and CI seems happy/green (for all the existing 64 bit systems) :)

Want me to push the 32<->64bit BP5 round-trip here or in a follow-up PR? I prefer a follow-up, because it will need a bit more logic/design discussions: #5121

eisenhauer:
Sorry I haven't gotten to digging into this yet, but soon!

pnorbert (approved):

PR #5121, opened 2026-07-13
BP5: 32<->64 Bit roundtrip

ax3l:
Portability: Ensure that BP5 files can be read and write, even if the producer is differing, e.g., from/to a 32/64bit system. Intentionally allow to truncate data fields >4GB on 32bit systems.

- [ ] rebase after #5104

cc @eisenhauer @pnorbert


eisenhauer:
@ax3l and @pnorbert,  I think this is ready for review.  It's a bit of a compromise.  The core idea is that we try to stay 64-bit in ADIOS core, while preserving size_t where it already exists in the bindings.  I think that gives us the most operational capacity (e.g. processing 64-bit sizes and offsets) with the least disruption.  I do have 32-bit output files in the archive, so we will know if we break the 64-bit version's ability to read those.  However, this doesn't include any 32-bit testing.  @ax3l, Norbert mentioned that you might be aiming at webassembly with this, rather than say a 32-bit host environment.  Is that the case?  Just thinking that we probably need to add a test environment, and the application domain that you're targeting might influence that choice...

pnorbert (approved):

PR #5122, opened 2026-07-13
Test: Attribute-Only Steps

ax3l:
Works well for both constant and modifyable attributes in BP5.

BP4 and BP3 do, as expected, only read back one step.

cc @pnorbert @franzpoeschel


eisenhauer (approved):

eisenhauer:
@pnorbert, I tweaked this, so now my approval is insufficient to merge...

pnorbert (approved):

PR #5124, opened 2026-07-14
Python docs: read semantics and deferred reads

eisenhauer:
Document some things that IMHO were unclear in the python API and give deferred reading more prominence in examples and in the interface (partly with the idea that coding agents will pick up these notes and be more likely to use the calls properly).

Read vs ReadRandomAccess semantics (Stream mode table, first __init__ docstring, step_selection is rra-only, FileReader loads all metadata at open) and when to use defer_read/read_complete, with a deferred-read example in the user guide.

### DIFF
--- a/docs/user_guide/source/api_python/python_example.rst
+++ b/docs/user_guide/source/api_python/python_example.rst
@@ -198,12 +198,12 @@ Python Read Random Access example
 Python Deferred (batched) Read example
 --------------------------------------
 
-Each synchronous ``read()`` call is a separate request, and when the data is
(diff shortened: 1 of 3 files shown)


pnorbert:
Thanks, I like this. But note that deferred read is not only for remote read. I can get 3-4x better read from my nvme locally by reading multiple variables at once (threading in BP5). AI should understand that too.

eisenhauer:
good point, reworded so the local threaded-read win is also included.

pnorbert (approved):

PR #5132, opened 2026-07-24
Support reading HDF5 files from TAR file.

pnorbert:
Refactored the remote connection preparation code from BP5Reader into Remote functions.

### DIFF
--- a/source/adios2/toolkit/remote/Remote.h
+++ b/source/adios2/toolkit/remote/Remote.h
@@ -20,11 +20,21 @@
 
 namespace adios2
 {
+
+struct RemoteSetup
+{
+    std::string hostName;
+    adios2::HostAccessProtocol protocol = HostAccessProtocol::Invalid;
+    adios2::XRootDTransferProtocol xrootdTransferProtocol = XRootDTransferProtocol::XRootD;
+    adios2::HostConfig *hostConfig = nullptr;
+};
+
 class Remote
 {
 
 public:
-    Remote(const adios2::HostOptions &hostOptions);
+    Remote();
+    Remote(const RemoteSetup &remoteSetup);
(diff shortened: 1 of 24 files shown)


pnorbert:
@eisenhauer This is for your review because I refactored the remote connection prep code from BP5Reader to Remote. Also, I couldn't test any of this with the xrootd connection only with the remote server.

Let's merge only once we are sure it works with xrootd too.

eisenhauer:
I'll take a look.  The sanitizers seem to be broken, so nothing is passing currently.

source/adios2/toolkit/remote/Remote.h
```diff
@@ -81,12 +91,27 @@ class Remote
...
+      If not found in environment, returned rs.protocol will be HostAccessProtocol::Invalid
+      and rs.hostName will be "localhost".
+*/
+RemoteSetup GetRemoteSetup(const std::string &remoteHost);
+
+/* Create/Get a Remote object connected to the target described in RemoteSetup */
+std::shared_ptr<adios2::Remote> GetRemote(const RemoteSetup &remoteSetup,
+                                          const std::string &RemoteFileName,
+                                          const adios2::Mode openMode, const bool rowMajorOrdering,
+                                          const Params remoteParams);
```
eisenhauer: remoteParams is potentially lots of strings, maybe it should be passed by reference here?

eisenhauer (dismissed):

source/adios2/toolkit/remote/Remote.h
pnorbert: yes, mistake. Merge your PR, then I will rebase and fix this too

eisenhauer (approved):

PR #5133, opened 2026-07-24
Reader-defined derived variables (local BP5 path)

eisenhauer:
IO::DefineReaderDerivedVariable defines an expression on a read IO over file variables; the reader computes it on Get. Parsed at define, resolved lazily against the file on Open, and kept out of the writer-side derived registry so a write engine sharing the IO never picks it up.

BP5Deserializer synthesizes a VarRec routed off the input variables' block structure (GetMetadataBase redirect), requiring congruent input shapes. Bindings for CXX, Python, C, and Fortran.

Also: reader-derived vars carry no precomputed min/max, and the existing "no stored min/max" path now reports an invalid range (min>max).

### DIFF
--- a/source/adios2/core/IO.cpp
+++ b/source/adios2/core/IO.cpp
@@ -1054,6 +1054,146 @@ VariableDerived &IO::DefineDerivedVariable(const std::string &name, const std::s
     }
     return variable;
 }
+
+void IO::DefineReaderDerivedVariable(const std::string &name, const std::string &expression)
+{
+    PERFSTUBS_SCOPED_TIMER("IO::DefineReaderDerivedVariable");
+
+    // Name shares the IO namespace with every variable. Reject the collisions we
+    // can see now; a clash with a file variable not yet installed is caught at
+    // resolve time, when creating the placeholder throws on the duplicate name.
+    if (m_ReaderDerivedVariables.find(name) != m_ReaderDerivedVariables.end() ||
+        m_VariablesDerived.find(name) != m_VariablesDerived.end() ||
+        m_Variables.find(name) != m_Variables.end())
+    {
+        helper::Throw<std::invalid_argument>("Core", "IO", "DefineReaderDerivedVariable",
+                                             "variable " + name + " already defined in IO " +
+                                                 m_Name);
+    }
+
+    // Parse now so syntax errors surface at define time. Input variables are not
+    // resolved here: on a read IO the file's variables need not exist until Open.
+    derived::ExprNode exprTree = detail::ParseToExprNode(expression);
+
+    ReaderDerivedVariable rec;
+    rec.Expression = expression;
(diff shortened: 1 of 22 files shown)


pnorbert (dismissed):

anagainaru:
Is this ready for review? I would like to test it before we merge.

eisenhauer:
> Is this ready for review? I would like to test it before we merge.

Yes, I think it is.  It could be better, particularly WRT shape-changing operators, but it's a start...

source/adios2/core/IO.cpp
```diff
@@ -1054,6 +1054,153 @@ VariableDerived &IO::DefineDerivedVariable(const std::string &name, const std::s
...
     return variable;
 }
+
+void IO::DefineReaderDerivedVariable(const std::string &name, const std::string &expression)
+{
+    PERFSTUBS_SCOPED_TIMER("IO::DefineReaderDerivedVariable");
+
+    // Name shares the IO namespace with every variable. Reject the collisions we
+    // can see now; a clash with a file variable not yet installed is caught at
+    // resolve time, when creating the placeholder throws on the duplicate name.
```
anagainaru, line 1072: We should do the same for the write time if that's not already the case

anagainaru (commented):
First pass looks good, I am wondering if we need a new function `DefineReadDerivedVariable` and not use the existing one `DefineDerivedVariable` that sees that we open the io for read (even though we call `Open` after we define the variable). I'll run this tomorrow with the other code I used to test the read side last time and will come back with an approve if everything looks good.

One other thought, what happened to all the other code that reads the expression from the variable name and computes the derived variable values, I will check tomorrow that we are not reimplementing the same thing for the two cases.

eisenhauer:
> First pass looks good, I am wondering if we need a new function `DefineReadDerivedVariable` and not use the existing one `DefineDerivedVariable` that sees that we open the io for read (even though we call `Open` after we define the variable). I'll run this tomorrow with the other code I used to test the read side last time and will come back with an approve if everything looks good.

The problem is that a single IO can be used for both read and write, and this may be a common pattern for example in a workflow where data is read, operated upon and then rewritten in some form.  The worry is that something intended as a reader-side derived variable (incurring costs only if you do a Get on it) might get interpreted as a writer-side variable (which has costs on every EndStep if its input variables are written).  Separating "reader-side" and "writer-side" derived variables at least gives us some isolation between intents and seemed to me to be a bit clearer than say, using the DerivedVarType parameter to do the same thing.

anagainaru (approved):
I ran it and it looks good, I checked to see how much it reimplements what was there for the writer side variables computed at read and it seems only the definition is replicated which is fine since I agree it would be a pain to implement (I didn't think someone would use the same io for both read and write). I think this is good to go

PR #5145, opened 2026-08-12
Migrate kwsys SystemTools file/env/dynamic-loading to C++17

vicentebolea:
Replace adios2sys::SystemTools file/path operations and Directory
listing with std::filesystem across core, engines, and toolkit
transports. Replace GetEnv/PutEnv/SplitString with small
adios2::helper equivalents. Replace DynamicLoader/ConvertToOutputPath
in adiosDynamicBinder.cpp with direct dlopen/LoadLibraryExW calls.

kwsys is not removed: bpls still uses CommandLineArguments (next PR),
and MD5 (BP5Helper.cpp, KVCacheCommon.cpp) is untouched.


vicentebolea:
Recreating from vicentebolea/ADIOS2 fork instead.

PR #5148, opened 2026-08-12
Replace kwsys CommandLineArguments in bpls with CLI11

vicentebolea:
Stacked on #5147.

Vendor CLI11 v2.7.2 (single-header, BSD-3) with an
ADIOS2_USE_EXTERNAL_CLI11 option, and rewrite bpls's argument parsing
against it. CLI11 combines short flags (-la == -l -a) and collects
positionals natively, so bpls's hand-rolled retry/explode logic around
kwsys's CommandLineArguments is dropped rather than ported.

kwsys is not removed: MD5 (BP5Helper.cpp, KVCacheCommon.cpp) still
depends on it.


vicentebolea:
Recreating with base=master to avoid pushing kwsys-cxx17-filesystem to origin.

PR #5150, opened 2026-08-12
ci: bump clang sanitizer to 22

vicentebolea:


eisenhauer (dismissed):

eisenhauer (approved):
Approving optimistically since it seems to have gotten further...

PR #5160, opened 2026-08-25
Derived vector JIT compilation

eisenhauer:
This PR does just-in-time dynamic code generation of derived variable expressions in at least some limited set of conditions.  It is currently not on by default, but only when the environment variable "DerivedFuse" is set.  When that *is* set, we generate a subroutine that does the complete expression element-by-element (as opposed to the current "interpreted" process where each operator is performed on whole array chunks, potentially leading to large temporaries and many trips through memory for the data).  

This code requires DILL's new Vector operations, currently supported on arm and x86_64 platforms.  If a derived expression does not satisfy eligibility requirements, we fall back to interpreted execution.
  - every operation is +  -  *  /  unary-negate  sqrt — including ^2, which becomes a multiply,
  - the expression is purely element-wise (no curl(), cross(), etc)
  - all inputs and the result are one type: all-float or all-double,
  - the CPU has the vector ISA (arm64: always; x86_64: AVX + FMA3).


pnorbert (approved):

anagainaru:
I would have liked to test it before merging

eisenhauer:
Sorry!  I didn't know if you were working yet...  Please eval, there's more to come.

PR #5161, opened 2026-08-27
Derived expression grammar rework

eisenhauer:
'/' always division, paths via bindings or backticks, named constants, attribute bindings (@attr), single-value variable inputs; drop dead slicing syntax

### DIFF
--- a/docs/user_guide/source/advanced/derived_variables.rst
+++ b/docs/user_guide/source/advanced/derived_variables.rst
@@ -33,6 +33,36 @@ There are currently two types of derived variables accepted by ADIOS2, one that 
                                 "magnitude(x, y)",
                                 adios2::DerivedVarType::StatsOnly);
 
+The expression consists of optional binding lines (``name = ...``) followed by the formula.
+Formula identifiers are simple names; ADIOS variable names containing ``/`` or ``.``
+(including leading-``/`` absolute paths) must either be bound to a simple name as above,
+or written inline between backticks (``` `var/Ux` + 1 ```).  Within a formula, ``/`` is
+always division: ``a/b`` divides the variables ``a`` and ``b``.
+
+A binding whose right-hand side is a number defines a named constant, usable anywhere in
+the formula and evaluated at definition time::
+
+   bpIO.DefineDerivedVariable("energy",
+                              "c = 2.99792458e8 \n"
+                              "ux = var/Ux \n uy = var/Uy \n uz = var/Uz \n m0 = var/m0 \n"
+                              "sqrt((ux*ux + uy*uy + uz*uz)*c^2 + (m0*c)^2)",
+                              adios2::DerivedVarType::StatsOnly);
+
+``@name`` refers to the single-element numeric attribute ``name``, either directly in the
+formula or through a binding.  ``@`` composes with every form a variable name can take:
+plain names inline, backticked names with special characters (``@`group/scale```), and
+both forms as binding right-hand sides.  The attribute must be defined before the derived variable,
+and its *current* value is used each time the expression is evaluated, so modifiable
+attributes take effect step by step::
+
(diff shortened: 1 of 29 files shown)


docs/user_guide/source/advanced/derived_variables.rst
```diff
@@ -33,6 +33,38 @@ There are currently two types of derived variables accepted by ADIOS2, one that
...
+A binding whose right-hand side is a number defines a named constant, usable anywhere in
+the formula and evaluated at definition time::
+
+   bpIO.DefineDerivedVariable("energy",
+                              "c = 2.99792458e8 \n"
+                              "ux = var/Ux \n uy = var/Uy \n uz = var/Uz \n m0 = var/m0 \n"
+                              "sqrt((ux*ux + uy*uy + uz*uz)*c^2 + (m0*c)^2)",
+                              adios2::DerivedVarType::StatsOnly);
+
+A binding whose right-hand side is ``@name`` takes its value from the numeric attribute
```
anagainaru: I don't think we can have an attribute and a variable with the same name so the @ convention should not necessary be needed. Is this a valid expression: "`@name` + 5"? It should be considering "`name` + 5" is valid.

source/adios2/toolkit/derived/ExprCodeStream.cpp
```diff
@@ -976,14 +984,15 @@ GetDims(const ExprCodeStream &cs,
...
+    // Find reference dims from the first non-scalar input variable
+    // (single-value inputs broadcast and contribute no shape)
     std::tuple<Dims, Dims, Dims> refDims = {{}, {}, {}};
     for (const auto &buf : cs.Buffers)
     {
         if (buf.IsInput)
         {
             auto it = nameToDims.find(buf.VarName);
-            if (it != nameToDims.end())
+            if (it != nameToDims.end() && !std::get<1>(it->second).empty())
```
anagainaru, line 1000: I would add a comment to what is the second element because we will def forget

anagainaru (commented):
Looks good but I still don't have a place to run things yet :) The testing you are adding should be enough. Three things:
- We need a way to give the expression and generate the testing automatically so we don't repeat so much code (+ doubling it for writer/reader side variables)
- It seems attributes cannot be inline variables "@attribute + 2". I think it should be allowed
- I assume this is valid but we do not test it "``var`` + 1" (variable not containing / )

docs/user_guide/source/advanced/derived_variables.rst
eisenhauer: Actually, variables and attributes have completely different namespaces, so you can have an attribute and a variable of the same name without error.  So we could get rid of the @ notation, but we'd have to decide which one "wins" if there are two of the same name.  (This might be a good time to say that we're not aggressive about throwing semantic errors in this.  You could issue a bind with a LHS that was a pre-existing variable.  You can issue multiple binds with the same LHS.  We should detect that in some pass and complain.  We currently do not.)

source/adios2/toolkit/derived/ExprCodeStream.cpp
eisenhauer: Agreed.

docs/user_guide/source/advanced/derived_variables.rst
anagainaru: I am fine with using @ for now to disambiguate which one they need. Once we make derived variables ON by default we should freeze all the changes to the expression.

pnorbert (dismissed):

anagainaru (approved):

eisenhauer:
@pnorbert The CMake changes mean that you'll need to re-review as Infrastructure...

pnorbert (approved):

Issue #5162, opened 2026-08-28
using bpls on a corrupted adios file leads to segfault


PrometheusPi:
One of our simulations tried writing an adios checkpoint, but could not finish. The file is this incomplete. Testing this using `pbls` leads to a segfault:

```
$ bpls checkpoint_100000.bp5/
Caught signal 11 (Segmentation fault: address not mapped to object at address 0x10)
==== backtrace (tid: 120627) ====
0 0x00000000004ee144 adios2::format::BP5Deserializer::GenerateReadRequests() ???:0
1 0x00000000002f78bc adios2::core::engine::BP5Reader::PerformLocalGets() ???:0
2 0x00000000002f8410 adios2::core::engine::BP5Reader::PerformGets() ???:0
3 0x00000000002f8c5c adios2::core::engine::BP5Reader::DoClose() ???:0
4 0x00000000001b5cb0 adios2::core::Engine::Close() ???:0
5 0x000000000047ebf0 adios2::utils::doList() ???:0
6 0x000000000047fd68 adios2::utils::bplsMain() ???:0
7 0x0000000000417bac main() ???:0
8 0x0000000000027540 __libc_start_call_main() ???:0
9 0x0000000000027618 __libc_start_main_alias_2() :0
10 0x0000000000417ef0 _start() ???:0
Segmentation fault (core dumped)
```

Would it be a better option to catch such errors and print to screen that the file is corrupted?

pnorbert:
Hi Richard, is this file on Frontier? Can we see it? It's a rare occasion where it got corrupted in a way we did not count for.

PrometheusPi:
No, the file was on Jülich Supercomputing Centre's (JSC) Jupiter system but was removed by now.
But I/colleagues surprisingly often encountered this before (e.g. also on LUMI, which is very similar to Frontier). 

I assume that if I run into this problem on Frontier, I should post the file here, right?

pnorbert:
Next time, can you save the md.0 mmd.0 md.idx files from a corrupted dataset and send it to me? Plus the output of `ls -l xxx.bp5` and `ls -ltr --time-style=+%s xxx.bp5`.

PrometheusPi:
A colleague of mine, @JessicaTiebel, could trigger a `bpls` segfault on JCS Jupiter again. 
[...]

PR #5166, opened 2026-08-28
Bp5 truncated/corrupt file recovery

eisenhauer:
This PR adds some truncation/corruption recovery testing tools.  These are not part of CI, but added here so they can be rerun occasionally if formats or methods change.  There are also bug-fixes for the issues that these tools uncovered.  Detections of truncated files, attempts at recovering what data can be provided to the user, exceptions when appropriate.  It also cleans up a bunch of hangs.  Turns out that when a writer dies without closing, the active flag is left on, so a reader will assume it's in a streaming-through-file situation and wait, used to be pretty much forever.  Now we're going to be more actively trying to detect that the file is really truncated, remember that we've determined that so that the next time we read we don't wait again, etc.  These fixes *may* fix issue #5162.  Can't really know without being able to duplicate it.

### DIFF
--- a/docs/user_guide/source/engines/bp5.rst
+++ b/docs/user_guide/source/engines/bp5.rst
@@ -58,6 +58,8 @@ This engine allows the user to fine tune the buffering operations through the fo
 
    1. **OpenTimeoutSecs**: (Streaming mode) Reader may want to wait for the creation of the file in ``io.Open()``. By default the Open() function returns with an error if file is not found.
 
+   1. **EOFWaitSecs**: How long a read of data the metadata refers to may wait on an end-of-file before failing, once the writer is known to be gone. The wait absorbs shared-filesystem consistency lag; on a file truncated by an interrupted writer, reads of the missing bytes fail cleanly after roughly 1.5x this many seconds. Recovery tools can set it low to fail fast.
+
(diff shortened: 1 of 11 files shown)


pnorbert:
The default 30 seconds for EOFWaitSecs is breaking the streaming equivalency of a code re SST vs BP5, doesn't it? It should be  hours not seconds for a default value. This was always a thorny issue, a rarely used feature vs hanging on a file that was not closed properly.

eisenhauer:
> The default 30 seconds for EOFWaitSecs is breaking the streaming equivalency of a code re SST vs BP5, doesn't it? It should be hours not seconds for a default value. This was always a thorny issue, a rarely used feature vs hanging on a file that was not closed properly.

The 30 seconds is unchanged in this PR, all we do here is actually make that value configurable.  But yeah, it's thorny.  I'd really love to have some kind of "I know I'm part of a streaming workflow, wait for me" indication that the Writer could provide, rather than relying upon the active flag, just because the latter is an unreliable indicator.  It would be a behaviour change for that rarely used feature, but it would help us behave better in the more common case that a writer died before it could close the file properly.

(I should add that the timeout specified is the max timeout in a progressive backoff.  If a read fails at EOF, we first try again in millisecs, backing off by a factor of two at each failure.  30 seconds is our max backoff, so we'll have a total of close to 45 seconds in the end.  I made it configurable because with the default it took 20 minutes to just to run my truncated file tests because every *@^*&#$%^ one of them had to wait 45 seconds to decide that, no, that file isn't really going to get bigger.)

pnorbert (approved):

PR #5171, opened 2026-09-02
Add configurable HTTPS campaign access

pnorbert:
Teach hosts.yaml and CampaignReader to select HTTPS endpoints and CA files, parse non-default ports, and verify TLS peers. This lets one campaign archive use Docker DNS internally and a trusted loopback endpoint from WSL without hostname aliases.

This was created for testing campaigns in hpc-campaign's container from the host environment, which exposed the simplicity of the current solution.


pnorbert:
Close this and add new PR

PR #5176, opened 2026-09-04
campaign: prepare simple-file reads for XRootD

pnorbert:
This PR adds the plumbing for reading arbitrary byte ranges through XRootD, but deliberately leaves the actual XRootD reader unimplemented.

Key changes:

- Adds Remote::OpenSimpleFile(..., Params) and the GetRemoteSimpleFile() factory for opening unstructured files rather than ADIOS datasets. The factory   supports native XRootD, HTTP/HTTPS/XrdCl, and SSH/EVPath using the existing host configuration logic. See source/adios2/toolkit/remote/Remote.cpp:323    and source/adios2/toolkit/remote/Remote.h:51.
- Refactors CampaignReader to use that factory. It now calls Read(offset, size, data), requires XRootD to return a non-null asynchronous handle, and   completes it with WaitForGet(). Unsupported raw reads produce an explicit error. See source/adios2/engine/campaign/CampaignReader.cpp:1576.
- Keeps SSH/EVPath behavior working and updates its unit test to go through the new factory. See testing/adios2/unit/TestRemote.cpp:56.
- Adds an unrelated safety improvement to campaign selection validation by avoiding overflow in start + count. See source/adios2/engine/campaign/CampaignReader.tcc:104.

For the XRootD implementer:

- Both XRootD classes now expose OpenSimpleFile(), currently just delegating to their normal Open().
- Both raw-read methods are still stubs returning nullptr:
    - Native SSI: source/adios2/toolkit/remote/XrootdRemote.cpp:419
    - HTTP/HTTPS/XrdCl: source/adios2/toolkit/remote/XrootdHttpRemote.cpp:540
- No files under source/utils/xrootd-plugin/ were changed. Therefore the server plugin still needs a proper range-read operation accepting a filename,  byte offset, and byte count.
- The native client should submit that operation similarly to Get(), set the destination and expected size, and return its promise as the handle.
- The HTTP client needs a matching URL/request encoding and an AsyncGet with expectedSize = Size.
- The HTTP-to-SSI decoder currently recognizes only get and batchget, so it must be extended for the new raw-read operation. See source/utils/xrootd-plugin/XrdHttpSsiHandler.cpp:305.
- Server-side validation should cover offset/count overflow, reads beyond EOF, short reads, file-open failures, and the existing single-response size  limit.
- Add native XRootD and HTTP/XrdCl tests; the commit only tests the SSH path.

In short: this commit defines the end-to-end API and consumer expectations. The remaining work is the XRootD client request construction, server/plugin  range-read handler, HTTP request encoding/decoding, and tests.

### DIFF
--- a/scripts/docker/spin-xrootd/xrootd-http.cfg
+++ b/scripts/docker/spin-xrootd/xrootd-http.cfg
@@ -22,8 +22,12 @@ oss.statlib -2 libXrdSsi.so
 # Load ADIOS2 SSI service plugin
 ssi.svclib /usr/local/lib64/libadios2_xrootd.so
 
-# Data directories served as plain files.  /data is the Spin mount; /srv is
-# where the hpc-campaign integration image places its campaign data.
+# Data directories served as plain files: /cfs is the NERSC community file
+# system as mounted on Spin (KSTAR data lives under /cfs/www), /data is the
+# image's generic mount point, /srv is where the hpc-campaign integration
+# image places its campaign data.
+ssi.fspath /cfs
+all.export /cfs r/o
 ssi.fspath /data
 all.export /data r/o
 ssi.fspath /srv
--- a/scripts/docker/spin-xrootd/xrootd-http.cfg
+++ b/scripts/docker/spin-xrootd/xrootd-http.cfg
@@ -6,8 +6,12 @@ all.adminpath /var/spool/xrootd
 all.pidpath /run/xrootd
 xrd.homepath /tmp
 
-# Enable SSI filesystem
-xrootd.fslib libXrdSsi.so
+# SSI stacked over the default file system ("default"): SSI requests go to
+# the ADIOS2 service, while paths listed under ssi.fspath are ordinary files
+# served by XRootD itself (HTTP Range, HEAD, stat).  Campaign image and text
+# replicas and tar archives are read that way; only ADIOS dataset queries
+# (URLs with an /_adios/ segment) reach the plugin.
+xrootd.fslib libXrdSsi.so default
 
 # Export path for SSI operations
 all.export /etc nolock r/w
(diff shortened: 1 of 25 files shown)


eisenhauer (dismissed):
When you get to the http stuff, don't forget the HEAD path!

pnorbert:
It is up to @eisenhauer to decide if he will go with this or needs a different approach for the XRootD plugin. This PR is one suggested approach to handle reading images and texts from remote with a unified interface that was with OpenSimpleFile(), GetRemoteSimpleFile() and Read().

pnorbert:
See https://hpc-campaign.readthedocs.io/en/latest/developer/index.html for building the docker container for all services and creating test campaigns for each access protocol. Then a single test for reading an image/text is:

`bpls -l ./testdata-xrootd-files.aca -d testdata/T_10_15_00000.png/800x800`
`bpls -l ./testdata-xrootd-files.aca -d testdata/readme`

eisenhauer:
So, several things in the commit that I added here.  First, it does revamp the URL pattern so that rather than all our SSI-plug-in serviced requests being differentiated by a /adios prefix, they now have a "/_adios/" as the first thing after the filename, and the regex looks for that.  This should simplify things for Pelican, but the main point is that we can then have normal "/filepath" byte range queries that come into xrootd or https.  We also add a mechanism to do those file range queries in the non-http XRootD configuratoin using XrdCl (which was already an implicit dependency of XrdSSI).

There was also some random bug cleanup, one of which ended as a different PR (python signed/unsigned char issue).

pnorbert:
Only @vicentebolea can approve this combo PR now

eisenhauer:
> Only @vicentebolea can approve this combo PR now

Hmm.  Poorly thought out on my part...

eisenhauer (approved):

eisenhauer:
Actually, @pnorbert, maybe you can approve, you're not the last pusher...

anagainaru (approved):
What a loophole :)

eisenhauer:
And I'm still going to have to override checks because of the Code Scanning thing...  Ah well.

PR #5181, opened 2026-09-11
Python: accept int8 or uint8 buffers for char variables in Get

eisenhauer:
On aarch64 Linux char is unsigned, so nanobind's dtype<char> is uint8 there while stream.py maps char to int8. Every read() of a char variable on that platform failed with "Type mismatch between Python buffer and incoming data". Seen on the hpc-campaign integration image (arm64): all TEXT datasets unreadable on every transport.

numpy has no char dtype and the bytes are identical either way, so Get now accepts either 8-bit integer buffer for char; all other types still require an exact match.

Adds a serial Stream test that round-trips int8 and uint8 arrays. The existing 8-bit round-trip tests are MPI-only and never ran on the aarch64 CI job, which is why this went unnoticed.

### DIFF
--- a/bindings/Python/nbEngine.cpp
+++ b/bindings/Python/nbEngine.cpp
@@ -175,6 +175,24 @@ void Engine::PerformDataWrite()
     m_Engine->PerformDataWrite();
 }
 
+namespace
+{
+// numpy has no char dtype. char is signed on x86 and unsigned on aarch64 Linux, so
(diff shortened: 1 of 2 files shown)


pnorbert (approved):

eisenhauer:
@vicentebolea Bypassing the rules because CodeQL isn't satisfied for reasons not related to this PR.

PR #5183, opened 2026-09-15
Fix Python detection for MPI pip builds

shimwell:
Would it be possible to add a check for `mpi4py` here when building the Python package with MPI support?

Currently, when `ADIOS2_USE_PIP=ON` and MPI is enabled, Python itself is detected but `mpi4py` is not checked. Since `ADIOS2_HAVE_Python` depends on `PythonModule_mpi4py_FOUND` for MPI builds, configuration completes with `MPI: ON` but `Python: OFF`. The resulting wheel therefore does not contain `adios2.bindings`.

The other Python configuration paths already check for `mpi4py`, so this adds the same check to the pip path.

I tested this by building a wheel with Python 3.12, MPICH and mpi4py. CMake reported both `MPI: ON` and `Python: ON`, and constructing `ADIOS(MPI.COMM_WORLD)` worked with both one and two MPI ranks.


eisenhauer (approved):
Generally I'd defer this to @vicentebolea, but he's out for a couple of weeks and this looks very straightfoward, so I'll approve.

eisenhauer:
And bypass merging because CodeQL check is broken.

shimwell:
Many thanks for the review and merge, much appreciated

